PR #4, opened 2023-09-05
Small adjustments to docs

maxschulz-COL:
## Description
- add docs link to package README
- add link to docs also to top-level README (prefer this for as long as we only have one package)
- fix broken link in docstring of `Navigation` model
- set version to `0.1.0b0`

## Checklist

## Types of changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,10 @@
 
 Vizro is a toolkit for creating modular data visualisation applications
 
+## Documentation
+
+See our [`docs`](https://vizro.readthedocs.io/en/latest/) folder for more details
+
 ## Contributors
 
 See the `authors` file for more details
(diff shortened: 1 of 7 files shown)


petar-qb (approved):

maxschulz-COL:
> Could you make one additional change inside the `contributing.md`? Under the section: "How to interact with the repository", inside the third step, maybe we should make a change: `pip install --editable` -> `pip install --editable .`

Hm I think this is a detail that we may want to omit. Ultimately we are not ecouraging the user to circumvent `pip`, but rather to understand in general what is going on behind the scenes. Wdyt?

petar-qb:
> Hm I think this is a detail that we may want to omit. Ultimately we are not ecouraging the user to circumvent pip, but rather to understand in general what is going on behind the scenes. Wdyt?

When I think more about it, probably makes sense. Leave as it is then 😄.

petar-qb (approved):

PR #5, opened 2023-09-05
Remove flaky part of the export data xslx unit test

petar-qb:
## Description
Flaky parts of the test are commented out.

## Checklist

## Types of changes

### DIFF
--- a/vizro-core/tests/unit/vizro/actions/test_export_data_action.py
+++ b/vizro-core/tests/unit/vizro/actions/test_export_data_action.py
@@ -1,7 +1,6 @@
 import pytest
 from dash._callback_context import context_value
 from dash._utils import AttributeDict
-from dash.dcc import send_data_frame
 
 import vizro.models as vm
 from vizro import Vizro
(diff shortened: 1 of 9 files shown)


vizro-core/tests/unit/vizro/actions/test_export_data_action.py
```diff
@@ -237,12 +237,16 @@ def test_multiple_targets_file_format_xlsx(self, callback_context_export_data, g
         )
         assert result["download-dataframe_scatter_chart"]["filename"] == scatter_chart_expected["filename"]
```
huong-li-nguyen, line 238: I've looked into this as well - there is not really a better way to do this check as the results do not seem to be deterministic when using the `to_excel` writer. Decoding and encoding the different base64 strings also do not help.

In this case, I would propose to remove the entire test, so the entire `test_multiple_targets_file_format_xlsx` function. I think it should be fine, because in the end we shouldn't test if the the `dcc.send_data_frame` works properly but what is important to us is that we pass on the **correct dataframe** to the Download component. 

Given that we test for exact content in all other tests using the `to_csv `writer (that does seem to be deterministic), I am fine with removing this test completely.

huong-li-nguyen (changes requested):

huong-li-nguyen (approved):

maxschulz-COL (approved):

PR #6, opened 2023-09-05
Use build unfctionality of RTD

maxschulz-COL:
## Checklist

## Types of changes

### DIFF
--- a/vizro-core/.readthedocs.yaml
+++ b/vizro-core/.readthedocs.yaml
@@ -9,11 +9,16 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
+  commands:
+    - pip install hatch
+    - cd vizro-core/ && hatch run docs:mkdocs build
+    - mkdir --parents $READTHEDOCS_OUTPUT
+    - mv vizro-core/site/ $READTHEDOCS_OUTPUT/html
 
-mkdocs:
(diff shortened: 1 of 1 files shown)


nadijagraca (approved):
Looks good. 

Love the interaction with rtd. :)

vizro-core/src/vizro/__init__.py
```diff
@@ -5,6 +5,6 @@
 
 __all__ = ["Vizro"]
 
-__version__ = "0.1.0"
+__version__ = "0.1.0dev0"
```
huong-li-nguyen, line 8: Do we not go live with 0.1.0? or why do we change to the dev version?

maxschulz-COL: We will, this is just for development hygiene. Since we are not live, we should aim to be clear in that what people see on github is not v0.1.0, but the development version leading up to it.

I think it is even more important for the schema, if someone was to download and use it, that person would not know that then the schema for the actualy version 0.1.0 would be different

huong-li-nguyen (approved):

petar-qb (approved):
Looks good to me!

nadijagraca (approved):

Joseph-Perkins (approved):

PR #10, opened 2023-09-05
Add tests for graph, parameter, filter, and types

lingyielia:
## Description
- [x] Additional tests for graph
- [x] Additional tests for parameter and filter
- [x] Additional tests for Types

## Checklist

## Types of changes


vizro-core/tests/unit/vizro/models/_components/test_graph.py
```diff
@@ -1,33 +1,90 @@
...
+        assert result.children.id == "text_graph"
+        assert result.children.className == "chart_container"
+        assert result.children.config == {
+            "autosizable": True,
+            "frameMargins": 0,
+            "responsive": True,
+        }
+        # result.children.figure.layout.template is ignored
+        assert isinstance(result, dcc.Loading)
+        assert isinstance(result.children.figure, go.Figure)
```
huong-li-nguyen, line 90: Would remove all of this and instead do a json comparison of the final component. This will then cover all of your checks above and more. The reason why we opted for this is because in very nested objects you would have numerous `isinstance` checks and equality checks. The json comparison is a more efficient way to do so.

Would first create a fixture of the expected result:

```
@pytest_fixture()
def expected_graph():
     return dcc.Loading(
            dcc.Graph(
                id="text_graph",
                figure=create_empty_fig("NO DATA"),
                config={
                    "autosizable": True,
                    "frameMargins": 0,
                    "responsive": True,
                },
                className="chart_container",
            ),
            color="grey",
            parent_className="chart_container",
        )
```

```suggestion
result = json.loads(json.dumps(graph.build(), cls=plotly.utils.PlotlyJSONEncoder))
expected = json.loads(json.dumps(expected_graph, cls=plotly.utils.PlotlyJSONEncoder))
assert result == expected
```

vizro-core/tests/unit/vizro/models/_components/test_graph.py
```diff
@@ -1,33 +1,90 @@
...
+            figure=px.scatter(
+                data_frame=px.data.gapminder(),
+                x="gdpPercap",
+                y="lifeExp",
+                size="pop",
+                color="continent",
+                hover_name="country",
+                size_max=60,
+            ),
+        )
```
huong-li-nguyen, line 76: Why not re-use the fixture we have?
```suggestion
    def test_graph_build(self, standard_px_chart):
        graph = vm.Graph(
            id="text_graph",
            figure=standard_px_chart
        )
```

vizro-core/tests/unit/vizro/models/_components/test_graph.py
```diff
@@ -1,33 +1,90 @@
...
+        with pytest.raises(ValidationError, match="must provide a valid CapturedCallable object"):
+            vm.Graph(
+                figure=standard_go_chart,
+            )
 
-def test_failed_graph_with_wrong_figure(standard_go_chart):
-    with pytest.raises(ValidationError, match="must provide a valid CapturedCallable object"):
-        vm.Graph(
-            figure=standard_go_chart,
+    @pytest.mark.parametrize("title, expected", [(None, 24), ("Test", None)])
```
huong-li-nguyen, line 28: Shall we add a test for the __getitem__ method of the Graph? Then I think we cover everything in the instantiation

vizro-core/tests/unit/vizro/models/_controls/test_parameter.py
```diff
@@ -78,6 +80,26 @@ def test_target_and_title_generation_valid(self, test_input, title):
...
+        [
+            (vm.Slider(min=0, max=1, value=0.8)),
+            (vm.RangeSlider(min=0, max=1, value=[0.2, 0.8])),
+            (vm.Checklist(options=["lifeExp", "gdpPercap", "pop"], value=["lifeExp"])),
+            (vm.Dropdown(options=["lifeExp", "gdpPercap", "pop"], multi=False, value="lifeExp")),
+            (vm.RadioItems(options=["lifeExp", "gdpPercap", "pop"], value="lifeExp")),
+        ],
+    )
+    def test_actions_generation_valid(self, test_input):
+        parameter = Parameter(targets=["scatter_chart.x"], selector=test_input)
```
huong-li-nguyen, line 94: Does this actually work?

You are providing numeric and categorical parameters to the same target. I am surprised that this doesn't throw an error?

huong-li-nguyen (changes requested):
Looks good to me, have a few suggestions but otherwise good to go for me 👍

vizro-core/tests/unit/vizro/models/test_types.py
```diff
@@ -148,12 +163,28 @@ def test_decorated_graph_function(self):
...
+    def test_decorated_df_standard_case(self, gapminder):
+        fig = decorated_graph_function_px(gapminder)
+        assert len(fig.data) > 0
+        assert fig.__class__ == _DashboardReadyFigure
+
+    def test_decorated_df_str(self):
+        fig = decorated_graph_function_px("gapminder")
+        assert fig == _DashboardReadyFigure()
+        assert len(fig.data) == 0
+
```
maxschulz-COL, line 175: Are these tests at the correct place / can you remind me again why we devided to test these?

maxschulz-COL (changes requested):
Thanks a lot! Have just minor suggestions aside from Ls comments

vizro-core/tests/unit/vizro/models/test_types.py
lingyielia: to cover this block of code:
```
                if isinstance(captured_callable["data_frame"], str):
                    # Enable running e.g. px.scatter("iris") from the Python API. Don't actually run the function
                    # because it won't get work as there's no data. It's vital we don't fetch data from the data manager
                    # yet either, because otherwise all lazy data will be loaded before the dashboard is started.
                    # This case is not relevant for the JSON/YAML API, which is handled separately through validation of
                    # CapturedCallable.
                    fig = _DashboardReadyFigure()
                else:
                    # Standard case for px.scatter(df: pd.DataFrame).
                    fig = func(*args, **kwargs)
                    fig.__class__ = _DashboardReadyFigure

File: vizro-core/tests/unit/vizro/models/_components/test_graph.py [hunk 6]
```diff
@@ -1,33 +1,90 @@
...
+
+class TestProcessFigureDataFrame:
+    def test_process_figure_data_frame_str_df(self, standard_px_chart_with_str_dataframe, gapminder):
+        data_manager["gapminder"] = gapminder
+        graph_with_str_df = vm.Graph(
+            id="text_graph",
+            figure=standard_px_chart_with_str_dataframe,
         )
+        assert data_manager._get_component_data("text_graph").equals(gapminder)
+        assert graph_with_str_df.figure["data_frame"] == "gapminder"
```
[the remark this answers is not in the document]
lingyielia: Not sure I understand the suggestion. Could you elaborate a bit?

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
```diff
@@ -182,3 +184,15 @@ def test_filter_build(self, test_column, test_selector):
...
+        "test_input,expected", [("country", vm.Dropdown), ("year", vm.RangeSlider), ("lifeExp", vm.RangeSlider)]
+    )
+    def test_set_actions(self, test_input, expected):
+        filter = vm.Filter(column=test_input)
+        model_manager["test_page"].controls = [filter]
+        filter.pre_build()
+        default_action = filter.selector.actions[0]
+        assert isinstance(default_action, ActionsChain)
+        assert isinstance(default_action.actions[0].function, CapturedCallable)
+        assert default_action.actions[0].id == f"filter_action_{filter.id}"
```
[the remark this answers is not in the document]
lingyielia: I think these are key items we wanted to ensure, like the tests for button

vizro-core/tests/unit/vizro/models/_components/test_graph.py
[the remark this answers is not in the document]
maxschulz-COL: Just to get the expected string from the fixture, but I think it is also a little overkill

vizro-core/tests/unit/vizro/models/_components/conftest.py
```diff
@@ -0,0 +1,50 @@
...
+import plotly.graph_objects as go
+import pytest
+from dash import dcc
+
+import vizro.plotly.express as px
+from vizro.models._components.graph import create_empty_fig
+
+
+@pytest.fixture
+def standard_px_chart_with_str_dataframe():
```
maxschulz-COL, line 11: I think the idea is that if the fixture is used only in one file, we can also put it in that file, so this one would go into `test_graph.py`, likewise for the others potentially

vizro-core/tests/unit/vizro/models/test_types.py
maxschulz-COL: Hm I see, good point. Is there a reason why we do not use the `Model`. I was a bit confused, because these tests are grouped under `TestModelFieldPython` and there was no `Model` in your tests.

Aside, what is `assert len(fig.data)` testing?

vizro-core/tests/unit/vizro/models/_controls/test_parameter.py
```diff
@@ -70,14 +70,32 @@ class TestPreBuildMethod:
...
 
+    @pytest.mark.parametrize(
+        "test_input",
+        [
+            (vm.Checklist(options=["lifeExp", "gdpPercap", "pop"], value=["lifeExp"])),
+            (vm.Dropdown(options=["lifeExp", "gdpPercap", "pop"], multi=False, value="lifeExp")),
+            (vm.RadioItems(options=["lifeExp", "gdpPercap", "pop"], value="lifeExp")),
+        ],
+    )
+    def test_set_actions_valid(self, test_input):
```
maxschulz-COL, line 89: Could we to round this of once and for all also split the parameter pre-build into functions, align the naming with filtes, and have the test refer to these splits.

I think if we do it now, we don't have to touch these again, which would be quite nice.

vizro-core/tests/unit/vizro/models/_components/test_graph.py
```diff
@@ -1,33 +1,108 @@
...
+        graph_with_df = vm.Graph(
+            id="text_graph",
+            figure=standard_px_chart,
+        )
+        assert data_manager._get_component_data("text_graph").equals(gapminder)
+        with pytest.raises(KeyError, match="'data_frame'"):
+            graph_with_df.figure["data_frame"]
+
+
+class TestGetitem:
```
maxschulz-COL, line 82: Is there a reason we have the tests for the `GetItem` in its own class, but the `Call` test in the instantiation? I would probably move them into a class `TestDunderMethodsGraph` or similar. Then its more visually clear

maxschulz-COL (commented):
Thanks a lot for incorporating the changes! 

Have a few more ideas, one is to align the parameter to the filter, which is a bit annoying new work, but then I think we are consistent which is going to be a big improvement for readability.

vizro-core/tests/unit/vizro/models/_components/conftest.py
huong-li-nguyen: All of the fixtures here would go into test_graph 👍

vizro-core/tests/unit/vizro/models/_components/test_graph.py
```diff
@@ -1,33 +1,108 @@
...
+            actions=[
+                vm.Action(function=export_data()),
+            ],
+        )
+        actions_chain = graph.actions[0]
+
+        assert graph.id == id
+        assert graph.type == "graph"
+        assert graph.figure == standard_px_chart._captured_callable
+        assert isinstance(actions_chain, ActionsChain)
```
huong-li-nguyen, line 39: The actions I would just leave as is, so no need to configure something here. The export_data example is actually quite confusing because we wouldn't configure that action on a Graph, but only on a Button or something similar. 

The action that would make sense with the Graph is the filter_interaction, but given that we thoroughly test them in the test_actions.py, I don't think you need to test these additionally

huong-li-nguyen (approved):
Minor comments, the rest is good-to-go for me

huong-li-nguyen (approved):

huong-li-nguyen (dismissed):

vizro-core/tests/unit/vizro/models/test_types.py
lingyielia: assert len(fig.data)  is testing whether there are data attached

huong-li-nguyen (dismissed):

huong-li-nguyen (dismissed):
@lingyielia , I've added a test to be consistent with testing the action_validator_factory function, see this PR: https://github.com/mckinsey/vizro/pull/20

It's covering a test that I previously asked you to remove. I think this way we can still test the main functionality we want to test (e.g. property is clickData, proper ActionChain and Action created via the validator, etc.) without having to provide an action function that doesn't make much sense to use in combination with the Graph. I've added a test function as a fixture there.

huong-li-nguyen (approved):

maxschulz-COL (approved):

PR #11, opened 2023-09-05
Introduce ActionLoop and Action.build method

petar-qb:
## Description
The PR is based on the following [conversation](https://github.com/mckinsey/vizx-hyphen/pull/404#discussion_r1312726706).
Here is a PoC where all Action model specific code is moved to the Action.build method.

### Current solution:
Much like the `Control.build` method builds a selector (creating selector callbacks and returning selector components to be displayed), now selectors (or any other model component) build their `actions` (creating an action-specific callback + returning the components needed for it). 
So, Action's callback is created every time the component, where it's defined, is built. Also Action's components are returned from the `Action.build` method together with component UI elements where it's defined.

### What is actually changed (looking at how the actions system works):
 - Every time a component (where the action is defined) is built, all action component states are reset. Some information might be lost (which isn't strictly a bad thing - we just need to conform to what we want the expected behaviour to be).
 - Some memory on the client side machine is saved because only the actions components from the current page are "visible" (loaded/stored on the browser).
 - It might be that the page rendering time is slowed down because page actions callbacks are created every time the page is opened.
 - Actions components are not globally visible which means that they can be used only from the actions within the currently opened page (it maybe makes sense).

### TO DO:
- [ ] Check is the statelessness preserved (especially because action callbacks are created dynamically each time when the component where it's defined is built).
- [ ] Check how to handle the case when some actions require a global component.

--- 

### Another possible solution (Open Question):
If we don't want to risk or spend time investigating further how the dashboard behaves with all the changes made (Action.build called within the `model components build` method), we can decide to call the Action.build for each action within the `Dashboard build` method. It means that all action components are stored in a global container and that every action callback is created only once when the application starts. 

## Checklist

## Types of changes


vizro-core/src/vizro/actions/_action_manager/_action_manager.py
```diff
@@ -0,0 +1,40 @@
+"""The action manager handles creation of all required actions callbacks."""
+
+from vizro.actions._action_manager._build_action_loop_callbacks import _build_action_loop_callbacks
```
huong-li-nguyen, line 3: Shall we move the action manager to vizro.managers? Or did you keep it here because the first resulted in lots of circular import issues?

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
```diff
@@ -0,0 +1,40 @@
...
+        """Builds a callback for each Action model in the Dashboard and gets all required components.
+
+        Returns:
+            All required components for each action model in the Dashboard.
+        """
+        return [
+            action_component
+            for _, action in model_manager._items_with_type(Action)
+            for action_component in action.build()
+        ]
```
huong-li-nguyen, line 40: What's the difference in getting the action components in _build_action_models and _build_action_loop? Both seem to return "all required action components"? Can we provide a more distinct description in the docstrings?

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
```diff
@@ -0,0 +1,40 @@
+"""The action manager handles creation of all required actions callbacks."""
+
+from vizro.actions._action_manager._build_action_loop_callbacks import _build_action_loop_callbacks
+from vizro.actions._action_manager._get_action_loop_components import _get_action_loop_components
+from vizro.managers import model_manager
+from vizro.models import Action
+
+
+class ActionManager:
```
huong-li-nguyen, line 9: Thanks for adding it in! I think this looks a lot cleaner now!

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
petar-qb: `_build_action_loop` builds callbacks and returns required components for smooth work of the `action loop`.
Otherwise, `_build_actions_models` builds callbacks and returns required components for each defined `Action model` in the system.

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @staticmethod
+    def _build_action_loop():
+        """Builds callbacks for the action loop and gets all required components.
+
+        Returns:
+            All required components for the action loop.
+        """
+        _build_action_loop_callbacks()
+        return _get_action_loop_components()
```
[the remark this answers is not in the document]
petar-qb: I decided on this implementation to follow the Model.build method implementation. Model.build methods create required callbacks + returns required components.

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
huong-li-nguyen: I still don't get it 😄  What does smooth work of the action loop mean? Can you give actual output examples? 
What are the components returned by `_build_action_loop` and the ones by` _build_action_models`? Is one of them just returning dcc.Stores? While the other returns actual dash components such as dcc.Download?

vizro-core/changelog.d/20230906_095422_petar_pejovic_action_build_method_inside_dashboard.md
maxschulz-COL, line 1: Why are there two scripts

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
petar-qb: Looking the architecture given below:
 - "trigger_to_global_store",
 - "gateway",
 - "update_remaining_actions",
 - "executor",
 - "after_action",
 - and "clientside_callback" 

are callbacks created inside the `_build_action_loop_callbacks` function. These callbacks implements "action loop" mechanism that enables actions to be executed one by one in a sequence. 

---

`_get_action_loop_components` function returns all required components so that the "action loop" can work. Here are some examples of these components (and there are many more):
```
        dcc.Store(id="empty_input_store"),
        dcc.Store(id="empty_output_store"),
        dcc.Store(id="action_finished"),
        dcc.Store(id="set_remaining"),
        dcc.Store(id="remaining_actions"),
        html.Div(id="cycle_breaker_div", style={"display": "hidden"}),
        dcc.Store(id="cycle_breaker_empty_output_store"),
```

These two functions are wrapped into the one function called: `_build_action_loop`. 

---

Otherwise, `_build_actions_models` triggers `Action.build` method for each action model in the system.
Looking the architecture given below:
 - "Action_1",
 - "Action_2",
 - and "Action_3"

are callbacks created by Action.build method (each Action.build method creates one of the listed callbacks ("Action_1", "Action_2", "Action_3",)).

---

Following the pattern like the Action.build method creates callbacks and returns some components to be displayed, `_build_action_loop` and `_build_actions_models` do the same. 


image

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
maxschulz-COL: Hm, I think we should rename it to `ActionsLoopEngine` or something, because it is not a manager. It doesn't really manage anything as it is a one way communication.

vizro-core/src/vizro/_vizro.py
```diff
@@ -51,7 +48,6 @@ def build(self, dashboard: Dashboard):
         self._pre_build()
 
         self.dash.layout = dashboard.build()
-        _build_app_callbacks()
```
maxschulz-COL, line 54: When is this now done?

maxschulz-COL (commented):
lgtm, have just a few small questions

vizro-core/changelog.d/20230906_095422_petar_pejovic_action_build_method_inside_dashboard.md
petar-qb: I removed unnecessary one. Thanks!

vizro-core/src/vizro/_vizro.py
petar-qb: Inside the Dashboard.build method. Line: `children=[*ActionManager._create_app_callbacks(), dash.page_container],`.

`ActionManager._create_app_callbacks()` builds all action callbacks and returns all action components.

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
huong-li-nguyen: Got it, thanks! I went into the methods and looked at the output as well - correct me here if I am wrong:

The `_build_actions_models` return actual output components (e.g. dcc.Download) which are targeted by the action functions provided to the models e.g. Button.

The  `_build_action_loop` returns hidden components (such as dcc.Store/html.Div) required to store relevant data for the action loop engine to work.

This is what I wanted to know, because in the docstring description it's not clear what is meant with "returning action components". I would make a clear distinction in the docstrings and return type hints, such that it's easier to follow!

If the return of these are always components, shall we re-name to `_get_action_loop_components` and `_get_action_model_components`?

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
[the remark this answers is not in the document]
huong-li-nguyen: Got it, fine for me to keep in both, but I think getting the components is the relevant bit here right? Would eventually rename the method then to `_get_action_loop_components`

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
```diff
@@ -0,0 +1,40 @@
...
+
+class ActionManager:
+    @classmethod
+    def build(cls):
+        """Builds callbacks for the action loop and for each Action in the Dashboard and returns their components.
+
+        Returns:
+            All required components for the action loop and for each Action in the Dashboard.
+        """
+        return cls._build_action_loop() + cls._build_action_models()
```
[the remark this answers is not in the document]
huong-li-nguyen: So sorry, have to correct myself here! Given that the output are action components hidden in the dashboard container,  it might be more suitable to call this `_get_all_action_components`. What do you think? Leave as is if you don't think it's an improvement 😄

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
huong-li-nguyen: I prefer ActionManager as I think it also fits better with the front-end world and the syntax they use, even if it's just calling static functions. I think at some point we do actually want to extend the responsibility of the ActionManager e.g. for registering custom action functions and then population the relevant mapping files. 

Just adding some examples here and their definition of the action manager, which I still find quite fitting:

- https://doc.babylonjs.com/typedoc/classes/BABYLON.ActionManager
"Action Manager manages all events to be triggered on a given mesh or the global scene."

- https://doc.qt.io/qtcreator-extending/core-actionmanager.html#details
"The action manager is the central bookkeeper of actions and their shortcuts and layout. It is a singleton containing mostly static functions."

huong-li-nguyen (approved):
Thanks for refactoring! Looks great to me in general - would be great if we can provide return type hints and more detailed docstring descriptions inside the ActionManager (see PR comments). The rest is good-to-go for me!

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
maxschulz-COL: Hm, I am still unsure. Yes it is used in other code bases, but so are other class names if one googles them.

The reason why I am unsure is because we seem to be picking a name that stands for a pattern in our code base. I find it confusing, because you attach expectations of certain behaviour (global variable, stores something, violates state) with that name.

If we are to extend this, we can always change the name, so why not pick a name that is more suitable now, and adjust that when we add things later. After all it is private (unlike our other managers, which are public). But then again, because it is private, the name also doesn't matter hugely, so it is not the most important question I think

petar-qb: It doesn't matter very because it's private, but have to agree with @maxschulz-COL. If we all agree, I would rename it to the `ActionsLoopEngine` or something similar.

huong-li-nguyen: I really don't like the Engine as a description because it has nothing to do with an engine for me😄 

 If you guys want to rename to something different than ActionManager - then maybe (I asked chat gpt):
- ActionBuilder
- ActionFactory
- ActionSequenceFactory
- ActionLoopFactory
- ActionComponentCreator

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
petar-qb: You're right, `_build_action_loop` and `_build_actions_models` do what you said. But, they also create callbacks.  The `_build_action_loop` also creates callbacks for the `action loop` and the `_build_actions_models` also creates callbacks for `each Action model` in the Dashboard. 

So, why I didn't separate `ActionManager._create_app_callbacks` into something like:
 -  `ActionManager._create_app_callbacks`,
 -  `ActionManager._get_actions_components`
?  

The reason is that I want to follow our way of managing models. Every model has a `build` method which actually:
 - creates a callback and
 - returns needed components
(e.g. Slider, RangeSlider, Action and so on)

So, I wanted to follow the same pattern in the ActionManager too, by "exposing" only the `build` (or some other name like `_create_app_callbacks`) method that actually do the same: 
 - creates needed callbacks and
 - returns needed components.   

Hence, I wouldn't change methods names. 

--- 

I will update docstrings, thanks for an advice.

---

Also, I don't think that return type can bring any useful info. Any of our model.build methods don't have return type, so Action.build neither. It could seem confusing because it would be something like (List[Union[[dcc.Download, html.Div, dcc.Store]]]).    

Also, I don't think that return type can yield useful information. None of our model.build methods have a return type, and neither does Action.build. Also, it might seem confusing because it would be something like `List[Union[dcc.Download, html.Div, dcc.Store]]` (actually it should be `List[Union[any_dcc_object, any_html_object]]`).

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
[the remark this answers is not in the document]
petar-qb: Let's continue the conversation about naming [here](https://github.com/mckinsey/vizro/pull/11#discussion_r1317593612)

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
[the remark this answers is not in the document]
petar-qb: Let's continue the conversation [here](https://github.com/mckinsey/vizro/pull/11#discussion_r1317593612)

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
petar-qb: My preferences are:
 - ActionLoop (calling like: ActionLoop.build or ActionLoop.create_app_callbacks),
 - ActionLoopFactory (calling like: ActionLoopFactory.build, ActionLoopFactory.create_app_callbacks)

maxschulz-COL: Either of those two would be fine for me. If we think we align enough with this: https://refactoring.guru/design-patterns/factory-method , then lets go for the second, if not, lets do the first

huong-li-nguyen: Then let's go with ActionLoop

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
huong-li-nguyen: - On the first point, I agree. If it creates the callbacks and gets the components, then let's leave as is
- Perfect, thanks for updating the docstrings!
- On the third point, I kind of disagree, because it would have helped me at least to understand what the different components being returned are. One of the reasons why we don't have it in our model.build is because it's obvious from the code what is being returned. Remember in the build method we e.g. directly `return html.Div(...)`.

Here, I had to run the functions and take a look at the output to understand that these two methods return different "action components". Remember that someone not familiar with the action code should optimally understand what's happening in the action loop. So while it might be obvious for you right now, it might help others by providing more detailed docs and type hints :) 

Isn't the return of this:
_get_action_loop_components -> List[Union[html.Div, dcc.Store]]
_build_action_models -> List[dcc.Download] (at least right now, this will probably be extended further)

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
petar-qb: Sounds like a plan 👍.

vizro-core/src/vizro/actions/_action_manager/_action_manager.py
petar-qb: You're right, I see. Let me implement it 😄.

huong-li-nguyen: If the return type hints lead to lots of mypy issues, that are difficult to resolve, I would also be fine with adding the expected return types in the docstrings at least. So if option 1 is difficult to implement/maintain, that would also be an alternative for me :)

petar-qb: I tried with the return type and as you guessed, I had a lot of trouble with mypy (), so I described it in docstrings.

PR #13, opened 2023-09-06
First version of building actions inside the Dashboard model

petar-qb:
## Description
Implementation of the "Another possible solution" described in the description [here](https://github.com/mckinsey/vizro/pull/11#issue-1882109782).

## Checklist

## Types of changes

### DIFF
--- a/vizro-core/src/vizro/models/_dashboard.py
+++ b/vizro-core/src/vizro/models/_dashboard.py
@@ -10,6 +10,7 @@ from dash import Input, Output, callback, html
 from pydantic import Field, validator
 
 from vizro._constants import MODULE_PAGE_404, STATIC_URL_PREFIX
+from vizro.managers import model_manager
 from vizro.models import VizroBaseModel
 from vizro.models._models_utils import _log_call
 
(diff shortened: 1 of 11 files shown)


huong-li-nguyen (commented):
As I've said before, I prefer this approach over the other one, because I would prefer to have all hidden components in one place. But we can discuss in detail in our catch-up call :)

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -110,3 +111,13 @@ def callback_update_theme(on: bool):
...
+
+    @staticmethod
+    def _build_actions():
+        from vizro.models import Action
+
+        return [
+            action_component
+            for _, action in model_manager._items_with_type(Action)
+            for action_component in action.build()
+        ]
```
huong-li-nguyen, line 123: This would optimally live in the ActionManager for me. And the Dashboard would then just call `ActionManager._build_actions()` instead of this being a `staticmethod` of the Dashboard. As this is not really the responsibility of the Dashboard.

huong-li-nguyen (commented):
Does it make sense to take the time now and introduce an ActionManager?

huong-li-nguyen (approved):

PR #14, opened 2023-09-06
Add unit test for slider

chiara-sophie:
## Description
Test for slider including
- Instantiation
- Build method
- Callback method

## Checklist

## Types of changes


vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -0,0 +1,124 @@
...
+    ):
+        slider = vm.Slider(min=0, max=10, value=1)
+        result = slider.update_slider_value(f"{slider.id}{trigger}", start_value, slider_value, input_store_value)
+
+        assert result == (expected_value,) * 3
+
+    @pytest.mark.parametrize(
+        "trigger, start_value, slider_value, input_store_value, expected_value",
+        [
+            ("_text_value", 3, 1, 1, 0),  # set new value by start
```
chiara-sophie, line 103: Is it intended to have a reverse ordered slider (from 10 to 0) with value set to 1 and when it is updated to 3, that it is set to 0 instead of 3? Maybe a more general question if reverse slider should be an option

huong-li-nguyen: No, that shouldn't be possible - good catch! You should add a @validator to the RangeSlider and check that the max value is higher than the min value, otherwise return a ValidationError

vizro-core/src/vizro/models/_components/form/slider.py
```diff
@@ -30,7 +30,9 @@ class Slider(VizroBaseModel):
     min: Optional[float] = Field(None, description="Start value for slider.")
     max: Optional[float] = Field(None, description="End value for slider.")
     step: Optional[float] = Field(None, description="Step-size for marks on slider.")
-    marks: Optional[Dict[str, float]] = Field(None, description="Marks to be displayed on slider.")
+    marks: Optional[Dict[str, float]] = Field(
```
maxschulz-COL, line 33: Why would this change happen, is this not related to some line length setting?

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -0,0 +1,142 @@
...
+        )
+
+        assert slider.min == min
+        assert slider.max == max
+        assert slider.step == step
+        assert slider.marks == marks
+        assert slider.value == value
+        assert slider.title == str(title)
+
+    def test_invalid_reverse_slider(self):
```
nadijagraca, line 53: Maybe change this to `test_create_slider_invalid_min_max` as to be more verbose.

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -0,0 +1,142 @@
...
+        assert slider.marks == {}
+
+    def test_create_slider_with_action(self):
+        slider = vm.Slider(actions=[vm.Action(function=export_data())])
+        slider_tac = slider.actions[0]
+        assert hasattr(slider, "id")
+        assert slider.type == "slider"
+        assert len(slider.actions) == 1
+        assert isinstance(slider_tac, ActionsChain)
+        assert slider_tac.trigger.component_property == "value"
```
huong-li-nguyen, line 71: Tests usually also serve as a documentation of how you can configure the relevant models, so I wouldn't use configuration examples that don't make much sense e.g. would you trigger exporting data via slider handles? 😄 

I would create a fake function as a fixture that you can re-use:

```
@capture("action")
def test_function():
    pass
```

And then you can test the other bits as expected

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -0,0 +1,142 @@
...
+            min=None,
+            max=None,
+            step=None,
+            marks=None,
+            value=None,
+            included=False,
+            className="slider_control_no_space",
+            id="slider_id",
+            persistence=True,
+        )
```
huong-li-nguyen, line 89: Would move this into a fixture to be consistent with the rest of the tests

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -0,0 +1,142 @@
...
+
+        result = json.loads(
+            json.dumps(component["slider_id"], cls=plotly.utils.PlotlyJSONEncoder)
+        )
+        expected = json.loads(
+            json.dumps(expected_slider, cls=plotly.utils.PlotlyJSONEncoder)
+        )
+
+        assert result == expected
+        assert component["slider_title"].children == "Test title"
```
huong-li-nguyen, line 99: The json comparison tests for this as well, so you don't have to
```suggestion
```

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -0,0 +1,142 @@
...
+        slider_tac = slider.actions[0]
+        assert hasattr(slider, "id")
+        assert slider.type == "slider"
+        assert len(slider.actions) == 1
+        assert isinstance(slider_tac, ActionsChain)
+        assert slider_tac.trigger.component_property == "value"
+
+
+class TestBuildMethod:
+    def test_slider_build(self):
```
huong-li-nguyen, line 75: I think I would take a more sophisticated example in here or have two tests, one which you already have (the basic default) and another one where you test the build of a slider with min, max, marks, title?

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -0,0 +1,142 @@
...
+        slider = vm.Slider(id="slider_id", title="Test title")
+        component = slider.build()
+
+        expected_slider = dcc.Slider(
```
nadijagraca, line 79: Can you create a fixture at the beginning of the file to use in the tests? 

Fixture can return all the components that are returned by slider.build() so Title, Slider, Input. Later on you could compare the entire `component` with the fixture, without referencing individual parts of slider.

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -0,0 +1,142 @@
...
+    )
+    def test_update_slider_value_triggered(  # noqa
+        self, trigger, start_value, slider_value, input_store_value, expected_value
+    ):
+        slider = vm.Slider(min=0, max=10, value=1)
+        result = slider.update_slider_value(
+            f"{slider.id}{trigger}", start_value, slider_value, input_store_value
+        )
+
+        assert result == (expected_value,) * 3
```
huong-li-nguyen, line 123: Shall we actually edit the input in the params for the test and then just do an assert of result == expected here?
```suggestion
        assert result == expected
```

huong-li-nguyen (changes requested):
Hey - thanks for adding these! 🚀 

Have a few more ideas what we could tests. I think we can split up the tests more and isolate the relevant bits we want to test

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
huong-li-nguyen: Let me actually create a separate PR where I'll add this for the rest of the form components and then you can re-use the fixture.

nadijagraca (changes requested):

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
huong-li-nguyen: @lingyielia , I'll add this to the test_graph.py as well. It should replace the tests on the actions, you've previously removed :)

huong-li-nguyen: Here is the PR to follow: https://github.com/mckinsey/vizro/pull/20 :)

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -0,0 +1,220 @@
...
+            ),
+            ({15: 15, 25: 25}, {"15": 15.0, "25": 25.0}),
+        ],
+    )
+    def test_valid_marks(self, marks, expected):
+        slider = vm.Slider(min=0, max=10, marks=marks)
+
+        assert slider.marks == expected
+
+    def test_invalid_marks(self):
```
nadijagraca, line 127: What happens when you provide both steps and marks to slider?

chiara-sophie: That is covered in `test_valid_marks_default`, but maybe it is not 100% clear, so I renamed it to `test_valid_marks_with_step`

vizro-core/src/vizro/models/_components/form/_form_utils.py
```diff
@@ -55,3 +55,27 @@ def validate_value(cls, value, values):
...
+
+
+def validate_max(cls, value, values):
+    """Reusable validator for the "max" argument for sliders."""
+    if value is None:
+        return value
+
+    if values["min"] is not None and value < values["min"]:
+        raise ValueError("Maximum value of slider is required to be larger than minimum value.")
+    return value
```
huong-li-nguyen, line 67:
```suggestion
def validate_max(cls, max, values):
    """Reusable validator for the "max" argument for sliders."""
    if max is None:
        return max

    if values["min"] is not None and max < values["min"]:
        raise ValueError("Maximum value of slider is required to be larger than minimum value.")
    return max
```

vizro-core/src/vizro/models/_components/form/_form_utils.py
```diff
@@ -55,3 +55,27 @@ def validate_value(cls, value, values):
...
+def validate_slider_value(cls, value, values):
+    """Reusable validator for the "value" argument for sliders."""
+    if value is None:
+        return value
+
+    if values["min"] is not None and not value >= values["min"]:
+        raise ValueError("Please provide a valid value larger than the minimum value.")
+
+    if values["max"] is not None and not value <= values["max"]:
+        raise ValueError("Please provide a valid value smaller than the maximum value.")
```
huong-li-nguyen, line 79: Do you just want to combine these and have one error message "Please provide a valid value between the min and max value."

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -0,0 +1,201 @@
...
+            match="Please provide a valid value larger than the minimum value.",
+        ):
+            vm.Slider(min=1, max=10, value=0)
+
+    def test_invalid_value_high(self):
+        with pytest.raises(
+            ValidationError,
+            match="Please provide a valid value smaller than the maximum value.",
+        ):
+            vm.Slider(min=1, max=10, value=11)
```
huong-li-nguyen, line 90: Would combine the error message here as proposed before and then parameterise the input going beyond the range of min/max?

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -0,0 +1,201 @@
...
+        with pytest.raises(
+            ValidationError,
+            match="Maximum value of slider is required to be larger than minimum value.",
+        ):
+            vm.Slider(min=10, max=0)
+
+    @pytest.mark.parametrize(
+        "value",
+        [
+            5,
```
huong-li-nguyen, line 67: Add negative number

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -0,0 +1,201 @@
...
+    )
+    def test_valid_marks(self, marks, expected):
+        slider = vm.Slider(min=0, max=10, marks=marks)
+
+        assert slider.marks == expected
+
+    def test_invalid_marks(self):
+        with pytest.raises(
+            ValidationError,
+            match="2 validation errors for Slider",
```
huong-li-nguyen, line 130: What is the validation error being returned? Is it understandable enough or do we need to create a validator such that users can understand the required format of the marks dict better?

huong-li-nguyen (commented):
Looks a lot better! Have a few comments, but I think afterwards we are ready to merge :)

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
chiara-sophie: It returns a summary of the 2 validaiton errors (e.g. `marks -> 0  value is not a valid float (type=type_error.float)`) - so I guess this is quite clear.

huong-li-nguyen: Yes! That works 👍

huong-li-nguyen (approved):

nadijagraca (approved):
Look good! 🚀 
Thank you for incorporating the feedback.

PR #16, opened 2023-09-06
Add unit tests for `Checklist`

huong-li-nguyen:
## Description
- Convert `validate_options_dict` and `validate_value` to re-usable validators
- Indicate `_set_actions` validator as private in all selectors
- Provide defaults to options and value in checklist directly
- Add unit tests for checklist

## Checklist

- [x] I have not referenced individuals, products or companies in any commits, directly or indirectly
- [x] I have not added data or restricted code in any commits, directly or indirectly
- [x] I have updated the docstring of any public function/class/model changed
- [x] I have added the PR number to the change description in the changelog fragment, e.g. `Enable feature XXX (#123)` (if applicable)

## Types of changes

- [x] Docs/refactoring (non-breaking change which improves codebase)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #17, opened 2023-09-06
Add unit tests for `RadioItems`

huong-li-nguyen:
## Description
- Fix default values for `options `and `value`
- Add re-usable validators to `RadioItems`
- Add unit tests for `RadioItems`

Note: Will merge this PR into main after this one https://github.com/mckinsey/vizro/pull/16 has been merged

## Checklist

- [x] I have added the PR number to the change description in the changelog fragment, e.g. `Enable feature XXX (#123)` (if applicable)

## Types of changes


vizro-core/tests/unit/vizro/models/_components/form/test_radioitems.py
```diff
@@ -0,0 +1,144 @@
...
+        [
+            (5, ["A", "B", "C"]),
+            (5, [{"label": "A", "value": "A"}, {"label": "B", "value": "B"}]),
+            ("D", ["A", "B", "C"]),
+            ("D", [{"label": "A", "value": "A"}, {"label": "B", "value": "B"}]),
+            (True, ["A", "B", "C"]),
+            (True, [{"label": "A", "value": "A"}, {"label": "B", "value": "B"}]),
+        ],
+    )
+    def test_create_radio_items_invalid_value_non_existing(self, test_value, options):
```
petar-qb, line 127: Can we preserve parameter order in those two tests (or 
 - `"options, test_value",` or 
 - `"test_value, options",`
 ?

vizro-core/tests/unit/vizro/models/_components/form/test_radioitems.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @pytest.mark.parametrize("options", [1, "A", True, 1.0])
+    def test_create_radio_items_invalid_options_type(self, options):
+        with pytest.raises(ValidationError, match="value is not a valid list"):
+            RadioItems(options=options)
+
+    def test_create_radio_items_invalid_options_dict(self):
+        with pytest.raises(
+            ValidationError,
+            match="Invalid argument `options` passed into Selector. Expected a dict with keys `label` and `value`.",
```
petar-qb, line 92: Once, we support a RadioItem to be configured on the right side of the screen (and not only as control selector), should we then change the word "Selector" to "component" or "model" inside the ValidationError message? 

If yes, should we do it now then?

petar-qb (dismissed):
A good way to extract validator helper functions. It looks much cleaner now 🚀.
I also like the tests that follow the `test_dropdown.py` implementation which I have already approved so +1 approval from me for this PR. 👍

vizro-core/tests/unit/vizro/models/_components/form/test_radioitems.py
huong-li-nguyen: That's a good question, I don't think we would call it like that though as these terms are also too generic. Let me just remove the "into Selector" part and then it should always be valid.

petar-qb (approved):
🥇

nadijagraca (approved):
I like how test_options are consistent among the tests.  I will coordinate with Chiara on slider tests as well.  🚀

PR #18, opened 2023-09-07
Consolidate conftest.py approach

huong-li-nguyen:
## Description
- Consolidate conftest.py creation approach aligned with A.
- Re-placed fixtures and deleted their `conftest.py` if only used within single file

In general our approach is to keep fixtures close to where they are being used:
- Add the test fixtures directly into the relevant `test_XXX.py` file if only used there
- If shared across several test files, add the fixtures in a `conftest.py` placed in the sub-directory
- Move one level up if the fixtures are shared across sub-directories etc.

## Checklist

- [x] I have added the PR number to the change description in the changelog fragment, e.g. `Enable feature XXX (#123)` (if applicable)

## Types of changes


petar-qb (approved):
Now everything is in its place. I double checked the tests/unit/vizro/actions folder and it already follows the practices from the the PR description.

vizro-core/tests/unit/vizro/conftest.py
```diff
@@ -25,20 +23,6 @@ def standard_px_chart(gapminder):
     )
 
 
-@pytest.fixture
```
maxschulz-COL, line 28: My only concern is, will these very standard fixtures be "found"?

maxschulz-COL (approved):
LGTM! Just have the concern that we will duplicate many fixtures if we put even the most general fixtures into files if they are currently only used once. But probably this is still the cleaner approach, so good to go!

vizro-core/tests/unit/vizro/conftest.py
huong-li-nguyen: Currently, there are not that many fixtures and we managed to re-use many after reminding people during PRs 😄 

I think we just need to double-check every now and then and consolidate

PR #20, opened 2023-09-08
Consolidate tests for action_validator_factory

huong-li-nguyen:
## Description
- Consolidate tests for `action_validator_factory` after several duplications, see this thread: https://github.com/mckinsey/vizro/pull/14#discussion_r1319517169
- Rewrote already existing test for `Button` to follow same test

Essentially, we shouldn't configure the components with an action function, that doesn't make much sense to be used in that component. It's safer to use a test function such that we can still replicate the behaviour.

## Checklist

## Types of changes

### DIFF
--- a/vizro-core/tests/unit/vizro/models/_components/form/test_checklist.py
+++ b/vizro-core/tests/unit/vizro/models/_components/form/test_checklist.py
@@ -6,7 +6,10 @@ import pytest
 from dash import dcc, html
 from pydantic import ValidationError
 
+from vizro.models._action._action import Action
+from vizro.models._action._actions_chain import ActionsChain
 from vizro.models._components.form import Checklist
+from vizro.models.types import CapturedCallable
 
 
 @pytest.fixture()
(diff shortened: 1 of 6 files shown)


vizro-core/tests/unit/vizro/models/_components/form/test_checklist.py
```diff
@@ -132,6 +135,19 @@ def test_create_checklist_invalid_value_format(self):
         with pytest.raises(ValidationError, match="value is not a valid list"):
             Checklist(value="A", options=["A", "B", "C"])
 
+    def test_set_action_via_validator(self, test_action_function):
```
chiara-sophie, line 138: Maybe a more general question: does it make sense to repeat the same test for each form component as this is testing the validator and nothing component specific?

chiara-sophie (approved):
LGTM! Will add the same test for Slider 👍

nadijagraca (approved):
Looks good. 👍

vizro-core/tests/unit/vizro/models/_components/form/test_checklist.py
huong-li-nguyen: I thought about that as well, the part that is component-specific is the `component.property` which differs per component and that is actually important to check, otherwise, no action function would work. I would also be fine with just keeping that assert line and removing the rest.

However, in the tests for the actions, we don't cover e.g. the creation of the ActionsChain yet. It's part of the next sprint, so as long as these checks are not covered elsewhere, I prefer to have them duplicated.

We can always remove later, when we cover these properly in the actions test 👍

PR #27, opened 2023-09-11
Automation of releases

maxschulz-COL:
## Description

**Automation of releases**
- only requires `hatch run prep-release` to prepare a release
- then PR needs to be raised and approved
- on merge, we tag main, publish on PyPI, create a github release and create a PR that bumps the version
- the version bump still needs to be approved manually

**Potential future improvements**
- auto-merge version bump, or even better, use `hatch-vcs`

## Checklist

## Types of changes


tools/check_package_release.py
```diff
@@ -0,0 +1,58 @@
...
+                if line.strip() == "NEW_RELEASE=true":
+                    double_release = True
+
+    with open(env_file, "a") as f:
+        f.write(f"NEW_RELEASE={new_release}\n")
+        if new_release == "true":
+            f.write(f"PACKAGE_NAME={package_name}\nPACKAGE_VERSION={package_version}\n")
+
+    if double_release:
+        sys.exit("Cannot release two packages at the same time. Please modify your PR.")
```
petar-qb, line 58: Do we need to check `if double_release` **before** we overwrite the `env_file`:
```
if double_release:
    sys.exit("Cannot release two packages at the same time. Please modify your PR.")

with open(env_file, "a") as f:
    f.write(f"NEW_RELEASE={new_release}\n")
    ...
```

tools/check_package_release.py
```diff
@@ -0,0 +1,58 @@
...
+    if "dev" not in package_version:
+        return True
+    else:
+        print(f"Skipped: {package_name} {package_version} is still under development")  # noqa: T201
+        return False
+
+
+if __name__ == "__main__":
+    """Check if a package needs to be released"""
+    new_release = "false"
```
petar-qb, line 32: Why are we processing the `new_release` value with a string instead of a bool.

.github/workflows/check-release.yml
```diff
@@ -0,0 +1,117 @@
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install requests
+          pip install --upgrade hatch
+
+      - name: Check which 📦 needs to be released
+        run: |
+          packages=("vizro-core") #  "vizro-core"
+
+          for package in "${packages[@]}"; do
```
petar-qb, line 27: You've prepared this configuration so it can handle a process with multiple packages. Do we already have the potential candidate that's not `vizro-core` and did you test it with multiple packages?

.github/workflows/check-release.yml
```diff
@@ -0,0 +1,117 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install hatch
+      - name: Bump version to next patch
+        run: |
+          formatted_date=$(date +"%Y-%m-%d/%H-%M-%S")
+          cd "${{ needs.check-version.outputs.package_name }}"
+          git checkout -b "release/version_bump_next_minor/${formatted_date}"
+          hatch version patch,dev
```
petar-qb, line 107: And what if we want to release a `minor` version rather than a `patch` version?

.github/workflows/check-release.yml
```diff
@@ -0,0 +1,117 @@
...
+          hatch run docs:changelog
+          hatch run schema
+          hatch run lint || hatch run lint
+          git config user.email "83698606+maxschulz-COL@users.noreply.github.com"
```
petar-qb, line 111: Are we okay with having the "maxschulz-COL" in the email here?

petar-qb (commented):
Awesome work 🚀 🚀!

vizro-core/docs/pages/development/contributing.md
```diff
@@ -145,6 +145,23 @@ Vizro's version is given by `__version__` in [`src/vizro/__init__.py`](https://g
 
 To build the source distribution and wheel, run `hatch build`.
 
+### Steps to create a Vizro release
+
+The following steps are required to create a release of Vizro. Note that this is only done by maintainers:
+
+1. Checkout a new branch with a suitable name, e.g. `git checkout -b release/v0.2.4`
```
l0uden, line 152: Does it mean that `prep-release` will create this version? I've created branch `release/v1.0.0`, but it created `0.1.1`

l0uden (changes requested):

tools/check_package_release.py
maxschulz-COL: I think it doesn't matter, but it may be clearer the other way round

tools/check_package_release.py
maxschulz-COL: good point - it was taken over from a script in kedro, let me try to change it

.github/workflows/check-release.yml
maxschulz-COL: Yes, the work of AX will be the next package, and I tested it with a hypothetical next package, but I expect that there will still be slight adjustments necessary when it comes to that

.github/workflows/check-release.yml
maxschulz-COL: Then we have to change the version manually, but this is likely happening only when a breaking PR hits main, so it should be done with that PR instead of right after release of the previous release

.github/workflows/check-release.yml
maxschulz-COL: I followed the github privacy guide, this is essentially the maximum you can do to hide your email. It is not my real email

.github/workflows/check-release.yml
```diff
@@ -0,0 +1,117 @@
...
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  check-version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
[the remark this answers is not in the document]
maxschulz-COL: Do I also update the `setup-python@v3` to `v4` or just the above?

.github/workflows/check-release.yml
```diff
@@ -0,0 +1,117 @@
...
+          pip install hatch
+      - name: Bump version to next patch
+        run: |
+          formatted_date=$(date +"%Y-%m-%d/%H-%M-%S")
+          cd "${{ needs.check-version.outputs.package_name }}"
+          git checkout -b "release/version_bump_next_minor/${formatted_date}"
+          hatch version patch,dev
+          hatch run docs:changelog
+          hatch run schema
+          hatch run lint || hatch run lint
```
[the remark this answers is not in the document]
maxschulz-COL: But the lint workflow does not modify the code right?

vizro-core/docs/pages/development/contributing.md
maxschulz-COL: No, the name is arbitrary, let me adjust the docs then

.github/workflows/check-release.yml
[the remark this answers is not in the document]
l0uden: It will modify the code, but what I'm thinking of now is why we even need `linting` here? This workflow will run only after merging to `main`, in this case all this checks should be green already. And also here's the logic if first lint fail then run it again and then push the changes is smth we should be avoid, because this check is present in the different workflow.

I think it is better to forget about my fist comment. What I'm thinking of is that this linting step is unnecessary here or anywhere (I saw it in `hatch.toml`) except lint workflow.

.github/workflows/check-release.yml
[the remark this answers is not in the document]
l0uden: Yes, it also has `v4` in other workflows. I didn't notice)

.github/workflows/check-release.yml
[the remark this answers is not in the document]
maxschulz-COL: The reason why I thought it was necessary is because we are modifying files (`hatch version` and `hatch schema`) and we want the new PR that is raised, to be merable immediately.

I tried this out now, and it seems like that it doesn't need linting, but it did for me before. Let me take it out and see

vizro-core/changelog.d/20230915_010825_maximilian_schulz_prepare_automation.md
```diff
@@ -0,0 +1,42 @@
...
+-->
+
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+<!--
+### Added
```
lingyielia, line 14: Do we want to insert content here?

vizro-core/docs/pages/development/contributing.md
```diff
@@ -145,6 +145,21 @@ Vizro's version is given by `__version__` in [`src/vizro/__init__.py`](https://g
 
 To build the source distribution and wheel, run `hatch build`.
 
+### Steps to create a Vizro release
+
+The following steps are required to create a release of Vizro. Note that this is only done by maintainers:
+
+1. Checkout a new branch with a suitable name, e.g. `git checkout -b release/vx.y.z` (note that the branch name does not influence the versioning!)
```
lingyielia, line 152: So the branch name would be matched with what's in the `__init__.py` right? Like currently it's `"0.1.1.dev0"` so the branch name should be release/v0.1.1

Shall we include the package name into the branch name as well? something like `release/vizro-core/v0.1.1`

vizro-core/docs/pages/development/contributing.md
```diff
@@ -145,6 +145,21 @@ Vizro's version is given by `__version__` in [`src/vizro/__init__.py`](https://g
...
+2. Run `hatch run prep-release` from the root folder of the package you want to release
+3. Check that the resulting `CHANGELOG.md` and `version` are as expected
+4. Commit all changes, raise a PR, and get approval from at least 2 maintainers
+5. After the PR is merged, also merge the automatically created version bump PR
+6. Verify that the newly released package on PyPI works as intended
+
+!!!note
+
+    If there are any changes to `main` after the release PR has been created and `hatch run prep-release` has been run, please restart the process as this may mess up the `CHANGELOG`.
```
lingyielia, line 161: Maybe the merging queue could be useful due to this potential conflicts? https://github.blog/2023-07-12-github-merge-queue-is-generally-available/

Just an idea, definitely no need to pivot our current approach now

lingyielia (changes requested):

vizro-core/changelog.d/20230915_010825_maximilian_schulz_prepare_automation.md
maxschulz-COL: I thought about it, but I think it is not really relevant to the user right?

vizro-core/docs/pages/development/contributing.md
maxschulz-COL: Good shout, I'll take a closer look. Thanks for sharing

tools/check_package_release.py
```diff
@@ -0,0 +1,58 @@
...
+        return False
+
+
+if __name__ == "__main__":
+    """Check if a package needs to be released"""
+    new_release = False
+    double_release = False
+    package_name = sys.argv[1]
+    package_version = sys.argv[2]
+
```
Anna-Xiong, line 36: if we do a check here similar check like:
```
    if len(sys.argv) != ARG_NUM:
        raise Exception("Usage: python extract_release_notes.py <filename> <heading>")
```

would the message be printed out if someone used it run when edit CICD process? if so, would it be helpful to have exception with usage instruction here as well

lingyielia (approved):
LGTM! Thanks for sorting this out 💯

petar-qb (approved):
Thanks for addressing and answering all the comments. Looks good!

vizro-core/docs/pages/development/contributing.md
```diff
@@ -145,6 +145,21 @@ Vizro's version is given by `__version__` in [`src/vizro/__init__.py`](https://g
 
 To build the source distribution and wheel, run `hatch build`.
 
+### Steps to create a Vizro release
+
+The following steps are required to create a release of Vizro. Note that this is only done by maintainers:
```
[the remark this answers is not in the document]
maxschulz-COL: tbh you are right - I removed it and added it in the maintainer development docs

petar-qb (approved):
👍

l0uden (dismissed):

tools/check_package_release.py
```diff
@@ -0,0 +1,55 @@
...
+import os
+import subprocess
+import sys
+
+import requests
+
+AVAILABLE_PACKAGES = ["vizro-core"]
+VERSION_MATCHSTR = r'\s*__version__\s*=\s*"(\d+\.\d+\.\d+)"'
+RESPONSE_ERROR = 404
+ARG_NUM = 3
```
petar-qb, line 11: Is it used anywhere?

petar-qb (approved):
One comment left. Besides, it looks good!

.github/workflows/check-release.yml
```diff
@@ -0,0 +1,110 @@
...
+      - name: Set outputs
+        id: version_check
+        run: |
+          echo "new_release=${{ env.NEW_RELEASE }}" >> $GITHUB_OUTPUT
+          echo "package_name=${{ env.PACKAGE_NAME }}" >> $GITHUB_OUTPUT
+          echo "package_version=${{ env.PACKAGE_VERSION }}" >> $GITHUB_OUTPUT
+    outputs:
+      new_release: ${{ steps.version_check.outputs.new_release }}
+      package_name: ${{ steps.version_check.outputs.package_name }}
+      package_version: ${{ steps.version_check.outputs.package_version }}
```
lingyielia, line 36: Like how the final solution laid out 👍

Anna-Xiong (approved):
LGTM! thanks for addressing the comments!

petar-qb (approved):
⚡

PR #1, opened 2023-09-04
Add guide to place/style images in Card

huong-li-nguyen:
## Description
- Add section to image placement and styling back to docs
- Replace images and references

## Checklist

- [x] I have not referenced individuals, products or companies in any commits, directly or indirectly
- [x] I have not added data or restricted code in any commits, directly or indirectly
- [x] I have updated the docstring of any public function/class/model changed
- [x] I have added the PR number to the change description in the changelog fragment, e.g. `Enable feature XXX (#123)` (if applicable)

## Types of changes

- [x] Docs/refactoring (non-breaking change which improves codebase)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #30, opened 2023-09-12
Vizro core tests range slider

nadijagraca:
## Description

1. Added test for range_slider. Tested:
  - Instantiation
  - Build method
  - Callback method

2. Added additional tests for slider component (test_step_precedence_over_marks and test_invalid_step)
3. Added step validator
4. Changed type for slider and range_slider marks argument

## Checklist

## Types of changes


vizro-core/src/vizro/models/_components/form/range_slider.py
```diff
@@ -30,19 +31,22 @@ class RangeSlider(VizroBaseModel):
     min: Optional[float] = Field(None, description="Start value for slider.")
     max: Optional[float] = Field(None, description="End value for slider.")
     step: Optional[float] = Field(None, description="Step-size for marks on slider.")
-    marks: Optional[Dict[str, float]] = Field(None, description="Marks to be displayed on slider.")
+    marks: Optional[Dict[float, str]] = Field(None, description="Marks to be displayed on slider.")
```
chiara-sophie, line 34: Makes much more sense 👍

chiara-sophie (commented):
Looks good to me. Some minor suggestions to be consistent with other tests 👍 🚀

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -0,0 +1,331 @@
...
+    )
+    def test_valid_min_max(self, min, max):
+        range_slider = vm.RangeSlider(min=min, max=max)
+        min = int(min) if isinstance(min, str) else min
+        max = int(max) if isinstance(max, str) else max
+
+        assert range_slider.min == min
+        assert range_slider.max == max
+
+    def test_invalid_min_max(self):
```
maxschulz-COL, line 160: As an overall guidance, ideally also changed for the slider tests, could we name the tests:

`test_validate_max_xyz`, where `xyz` could indeed be `invalid` or so. This would make it easier to see which validator you are even trying to trigger

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -0,0 +1,331 @@
...
+        with pytest.raises(
+            ValidationError, match="Maximum value of slider is required to be larger than minimum value."
+        ):
+            vm.RangeSlider(min=10, max=0)
+
+    @pytest.mark.parametrize("value", [None, [1, 2], [0.1, 1.1], [-10, 10], [10, 10]])
+    def test_valid_value(self, value):
```
maxschulz-COL, line 167: Does it make sense to have also other valid options? For example would a list of strings be coerced?

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -0,0 +1,331 @@
...
+            ([1, 2, 3], "ensure this value has at most 2 items"),
+        ],
+    )
+    def test_invalid_value(self, value, match):
+        with pytest.raises(ValidationError, match=match):
+            vm.RangeSlider(min=0, max=10, value=value)
+
+    @pytest.mark.parametrize(
+        "step",
+        [1, 2.5, 10, None],
```
maxschulz-COL, line 190: again, are further options valid?

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -0,0 +1,331 @@
...
+    )
+    def test_valid_step(self, step):
+        range_slider = vm.RangeSlider(min=0, max=10, step=step)
+
+        assert range_slider.step == step
+
+    def test_invalid_step(self):
+        with pytest.raises(ValidationError, match="The step value of the slider must be less than the max value."):
+            vm.RangeSlider(min=0, max=10, step=11)
```
maxschulz-COL, line 199: In relation to the other comment: what is `min` was -4, this is valid no?

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -0,0 +1,331 @@
...
+        slider = vm.RangeSlider(min=0, max=10, marks=marks, step=step)
+
+        assert slider.marks == expected_marks
+        assert slider.step == step
+
+    @pytest.mark.parametrize(
+        "marks, expected",
+        [
+            ({i: str(i) for i in range(0, 10, 5)}, {i: str(i) for i in range(0, 10, 5)}),
+            ({15: 15, 25: 25}, {15.0: "15", 25.0: "25"}),
```
maxschulz-COL, line 219: What happens if I flip the string and the float?

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -0,0 +1,331 @@
...
+
+    @pytest.mark.parametrize(
+        "trigger_id, min, max, input_store, value, expected_value",
+        [
+            ("_start_value", 0, 10, [1, 9], [2, 3], (0, 10, [0, 10], (0, 10))),
+            ("_input_store", 0, 10, [1, 9], [2, 3], (1, 9, [1, 9], (1, 9))),
+            ("_end_value", 2, 8, [1, 9], [2, 3], (2, 8, [2, 8], (2, 8))),
+            ("", 2, 8, [1, 9], [2, 3], (2, 8, [2, 8], (2, 8))),
+            ("_start_value", 0, 0, [1, 9], [2, 3], (0, 0, [0, 0], (0, 0))),
+            ("_start_value", 0, 10, [1, 9], [-2, 12], (0, 10, [0, 10], (0, 10))),
```
maxschulz-COL, line 295: This is not really readable, I would be more verbose using a comment:

e.g change start value (left handle) will do xyz because

So one reading these understands the cases

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -0,0 +1,331 @@
...
+        )
+        assert result == expected_value
+
+    @pytest.mark.parametrize(
+        "trigger, start_value, end_value, slider_value, input_store_value",
+        [
+            ("_text_value", None, 9, [1, 9], [1, 9]),
+        ],
+    )
+    def test_update_slider_invalid(self, trigger, start_value, end_value, slider_value, input_store_value):
```
maxschulz-COL, line 317: Could these ever happen? Like in principle, would a user be able to trigger this? If not then we don't really need to test for it. It is more important to think of every possible situation that could happen, and check if it works

maxschulz-COL (changes requested):
Thanks a lot 🚀  More tested every day :)

vizro-core/src/vizro/models/_components/form/range_slider.py
```diff
@@ -30,19 +31,22 @@ class RangeSlider(VizroBaseModel):
...
     # Re-used validators
+    _validate_max = validator("max", allow_reuse=True)(validate_max)
+    _validate_value = validator("value", allow_reuse=True)(validate_slider_value)
+    _validate_step = validator("step", allow_reuse=True)(validate_step)
     _set_actions = _action_validator_factory("value")
 
     @validator("marks", always=True)
     def set_default_marks(cls, v, values):
-        return v if values["step"] is None else {}
+        return v if values.get("step") is None else {}
```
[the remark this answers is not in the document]
nadijagraca: Good point. I've added re-usable validator.

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
nadijagraca: Yes, I've made changes to the validator to account for that.

vizro-core/src/vizro/models/_components/form/range_slider.py
```diff
@@ -131,3 +128,20 @@ def update_slider_values(start, end, slider, input_store):
             className="selector_container",
             id=f"{self.id}_outer",
         )
+
+    def _update_slider_values(self, trigger_id, start, end, slider, input_store, value):
```
maxschulz-COL, line 132: Mea culpa, I think its better to rename start and end to start_txt and end_txt, correct right?

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -0,0 +1,346 @@
...
+        assert range_slider.value is None
+        assert range_slider.title is None
+        assert range_slider.actions == []
+
+    def test_create_range_slider_mandatory_and_optional(self):
+        range_slider = vm.RangeSlider(
+            min=0, max=10, step=1, marks={}, value=[1, 9], title="Test title", id="range_slider_id"
+        )
+
+        assert range_slider.min == 0
```
maxschulz-COL, line 139: Minor thing, but could we have similar order?

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -0,0 +1,346 @@
...
+        [
+            (
+                "_start_value",
+                2,
+                10,
+                [3, 10],
+                [1, 10],
+                None,
+                (2, 10, [2, 10], (2, 10)),
+            ),  # change start value by left handle
```
maxschulz-COL, line 311:
```suggestion
            (  # user changes left text field, slider and store update
                "_start_value",
                2,
                10,
                [3, 10],
                [1, 10],
                None,
                (2, 10, [2, 10], (2, 10)),
            ),
```
Could we try to do it in the above style, I still don't find it clear enough. Part of that has to do with maybe renaming start to start_txt or so (if that's correct).

In this example, are [3,10] and [1,10] correct? Why would they ever be different

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -0,0 +1,346 @@
...
+            ("_input_store", 0, 10, [3, 10], [1, 9], None, (1, 9, [1, 9], (1, 9))),  # set new value by input store
```
maxschulz-COL, line 312: Can we put them in the above format, then it's easier to read the whole test

maxschulz-COL: Maybe explain here that user refreshes page, so xyz..

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -0,0 +1,346 @@
...
+            ("_end_value", 0, 7, [3, 7], [1, 7], None, (0, 7, [0, 7], (0, 7))),  # change start value by right handle
```
maxschulz-COL, line 313: Is that correct? if i understood the first example correctly, this would be: user changes right text field no?

maxschulz-COL (commented):
Thanks a lot for adressing the comments! Have just some more minor suggestions. One ask, could you go through the PR comments and mark the things as resolved that you have adressed. Makes it easier to re-check 🙏

vizro-core/src/vizro/models/_components/form/_form_utils.py
```diff
@@ -72,9 +72,29 @@ def validate_slider_value(cls, value, values):
...
+    """Reusable validator for the "step" argument for sliders."""
+    if step is None:
+        return step
+    min_value = values["min"]
+    max_value = values["max"]
+
+    if max_value is not None and step > (max_value - min_value):
+        raise ValueError(
+            "The step value of the slider must be less than or equal to the difference between max and min."
+        )
```
chiara-sophie, line 95: I would make that consitent with the rest: 
```suggestion
    if values["max"] is not None and step > (values["max"] - values["min"]):
        raise ValueError(
            "The step value of the slider must be less than or equal to the difference between max and min."
        )
```

vizro-core/src/vizro/models/_components/form/_form_utils.py
```diff
@@ -72,9 +72,29 @@ def validate_slider_value(cls, value, values):
...
+    return step
+
+
+def set_default_marks(cls, v, values):
+    return v if values.get("step") is None else {}
```
chiara-sophie, line 100: Same here: 
```suggestion
    return v if values["step"] is None else {}
```

chiara-sophie (commented):
Looks good! And thanks for also including the changes to slider for consitency. 👍

maxschulz-COL (approved):
Awesome, good to go for me 🚀

chiara-sophie (approved):
Great! Looks good 🚀

PR #32, opened 2023-09-12
Disable creation of accordion navigation for single page

huong-li-nguyen:
## Description
- Disable creation of accordion navigation if only single page is provided
- Add test for disabling accordion navigation
- Remove redundant tests from test_navigation and test_accordion

## Screenshot

**One page only:**
Screenshot 2023-09-12 at 17 57 39

Note: Talked w/ J, and for now we've decided that in case the user does not provide a dashboard title, controls and one page only, the navigation bar will still be created. In the future, we could consider removing that automatically as well.

## Checklist

## Types of changes


vizro-core/tests/unit/vizro/models/_navigation/test_accordion.py
```diff
@@ -41,46 +41,29 @@ def test_field_invalid_pages_empty_list(self):
...
-        result_navigation = Accordion(pages=accordion_pages, id="accordion_list")
-        expected_navigation = accordion_from_page_as_list
-
-        # setting accordion id to fix the random id generation
-        expected_navigation.children.id = "accordion_list"
-
-        result = json.loads(json.dumps(result_navigation.build(), cls=plotly.utils.PlotlyJSONEncoder))
-        expected = json.loads(json.dumps(expected_navigation, cls=plotly.utils.PlotlyJSONEncoder))
-
-        assert result == expected
```
huong-li-nguyen, line 86: This is removed as this is covered by `test_accordion_build_default`

vizro-core/tests/unit/vizro/models/_navigation/test_navigation.py
```diff
@@ -49,52 +49,13 @@ def test_field_invalid_pages_input(self, pages, expected_error):
...
-        result_navigation = vm.Navigation(pages=pages_as_dict)
-        result_navigation.pre_build()
-
-        # setting accordion id to fix the random id generation
-        result_navigation._selector.id = "accordion_dict"
-
-        result = json.loads(json.dumps(result_navigation.build(), cls=plotly.utils.PlotlyJSONEncoder))
-        expected = json.loads(json.dumps(accordion_from_pages_as_dict, cls=plotly.utils.PlotlyJSONEncoder))
-
+    @pytest.mark.parametrize("pages", [["Page 1", "Page 2"], {"Page 1": ["Page 1"], "Page 2": ["Page 2"]}, None])
```
huong-li-nguyen, line 52: @nadijagraca I've reviewed the tests again and removed all of these here, as we already tested all of these in the `test_accordion.py`. So, these are essentially duplicates. 

The only test we need is that the `accordion.build()` returns the same as the `navigation.build()`, similar to how we have it for our tests in `test_filter.py`. The only test that would be worth adding here is if, at some point, we make the Accordion public. In this case, we should test for the use case of precedence, so if pages are provided to Navigation and Accordion, the configuration for Accordion will have precedence. 

However, given that this is private for now, this is not required yet.

vizro-core/src/vizro/models/_navigation/_accordion.py
```diff
@@ -65,6 +65,10 @@ def _create_default_accordion(self):
 
         accordion_items = [self._create_accordion_item(accordion_buttons=accordion_buttons)]
 
+        # Don't create accordion navigation if there is only one page and one accordion item
+        if len(accordion_buttons) == len(accordion_items) == 1:
+            return None
```
nadijagraca, line 70: Should we enable user to create one page navigation if they explicitly define navigation in the dashboard?

vizro-core/tests/unit/vizro/models/_navigation/test_accordion.py
```diff
@@ -41,46 +41,29 @@ def test_field_invalid_pages_empty_list(self):
...
-        component = accordion_div.build()
-
-        result = json.loads(json.dumps(component, cls=plotly.utils.PlotlyJSONEncoder))
+        accordion = Accordion(pages=pages_as_dict, id="accordion_dict").build()
+        result = json.loads(json.dumps(accordion, cls=plotly.utils.PlotlyJSONEncoder))
         expected = json.loads(json.dumps(accordion_from_pages_as_dict, cls=plotly.utils.PlotlyJSONEncoder))
-
         assert result == expected
 
+    def test_accordion_build_single_page_accordion(self, single_accordion_single_page):
```
nadijagraca, line 63: Are we using `single_accordion_single_page` fixture in this test?

nadijagraca (dismissed):
Really great addition (or removal :D). I just have two small questions, otherwise looks really good. 🚀

vizro-core/tests/unit/vizro/models/_navigation/test_accordion.py
huong-li-nguyen: Ah, you are right! Good catch, I added that before i just created the accordion in the test.

vizro-core/src/vizro/models/_navigation/_accordion.py
huong-li-nguyen: No, I don't think we should. It's a visual guideline enforced by us. The navigation doesn't have any functional purpose if there is only one page. It only adds visual noise which we should prevent from a design perspective. 

However, if for some reason many people request it, we can always add it in, but I would not enable it now

lingyielia (approved):
LGTM!

PR #34, opened 2023-09-13
Optimizing the client-server communication

petar-qb:
## Description
WIP

Progress:
- [x] Implement action loop `trigger_to_global_store` as clientside_callback.
- [x] Merge `gateway` , `set_remaining` and `executor` into one single callback.
- [ ] Rewrite the merged (new gateway) callback as clientside_callback.
- [x] Merge after_action and clientside_callback into one callback (maybe even clientside_callback).
- [x] Overwrite RangeSlider, Slider, ThemeSelector and other callbacks in the system to clientside_callback.
- [x] Try to merge two callbacks where input is theme_selector. (IMPOSIBLE because Patch can’t be in the clientside_callback) - actually it isn’t needed.
 

## Screenshot
Action loop architecture for the current system:
image

Action loop architecture for the optimised system (after we rewrite the gateway callback into the client-side one):
image

## To be discussed (not the part of the current PR):
- Maybe save `vizro.themes` on the frontend side as well? Then convert update_graph_theme to clienside callback.
- Try without dash.pages / or decorate (wrap) it.
- At the user's request, enable keeping original data on the frontend, instead of in the DataManager.
- Or maybe keep filtered data in the browser always!
- Put some kind of `@lru_cache`
 

## Checklist

## Types of changes

### DIFF
--- a/vizro-core/examples/default/app.py
+++ b/vizro-core/examples/default/app.py
@@ -454,6 +454,16 @@ def create_country_analysis():
                             targets=["bar_country"],
                         )
                     ),
+                    vm.Action(
+                        function=export_data(
+                            targets=["bar_country"],
+                        )
+                    ),
+                    vm.Action(
+                        function=export_data(
+                            targets=["bar_country"],
+                        )
+                    ),
                 ],
             ),
         ],
@@ -516,22 +526,20 @@ def create_home_page():
     return page_home
 
 
-create_country_analysis()
-
 dashboard = vm.Dashboard(
     pages=[
         create_home_page(),
         create_variable_analysis(),
         create_relation_analysis(),
         create_continent_summary(),
-        # create_country_analysis(),
+        create_country_analysis(),
     ],
-    navigation=vm.Navigation(
-        pages={
-            "Analysis": ["Homepage", "Variable Analysis", "Relationship Analysis", "Country Analysis"],
-            "Summary": ["Continent Summary"],
-        }
-    ),
+    # navigation=vm.Navigation(
+    #     pages={
+    #         "Analysis": ["Homepage", "Variable Analysis", "Relationship Analysis", "Country Analysis"],
+    #         "Summary": ["Continent Summary"],
+    #     }
+    # ),
 )
 
 if __name__ == "__main__":
--- a/vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
+++ b/vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
@@ -41,7 +41,7 @@ def _build_action_loop_callbacks() -> None:
         # Callback that enables gateway callback to work in the multiple page app
         clientside_callback(
             """
-            function(input, data) {
+            function trigger_to_global_store(input, data) {
                 return data;
             }
             """,
@@ -113,7 +113,7 @@ def _build_action_loop_callbacks() -> None:
     # Callback that triggers the next iteration
     clientside_callback(
         """
-        function(data) {
+        function after_action_cycle_breaker(data) {
             document.getElementById("cycle_breaker_div").click()
             return [];
         }
--- a/vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
+++ b/vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
@@ -61,6 +61,49 @@ def _build_action_loop_callbacks() -> None:
             )
         )
 
+    # TODO: need to be finished
+    # clientside_callback(
+    #     """
+    #     function gateway(remaining_actions, cycle_breaker_div, ...gateway_triggers) {
+    #         var ctx_triggered, triggered_actions_chains_ids, actions_chain_to_trigger, final_action_sequence, next_action, output_list, trigger_next;
+    #
+    #         ctx_triggered = dash_clientside.callback_context.triggered
+    #         triggered_actions_chains_ids = []
+    #
+    #         console.log("ctx_triggered")
+    #         console.log(ctx_triggered)
+    #
+    #         for (let i = 0; i < ctx_triggered.length; i++) {
+    #             triggered_actions_chains_ids.push(JSON.parse(ctx_triggered[i]['prop_id'].split('.')[0])['trigger_id']);
+    #         }
+    #
+    #         console.log("triggered_actions_chains_ids");
+    #         console.log(triggered_actions_chains_ids);
+    #
+    #         if (trigger_id === 'cycle_breaker_div') {
+    #             remaining_actions = remaining_actions.split(1);
+    #         }
+    #         else {
+    #
+    #         }
+    #
+    #         console.log("remaining_actions");
+    #         console.log(remaining_actions);
+    #         console.log("cycle_breaker_div");
+    #         console.log(cycle_breaker_div);
+    #         console.log("gateway_triggers");
+    #         console.log(gateway_triggers);
+    #         console.log('');
+    #     }
+    #     """,
+    #     # output=gateway_outputs,
+    #     output=[Output("remaining_actions", "data")] + [Output({"type": "action_trigger", "action_name": action.id}, "data") for action in actions],
+    #     # inputs=gateway_inputs,
+    #     inputs=[State("remaining_actions", "data"), Input("cycle_breaker_div", "n_clicks")] + gateway_inputs["gateway_triggers"],
+    #     prevent_initial_call=True,
+    # )
+
+    # TODO: Move it to the clientside callback
     @callback(
         output=gateway_outputs,
         inputs=gateway_inputs,
--- a/vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
+++ b/vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
@@ -23,10 +23,28 @@ def _build_action_loop_callbacks() -> None:
     actions_chains = _get_actions_chains_on_registered_pages()
     actions = _get_actions_on_registered_pages()
 
-    gateway_triggers: List[Input] = []
+    if not actions_chains:
+        return
+
+    gateway_inputs: Dict[str, Any] = {
+        "gateway_triggers": [],
+        "cycle_breaker_div": Input("cycle_breaker_div", "n_clicks"),
+        "remaining_actions": State("remaining_actions", "data"),
+
+    }
+    gateway_outputs: Dict[str, Any] = {
+        "action_triggers": [Output({"type": "action_trigger", "action_name": action.id}, "data") for action in actions],
+        "remaining_actions": Output("remaining_actions", "data"),
+    }
+
     for actions_chain in actions_chains:
-        # TODO: Potentially convert to clientside callback
-        @callback(
+        # Callback that enables gateway callback to work in the multiple page app
+        clientside_callback(
+            """
+            function(input, data) {
+                return data;
+            }
+            """,
             Output({"type": "gateway_input", "trigger_id": actions_chain.id}, "data"),
             Input(
                 component_id=actions_chain.trigger.component_id,
@@ -35,148 +53,72 @@ def _build_action_loop_callbacks() -> None:
             State({"type": "gateway_input", "trigger_id": actions_chain.id}, "data"),
             prevent_initial_call=True,
         )
-        def trigger_to_global_store(_, data):
-            return data
 
-        gateway_triggers.append(
+        gateway_inputs["gateway_triggers"].append(
             Input(
                 component_id={"type": "gateway_input", "trigger_id": actions_chain.id},
                 component_property="data",
             )
         )
 
-    # TODO: don't create any components or callback if there's no actions configured
-    if not gateway_triggers:
-        gateway_triggers.append(Input("empty_input_store", "data"))
-
-    action_triggers = [Output({"type": "action_trigger", "action_name": action.id}, "data") for action in actions]
-    if not action_triggers:
-        action_triggers.append(Output("empty_output_store", "data", allow_duplicate=True))
-
     @callback(
-        Output("set_remaining", "data"),
-        gateway_triggers,
+        output=gateway_outputs,
+        inputs=gateway_inputs,
         prevent_initial_call=True,
     )
-    def gateway(*gateway_triggers: List[dcc.Store]) -> List[Optional[str]]:
-        """Determines the final sequence of actions to be triggered.
+    def gateway(**inputs: Dict[str, Any]):
+        """GATEWAY."""
+        remaining_actions = inputs["remaining_actions"]
+        if ctx.triggered_id == "cycle_breaker_div":
+            remaining_actions = remaining_actions[1:]
+        else:
+            triggered_actions_chains_ids = [
+                json.loads(triggered["prop_id"].split(".")[0])["trigger_id"] for triggered in ctx.triggered
+            ]
 
-        Args:
-            gateway_triggers: Each 'gateway_trigger' (ctx.triggered_id) provides the 'id' (or trigger_id) from the
-                'actions_chain' that should be executed.
+            # Trigger only the on_page_load action if exists.
+            # Otherwise, a single regular (non on_page_load) action is triggered
+            actions_chain_to_trigger = next(
+                (
+                    actions_chain_id
+                    for actions_chain_id in triggered_actions_chains_ids
+                    if ON_PAGE_LOAD_ACTION_PREFIX in actions_chain_id
+                ),
+                triggered_actions_chains_ids[0],
+            )
+            logger.debug("=========== ACTION ===============")
+            logger.debug(f"Triggered component: {triggered_actions_chains_ids[0]}.")
+            final_action_sequence = [
+                {"Action ID": action.id, "Action name": action_functions[action.function._function]}
+                for action in model_manager[actions_chain_to_trigger].actions  # type: ignore[attr-defined]
+            ]
+            logger.debug(f"Actions to be executed as part of the triggered ActionsChain: {final_action_sequence}")
+            remaining_actions = [action_dict["Action ID"] for action_dict in final_action_sequence]
 
-        Returns:
-            List of final action sequence names which need to be triggered in order.
-
-        Raises:
-            PreventUpdate:
-                If screen with triggers is rendered but component isn't triggered.
-        """
-        triggered_actions_chains_ids = [
-            json.loads(triggered["prop_id"].split(".")[0])["trigger_id"] for triggered in ctx.triggered
-        ]
-
-        # Trigger only the on_page_load action if exists.
-        # Otherwise, a single regular (non on_page_load) action is triggered
-        actions_chain_to_trigger = next(
-            (
-                actions_chain_id
-                for actions_chain_id in triggered_actions_chains_ids
-                if ON_PAGE_LOAD_ACTION_PREFIX in actions_chain_id
-            ),
-            triggered_actions_chains_ids[0],
-        )
-        logger.debug("=========== ACTION ===============")
-        logger.debug(f"Triggered component: {triggered_actions_chains_ids[0]}.")
-        final_action_sequence = [
-            {"Action ID": action.id, "Action name": action_functions[action.function._function]}
-            for action in model_manager[actions_chain_to_trigger].actions  # type: ignore[attr-defined]
-        ]
-        logger.debug(f"Actions to be executed as part of the triggered ActionsChain: {final_action_sequence}")
-        return [action_dict["Action ID"] for action_dict in final_action_sequence]
-
-    @callback(
-        Output("remaining_actions", "data"),
-        Input("cycle_breaker_div", "n_clicks"),
-        Input("set_remaining", "data"),
-        State("remaining_actions", "data"),
-        prevent_initial_call=True,
-    )
-    def update_remaining_actions(
-        action_finished: Optional[Dict[str, Any]],
-        set_remaining: List[str],
-        remaining_actions: List[str],
-    ) -> List[str]:
-        """Updates remaining action sequence that should be performed.
-
-        Args:
-            action_finished:
-                Input that signalise action callback has finished
-            set_remaining:
-                Input that pass action sequence set in 'gateway' callback
-            remaining_actions:
-                State represents remaining actions sequence
-        Returns:
-            Initial or diminished list of remaining actions needed to be triggered.
-        """
-        # Propagate sequence of actions from gateway callback
-        triggered_id = ctx.triggered_id
-        if triggered_id == "set_remaining":
-            return set_remaining
-        # Pop first action
-        if triggered_id == "cycle_breaker_div":
-            return remaining_actions[1:]
-        return []
-
-    @callback(
-        *action_triggers,
-        Input("remaining_actions", "data"),
-        prevent_initial_call=True,
-    )
-    def executor(remaining_actions: List[str]) -> List[Any]:
-        """Triggers callback of first action of remaining_actions list.
-
-        Args:
-            remaining_actions:
-                Action sequence needed to be triggered.
-
-        Returns:
-            List of dash.no_update objects for all outputs except for next action.
-
-        Raises:
-            PreventUpdate:
-                If there is no more remaining_actions needs to be triggered.
-        """
         if not remaining_actions:
             raise PreventUpdate
 
         next_action = remaining_actions[0]
-        output_list = ctx.outputs_list if isinstance(ctx.outputs_list, list) else [ctx.outputs_list]
+        output_list = ctx.outputs_grouping["action_triggers"]
 
         # Return dash.no_update for all outputs except for the next action
         trigger_next = [no_update if output["id"]["action_name"] != next_action else None for output in output_list]
         logger.debug(f"Starting execution of Action: {next_action}")
-        return trigger_next
 
-    # Callback called after an action is finished
-    @callback(
-        Output("cycle_breaker_div", "children"),
-        Input("action_finished", "data"),
-        prevent_initial_call=True,
-    )
-    def after_action(*_) -> None:
-        """Triggers clientside callback responsible for starting a new iteration."""
-        logger.debug("Finished Action execution.")
+        return {
+            "action_triggers": trigger_next,
+            "remaining_actions": remaining_actions
+        }
 
     # Callback that triggers the next iteration
     clientside_callback(
         """
-        function(children) {
+        function(data) {
             document.getElementById("cycle_breaker_div").click()
-            return children;
+            return [];
         }
         """,
         Output("cycle_breaker_empty_output_store", "data"),
-        Input("cycle_breaker_div", "children"),
+        Input("action_finished", "data"),
         prevent_initial_call=True,
     )
--- a/vizro-core/src/vizro/models/_components/form/range_slider.py
+++ b/vizro-core/src/vizro/models/_components/form/range_slider.py
@@ -1,6 +1,6 @@
 from typing import Dict, List, Literal, Optional
 
-from dash import Input, Output, State, callback, callback_context, dcc, html
+from dash import Input, Output, State, clientside_callback, callback_context, dcc, html
 from pydantic import Field, validator
 
 from vizro.models import Action, VizroBaseModel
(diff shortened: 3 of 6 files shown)


vizro-core/examples/default/app.py
```diff
@@ -516,22 +526,20 @@ def create_home_page():
...
-            "Analysis": ["Homepage", "Variable Analysis", "Relationship Analysis", "Country Analysis"],
-            "Summary": ["Continent Summary"],
-        }
-    ),
+    # navigation=vm.Navigation(
+    #     pages={
+    #         "Analysis": ["Homepage", "Variable Analysis", "Relationship Analysis", "Country Analysis"],
+    #         "Summary": ["Continent Summary"],
+    #     }
+    # ),
```
huong-li-nguyen, line 542: I guess this is just for testing, but don't forget to clean it up when merging :)

vizro-core/src/vizro/models/_components/form/range_slider.py
```diff
@@ -54,32 +54,51 @@ def build(self):
...
-            else:
-                start_text_value, end_text_value = input_store if input_store is not None else value
-
-            start_value = min(start_text_value, end_text_value)
-            end_value = max(start_text_value, end_text_value)
-            start_value = max(self.min, start_value)
-            end_value = min(self.max, end_value)
-            slider_value = [start_value, end_value]
-            return start_value, end_value, slider_value, (start_value, end_value)
+        clientside_callback(
```
huong-li-nguyen, line 65: What's the advantage of moving these to a clientside_callback right now? Asking because currently in its current form it's very easy to unit test these callbacks as you can separate function from the calback. In the clientside_callback this seems more difficult to do now (See the latest added tests for sliders on this)

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -99,12 +99,15 @@ def build(self):
 
     @staticmethod
     def _update_theme():
-        @callback(
+        clientside_callback(
```
huong-li-nguyen, line 102: Same question on what the advantage of doing this is

huong-li-nguyen (commented):
Took a quick look - the example dashboard definitely feels quicker! 🚀 🐎  Great job, Petar!

I have a few open questions, some suggestion but would have to take a detailed look when I am back! If the PR is ready by this week, I'll leave the review to Max or Lingyi :)
Otherwise i'll take a look when I'm back

vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
```diff
@@ -23,10 +23,28 @@ def _build_action_loop_callbacks() -> None:
...
+    gateway_outputs: Dict[str, Any] = {
+        "action_triggers": [Output({"type": "action_trigger", "action_name": action.id}, "data") for action in actions],
+        "remaining_actions": Output("remaining_actions", "data"),
+    }
+
     for actions_chain in actions_chains:
-        # TODO: Potentially convert to clientside callback
-        @callback(
+        # Callback that enables gateway callback to work in the multiple page app
+        clientside_callback(
```
huong-li-nguyen, line 42: By the way, is there any good way of testing these clientside callbacks? It seems more difficult given that the function is wrapped inside and one cannot refactor it out. This is how we currently test our callbacks: https://stackoverflow.com/questions/67638178/how-to-write-testcases-for-plotly-dash-callback-function

Is there anything similar we could do?

vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
```diff
@@ -23,10 +23,28 @@ def _build_action_loop_callbacks() -> None:
...
+    gateway_inputs: Dict[str, Any] = {
+        "gateway_triggers": [],
+        "cycle_breaker_div": Input("cycle_breaker_div", "n_clicks"),
+        "remaining_actions": State("remaining_actions", "data"),
+
+    }
+    gateway_outputs: Dict[str, Any] = {
+        "action_triggers": [Output({"type": "action_trigger", "action_name": action.id}, "data") for action in actions],
+        "remaining_actions": Output("remaining_actions", "data"),
+    }
```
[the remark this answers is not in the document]
petar-qb: This will be removed as the 'gateway' callback will also be overwritten in the clientside_callback.

vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
```diff
@@ -35,148 +53,115 @@ def _build_action_loop_callbacks() -> None:
...
-            for action in model_manager[actions_chain_to_trigger].actions  # type: ignore[attr-defined]
-        ]
-        logger.debug(f"Actions to be executed as part of the triggered ActionsChain: {final_action_sequence}")
-        return [action_dict["Action ID"] for action_dict in final_action_sequence]
-
+    # TODO: need to be finished
+    # clientside_callback(
+    #     """
+    #     function gateway(remaining_actions, cycle_breaker_div, ...gateway_triggers) {
+    #         var ctx_triggered, triggered_actions_chains_ids, actions_chain_to_trigger, final_action_sequence, next_action, output_list, trigger_next;
```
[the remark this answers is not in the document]
petar-qb: I didn't succeed to finish overwriting the 'gateway' callback into the clientside one so I just left it commented. I'm going to rewrite it (as the TODO line above says) today and save one additional unnecessary request to the server side.

vizro-core/src/vizro/models/_components/form/range_slider.py
petar-qb: In the current implementation, when the user changes the `Input ID_start_value`, a request is sent to the server and the server then calculates the new values for the `slider` and `Input ID_end_value` components. It took about 350ms and the filtering action was delayed by 350ms accordingly.

Implemented as pure client-side javascript code, this process is calculated in the user's browser and takes a few milliseconds to calculate.

---

> In the clientside_callback this seems more difficult to do now (See the latest added tests for sliders on this)

Good point! Let me see how we can test it.

vizro-core/src/vizro/models/_dashboard.py
petar-qb: It seems strange to do a calculation like:
```
if on == True:
      return 'vizro_dark'
else:
      return 'vizro_light'
```
on the server side. This means that if users change the theme, they have to wait about 400 ms for a response. By moving the callback to the client side, this change happens instantly.

Ideally we should define callbacks on the server-side only if they have to communicate with data_manager, model_manager or some other service that's hosted on the server together with the vizro app.

vizro-core/src/vizro/models/_components/form/range_slider.py
```diff
@@ -54,32 +54,51 @@ def build(self):
...
+                start_value = Math.max(self_data["min"], start_value);
+                end_value = Math.min(self_data["max"], end_value);
+                slider_value = [start_value, end_value];
+              
+                return [start_value, end_value, slider_value, [start_value, end_value]];
+            }
+            """,
+            output=output,
+            inputs=inputs,
+        )
```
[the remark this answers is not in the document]
petar-qb: Thanks for heads up. I'll fix it.

vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
petar-qb: Let's continue this conversation in the thread [here](https://github.com/mckinsey/vizro/pull/34/files#r1324696236).

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -187,40 +196,3 @@ def test_slider_build(self, expected_slider):
...
-        ],
-    )
-    def test_update_slider_invalid(self, trigger, start_value, slider_value, input_store_value):
-        slider = vm.Slider(min=0, max=10, value=1)
-
-        with pytest.raises(
-            TypeError,
-            match="'>' not supported between instances of 'NoneType' and 'float'",
-        ):
-            slider._update_slider_value(f"{slider.id}{trigger}", start_value, slider_value, input_store_value)
```
petar-qb, line 226: Removed because a callback implemented in javascript cannot be tested with pytest. I'm currently trying to find a better way to test all those clientside_callbacks. 

If you have any proposal or conclusion, feel free to express it [here](https://github.com/mckinsey/vizro/pull/34/files#r1324696236). 😄

vizro-core/src/vizro/models/_components/form/range_slider.py
```diff
@@ -54,32 +54,55 @@ def build(self):
...
 
         return html.Div(
             [
+                dcc.Store(
+                    f"{self.id}_callback_data",
+                    storage_type="local",
```
lingyielia, line 99: what's the implication when `storage_type="local"`? 

I see

> storage_type (a value equal to: 'local', 'session' or 'memory'; default 'memory'): The type of the web storage. memory: only kept in memory, reset on page refresh. local: window.localStorage, data is kept after the browser quit. session: window.sessionStorage, data is cleared once the browser quit.

from https://dash.plotly.com/dash-core-components/store

Would it keep stale data in the Store?

lingyielia (changes requested):

lingyielia:
Another future use case relevant to this topic: when a dashboard needs to connect to the database and pull data from it, 

> Clientside callbacks are not possible if you need to refer to global variables on the server or a DB call is required.

https://dash.plotly.com/clientside-callbacks

How should we construct the action chain if there is a DB call involved

vizro-core/src/vizro/models/_components/form/range_slider.py
petar-qb: As it's explained in the link you sent, data will be preserved in the store if users refresh the currently opened page in the app or even if they close and reopen a browser. But, data in the `f"{self.id}_callback_data"` can't be changed (once it's returned from the server) so it doesn't make sense to keep in the `window.localStorage`. 


I removed the `storage_type` here and in the `Slider` as well. Thanks for the heads up 🚀.

petar-qb:
@lingyielia 
> How should we construct the action chain if there is a DB call involved

Only the internal mechanism (the mechanism that allows the action loop to work) is implemented using clientside-callbacks. If someone needs a DB connection, they can create a custom action or we can release a predefined action for that (either way it's an Action). Remember that Action callbacks are `blue rectangles` in the architecture sketch (look Description) which means that they are not client-side callbacks, but regular dash.callbacks that "live in the server".

vizro-core/changelog.d/20230918_111226_petar_pejovic_optimise_app_rendering_time.md
```diff
@@ -0,0 +1,43 @@
...
+- A bullet item for the Added category.
+
+-->
+<!--
+
+### Changed
+
+- Optimise the client-server communication ([#34](https://github.com/mckinsey/vizro/pull/34))
+
+-->
```
maxschulz-COL, line 25:
```suggestion

### Changed

- Optimise the client-server communication ([#34](https://github.com/mckinsey/vizro/pull/34))
    - eliminate most server side callback in favour of client-side callbacks
    - logging information now only displayed for action function carried out (no trigger or finished information)

```

maxschulz-COL: Do not forget the comment out, and also to check whether my indentations actually play out nicely

vizro-core/src/vizro/models/_components/form/range_slider.py
maxschulz-COL: Two suggestions/thoughts that you may have considered already, but let's discuss (on the basis of info here: https://dash.plotly.com/clientside-callbacks):

- is it an option to use the assets folder? I assume not, but let's discuss
- have we considered a compromise for client-side callbacks:

As for the second, it could be advantageous to not go all the way with optimisations if it improves readability and testability. The say
> When the overhead cost of a callback becomes too great and no other optimization is possible, the callback can be modified to be run directly in the browser instead of a making a request to Dash.
which kind of hints that this is the last resort. Potential options I see:
- grouping the callbacks might help already a lot
- moving only some callbacks to client-side, and leaving the most complex code server side (have my eye slightly on the `function gateway` here)

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -292,81 +308,3 @@ def test_range_slider_build_with_optional(self, expected_range_slider_with_optio
         expected = json.loads(json.dumps(expected_range_slider_with_optional, cls=plotly.utils.PlotlyJSONEncoder))
 
         assert result == expected
-
-
-class TestCallbackMethod:
```
maxschulz-COL, line 297: :(

maxschulz-COL (commented):
Thanks for the optimizations! Started a discussion on a few routes we can take to move forward

vizro-core/src/vizro/models/_components/form/range_slider.py
petar-qb: > is it an option to use the assets folder? I assume not, but let's discuss

Moving the javascript function into separate `static/js/` files looks much cleaner (functions become even more testable then).

> have we considered a compromise for client-side callbacks:
 
I considered it myself 😄. There are two reasons why the `gateway' callback should be implemented in javascript in my opinion:
   - Significant performance improvement. See `Speed-up' section in the PR description. If we implement the `gateway` callback as the server-side one, it would add about 400-800ms` to all the results in the `Optimised system` row. As a dashboard user, there is a big difference for me between 500ms and 1200ms waiting for filter results.
   - Another reason is that is much easier and more stable to implement in javascript cases such as (will be implemented next sprint): 
     - Handling handle when a new action chain is started before the previous one has finished.
     - Handling when number of actions in the queue (remaning_actions store) reached maximum limit (e.g. 100 actions).

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
petar-qb: I know :/

Still, if we move javascript code into separate `statis/js/` files, it seems feasible to write unit tests for these functions as well.

vizro-core/src/vizro/models/_components/form/range_slider.py
maxschulz-COL: Could we move this to a `js` file then? I think then we can far more easily implement unit tests

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
maxschulz-COL: We may want to go this route: https://jest-community.github.io/jest-pytest/docs/overview.html

vizro-core/src/vizro/models/_components/form/range_slider.py
petar-qb: Let's continue conversation about unit testing [here](https://github.com/mckinsey/vizro/pull/34#discussion_r1329704237).

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
petar-qb: I found several ways to test javascript functions. 
- Option 1 - `pytest`:
    - By using standard python lib called `subprocess` that triggers pure javascript test functions without any javascript test frameworks. 
    - Faster to implement and incorporate into existing project and CI process.
    - We don’t need to learn a new javascript framework
    - We can utilise all pytest options like test parametrisation.
    - But it’s a bad practice (hacky way to test it) especially for the open sourced project.
- Option 2 - `jest`:
    - By using jest javascript framework.
    - Not-to-easy to implement and incorporate into existing project and CI process (Node.js & npm have to be installed and configured).
    - We need to learn a new js framework.
    - But it’s a good practice especially for the open sourced project.
- Option 3 - pytest-jest:
    - The package has 40 stars in github and no documentation.

My preference is the second option because it's the only one that looks reliable and can meet the needs of the project (especially once the number of clientside callback functions grows over time). 
In that case I'd leave these changes for the separate PR, because the `node.js` & `npm` have to be installed, configured locally and for the CI process. However, I guess there is no big rush for it since these callbacks are already tested with UI tests (If I'm correct @l0uden)?

l0uden: `However, I guess there is no big rush for it since these callbacks are already tested with UI tests (If I'm correct @l0uden)?`

Sure, integration tests cover working of filter, parameters, interactions etc. on the UI level.

maxschulz-COL: Thanks a lot @petar-qb for the summary. I am still undecided...

Option 1 - the big advantage I see is that we do not leave our python first approach, and we could continue to cover everything with `hatch` 

Option 2 - I agree that this is the clean way to do it, but we would also loose a lot, by not relying on hatch alone to manage environments.

Option 3 - Agreed, let's not do that.

Can you maybe to a quick spike on option 2 that checks (no need to run complicated tests, just a dummy one):
- is it possible to create a command `hatch run test-js` that either runs those tests OR tells you exactly what you need to install how to get it run
- check to what degree this crosses with `hatch` in general

petar-qb: @maxschulz-COL here is a first version of javascript testing setup. I said first version because probably the output from `hatch run test-js` will be changed 😄. 

I have one UX improvement and will configure it if you agree with. If the all the Node.js/npm setup steps are successfully configured, does it make sense to run js tests on `hatch run test-js`. And if tests can't be run, that `hatch run test-js` shows guidelines on how to setup it? 

I have one user experience improvement and I will configure it if you agree. If all Node.js/npm setup steps are successfully configured, does it make sense to run js tests on `hatch run test-js`. And if the tests can't run, does that `hatch run test-js` show guidance on how to set it up?

maxschulz-COL: Yes, that is how I imagine it, if that's possible

maxschulz-COL: Ok I followed the guide, and I was easily able to run the tests I think. Now I think we should just make it so that `hatch run test-js` runs the tests if your setup is working.

If the hatch script becomes unwiedy, we may want to create a script for this!

vizro-core/hatch.toml
```diff
@@ -65,6 +65,7 @@ test-integration = [
   "pytest tests/integration --headless -k yaml {args}"
 ]
 test-unit = "pytest tests/unit {args}"
+test-js = "echo '' && echo 'To run tests for the javascript files you need to:' && echo '1. Install Node.js by downloading the LTS version at https://nodejs.org/en/download' && echo '2. Run command: \"npm install\" (to install npm dependencies)' && echo '3. Run command: \"npx jest\" (to run javascript tests)' && echo ''"
 update-snyk-requirements = "python ../tools/generate_snyk_requirements.py {args}"
```
maxschulz-COL, line 69: As discussed, this should cater for both cases, if setup is complete, then run tests, if not, show these instructions

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
petar-qb: I just started to add separate bash script for handling it inside the `vizro-core/tools`.

vizro-core/.babelrc
maxschulz-COL, line 1: Can we hide the 4 files in a folder, or do you think they should be at the root?

vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js
maxschulz-COL, line 1: Does it make sense to mimic the folder structure of `src` and add them precisely in the matching file?

vizro-core/.babelrc
petar-qb: Did it 😄.

vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js
petar-qb: Hmm, `css` folder doesn't mimic the same folder structure pattern of `src` and it's more like flat folder structure. I divided `js` scripts into `actions` and `module` and that's enough for now by my opinion. 

If you think there's a room for improvement I'd like to implement it.  
P.S. I think that the `tests/js/` folder structure has to follow the `src/vizro/static/js` folder structure. WDYT?

vizro-core/hatch.toml
```diff
@@ -55,7 +55,8 @@ schema = "python schemas/generate.py {args}"
 secrets = "pre-commit run gitleaks --all-files"
 test = [
   "test-unit",
-  "test-integration"
+  "test-integration",
+  "test-js"
```
petar-qb, line 59: Is it okay? @maxschulz-COL 
Do I also need to modify the `vizro-core/docs/pages/development/contributing.md` file accordingly?

maxschulz-COL: Yes please, would be great if you could update it for that `hatch` command, and ideally double check if that section is still up to date in general

vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js
maxschulz-COL: Ok let's go with the simplified form, I think that should suffice, and yes, the tests should then mimic that structure

vizro-core/changelog.d/20230918_111226_petar_pejovic_optimise_app_rendering_time.md
```diff
@@ -0,0 +1,45 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category.
+
+-->
+
+### Changed
+
+- Optimise the client-server communication ([#34](https://github.com/mckinsey/vizro/pull/34))
```
maxschulz-COL, line 22:
```suggestion
- Optimize the client-server communication ([#34](https://github.com/mckinsey/vizro/pull/34))
``

File: vizro-core/changelog.d/20230918_111226_petar_pejovic_optimise_app_rendering_time.md [hunk 15]
```diff
@@ -0,0 +1,45 @@
...
+  - Eliminate most server side callback in favour of client-side callbacks
```
maxschulz-COL, line 23:
```suggestion
  - Eliminate most server side callbacks in favour of client-side callbacks
```

vizro-core/changelog.d/20230918_111226_petar_pejovic_optimise_app_rendering_time.md
```diff
@@ -0,0 +1,45 @@
...
+  - Add tests for client-side callbacks written in Node.js framework called `jest`.
+  - Add hatch command `hatch run test-js` that runs unit tests written in the jest.
```
maxschulz-COL, line 25:
```suggestion
  - Add hatch command `hatch run test-js` that runs unit tests written in `jest`.
```

vizro-core/docs/pages/development/contributing.md
```diff
@@ -47,13 +47,19 @@ If no example is specified (`hatch run example`) then the [default example](http
 
 ## Testing
 
-Tests are handled using the [`pytest`](https://docs.pytest.org/) framework, and test environments are managed by Hatch. To run all tests, run
+Tests are handled using the [`pytest`](https://docs.pytest.org/) and [`jest`](https://jestjs.io/) frameworks, and test environments are managed by Hatch. To run all tests, run
```
maxschulz-COL, line 50:
```suggestion
Tests are handled using the [`pytest`](https://docs.pytest.org/) and [`jest`](https://jestjs.io/) frameworks, and test environments are managed by Hatch. To run all python tests, run
```

vizro-core/docs/pages/development/contributing.md
```diff
@@ -47,13 +47,19 @@ If no example is specified (`hatch run example`) then the [default example](http
...
 
 ```console
 hatch run test
 ```
 
-To run only unit tests, run `hatch run test-unit` and for integration tests only, run `hatch run test-integration`. Note that the latter executes all examples in separate testing sessions due to [this](https://github.com/mckinsey/vizro/issues/10) issue by providing the `-k` tag per example.
+To run only unit tests, run `hatch run test-unit`, and for integration tests only run `hatch run test-integration`.
```
maxschulz-COL, line 56:
```suggestion
To run only python unit tests, run `hatch run test-unit`, and for python integration tests only run `hatch run test-integration`.
```

vizro-core/docs/pages/development/contributing.md
```diff
@@ -47,13 +47,19 @@ If no example is specified (`hatch run example`) then the [default example](http
...
+
+To run only jest unit tests for javascript functions, run `hatch run test-js`.
+Note that Node.js is required to run tests written in jest framework. If you don't have Node.js installed, guidelines on how to install Node.js will appear when you run the command: `hatch run test-js`.
+Otherwise, if Node.js is installed, then the same command (`hatch run test-js`) runs jest unit tests.
```
maxschulz-COL, line 60:
```suggestion
To run only jest unit tests for javascript functions, run `hatch run test-js`.
Note that Node.js is required to run tests written in the jest framework. If you don't have `Node.js` installed, guidelines on how to install Node.js will appear when you run the command: `hatch run test-js`.
Otherwise, if `Node.js` is installed, then the same command (`hatch run test-js`) runs jest unit tests.
```

vizro-core/docs/pages/development/contributing.md
```diff
@@ -47,13 +47,19 @@ If no example is specified (`hatch run example`) then the [default example](http
...
+
+For unit and integration tests, note that the latter executes all examples in separate testing sessions due to [this](https://github.com/mckinsey/vizro/issues/10) issue by providing the `-k` tag per example.
```
maxschulz-COL, line 62:
```suggestion
For integration tests, all examples are executed in separate testing sessions due to [this](https://github.com/mckinsey/vizro/issues/10) issue by providing the `-k` tag per example.
```

vizro-core/hatch.toml
```diff
@@ -64,6 +65,7 @@ test-integration = [
   "pytest tests/integration --headless -k json {args}",
   "pytest tests/integration --headless -k yaml {args}"
 ]
+test-js = "./tools/run_jest.sh"
```
maxschulz-COL, line 68: Should we enable passing arguments like for `test-unit`?

vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
```diff
@@ -1,31 +1,30 @@
...
 
 def _build_action_loop_callbacks() -> None:
     """Creates all required dash callbacks for the action loop."""
-    # TODO - Reduce the number of the callbacks in the action loop mechanism
     actions_chains = _get_actions_chains_on_registered_pages()
     actions = _get_actions_on_registered_pages()
 
-    gateway_triggers: List[Input] = []
+    if not actions_chains:
+        return
```
maxschulz-COL, line 21: (for later) Why has that appeared?

vizro-core/src/vizro/models/_components/form/range_slider.py
```diff
@@ -60,23 +60,30 @@ def build(self):
...
         return html.Div(
             [
+                dcc.Store(
+                    f"{self.id}_callback_data",
+                    data={
+                        "id": self.id,
+                        "min": self.min,
+                        "max": self.max,
+                    },
+                ),
```
maxschulz-COL, line 86: (for later) Why is this now needed?

vizro-core/src/vizro/static/js/clientside_functions.js
maxschulz-COL, line 1: (for later) Please elaborate a little on this

vizro-core/package.json
maxschulz-COL, line 1: (for alter) Please elaborate a little on this

maxschulz-COL (commented):
LGTM!

Have a few very small suggestions that we should do before merging. Anything marked as (for later) please ignore until we have merged.

.gitignore
```diff
@@ -1,4 +1,8 @@
 
+# npm node_modules
+package-lock.json
```
pruthvip15, line 3: Hi [nit picking|optional], I was just glancing over this.
 its a practice to commit the lock file also. all the projects do commit their lock file also. wdyt ?

vizro-core/docs/pages/development/contributing.md
petar-qb: Isn't that `hatch run test` runs all tests (python and js)?

I implemented like that in the hatch.toml file:
```
test = [
  "test-unit",
  "test-integration",
  "test-js"
]
```

.gitignore
petar-qb: Good spot, you're right. 👍

vizro-core/docs/pages/development/contributing.md
petar-qb: Since I supported passing arguments to the jest through the command `hatch run test-js -all`, I removed calling `test-js` with the command `hatch run test` because `jest` has different arguments and flags than `pytest`.

Does that make sense for you?

maxschulz-COL: You are right... Hm, ok let's keep it then

vizro-core/src/vizro/_vizro.py
```diff
@@ -88,7 +88,7 @@ def _append_styles(walk_dir: str, url_prefix: str) -> Tuple[List[str], List[str]
             path = os.path.join("/" + url_prefix, base, f) if base else os.path.join("/" + url_prefix, f)
             extension = os.path.splitext(f)[1]
             if extension == ".js":
-                _vizro_js.append(path)
+                _vizro_js.append({"src": path, "type": "module"})
```
maxschulz-COL, line 91: @huong-li-nguyen Can you double check if this is in line with what we previously intended. Was not so involved in this

huong-li-nguyen: What does this do? As far as I remember we initially implemented this as the assets couldn't be read in properly on some of the hosted versions. For me it's fine as this works as before and the js files live in the static folder. 

@Anna-Xiong - can you double-check Petars changes to the initial function you wrote?

maxschulz-COL: @Anna-Xiong Or could you also have a look?

petar-qb: This change was made to enable import/export features in `js` scripts. Without setting those scripts as modules (with `"type": "module"`) I had to push all the `js` files into the one big file (which doesn't look pretty).

vizro-core/hatch.toml
```diff
@@ -64,6 +64,7 @@ test-integration = [
   "pytest tests/integration --headless -k json {args}",
   "pytest tests/integration --headless -k yaml {args}"
 ]
+test-js = "./tools/run_jest.sh {args}"
```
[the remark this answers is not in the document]
petar-qb: We should keep `test-js` script separate from the `test` script since `test-unit` and `test-integration` lie on the `pytest`. The reason is that the users can pass arguments like `hatch run test -vv` and `test-js`, that actually runs command `npx jest`, doesn't support `-vv` flag.

lingyielia (approved):
LGTM! See great speed improvement

maxschulz-COL (approved):
Lgtm now!

vizro-core/src/vizro/_vizro.py
Anna-Xiong: I am not so sure about js here, have you tested adding js @petar-qb ? 
The thing is that in source code i did not see adding this type and module, but maybe the js has been created into another object

huong-li-nguyen: @maxschulz-COL @petar-qb - did you guys follow up on this one before the merge? Otherwise please double-check

petar-qb: This function returns only the `css` and `js` files from the `src/vizro/static/` folder. There won't be any problems because `src/vizro/static` contains set of css files and only javascript files added by me. A dashboard creator cannot customise `src/vizro/static` folder because additionally added `css/js` files (added by dashboard creator) are added through the `assets` folder instead (`src/vizro/assets/`).

vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
petar-qb: This prevents the creation of action loop callbacks if there are no actions defined in the system. It only covers the edge case if users create a dashboard without charts, filters, parameters...

vizro-core/src/vizro/models/_components/form/range_slider.py
petar-qb: When the RangeSlider callback was implemented as a server-side callback, it could use global variables from the class instance such as: self.id, self.min, self.max...

Since the current solution is implemented as a client-side javascript callback, this means that the callback cannot access global or class instance variables. This means that we had to save this information on the client side (hide it in dcc.Store).

vizro-core/src/vizro/static/js/clientside_functions.js
petar-qb: This file creates the mapper of javascript functions where:
- key is the name of the function and 
- value is callable function written in javascript.  
(once the app is started, the mapper is created and it lives on the client-side).

Those functions are accessible from any point of code on the client-side. The mapper serves javascript functions for all `dash.clientside_callback` in the app. A function can be accessed using the `dash.ClientsideFunction` abstract. 

```
clientside_callback(
    ClientsideFunction(
        namespace="clientside", 
        function_name="update_range_slider_values"
    ),
    output=output,
    inputs=inputs,
)
```

vizro-core/package.json
petar-qb: The `package.json` file is the heart of any Node project. It records important metadata about a project which is required to install dependencies and run scripts.

Our dependencies are:
 - `jest` - javascript framework that enables us to write unit tests for javascript functions,
 - `babel` - javascript transcompiler that allows us to take advantage of the newest features of the language (such as import/export syntax feature).

PR #63, opened 2023-09-26
Introduce coverage settings and failure threshold

maxschulz-COL:
## Description
- move coverage settings to package `pyproject.toml` (they weren't doing anything before)
- slight adjustments and introduction of minimum coverage
- set minimum coverage to **82%**

**Note**
- Took out tests from coverage, as I personally find it more misleading than helpful, it dilutes the true coverage (83% without tests vs 97% with tests)
However, it was interesting to see that some tests seem to not be executed (when running coverage on test files as well), this needs further investigation (created a ticket)

## Checklist

## Types of changes

## Notice

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,22 +12,6 @@ exclude_dirs = ["tests"]
 line-length = 120
 target-version = ["py38"]
 
-[tool.coverage.paths]
-tests = ["tests"]
-vizro = ["src/vizro"]
-
-[tool.coverage.report]
-exclude_lines = [
-  "no cov",
-  "if __name__ == .__main__.:",
-  "if TYPE_CHECKING:"
-]
-
-[tool.coverage.run]
-branch = true
-parallel = true
-source_pkgs = ["vizro", "tests"]
-
 [tool.mypy]
 # strict checks : strict = true
 check_untyped_defs = true
--- a/vizro-core/changelog.d/20230926_113052_maximilian_schulz_introduce_coverage.md
+++ b/vizro-core/changelog.d/20230926_113052_maximilian_schulz_introduce_coverage.md
@@ -1,0 +1,42 @@
+<!--
+A new scriv changelog fragment.
+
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -12,22 +12,6 @@ exclude_dirs = ["tests"]
 line-length = 120
 target-version = ["py38"]
 
-[tool.coverage.paths]
-tests = ["tests"]
-vizro = ["src/vizro"]
```
huong-li-nguyen, line 17: Is there a reason why we move this to the pyproject.toml of vizro-core instead of adapting the path here? So e.g. `vizro-core= ["vizro-core/src/vizro"]`. I thought the advantage of having it here is that we then configure the test coverage requirements for all sub-packages here

maxschulz-COL: But the test coverage requirements are not not global, but rather per package. Some of the settings could potentially act globally, but some definitely cannot. E.g. the `fail_under` should be done per package.

huong-li-nguyen: Ah, so we cannot enforce a min. test coverage of 80% for all our sub-packages in the same pyproject.toml?

huong-li-nguyen (dismissed):

pyproject.toml
maxschulz-COL: I think in principle this could maybe be possible, but I think it would be better not too.

Say `vizro-core` is really mature, and we have a minimum coverage of 95%, but then there is a new package with 60% (which is fine to start with maybe in that case), what do we do.

The current idea is to always increase the minimal coverage when we test more, and just ensure it doesn't get lower after that

maxschulz-COL:
> How do you mean that "Taking out tests from coverage" dilutes the true coverage?

First there was a small typo in my code, which is now fixed, but now you can see the difference if you take out the `path` and also the `source_pkgs` specification. Then it runs the coverage on tests as well, and you get 97%

petar-qb (approved):
Looks good!

huong-li-nguyen (approved):

PR #59, opened 2023-09-25
Improve duplicate ID error message to guide people using Jupyter notebooks

maxschulz-COL:
## Description
- add additional warnings in case of duplicate IDs in `model_manager` in case this occurs in Jupyter notebooks
- adresses some issues described in https://github.com/mckinsey/vizro/pull/56 (check here also for a pathway to future improvements)

## Checklist

- [x] I have not referenced individuals, products or companies in any commits, directly or indirectly
- [x] I have not added data or restricted code in any commits, directly or indirectly
- [x] I have updated the docstring of any public function/class/model changed
- [x] I have added the PR number to the change description in the changelog fragment, e.g. `Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))` (if applicable)
- [x] I have added tests to cover my changes (if applicable)

## Types of changes

- [x] Docs/refactoring (non-breaking change which improves codebase)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

  - I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
  - I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorised to submit this contribution on behalf of the original creator(s) or their licensees.
  - I certify that the use of this contribution as authorised by the Apache 2.0 license does not violate the intellectual property rights of anyone else.

PR #74, opened 2023-09-28
Enable highlighting of active page button and refactor `accordion.build()`

nadijagraca:
## Description
- **Visual fixes:**
  - Enable highlighting of active accordion button by providing page_id
  - Change the collapsible default behavior of the accordion to always be open
  - Align accordion title and page title
  - Remove highlighting of active accordion group title
- **Optimise for-loop for the creation of the AccordionButtons to loop through the provided pages** and then check if it exists in the dash page registry, required:
   - Move the creation of dash.page_registry to `Dashboard.pre_build`
   - Update `_validate_pages` validator to take registered pages from ModelManager as `dash.page_registry` is not available at point of validation
- **Remove Accordion().build() from Page.build()**, required:
   - Add validator for `dashboard.navigation` to default to `Navigation()`
   - Update all unit tests respectively

## Screenshot
Screenshot 2023-09-28 at 14 36 30

## Checklist

## Types of changes

## Notice


vizro-core/changelog.d/20230928_143738_nadija_ratkusic_graca_highlight_active_accordion_item.md
```diff
@@ -0,0 +1,41 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+
+### Fixed
+
+- Fixed highlighting for active accordion item and arranged a setup to maintain the expanded state of the parent accordion item when page is active. ([#74](https://github.com/mckinsey/vizro/pull/74))
```
huong-li-nguyen, line 34:
```suggestion
- Add highlighting of accordion of selected page and change the collapsible default behavior of accordion. ([#74](https://github.com/mckinsey/vizro/pull/74))
```

vizro-core/src/vizro/models/_navigation/_accordion.py
```diff
@@ -26,17 +26,17 @@ class Accordion(VizroBaseModel):
     _validate_pages = validator("pages", allow_reuse=True, always=True)(_validate_pages)
 
     @_log_call
-    def build(self):
+    def build(self, page_id):
```
huong-li-nguyen, line 29: @antonymilne - we usually don't have any arguments for the build method. I don't know if we should deviate from that. Here, it is required to know what the currently active page is such that a different CSS class is provided in the accordion.

Is this better stored in a PrivateAttr or a dcc.store, as knowing the currently selected page seems helpful in several places?

vizro-core/src/vizro/models/_navigation/_accordion.py
```diff
@@ -77,23 +77,24 @@ def _create_default_accordion(self):
                     class_name="accordion",
                     persistence=True,
                     persistence_type="session",
+                    always_open=True,
```
huong-li-nguyen, line 80: I definitely prefer to have this setting, but it deviates a bit from the behaviour we've discussed with J. Let's just quickly align with him in our catch-up

huong-li-nguyen (commented):
Looks great to me in general! Thanks for fixing all the CSS 🚀 

Would align with @antonymilne regarding the build argument or whether we should actually store these as a PrivateAttr or dcc.store and then I think we're good!

vizro-core/src/vizro/models/_navigation/_accordion.py
```diff
@@ -26,17 +26,17 @@ class Accordion(VizroBaseModel):
...
+        return self._create_default_accordion(page_id=page_id)
 
-    def _create_accordion_buttons(self, accordion_pages):
+    def _create_accordion_buttons(self, accordion_pages, page_id):
         return [
             dbc.Button(
                 children=[page["name"]],
                 key=page["relative_path"],
-                className="accordion-item-button",
+                className="accordion-item-button-active" if page_id == page["name"] else "accordion-item-button",
```
huong-li-nguyen, line 39: Instead of switching between different CSS classes, we can just use the existing class and leverage its state. For this you need to use the active parameter.

```suggestion
                className="accordion-item-button",
                active=True if page_id == page["name"] else False,
```

I'll comment on the accordion.css file how you would have to change the CSS properties there accordingly

vizro-core/src/vizro/static/css/accordion.css
```diff
@@ -113,6 +113,11 @@ div.page_container .accordion-item-button {
   line-height: var(--text-size-03);
 }
 
+.accordion-item-button-active {
+  background-color: var(--state-overlays-selected);
+  color: var(--text-primary);
+}
+
```
huong-li-nguyen, line 120:
```suggestion
.accordion-item-button.btn.btn-primary.active {
  background-color: var(--state-overlays-selected);
  color: var(--text-primary);
}
```

vizro-core/src/vizro/static/css/accordion.css
```diff
@@ -102,7 +101,8 @@
   gap: var(--spacing-01);
 }
 
-div.page_container .accordion-item-button {
+div.page_container .accordion-item-button,
+.accordion-item-button-active {
```
huong-li-nguyen, line 105:
```suggestion
div.page_container .accordion-item-button {
```

petar-qb (approved):
I have nothing to add. Looks good! 👍

huong-li-nguyen:
> I have nothing to add. Looks good! 👍

Perfect, thanks! @antonymilne had a few improvement suggestions I have implemented, so please take another look. :) 

The relevant ones are:
- Optimise for-loop for the creation of the AccordionButtons to loop through the provided pages and then check if it exists in the dash page registry (previously implemented the other way around), required:
   - Move the creation of dash.page_registry to Dashboard.pre_build
   - Update `_validate_pages` validator to take registered pages from ModelManager as dash.page_registry is not available at point of validation
- **Remove Accordion().build() from Page.build()**, required:
   - Add validator for `dashboard.navigation` to default to `Navigation()`

vizro-core/tests/unit/vizro/models/test_page.py
```diff
@@ -101,6 +101,8 @@ def test_action_auto_generation_valid(self, standard_px_chart):
 # TODO: Add unit tests for private methods in page build
 def test_page_build_left_side_removed(standard_px_chart):
     page = vm.Page(title="Single Page", components=[vm.Graph(id="scatter_chart", figure=standard_px_chart)])
-    vm.Dashboard(pages=[page]).build()
+    dashboard = vm.Dashboard(pages=[page])
+    dashboard.pre_build()
+    dashboard.navigation.pre_build()
```
huong-li-nguyen, line 106: @antonymilne - I've implemented the changes we've discussed and I think they are an improvement to what we've had before (Navigation being dependent on `Dashboard.build()`). 

However, this set-up still bugs me 😄 It's related to the fact that we build the navigation in the Page, and the navigation now depends on the dashboard.pre_build(). So this page.build() test currently requires a pre-build of the dashboard and the navigation (otherwise it cannot find the private _selector attribute).

I think for now that's fine, as these steps will always be executed before the `page.build()`. However, I am sure we can solve this by moving the `navigation.pre-build()/build()` to the dashboard. Just wanted to emphasize above as it plays into our discussion enabling arguments inside the `build()` methods.

petar-qb (approved):
I like to see that dash.page_registry moved to the Dashboard._pre_build method. It looks much cleaner now.

vizro-core/tests/unit/vizro/models/test_page.py
antonymilne: Agree with all this. There's actually one potentially problem which we should check: does `dashboard.pre_build` need to be run before `navigation.pre_build` in order to work? Or if you swap those round is it still ok?

Would things be neater if we moved the registration of pages into a `Dashboard` validator so that it's executed before any pre-build? We should think carefully before doing that though because it would be very easy to have a contaminated Dash page registry unless we empty it out first (same problem as we have with the model manager).

antonymilne (commented):
Part of me actually still prefers to use a clientside callback for this ([like Dash pages themselves do to update the title](https://github.com/plotly/dash/blob/d9b8d17e3fae5c7f9f342e480a85e782f6289fe8/dash/dash.py#L2143-L2151)), but I trust @petar-qb's instinct here saying that we should avoid that. I don't much like having an argument in `build`, but I always thought we might have to do that somewhere in our codebase eventually. So let's go for it 🙂

Possibly arguments in other `build` methods will help untangle other pieces like the placement of the code that currently lives in `Page`. But in general let's still try to avoid adding arguments in `build` unless it really helps a lot, like it does here.

Generally this looks great and I'm happy to approve once a couple of things have been checked! 🙏

vizro-core/tests/unit/vizro/models/_navigation/test_accordion.py
```diff
@@ -37,31 +37,31 @@ def test_field_invalid_pages_empty_list(self):
...
 
-@pytest.mark.usefixtures("dashboard_build")
+@pytest.mark.usefixtures("dashboard_prebuild")
 class TestAccordionBuild:
     """Tests accordion build method."""
 
     @pytest.mark.parametrize("pages", [["Page 1", "Page 2"], None])
     def test_accordion_build_default(self, pages, accordion_from_page_as_list):
-        accordion = Accordion(pages=pages, id="accordion_list").build()
+        accordion = Accordion(pages=pages, id="accordion_list").build(active_page_id="Page 1")
```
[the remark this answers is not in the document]
huong-li-nguyen: In the first three unit tests, we need the `active_page_id ` as we check in the json comparison for equality, and there, the active parameter is turned on for the Accordion Button of Page 1. I've removed the parameter from all other unit tests, though, where it's unnecessary 👍 

These tests also already suffice for me to check if the highlighting works. We only need to know if the active parameter in the AccordionButton is True. The rest is done via pure CSS. I don't think we need to test if the CSS attributes work, or do we? Other than that, our front-end tests might be able to capture it via screenshots!

This is a good read for the PlotlyJSONEncoder - we've decided to implement it following that approach: https://community.plotly.com/t/unit-test-dash-html-components-and-dash-core-components/9104

vizro-core/src/vizro/models/_navigation/_accordion.py
```diff
@@ -26,24 +26,24 @@ class Accordion(VizroBaseModel):
...
-            dbc.Button(
-                children=[page["name"]],
-                key=page["relative_path"],
-                className="accordion-item-button",
-                href=page["relative_path"],
+    def _create_accordion_buttons(self, pages, active_page_id):
+        """Creates a button for each provided page that is registered."""
+        accordion_buttons = []
+        for page_id in pages:
+            page = dash.page_registry[page_id]
```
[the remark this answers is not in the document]
huong-li-nguyen: Yes! So if you try to use a page in the `Navigation` that is registered in the ModelManager but not in the dash.page_registry(), you will get a validation error before the dashboard build:

Screenshot 2023-10-04 at 16 19 29

```
dashboard = vm.Dashboard(
    pages=[
        #create_home_page(),
        create_variable_analysis(),
        create_relation_analysis(),
        create_continent_summary(),
        create_country_analysis(),
    ],
    navigation=vm.Navigation(
        pages={
        }
    ),
)
```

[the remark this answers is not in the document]
huong-li-nguyen: Ahhhh, just double-checked - the validations don't capture it. They only capture validations errors if you use pages in the Navigation that are completely unknown or missing.

Let me add a validation check in there

vizro-core/tests/unit/vizro/models/test_page.py
huong-li-nguyen: The order here doesn't matter, the test will still pass if we switch the order. I think it's because we are re-using the same validator in the Navigation and Accordion, so either way the pages will get auto-populated if no Navigation is provided.

Given the order above doesn't matter - shall we leave as is or shall I try to move it to a validator? We actually had a problem of contaminated dash page registry leading to test interdependence which was fixed in this PR: https://github.com/mckinsey/vizro/pull/84

However, I had to remove each page individually as initiating with an empty OrderedDict again left the dash.page_registry permanently empty.

huong-li-nguyen: **EDIT:** Seems like we can't move the registration of pages into a Dashboard validator, you'll get the error `"dash.exceptions.PageError: `dash.register_page()` must be called after app instantiation"`. So I guess it can only live in the pre-build or build?

Screenshot 2023-10-04 at 18 19 54

vizro-core/tests/unit/vizro/models/_navigation/test_accordion.py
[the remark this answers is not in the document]
antonymilne: Ah ok, I didn't realise that the tests were already actually testing that the right page was being set to `active=True`. That is definitely sufficient then, no need to test the CSS attributes at all.

Thanks for the link on the testing strategy, I'll check it out!

vizro-core/tests/unit/vizro/models/test_page.py
antonymilne: Ah yes, that is a problem. Possibly we should actually treat the `dash.register_page` as something quite special that lives in `Vizro.build` itself:
```
    def build(self, dashboard: Dashboard):
        for page in dashboard.pages:
             dash.register_page(...)
        self._pre_build()
        ....
```

Let's not do it now, given that it seems to work ok in `pre_build` and the order doesn't seem to matter.

This does feel a bit fragile though, and I can imagine it inadvertently introducing bugs in future - we always need to stay careful to make sure that `pre_build` methods can run in any order with the same result, and that is easy to forget. There seem to be a few problems with page registration and model manager anyway (e.g. #59), so let's bear in mind the possibility of moving this bit to `Vizro.build` when thinking about those since it might improve things overall.

I have some comments on #84 but let's save that for another conversation 🙂

vizro-core/src/vizro/models/_navigation/_accordion.py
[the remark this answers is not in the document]
antonymilne: Hmm this whole navigation/page registration/model manager thing is clearly quite tricky and I'll need to spend some more time looking through the validator and rest of the code properly to understand fully.

This looks like a good solution for now, but I wonder if moving the `dash.register_page` even earlier in the process into `Vizro.build` itself will be a better solution ultimately as in my other comment 🤔 This way we wouldn't need to use the model manager to access pages at all (?), which might be a lot nicer.

vizro-core/src/vizro/models/_navigation/_accordion.py
```diff
@@ -26,20 +26,23 @@ class Accordion(VizroBaseModel):
...
 
     def _create_accordion_buttons(self, pages, active_page_id):
         """Creates a button for each provided page that is registered."""
         accordion_buttons = []
         for page_id in pages:
-            page = dash.page_registry[page_id]
+            try:
+                page = dash.page_registry[page_id]
+            except KeyError:
+                raise KeyError(f"Page with ID {page_id} cannot be found. Please add the page to `Dashboard.pages`")
```
antonymilne, line 39:
```suggestion
            except KeyError as exc:
                raise KeyError(f"Page with ID {page_id} cannot be found. Please add the page to `Dashboard.pages`") from exc
```

Very minor, but it's generally nice to chain these sort of exceptions.

antonymilne (approved):
This is great, thank you for all your iteration on it! We should discuss #84 briefly and then also consider whether to do further refactoring to move the `dash.register_page`, but I'll need to spend a while longer getting up to speed on the existing code for that. For now I think this solution should be fine so very happy to have it merged 👍 

Outside all the discussion on where to put page registration and whether `build` methods should take arguments I'm super happy to have the improvement to highlight the active page because this was actually my wife's first comment when I showed her the new demo also 😀

vizro-core/src/vizro/models/_navigation/_accordion.py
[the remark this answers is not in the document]
huong-li-nguyen: I would definitely prefer that! I don't think the `_validate_pages` validator should use the model manager at all if we currently register all pages (whether in the dashboard or not). But I understand why Nadija wrote it that way, as I tried different approaches as mentioned above, but only the ModelManager seems to have the information at that point of time 😅 

Evaluating whether we can move the `dash.register_page` to `Vizro.build` definitely makes sense to me!

vizro-core/tests/unit/vizro/models/test_page.py
huong-li-nguyen: I have a few questions on that one, but I'll add it to the list of minor topics to catch-up on! Will leave this conversation open, so we can go back :)

PR #80, opened 2023-09-29
Add unit tests for `Action` model

petar-qb:
## Description

- [x] Tests for Action model are implemented. 
- [x] The code coverage is increased (from 83% -> 87%).  
- [x] Action.build method is fragmented into few separate private methods which makes the Action model more testable.  
- [x] Two bugs fixed:
  - [x] BUG: Custom action callbacks cannot return a single or list/tuple of values (Only dictionary is supported). Now this restriction is removed and it's possible to return single/list/tuple of values and output is properly handled. 
  - [x] BUG: _get_action_callback_mapping returns empty dictionary by default when argument is component. Default value for component argument should be an empty list. 
- [x] Raising the custom exception if action callback return value length doesn't match the number of outputs. 
- [x] Some return value type hints added.
- [x] Some linting issues fixed. 
- [x] Packing Action model and Action loop components into html.Div.  

## Checklist

## Types of changes

## Notice


vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -59,6 +58,57 @@ def build(self):
...
+
+        # Map returned values to dictionary format where None belongs to the "action_finished" output
+        return {"action_finished": None, **dict(zip(outputs, return_value))}
+
+    @_log_call
+    def build(self):
+        """Builds a callback for the Action model and returns required components for the callback.
+
+        Returns:
+            List of required components for the Action model added in the `Dashboard` level e.g. List[dcc.Download].
```
huong-li-nguyen, line 103:
```suggestion
            List of required components (e.g. dcc.Download) for the Action model added to the `Dashboard` container.
```

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -59,6 +58,57 @@ def build(self):
...
+        """
+        # callback_inputs, and callback_outputs are "dash.State" and "dash.Output" objects made of three parts:
+        # 1. User configured inputs/outputs - for custom actions,
+        # 2. Vizro configured inputs/outputs - for predefined actions,
+        # 3. Hardcoded inputs/outputs - for custom and predefined actions (enable callbacks to live inside Action loop).
+        # action_components represents return value of the build method.
```
huong-li-nguyen, line 109: Would add this as a docstring to the private `_get_callback_mapping()` instead

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -59,6 +58,57 @@ def build(self):
...
+            )
+
+        if isinstance(return_value, dict):
+            return {"action_finished": None, **return_value}
+
+        # If return_value is a single element, ensure return_value is a list
+        if not isinstance(return_value, (list, tuple)):
+            return_value = [return_value]
+
+        # Map returned values to dictionary format where None belongs to the "action_finished" output
```
huong-li-nguyen, line 95:
```suggestion
```

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -59,6 +58,57 @@ def build(self):
...
+                f"Number of action's returned elements: {return_value_len} does not match the number"
+                f" of action's defined outputs: {len(outputs)}."
+            )
+
+        if isinstance(return_value, dict):
+            return {"action_finished": None, **return_value}
+
+        # If return_value is a single element, ensure return_value is a list
+        if not isinstance(return_value, (list, tuple)):
+            return_value = [return_value]
```
huong-li-nguyen, line 93: Shall we change the order here? Otherwise this might easily break as this currently depends on a specific code sequence to be executed.

When would the return_value be a tuple? I thought this would only return a list or dict?

```suggestion
        # If return_value is a single element, ensure return_value is a list
        if not isinstance(return_value, (list, tuple, dict)):
            return_value = [return_value]
        if isinstance(return_value, dict):
            return {"action_finished": None, **return_value}
        return {"action_finished": None, **dict(zip(outputs, return_value))}
```

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -0,0 +1,205 @@
...
+            outputs=outputs,
+        )
+        callback_inputs, callback_outputs, action_components = action._get_callback_mapping()
+        assert callback_inputs == {
+            **{
+                f'{input.split(".")[0]}_{input.split(".")[1]}': dash.State(input.split(".")[0], input.split(".")[1])
+                for input in inputs
+            },
+            "trigger": dash.Input({"action_name": action.id, "type": "action_trigger"}, "data"),
+        }
```
huong-li-nguyen, line 148: Shall we move the expected result into fixtures? Would be more consistent with the rest of our unit tests. Same comment for code sections below

huong-li-nguyen (commented):
Could only take a quick look. Looks good to me in general 👍  Only have a few comments!

vizro-core/changelog.d/20230927_081254_petar_pejovic_actions_unit_tests.md
```diff
@@ -0,0 +1,42 @@
...
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category.
+
```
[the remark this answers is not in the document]
maxschulz-COL: Then I think we should not mention it, it would just be confusing

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -0,0 +1,205 @@
...
+            "compo-nent.property",
+            "component.property_1",
+        ],
+    )
+    def test_outputs_invalid(self, outputs, test_action_function):
+        with pytest.raises(Exception):
+            Action(function=test_action_function, inputs=[], outputs=outputs)
+
+
+class TestActionBuild:
```
maxschulz-COL, line 106: Are we only testing this case, as teh creation of the callbacks is hard to test without lower level private functions?

vizro-core/tests/unit/vizro/models/_action/test_action.py
maxschulz-COL: Agreed, can we also remove some of the computation from the output? I would prefer to have a clear hard-coded dict as expected outcome:

1. This keeps tests pure, as in they don't mix function and result
2. it makes them far more readable as function documentation

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -0,0 +1,205 @@
...
+                    output.split(".")[0], output.split(".")[1], allow_duplicate=True
+                )
+                for output in outputs
+            },
+            "action_finished": dash.Output("action_finished", "data"),
+        }
+        assert action_components == []
+
+    @pytest.mark.parametrize(
+        "custom_action_function, callback_context_outputs_grouping, expected_function_return_value",
```
maxschulz-COL, line 161: I found this unintuitive to read. Source of my confusion (and maybe we can make this better if you agree that the points are valid):

1. `custom_action_function` does not look like a function until realizing later how it handle's `requests`. Maybe a renaming might be useful, smth like `custom_action_function_mock_return`
2. Similar for the `callback_context_outputs_grouping`, solution potentially similar to above - for filter tests we chose e.g. `callback_context_filter_continent` so that the parametrization is a lot more intuitive --> can we achieve a name that suggests what this context is parametrized for?

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -0,0 +1,205 @@
...
+            ({"component_1_property": "new_value"}, ["component_1_property", "component_2_property"]),
+            ({"component_1_property": "new_value", "component_2_property": "new_value_2"}, ["component_1_property"]),
+        ],
+        indirect=True,
+    )
+    def test_action_callback_function_return_value_invalid(
+        self, custom_action_function, callback_context_outputs_grouping
+    ):
+        action = Action(function=custom_action_function())
+        with pytest.raises(ValueError):
```
maxschulz-COL, line 204: Can we be more specific about the correct error message, and potentially annotate the above test cases as to what they represent. I would prefer not to have a lot of cases that are ultimately never imaginable

maxschulz-COL (changes requested):
Thanks a ton, we are on our path to 💯 . Some notes, let me know if you disagree

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -59,6 +58,57 @@ def build(self):
...
+        return_value_len = (
+            1 if not hasattr(return_value, "__len__") or isinstance(return_value, str) else len(return_value)
+        )
+        outputs = list(ctx.outputs_grouping.keys())
+        outputs.remove("action_finished")
+        if len(outputs) != return_value_len:
+            raise ValueError(
+                f"Number of action's returned elements: {return_value_len} does not match the number"
+                f" of action's defined outputs: {len(outputs)}."
+            )
```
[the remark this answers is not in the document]
petar-qb: With raising this custom error I just wanted to hide "action_finished" internal output from this error message.

Here's an example:
I created the custom action with 1 defined output (export_button_children) and its function that returns 0 elements:

```
@capture("action")
def custom_action(export_button_children):
    return

vm.Action(
    function=custom_action(),
    inputs=["export_button.children"],
    outputs=["export_button.children"],
)               
```

Without custom error handling I got the following error from Dash: 

image

With custom error handling code I got the following error from Vizro:
image

vizro-core/src/vizro/models/_action/_action.py
petar-qb: > When would the return_value be a tuple? I thought this would only return a list or dict?

Here is an example of the function that returns a tuple, we also want to support:
```
def fun():
    return 1, 2
```

> Maybe we could even push this further, to make it check for an iterable?

I tried, but string is also iterable object and we need to treat strings as single objects and convert them into a list.

vizro-core/src/vizro/actions/_callback_mapping/_get_action_callback_mapping.py
```diff
@@ -42,5 +44,8 @@ def _get_action_callback_mapping(action_id: ModelID, argument: str) -> Union[Dic
         },
         _on_page_load.__wrapped__: {"inputs": _get_action_callback_inputs, "outputs": _get_action_callback_outputs},
     }
-    action_call = action_callback_mapping.get(action_function, {}).get(argument, {})
-    return action_call if isinstance(action_call, dict) else action_call(action_id=action_id)
+    action_call = action_callback_mapping.get(action_function, {}).get(argument)
+    default_value: Union[List[dcc.Download], Dict[str, Union[Input, State, Output]]] = (
+        [] if argument == "components" else {}
+    )
```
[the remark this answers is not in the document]
petar-qb: It seems like leaky abstraction even more to me that the function returns `{}` if the argument is "components". If some function requires "components", than it expects a list as a response. (empty or not-empty list, but still a list object). 

Let's imagine that there are many places from where the `_get_action_callback_mapping` is called with argument="components". It would mean that this case should be handled on every call. e.g.:

```
components = _get_action_callback_mapping(
    action_id="ACTION_ID",
    argument="components"
) or []  # 'or []' should be places on every _get_action_callback_mapping call.
```

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -59,6 +58,57 @@ def build(self):
...
+
+        if isinstance(return_value, dict):
+            return {"action_finished": None, **return_value}
+
+        # If return_value is a single element, ensure return_value is a list
+        if not isinstance(return_value, (list, tuple)):
+            return_value = [return_value]
+
+        # Map returned values to dictionary format where None belongs to the "action_finished" output
+        return {"action_finished": None, **dict(zip(outputs, return_value))}
```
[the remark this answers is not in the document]
petar-qb: For upcoming custom actions we want to support few different ways of defining return values (as Dash already does):
```
    @capture("action")
    def custom_action(input: int):x
        return input+1, input+2   # return value as a tuple 
        return [input+1, input+2]   # return value as a list
        return {"output_2_children": input+2, "output_1_children": input_1}.   # return value as a dict
``` 

So we want to support return value to be:
- single element,
- tuple of elements,
- list of elements,
- dictionary.

Regarding that we add "action_finished" internal `dash.Output`, we always pack the users custom action function return value to the dictionary format so we don't mess the order of outputs.

vizro-core/src/vizro/models/_action/_action.py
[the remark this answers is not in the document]
maxschulz-COL: Just for my understanding, the above custom action **function** you configured has no return, but in the action model expects one output. Hence it raises the custom error?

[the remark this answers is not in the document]
petar-qb: Yes, exactly.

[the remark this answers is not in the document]
maxschulz-COL: I think its a good idea then, but happy to follow Lis idea of moving it into a utils function

vizro-core/changelog.d/20230927_081254_petar_pejovic_actions_unit_tests.md
```diff
@@ -0,0 +1,42 @@
+<!--
```
antonymilne, line 1: Should this file exist if it has nothing in it? (Sounds like a deep philosophical question...)

vizro-core/pyproject.toml
```diff
@@ -56,7 +56,7 @@ exclude_lines = [
   "if __name__ == .__main__.:",
   "if TYPE_CHECKING:"
 ]
-fail_under = 82
+fail_under = 86
```
antonymilne, line 59: Does this coverage constraint just look at unit tests now @maxschulz-COL?

vizro-core/src/vizro/actions/_callback_mapping/_get_action_callback_mapping.py
```diff
@@ -18,7 +18,9 @@
 from vizro.managers._model_manager import ModelID
 
 
-def _get_action_callback_mapping(action_id: ModelID, argument: str) -> Union[Dict[str, Union[Input, State, Output]]]:
+def _get_action_callback_mapping(
+    action_id: ModelID, argument: str
+) -> Union[List[dcc.Download], Dict[str, Union[Input, State, Output]]]:
```
antonymilne, line 23: All of `Union[Input, State, Output]` are just subclasses of `DashDependency`, so you could just make it that if you prefer.

Or if you think it's clearer this way then fine, up to you 🙂

vizro-core/src/vizro/actions/_callback_mapping/_get_action_callback_mapping.py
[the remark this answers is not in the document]
antonymilne: I'm not sure I like the dictionary structure in general here, but for now it looks ok I think.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -59,6 +69,50 @@ def build(self):
             "action_finished": Output("action_finished", "data", allow_duplicate=True),
         }
 
+        action_components = _get_action_callback_mapping(
+            action_id=self.id, argument="components"  # type: ignore[arg-type]
```
antonymilne, line 73: Can we get rid of the `ignore` here? Do you just need to do `ModelID(self.id)`?

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -59,6 +69,50 @@ def build(self):
...
+        # Raising the custom exception if return value length doesn't match the number of outputs
+        return_value_len = (
+            1 if not hasattr(return_value, "__len__") or isinstance(return_value, str) else len(return_value)
+        )
+        outputs = list(ctx.outputs_grouping.keys())
+        outputs.remove("action_finished")
+        if len(outputs) != return_value_len:
+            raise ValueError(
+                f"Number of action's returned elements: {return_value_len} does not match the number"
+                f" of action's defined outputs: {len(outputs)}."
```
antonymilne, line 96:
```suggestion
                f"Number of action's returned elements ({return_value_len}) does not match the number"
                f" of action's defined outputs ({len(outputs)})."
```

vizro-core/src/vizro/models/_action/_action.py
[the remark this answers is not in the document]
antonymilne: Do we actually want to allow all these possibilities even if they're possible in Dash? Seems like dictionary is the most general one and doesn't leave any ambiguity in ordering arguments (plus we should allow single element so you don't need to write a dictionary just for one thing).

I remember thinking about this before and not sure whether we discussed it or not but interested in your thoughts on when the different output formats are useful. It's fine by me if we do want to allow the others, just would be good to understand. Tuple of elements makes sense to me because it's how you write normal Python functions; list of elements feels like kind of a weird Dash thing.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -59,6 +69,50 @@ def build(self):
...
+            return {"action_finished": None, **return_value}
+
+        return {"action_finished": None, **dict(zip(outputs, return_value))}
+
+    @_log_call
+    def build(self):
+        """Builds a callback for the Action model and returns required components for the callback.
+
+        Returns:
+            List of required components (e.g. dcc.Download) for the Action model added to the `Dashboard` container.
```
antonymilne, line 112: Is it possible/desirable to wrap these into a single `html.Div` or something else container? Until now we've had a policy that `build` should always return a single component and not a list of components. I know `Action` is a bit different, but it would be nice if we could maintain this convention if it's easy to do so.

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -0,0 +1,246 @@
...
+        for output in request.param
+    }
+
+
+@pytest.fixture
+def callback_context_set_outputs_grouping(request):
+    """Mock dash.callback_context that represents outputs grouping for custom action."""
+    outputs = request.param
+
+    outputs_grouping = {output: None for output in outputs}
```
antonymilne, line 47:
```suggestion
    outputs_grouping = dict.fromkeys(outputs)
```

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -0,0 +1,246 @@
...
+        "inputs",
+        [
+            "string",
+            "component_property",
+            "compo-nent.property",
+            "component.property_1",
+        ],
+    )
+    def test_inputs_invalid(self, inputs, test_action_function):
+        with pytest.raises(ValidationError, match=".* value is not a valid list .*"):
```
antonymilne, line 103:
```suggestion
        with pytest.raises(ValidationError, match="value is not a valid list"):
```
Just a simple text match will work here in the regex. Also in the next test.

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -0,0 +1,246 @@
...
+            ({"component_1_property": "new_value", "component_2_property": "new_value_2"}, ["component_1_property"]),
+        ],
+        indirect=True,
+    )
+    def test_action_callback_function_return_value_invalid(
+        self, custom_action_function_mock_return, callback_context_set_outputs_grouping
+    ):
+        action = Action(function=custom_action_function_mock_return())
+        with pytest.raises(
+            ValueError,
```
antonymilne, line 243: Could this be a `ValidationError`?

antonymilne (approved):
I'm going to need to spend much more time understanding all the actions code and the tests that already exist because there's a lot of code I'm not familiar with yet. But this looks like amazing work! Great job increasing the test coverage on things which were not easy to test! 💯 Just have a few minor suggestions and questions for my own understanding.

vizro-core/changelog.d/20230927_081254_petar_pejovic_actions_unit_tests.md
petar-qb: If you try, one of the Github checks will fail, so this file should exist 😄. (Just kidding) 
From a philosophical point of view it sounds fine to me to get contributors to create this even if they don't change anything worth mentioning. I see this compulsion as a reminder to contributors to be aware that they need to describe the changes if they are worth mentioning.
What do you think @maxschulz-COL?

maxschulz-COL: Yes! The idea @antonymilne is that its obligatory to create such a file to be aware that it may need writing. So far I think it has been a positive influence, and since its really easy to create (`hatch run docs:changelog`) its not really a burden

maxschulz-COL (approved):
Looks good from my side! Thanks a ton! If you could add the validator test for the validator introduced in my PR, that would be heroic, but not required :)

vizro-core/changelog.d/20230927_081254_petar_pejovic_actions_unit_tests.md
maxschulz-COL: Yup, and maybe one extra cool bit is that we have an unreleased section at the op: https://github.com/mckinsey/vizro/blob/main/vizro-core/CHANGELOG.md#unreleased

There we refer to the folder containing the fragments, so people can browse what branch/PR introduced what chagnes.

vizro-core/src/vizro/models/_action/_action.py
antonymilne: Oooh I see, that is a bit ugly. I think `ModelID(str(self.id))` is still better than the ignore though 🙂

vizro-core/changelog.d/20230927_081254_petar_pejovic_actions_unit_tests.md
antonymilne: Ah ok, nice! Understanding how the changelog works is on my list of things to catch up on but this sounds really great 👍

vizro-core/tests/unit/vizro/models/_action/test_action.py
petar-qb: How do you mean? To raise a `pydantic.ValidationError`?
It doesn't seem that worthwhile to me because the  `pydantic.ValidationError` requires a Model in the constructor. 
In that case I'd need to create a new model (e.g. LengthValidatior(element_1, element_2)) that has two arguments, and a validator that asserts `assert len(element_1) == len(element_2)`.

vizro-core/src/vizro/models/_action/_action.py
petar-qb: Let's continue this conversation [here](https://github.com/mckinsey/vizro/pull/80#discussion_r1348562316)

vizro-core/src/vizro/models/_action/_action.py
petar-qb: I managed to do it and now it looks much more consistent 😄. Also, the ActionLoop components are now bounded by an `html.Div` object and we can say that every application component (even every dcc.Store) has a parent Div.

vizro-core/src/vizro/models/_action/_action.py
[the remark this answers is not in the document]
petar-qb: Valid point. I concur that returning values like `dictionary`, `single element`, and `tuple` is clear. However, I'm now hesitant about using a `list` as a return value due to potential ambiguity. The question arises: Should we interpret the returned list as a single element, or is it meant to be treated differently? 

Since this logic is closely related to `custom actions` that are **not officially released**, do you mind if we continue this conversation in that PR (will open in October)? I'm also happy to re-implement the `list' handling in this PR if we have a strong preference for how to handle them.

vizro-core/tests/unit/vizro/models/_action/test_action.py
antonymilne: Ah yes, sorry - I misread the test. I thought the `Action()` was inside the `pytest.raises` so would be raising `ValidationError`. Ignore me.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -130,4 +158,12 @@ def build(self):
...
-        return action_components
+        # return action_components
+        return (
+            html.Div(
+                children=action_components,
+                id=f"{self.id}_action_model_components_div",
+            )
+            if action_components
+            else None
+        )
```
antonymilne, line 169: Thanks for making this change - I think it's much better now it's consistent 👍 

Just one more thing to make it even more consistent: I don't think we have any other `build` methods that can return `None` (is that right @huong-li-nguyen?). So let's try to avoid it here also if we can:
```suggestion
        return html.Div(
                children=action_components,
                id=f"{self.id}_action_model_components_div",
            )
```
or possibly `html.Div(children=None)`, whatever works.

It means we have a redundant empty div, but if that doesn't cause any problems then I think this is preferable.

FYI the basis for these changes is polymorphism, which here basically means to maintain a single consistent contract (function signature and return types) across all implementations of `build`. Currently that (very informal) contract states that the return type is always a single Dash component and cannot be `None`. The simpler we can keep the contract (so no `Optional` or `Union` types), the better (e.g. for type-checking, fewer bugs, ease of understanding).

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -113,18 +116,43 @@ def test_inputs_invalid(self, inputs, test_action_function):
...
+        ):
+            Action(function=export_data(file_format="invalid_file_format"))
+
+    def test_export_data_xlsx_without_required_libs_installed(self, monkeypatch):
+        # Mock only if the 'importlib.util.find_spec(argument)' argument is "openpyxl" or "xlsxwriter".
+        monkeypatch.setattr(
+            importlib.util,
+            "find_spec",
+            lambda arg: None if arg in ["openpyxl", "xlsxwriter"] else importlib.util.find_spec(arg),
+        )
```
antonymilne, line 141: Let's try to make this test more "black box" by doing it in a way that doesn't directly rely on the implementation. So basically try to monkeypatch at a higher level, something like this:
```
monkeypatch.setitem(sys.modules, "openpyxl", None)
monkeypatch.setitem(sys.modules, "xlswriter", None)
```

antonymilne (approved):
A couple more comments but still looks good 🙂

vizro-core/src/vizro/models/_action/_action.py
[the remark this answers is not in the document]
antonymilne: Very happy to continue the conversation elsewhere - it's not urgent at all and not related to this PR, I just commented on it since I happened to see it here 🙂

maxschulz-COL (approved):
LGTM! If we could still enable the higher level mock that would be great 💪

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: Yes, that's right. Currently, every `build` method returns a component. However, what we do have in a few places is something like below where components of the children can return None instead of an `html.Div`. This is required such that the `html.Div` doesn't add unnecessary space, which could mess up our padding/margins. This should hopefully still be in line with what you mention above @antonymilne because all these inner components that could potentially return None are created within the return component, so shouldn't affect the type checks etc.

```
A = html.Div(children=[a_children]) if a_children or None
return html.Div(
                children=[a, b, c],
                id=f"{self.id}_action_model_components_div",
            )
```

@petar-qb , if you implement the above - can you double-check that that `html.Div `doesn't add any spacing? Given that this `html.Div` doesn't contain any visualizable components, it should be invisible (so no added spacings etc.). You might have to provide a className to the html.div called "action_components_container_outer" and then provide a width and height of 0?

petar-qb: I've already checked it and it doesn't add any spacing because it doesn't contain any visible objects in it.

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -0,0 +1,288 @@
...
+    def test_export_data_file_format_invalid(self):
+        with pytest.raises(
+            ValueError, match='Unknown "file_format": invalid_file_format.' ' Known file formats: "csv", "xlsx".'
+        ):
+            Action(function=export_data(file_format="invalid_file_format"))
+
+    def test_export_data_xlsx_without_required_libs_installed(self, monkeypatch):
+        monkeypatch.setitem(sys.modules, "openpyxl", None)
+        monkeypatch.setitem(sys.modules, "xlswriter", None)
+        # monkeypatch.setattr(
```
maxschulz-COL, line 149: Do we not want to eremove the commented out code @petar-qb ?

petar-qb: Sure, my mistake 😞.
Let's patch it in upcoming PR.

antonymilne: Also `# return action_components` in _action.py 🙂

PR #84, opened 2023-10-01
Fix unit test interdependence 

huong-li-nguyen:
## Description
- Fix unit test interdependence caused by dash.page_registry not being cleaned up after dashboard.build
  - Replace app.build with dashboard.build in unit tests for Accordion / Navigation /Dashboard
  - Add unit test `test_page_build_left_side_removed` (could previously not be added due to interdependence)
- Refactor `Accordion` model (as part of an investigation on whether we can get rid of the dashboard.build dependence)

## Checklist

## Types of changes

## Notice


vizro-core/tests/unit/vizro/models/test_page.py
```diff
@@ -98,3 +99,8 @@ def test_action_auto_generation_valid(self, standard_px_chart):
 
 
 # TODO: Add unit tests for private methods in page build
+def test_page_build_left_side_removed(standard_px_chart):
+    page = vm.Page(title="Single Page", components=[vm.Graph(id="scatter_chart", figure=standard_px_chart)])
+    vm.Dashboard(pages=[page]).build()
+    assert ("className='left_side'" not in str(page.build())) is True
+    del dash.page_registry["Single Page"]
```
maxschulz-COL, line 106: Is there a way to clear the entire registry? In that case it might make sense to have a fixture for that, that we can attach to any test that needs it, instead of doing it manually at the end

maxschulz-COL (dismissed):
Lgtm! Thanks for tackling this

vizro-core/src/vizro/models/_navigation/_accordion.py
```diff
@@ -84,32 +75,24 @@ def _create_default_accordion(self):
             id=f"{self.id}_outer",
         )
 
-    def _create_accordion(self):
-        accordion_items = []
+    def _create_default_accordion(self):
+        """Creates a default accordion with all pages provided to the Dashboard."""
+        registered_pages = [page for page in dash.page_registry.keys() if page != MODULE_PAGE_404]
```
petar-qb, line 80: Does it make sense to define `self.pages` in `Accordion._pre_build` method (based on custom `self.pages` or `dash.page_registry` for default accordion creation) and then to reuse the `self.pages` in methods:
 - _create_default_accordion,
 - _create_custom_accordion,
 - _create_accordion_buttons,

In that way we don't need to calculate pages (from `dash.page_registry.keys()`) each time the navigation is built (and it's built every time the page is changed).

Also, then we don't need to propagate pages object to the `_create_accordion_buttons` every time.

vizro-core/src/vizro/models/_navigation/_accordion.py
```diff
@@ -40,32 +44,19 @@ def _create_accordion_buttons(self, accordion_pages):
...
-            )
-            for page in dash.page_registry.values()
-            if page["module"] != MODULE_PAGE_404
-        ]
-
-        accordion_items = [self._create_accordion_item(accordion_buttons=accordion_buttons)]
-
-        # Don't create accordion navigation if there is only one page and one accordion item
+    def _get_accordion_container(self, accordion_items, accordion_buttons):
+        # Return no container if there is only one page in the dashboard
```
petar-qb, line 59: Could it happen that `len(accordion_buttons)` is equal to 1 and that `len(accordion_items)` isn't? 
It looks like we can replace the `if` statement below with the following:
`if len(accordion_buttons) == 1:`

petar-qb (approved):
I really like the refactored code. It's so easy to read and understand now.

vizro-core/tests/unit/vizro/models/test_page.py
huong-li-nguyen: Unfortunately not! But this is what I initially tried e.g. `dash.page_registry = OrderedDict()` - however, this permanently clears the dash.page_registry and we end up having interdependence between the tests again.

vizro-core/src/vizro/models/_navigation/_accordion.py
huong-li-nguyen: That check wouldn't suffice! For example, if you provide the navigation as below - you would have ` len(accordion_buttons) = 1` while `len(accordion_items) = 2`. In this case, I don't want to return the Accordion. So the check that both accordion_buttons and accordion_items are of length 1 are required to ensure that only one page exists in the dashboard :

```
 navigation=vm.Navigation(
        pages={
            "Analysis": ["Variable Analysis"],
        }
    ),
```

vizro-core/changelog.d/20231001_000340_huong_li_nguyen_fix_test_interdependence.md
```diff
@@ -0,0 +1,41 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+
+### Fixed
+
+- Fix unit test interdependence issue due to shared dash.page_registry ([#84](https://github.com/mckinsey/vizro/pull/84))
```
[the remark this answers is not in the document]
maxschulz-COL: Overall I think we should not try to be ironclad about any rule we come up with. If it's very important I would mention it either way. Most importantly the changelog is for users and contributors alike to understand what changed between versions.

In general I think it is more important for users though, as they are more likely to compare versions and decide which one to use. So would always be more in favour of keeping keeping the focus on issues that concern/affect the user.

So this one is probably a 50/50, I personally would not really care to read it in a changelog, but I can see a case where this could be relevant. But most of the time it feels like clogging.

vizro-core/src/vizro/models/_navigation/_accordion.py
huong-li-nguyen: I like the push! It gave me another idea, so I am deviating a bit from your proposal, but you got me inspired 😄 

I've moved the default page call with `dash.page_registry.keys()` to the validator `_validate_pages`. Having it in the validator instead of the pre-build has the advantage of re-using that validator in the Navigation model and other subsequent navigation-type models, so we're not repeating the same code in several pre-builds. It also makes a pre-build method redundant.

Because of that change, we can eliminate the `_create_default_accordion` private method!

vizro-core/changelog.d/20231001_000340_huong_li_nguyen_fix_test_interdependence.md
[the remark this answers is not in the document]
huong-li-nguyen: Fully agree! I would leave this in then

vizro-core/src/vizro/models/_navigation/_accordion.py
petar-qb: Happy to see latest changes 😄.

maxschulz-COL (dismissed):
I like the new idea of moving it to the validator, this improved the code readibility a lot! And we can remove the registry access in one place only if we ever want to

vizro-core/src/vizro/models/_navigation/_accordion.py
```diff
@@ -40,33 +42,20 @@ def _create_accordion_buttons(self, accordion_pages):
                 href=page["relative_path"],
             )
             for page in dash.page_registry.values()
-            for accordion_page in accordion_pages
-            if accordion_page == page["module"] and page["module"] != MODULE_PAGE_404
+            if page["name"] in pages
```
petar-qb, line 45: Since you moved dash registry pages into the validator, does it mean we can use `self.pages` here?

vizro-core/src/vizro/models/_navigation/_accordion.py
```diff
@@ -40,33 +42,20 @@ def _create_accordion_buttons(self, accordion_pages):
...
-            if page["module"] != MODULE_PAGE_404
-        ]
-
-        accordion_items = [self._create_accordion_item(accordion_buttons=accordion_buttons)]
-
-        # Don't create accordion navigation if there is only one page and one accordion item
-        if len(accordion_buttons) == len(accordion_items) == 1:
+    def _get_accordion_container(self, accordion_items, accordion_buttons):
+        # Return no container if there is only one page in the dashboard or no pages exist
+        if (len(accordion_buttons) == len(accordion_items) == 1) or not accordion_buttons:
```
petar-qb, line 58: 🚀

petar-qb (dismissed):
I shared a little note [here](https://github.com/mckinsey/vizro/pull/84#discussion_r1342513902), but feel free to overlook it and go ahead with the merge if it doesn't quite resonate with you 😄.

maxschulz-COL (approved):

petar-qb (approved):

PR #88, opened 2023-10-02
Enable Button as ControlType

huong-li-nguyen:
## Description
- Add Button to ControlType

## Checklist

## Types of changes

## Notice


maxschulz-COL (commented):
Given that we have significantly extended our API, we should probably add this new type tp our user guides! It may duplicate our actions, but maybe we can find a good way to do so

huong-li-nguyen:
> Given that we have significantly extended our API, we should probably add this new type tp our user guides! It may duplicate our actions, but maybe we can find a good way to do so

Hmm.. I see your point but we would essentially duplicate our user guide on the Button (that is currently living under components) under controls as well. We could also just add it in as a section and then reference the content in components?

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -87,7 +87,7 @@
   width: 336px;
 }
 
-.keyline {
+hr {
```
petar-qb, line 90: Does it make sense to add a class (eg "vizro-hr") to every `html.Hr()` we use and target them with the `hr.vizro-hr` css selector. It might help if users add custom components that contain `html.Hr()` that are not affected by Vizro css.

petar-qb (approved):
Glad to see that ControlType supports vm.Button.
I left one comment, but other than that, the PR looks good!

vizro-core/src/vizro/static/css/layout.css
huong-li-nguyen: In this case, the other way around makes more sense :) In general, we should provide the visual default for the user to maintain visual consistency across the dashboards with our components. Hence, we define the default styling for all horizontal rulers via `hr`.

If a user wants to deviate from the default styling for a specific ruler, they will provide a class name to the ruler e.g. `html.Hr(className="my-ruler")` and provide their custom styling via `hr.my-ruler {}`.

Users who want to deviate from the default styling for all rulers can overwrite the `hr {}` in their custom assets folder. So here, we do not need to provide a className.

petar-qb: Both approaches are correct, we just need to align and be consistent in the approach we want to push.

antonymilne: (unresolving just so the comment shows, not because there's anything more to resolve here)

huong-li-nguyen: We should be good! We already followed that approach for all typography elements and links, the horizontal ruler was the only element where we didn't follow it which was now fixed in this PR https://github.com/mckinsey/vizro/pull/91 :)

petar-qb: Glad to see we're following this approach everywhere 😄.

huong-li-nguyen:
Closing due to decision to not add Button as an officially supported ControlType and instead go via the route of custom components such that the user can add the Button as a ControlType themselves via `vm.Page.add_type("controls", vm.Button)`

Issue #94, opened 2023-10-04
Allow Parameters to have optional selection


maxschulz-COL:
### What's the problem this feature will solve?

Currently when using parameters it is impossible to use `None` as this is forbidden by pydantic types

### Describe the solution you'd like

Ability to allow parameters to take `None` as value

### Alternative Solutions

Pydantic validation forbids any feasible work arounds

### Additional context

None required

### Code of Conduct

- [X] I agree to follow the  [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

PR #95, opened 2023-10-04
Allow None/Optional parameters

maxschulz-COL:
## Description
- This PR adresses the https://github.com/mckinsey/vizro/issues/94
- New behaviour: Allows `NONE` (the string) as a valid parameters option, which gets converted into `None`
- Additional feature: allows dropdown to be clearable

## Checklist

## Types of changes

## Notice


vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -34,6 +36,17 @@ def _get_component_actions(component) -> List[Action]:
...
 
 
+def _validate_selector_value_NONE(value: Union[SingleValueType, MultiValueType]) -> ValidatedNoneValueType:
+    validated_value: ValidatedNoneValueType = value
+    if value == [NONE_OPTION]:
+        validated_value = [None]
+    elif value == NONE_OPTION:
+        validated_value = None
+    elif isinstance(value, list) and len(value) > 1 and NONE_OPTION in value:
+        validated_value = [i for i in value if i != NONE_OPTION]  # type: ignore[assignment]
```
petar-qb, line 46: I don't understand why the NONE_OPTION is removed from the list of values in the statement:
```
elif isinstance(value, list) and len(value) > 1 and NONE_OPTION in value:
        validated_value = [i for i in value if i != NONE_OPTION]
```

Isn't that only NONE_OPTION should be replaced with `None` value in a list, as:
```
elif isinstance(value, list) and len(value) > 1 and NONE_OPTION in value:
        validated_value = [i for i in value if i != NONE_OPTION else None]
```

I see it's already covered by unit tests as well, but still don't understand the reason.

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -112,20 +125,20 @@ def _get_parametrized_config(targets: List[str], parameters: List[CallbackTrigge
...
-        for ctx_trigger_dict in parameters:
-            selector_value = ctx_trigger_dict["value"]
-            selector_actions = _get_component_actions(model_manager[ctx_trigger_dict["id"]])
+        for ctd in parameters:
+            selector_value = ctd["value"]
+            if hasattr(selector_value, "__iter__") and ALL_OPTION in selector_value:  # type: ignore[operator]
+                selector: SelectorType = model_manager[ctd["id"]]
+                selector_value = selector.options
+            selector_value = _validate_selector_value_NONE(selector_value)
+            selector_actions = _get_component_actions(model_manager[ctd["id"]])
```
petar-qb, line 134: I 👍 it!

vizro-core/src/vizro/models/_controls/parameter.py
```diff
@@ -27,7 +27,8 @@ class Parameter(VizroBaseModel):
     Args:
         type (Literal["parameter"]): Defaults to `"parameter"`.
         targets (List[str]): Targets in the form of `<target_component>.<target_argument>`.
-        selector (Optional[SelectorType]): See [SelectorType][vizro.models.types.SelectorType]. Defaults to `None`.
+        selector (SelectorType): See [SelectorType][vizro.models.types.SelectorType]. Converts selector value
+            `NONE` (str) into `None` (value) to allow optional parameters.
```
petar-qb, line 31:
```suggestion
        selector (SelectorType): See [SelectorType][vizro.models.types.SelectorType]. Converts selector value
            "NONE" (str) into `None` (value) to allow optional parameters.
```

vizro-core/changelog.d/20231004_184107_maximilian_schulz_allow_none_parameters.md
```diff
@@ -0,0 +1,41 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+
+### Added
+
+- Passing `NONE` (`str`) to options of `Parameter` selectors pass `None` (`value`) to underlying component. This allows optional parameters. ([#95](https://github.com/mckinsey/vizro/pull/95))
```
petar-qb, line 16:
```suggestion
- Passing "NONE" (`str`) to options of `Parameter` selectors pass `None` (`value`) to underlying component. This allows optional parameters. ([#95](https://github.com/mckinsey/vizro/pull/95))
```

vizro-core/src/vizro/actions/_actions_utils.py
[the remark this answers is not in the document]
maxschulz-COL: i can try this, but I fear mypy will complain. Let's see

maxschulz-COL: The reason is that `NONE` is not a valid value, so it would either need to be removed, or replaced.

It is replaced in the cases where it is the single option, but removed in the case where it is among others, as I believe this would be the intuitive case: None + smth = smth.

Do you think we would want to make it so that users can actively use ["foo", "bar",None] or should that not represent ["foo", "bar"]?

vizro-core/src/vizro/models/_controls/parameter.py
antonymilne:
```suggestion
        selector (SelectorType): See [SelectorType][vizro.models.types.SelectorType]. Converts selector value
            `"NONE"` into `None` to allow optional parameters.
```

vizro-core/src/vizro/actions/_actions_utils.py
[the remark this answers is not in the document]
antonymilne: Agree with @petar-qb here and actually I'd go one step further and just do `return None` etc. directly in the control statements without making the `validated_value` variable at all.

The `and len(value) > 1 and NONE_OPTION in value` is actually redundant here but maybe makes things more explicit.

Is there any chance we could end up with `value = [NONE_OPTION, NONE_OPTION]`? This case would currently end up as `[]` when maybe it should instead be `[None]`.

antonymilne (approved):
Generally looks great, just a few small comments! ⭐

vizro-core/src/vizro/actions/_actions_utils.py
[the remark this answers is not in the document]
maxschulz-COL: Ok great points! Implemented a second version

vizro-core/tests/unit/vizro/actions/test_parameter_action.py
```diff
@@ -137,6 +174,35 @@ def test_one_parameter_one_target(
...
+        self,
+        callback_context_parameter_hover_data,
+        target_scatter_parameter_hover_data,
+    ):
+        # Creating and adding a Parameter object to the existing Page
+        y_parameter = vm.Parameter(
+            id="test_parameter",
+            targets=["scatter_chart.hover_data"],
+            selector=vm.Dropdown(
+                id="hover_data_parameter", multi=True, options=["None", "lifeExp", "pop", "gdpPercap"], value=["None"]
```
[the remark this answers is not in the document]
maxschulz-COL: No, you are correct, corrected now!

vizro-core/tests/unit/vizro/actions/test_parameter_action.py
```diff
@@ -71,6 +78,36 @@ def callback_context_parameter_y(request):
     return context_value
 
 
+@pytest.fixture
```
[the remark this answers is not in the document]
maxschulz-COL: They are a tiny bit janky, but I think they are nonetheless really cool as they allow us to test actions properly carried out

vizro-core/tests/unit/vizro/actions/test_parameter_action.py
```diff
@@ -137,6 +174,35 @@ def test_one_parameter_one_target(
         assert result["scatter_chart"] == target_scatter_parameter_y
         assert "box_chart" not in result
 
+    @pytest.mark.parametrize(
+        "callback_context_parameter_hover_data, target_scatter_parameter_hover_data",
+        [(["NONE", "pop"], ["pop"]), (["NONE"], [None])],
```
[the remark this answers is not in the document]
maxschulz-COL: Regarding "NONE", this is covered in the test above.

As for more tests, great idea, I also introduced other tests, e.g. what if there is just `["ALL"].

The (hopefully) intuitive idea is that it displays the largest union. Currently having `["ALL", "foo"]` resolves to `["foo","bar"]` and so I think it is natural to have `["NONE", "ALL"]` also resolve to `["foo","bar"]`.

vizro-core/tests/unit/vizro/actions/test_parameter_action.py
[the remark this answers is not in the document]
antonymilne: Was the test definitely testing the right thing then? Because It seemed to pass before when you had this as `"None"`.

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -37,14 +37,13 @@ def _get_component_actions(component) -> List[Action]:
...
-        validated_value = None
-    elif isinstance(value, list) and len(value) > 1 and NONE_OPTION in value:
-        validated_value = [i for i in value if i != NONE_OPTION]  # type: ignore[assignment]
-    return validated_value
+    if value == NONE_OPTION:
+        return None
+    elif value == [NONE_OPTION]:
+        return [None]
+    elif isinstance(value, list):
+        return [i for i in value if i != NONE_OPTION] or [None]  # type: ignore[list-item, return-value]
```
antonymilne, line 45: I believe the `value == [NONE_OPTION]` case is now contained in the one that follows? So probably not required as a separated `elif` unless you think it's clearer this way.

antonymilne (approved):

vizro-core/tests/unit/vizro/actions/test_parameter_action.py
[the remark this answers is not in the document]
maxschulz-COL: Yes, because it didn't matter in the test cases used. The callback context faked that the user would select `"NONE"`, at which point the original options or value do not matter.

Having introduced `"ALL"` this changed, and the error showed, so I fixed it :)

vizro-core/src/vizro/actions/_actions_utils.py
maxschulz-COL: Hm it is clearer, but I think we can skip it, makes it shorter which is good :)

vizro-core/src/vizro/actions/_actions_utils.py
petar-qb: Oke, I see. We handle "NONE" in the opposite way we handle "ALL" option which absolutely makes sense to me. 

(more like a reminder and heads up then a real question 😄) ⬇️ 
I'm just wondering is it possible to send something like `[1, 2, None, 4]` from Parameter (e.g. vm.Dropdown) to `chart.x` target or `['blue', 'red', None, 'purple']` from another Parameter to `chart.mark.color` target? (probably we encounter similar problem with the "ALL" option as well).

And if it's impossible, are we aware of that limitation?

petar-qb (approved):
Really nice job. ✅  

I just left one heads-up in a [comment](https://github.com/mckinsey/vizro/pull/95#discussion_r1348598902).

vizro-core/src/vizro/actions/_actions_utils.py
maxschulz-COL: As per above, I think if I understand correctly what you mean, it will not be possible.

Issue #86, opened 2023-10-02
Absence of openpyxl library causes console errors


l0uden:
### Description

When trying to export xlsx file using actions, console error occurs

```
Traceback (most recent call last):
  File "/Users/o/PycharmProjects/vizro/vizro-core/src/vizro/models/_action/_action.py", line 80, in callback_wrapper
    return_value = self.function(**inputs) or {}
  File "/Users/o/PycharmProjects/vizro/vizro-core/src/vizro/models/types.py", line 59, in __call__
    return self.__function(**bound_arguments.arguments)
  File "/Users/o/PycharmProjects/vizro/vizro-core/src/vizro/actions/export_data_action.py", line 61, in export_data
    callback_outputs[f"download-dataframe_{target_id}"] = dcc.send_data_frame(
  File "/Users/o/anaconda3/envs/w3af/lib/python3.8/site-packages/dash/dcc/express.py", line 91, in send_data_frame
    return _data_frame_senders[name](writer, filename, type, **kwargs)
  File "/Users/o/anaconda3/envs/w3af/lib/python3.8/site-packages/dash/dcc/express.py", line 32, in send_bytes
    content = src if isinstance(src, bytes) else _io_to_str(io.BytesIO(), src, **kwargs)
  File "/Users/o/anaconda3/envs/w3af/lib/python3.8/site-packages/dash/dcc/express.py", line 58, in _io_to_str
    writer(data_io, **kwargs)
  File "/Users/o/anaconda3/envs/w3af/lib/python3.8/site-packages/pandas/core/generic.py", line 2252, in to_excel
    formatter.write(
  File "/Users/o/anaconda3/envs/w3af/lib/python3.8/site-packages/pandas/io/formats/excel.py", line 934, in write
    writer = ExcelWriter(  # type: ignore[abstract]
  File "/Users/o/anaconda3/envs/w3af/lib/python3.8/site-packages/pandas/io/excel/_openpyxl.py", line 56, in __init__
    from openpyxl.workbook import Workbook
ModuleNotFoundError: No module named 'openpyxl'
```

### vizro version

0.1.3

### Python version

3.8

### OS

Mac OS

### How to Reproduce

1. Install vizro package into clean env
2. Add export xlsx action to the dashboard
3. Run export action

### Code of Conduct

- [X] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

antonymilne:
[...]

PR #97, opened 2023-10-05
Introduce validator to raise ModuleNotFoundError at dashboard creation

maxschulz-COL:
## Description
- adresses https://github.com/mckinsey/vizro/issues/86
- @petar-qb It might make sense to add the test for this only after you merged the PR on the action model, or directly add it to that PR

## Checklist

## Types of changes

## Notice


vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -35,6 +36,16 @@ class Action(VizroBaseModel):
...
     )
 
+    @validator("function")
+    def check_export_imports(cls, function):
+        if (
+            function._CapturedCallable__function.__name__ == "export_data"
+            and function._arguments.get("file_format") == "xlsx"
+        ):
+            if importlib.util.find_spec("openpyxl") is None and importlib.util.find_spec("xlsxwriter") is None:
+                raise ModuleNotFoundError("You must install either openpyxl or xlsxwriter to export to xlsx format.")
```
chiara-sophie, line 46:
```suggestion
                raise ModuleNotFoundError("Either openpyxl or xlsxwriter must be installed to export to xlsx format.")
```

chiara-sophie (commented):
Looks good! 👍 Just an idea on how I would rewrite the Error message.

petar-qb:
The link from the PR Description is wrong, so can you just update it? 😄.

vizro-core/src/vizro/models/_action/_action.py
maxschulz-COL: Hm, I think active is perferrable over passive no?

chiara-sophie: I prefer the passive, because you must seems quite harsh - but i guess it is personal preference :D

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -35,6 +36,17 @@ class Action(VizroBaseModel):
         regex="^[a-zA-Z0-9_]+[.][a-zA-Z_]+$",
     )
 
+    # TODO: Abstract away once pattern becomes clear regarding function/import checks
+    @validator("function")
+    def check_export_imports(cls, function):
+        if (
+            function._CapturedCallable__function.__name__ == "export_data"
+            and function._arguments.get("file_format") == "xlsx"
```
antonymilne, line 44:
```suggestion
            function._function.__name__ == "export_data"
            and function._arguments.get("file_format") == "xlsx"
```

The `CapturedCallable._function` property is there to avoid needing to access the name-mangled `__function`. Accessing something that is name-mangled like that is very unlikely to be the right thing to do, even in a private function.

I don't much like the use of `function._arguments.get` (as per the comments on the `_arguments` property I hope to remove it some time if feasible) but I see why you've done it since I know `function.get` won't itself work. 

We should probably [`apply_defaults`](https://docs.python.org/3/library/inspect.html#inspect.BoundArguments.apply_defaults) in `CapturedCallable.__getitem__` or `CapturedCallable.__init__` to get things like this working more nicely, but that's way outside the scope of this PR and would need some extra thought and testing, so let's not do here.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -35,6 +36,17 @@ class Action(VizroBaseModel):
...
+        ):
+            if importlib.util.find_spec("openpyxl") is None and importlib.util.find_spec("xlsxwriter") is None:
```
antonymilne, line 46: Do you have any reference on why to check for modules this way? I've never seen it before and the Python docs are not particularly enlightening.

What I was expecting/probably more conventional was `import` wrapped in `try/except`, but I expect the `importlib` version is fine if it works.

vizro-core/changelog.d/20231005_145819_maximilian_schulz_86_absence_of_openpyxl_library_causes_console_errors.md
```diff
@@ -0,0 +1,41 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+
+### Added
+
+- Raise `ModuleNotFoundError` in case the `export_data` action is used, the export type is `xlsx`, but neither `openpyxl` or `xlsxwriter` are installed. ([#97](https://github.com/mckinsey/vizro/pull/97))
```
antonymilne, line 16:
```suggestion
- Raise `ModuleNotFoundError` in case the `export_data` action is used with `file_format="xlsx"` and neither `openpyxl` nor `xlsxwriter` are installed ([#97](https://github.com/mckinsey/vizro/pull/97))
```
(Do we have a consistent use of full stops on these changelog snippets? Generally seems like we don't include them? Very pedantic I know but would be tidy...)

vizro-core/src/vizro/models/_action/_action.py
maxschulz-COL: Good point - so this was how I was implementing it first, but then `ruff` complained and suggested this alternative way, which I agree is the cleaner way.

https://docs.astral.sh/ruff/rules/unused-import/

vizro-core/changelog.d/20231005_145819_maximilian_schulz_86_absence_of_openpyxl_library_causes_console_errors.md
maxschulz-COL: We haven't made any guideline on this yet... It can be altered in the release PR, when all the fragments are collected together. For me that is acceptable for now.

The only guidelines we follow so far:
- focus on what would be important primarily to a user
- use active language (add, change, guide, raise...)

vizro-core/src/vizro/models/_action/_action.py
maxschulz-COL: Ah good point! Yes it didn't feel right at all! Just had forgotten about how to do it the proper way. Thanks!

Suggestion makes sense, shall I leave a note here? I guess it will be found automatically once we remove it, so maybe not necessary

l0uden (approved):
Cool!
Tested it. 
Approve from my side.

vizro-core/src/vizro/models/_action/_action.py
antonymilne: I see we also have a `TODO` on the one other call to `_arguments`, so let's put one in here as well just so we've got it everywhere a developer might be looking.

vizro-core/src/vizro/models/_action/_action.py
antonymilne: Ah nice 👍 Thanks for the information.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -35,6 +36,17 @@ class Action(VizroBaseModel):
         regex="^[a-zA-Z0-9_]+[.][a-zA-Z_]+$",
     )
 
+    # TODO: Abstract away once pattern becomes clear regarding function/import checks
```
antonymilne, line 39: This comment is very vague and I think should be clearer. Imagine you're a junior developer who wasn't part of the conversation you and I had yesterday - what is actually undesirable with the code as it exists and what might possible solutions be?

Problem: generic `Action` model shouldn't depend on details of particular actions like export_data.
Possible solutions:  make a generic mapping of action functions to validation functions or the imports they require, and make the code here look up the appropriate validation using the function as key.

antonymilne (approved):
Just one suggestion but looks good 👍

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -35,6 +36,17 @@ class Action(VizroBaseModel):
...
+    @validator("function")
+    def check_export_imports(cls, function):
+        if (
+            function._function.__name__ == "export_data"
+            and function._arguments.get("file_format") == "xlsx"
+        ):
+            if importlib.util.find_spec("openpyxl") is None and importlib.util.find_spec("xlsxwriter") is None:
```
petar-qb, line 46: We had the discussion about predefined actions error handling few months ago (and we are aware we still don't have the right answer and solution). I don't have a strong preference about where these errors should be handled. There are two options:
 - In the Action model (error will be raised as soon the Action model is initialised).
 - In the predefined action function body (error will be dynamically raised when the action is triggered from the UI). 

There's just one thing I have a strong preference and it is that we need to be consistent.

There are already few exceptions handling defined in the predefined action functions and the best example is the one from the predefined `export_data` action where we handle "Unknown file_format":

```
        if file_format == "csv":
            writer = data_frames[target_id].to_csv
        elif file_format == "xlsx":
            writer = data_frames[target_id].to_excel
        else:
            raise ValueError(f'Unknown "file_format": {file_format}.' f' Known file formats: "csv", "xlsx".')
```
 
This exception could also be handled in a similar way the required installed library is (in the Action model validator).

Do you guys think this error is more specific than "Unknown file_format" is and that it should be handled here (in a different way)?

petar-qb (approved):

vizro-core/src/vizro/models/_action/_action.py
maxschulz-COL: Hm that's a very valid point. I just tried it out, and it is a very similar case to the one we just implemented. My gut feeling is also that we should be consistent, and having the wrong parameters feels like a similar level to not having a module installed...

@antonymilne: What do you think? Should we address this immediately, or shall we merge as is to come back to this later?

vizro-core/src/vizro/models/_action/_action.py
maxschulz-COL: Added it to the `TODO` in `_arguments`, there was no `TODO` in `CapturedCallable.__getitem__`, so I left that one out.

vizro-core/src/vizro/models/_action/_action.py
maxschulz-COL: @petar-qb We have moved this both to the Action now (including a TODO on how to improve that).

Could we cover this validator in the Action model unit tests?

petar-qb: That's great 🥇.  

The PR is ready for merging (just remove or comment out the `tests/unit/vizro/actions/test_export_data_action.py::TestExportData::test_invalid_file_format`, and I'll incorporate this test in my PR).

I also think we need to talk more about action function arguments validation (really not urgent), because the topic is slightly complicated then it seems on the first glance.

PR #113, opened 2023-10-16
Provide CSS ID to outer div containers

huong-li-nguyen:
## Checklist

## Types of changes

## Notice

### DIFF
--- a/vizro-core/src/vizro/models/_navigation/_accordion.py
+++ b/vizro-core/src/vizro/models/_navigation/_accordion.py
@@ -76,7 +76,7 @@ class Accordion(VizroBaseModel):
                 html.Hr(),
             ],
             className="nav_panel",
-            id=f"{self.id}_outer",
+            id="nav_panel_outer",
         )
 
     def _create_accordion(self, active_page_id):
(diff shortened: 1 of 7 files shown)


vizro-core/src/vizro/models/_navigation/_accordion.py
```diff
@@ -76,7 +76,7 @@ def _get_accordion_container(self, accordion_items, accordion_buttons):
                 html.Hr(),
             ],
             className="nav_panel",
-            id=f"{self.id}_outer",
+            id="nav_panel_outer",
```
maxschulz-COL, line 79: Is this so they can be adressed all together, and not require a one-by-one treatment?

maxschulz-COL (approved):
Lg from myside

vizro-core/src/vizro/models/_navigation/_accordion.py
huong-li-nguyen: The nav panel is always unique per page, so they don't require an additional provision of a unique ID. This also makes it easier for us and the user to affect the container, as the ID is always the same. Previously users wouldn't even be able to provide an ID to the nav_panel without accessing our private models. 

If at some point we could have several nav panels within the same page, we would need the ID in there, but currently, this is not planned.

nadijagraca: Following that logic, I would need to add "nav_bar_outer" to the nav_bar to my icon pr

nadijagraca (approved):
Looks good to me.

PR #114, opened 2023-10-16
Enable dash datatable

maxschulz-COL:
## Description
- Implementation of the `vm.Table` model, alongside the shipped standard table function
- implementation of `vizro.charts.tables.dash_data_table`

Further refactorings include:
- multiple abstractions of implementation specific details
- correct typing for `ModelID`
- moving theme update logic out of page model

Things to be added once implementation is agreed upon:
- [x] Unit tests
- [x] Documentation

**EDIT**

Main open questions - may not all need to be adressed in this PR:
- [x] Correct abstraction level for component properties - see in parts https://github.com/mckinsey/vizro/pull/114#discussion_r1365498632
- [x] Correct fix for the outcome of `bind_partials` - see https://github.com/mckinsey/vizro/pull/114#discussion_r1364523451 (fixed in other PR)
- [x] Correct input path for `dash_data_table` and correct name thereof - see https://github.com/mckinsey/vizro/pull/114#discussion_r1364532187
- [x] Correct overall renaming of things - see https://github.com/mckinsey/vizro/pull/114#discussion_r1365174688

**EDIT 2**
PR almost finished, tasks left to do:
- correct place for `ModelID`
- @nadijagraca removal of `!important`

## Screenshot
Screenshot 2023-10-25 at 19 34 23

## Checklist

## Types of changes

## Notice


vizro-core/src/vizro/models/_page.py
```diff
@@ -132,11 +132,11 @@ def build(self):
...
-    def _update_graph_theme(self):
+    def _update_graph_theme(self):  # LN: needs to be refactored so plotly-independent or extendable - DONE
         outputs = [
-            Output(component.id, "figure", allow_duplicate=True)
+            component._get_update_graph_theme_output()
             for component in self.components
-            if isinstance(component, Graph)
+            if hasattr(component, "_get_update_graph_theme_output")
         ]
         if outputs:
```
huong-li-nguyen, line 141: @maxschulz-COL  - can this be moved completely out of the page model? Asking because this is currently affecting the creation of tabs/containers in a draft PR I have. Optimally, this should live in the graph itself, as these containers shouldn't need to be responsible for the theme update in their components.

maxschulz-COL: hm i can try

maxschulz-COL: Yes, this is actually quite easy I think, and definitely the right call!

vizro-core/examples/default/app.py
maxschulz-COL, line 1: Will be changed back after, this is just for showcase purpose.

Does it make sense to add the table to the demo, given that its such a prominent component?

vizro-core/pyproject.toml
```diff
@@ -23,11 +23,11 @@ dependencies = [
   "pandas",
   "pydantic>=1.10.13, <2",  # must be synced with pre-commit mypy hook
   "dash_daq",
+  "dash_ag_grid",
```
maxschulz-COL, line 26: Take out after PR review

vizro-core/src/vizro/static/css/dropdown.css
```diff
@@ -131,11 +138,12 @@ div.nav_panel a {
 
 /** spacing for tag elements only when elements are present inside
 wrapper **/
+/* Padding allows elements not to jump but may be too large */
 .Select--multi.has-value .Select-multi-value-wrapper {
   display: flex;
   flex-wrap: wrap;
   gap: 4px;
-  padding: 4px 8px;
+  padding: 13px 8px;
```
maxschulz-COL, line 146: @nadijagraca Can you help me in fixing this behaviour: the current setting stops the jumping of the dropdown, but now the padding is a little too large for my taste. Have not been able to figure out why

vizro-core/src/vizro/charts/tables/dash_table.py
```diff
@@ -0,0 +1,14 @@
+import pandas as pd
+from dash import dash_table
+
+
+from vizro.models.types import capture
+
+@capture("table")
+def dash_data_table(data_frame: pd.DataFrame, **kwargs):
```
antonymilne, line 8: Still to discuss some time: what should we call this? `table` or `data_table` or `dash_data_table` are the options I think.

antonymilne (commented):
I just did an initial read through and it definitely feels like we're on the right track here 👍 Very nice job abstracting away all the right pieces. My only major concern is the captured callable stuff we discussed earlier, so would be good to get to the bottom of that.

Ideally the mypy tidying would go in a separate PR. Normally I don't mind cheekily sneaking a small unrelated change or two into a PR but here it causes quite a widespread set of changes to what is an important PR that adds a big new feature, so it's a bit much here. But it's probably more effort than it's worth to split off those changes to a separate branch so don't worry about it now.

vizro-core/examples/default/app.py
antonymilne: I think yes. More generally though I think we're getting to the point where we need to figure out what the demo is for, since we have a few different things we're interested in:
1. Trying out features while we develop stuff like you do here 
2. The nice gapminder example
3. The public demo
4. A showcase of all the different components

At the moment these are all served by the same `default/app.py` but that's probably not good in the longer term and we need some new example dashboards. We should also re-consider whether we want the yaml/json/dict versions and what they should contain.

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -192,12 +200,12 @@ def _get_modified_page_charts(
...
-            **parameterized_config[target],
-        ).update_layout(template="vizro_dark" if ctd_theme["value"] else "vizro_light")
-        for target in targets
-    }
+    outputs = {}
+    for target in targets:
+        outputs[target] = model_manager[target](  # type: ignore[operator]
+            data_frame=filtered_data[target], **parameterized_config[target]
+        )
+        if hasattr(outputs[target], "update_layout"):
```
antonymilne, line 208: Not sure I like this, but it's probably ok, beyond the scope of this PR anyway and is probably just I don't yet understand how the themes and actions currently work...

But in short, why do we need to `update_layout` here as well as the `_update_theme` callback you moved to `Graph`?

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -0,0 +1,83 @@
...
+        if arg_name == "type":
+            return self.type
+        return self.table[arg_name]
+
+    @_log_call
+    def build(self):
+        data = data_manager._get_component_data(self.id)  # type: ignore
+        kwargs = self.table._arguments.copy()
+        kwargs.pop("data_frame", None)
+        return html.Div(self.table._function(data_frame=data, **kwargs), id=self.id)
```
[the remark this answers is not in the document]
antonymilne: In fact could this be the reason for the apparent bug with the captured callable?

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -113,17 +113,25 @@ def _update_nested_graph_properties(graph_config: Dict[str, Any], dot_separated_
...
+def _get_parametrized_config(
+    targets: List[ModelID], parameters: List[CallbackTriggerDict]
+) -> Dict[ModelID, Dict[str, Any]]:
     parameterized_config = {}
     for target in targets:
         # TODO - avoid calling _captured_callable. Once we have done this we can remove _arguments from
         #  CapturedCallable entirely.
-        graph_config = deepcopy(model_manager[target].figure._arguments)  # type: ignore[index, attr-defined]
+        graph_config = (
+            deepcopy(model_manager[target].figure._arguments)  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
antonymilne: Ah now I realise these are not `_output_property` but in fact refer to the names of the fields that contain the captured callable.

Shall we have another field in `Graph` and `Table` that sets `_callable_component`?

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -0,0 +1,83 @@
...
+
+    Args:
+        type (Literal["table"]): Defaults to `"table"`.
+        table (CapturedCallable): Table like object to be displayed. Current choices include:
+            [`dash_table.DataTable`](https://dash.plotly.com/datatable).
+        actions (List[Action]): See [`Action`][vizro.models.Action]. Defaults to `[]`.
+    """
+
+    type: Literal["table"] = "table"
+    table: CapturedCallable = Field(..., description="Table to be visualized on dashboard")
```
antonymilne, line 27: This field will need an `import_path` when we've worked out where to put the table function(s).

antonymilne:
@maxschulz-COL @huong-li-nguyen @petar-qb 

[178559e](https://github.com/mckinsey/vizro/pull/114/commits/178559eec1260e2161b66714ae047425f710635c) is just a very rough and hacky demonstration of something, not to be included in this PR (I reverted it in the next commit).

A good way to check that all the code that needed to be made general is fully generalised is to try and add another component other than `Graph` and `Table` and see if it works. So I wanted to see what it would take to adapt the `Card.text` into a model that is reactive to filters and parameters in the same way that `Graph.figure` and `Table.table` are.

Why is this of interest?
1. It shows the bits of code that are not yet fully general as they are hardcoded to handle only `Graph` and `Table`, since I needed to adapt them to also handle `Card`. I could only see one such case which I spotted before: https://github.com/mckinsey/vizro/pull/114#discussion_r1364525910
2. It demonstrates a possible future route where we could make more stuff reactive if we want to do so. This was always meant to be possible (one of the reasons for using a general `CapturedCallable` rather than something hardcoded to only handle graphs) but I didn't actually show anyone yet.

Two questions I'm interested in now:
1. Does the same thing work already if I try to use `Card.text` as input? e.g. by setting `_input_property = "n_clicks"` and seeing if I can use the card as an input object just by clicking on it. Silly example obviously, but would again demonstrate whether the current solution is fully general or not
2. Without making these changes to the `Card`, how would I achieve the same behave as in my demo just using a custom action?

image

antonymilne:
Just from quickly playing around with the the example dashboard here, I prefer the AG Grid. I wonder whether we should be recommending that instead of Dash data table as the default table for people to use in the (I guess) most common case that someone just wants to draw a nice table?
* for small screen sizes the table has its own little scroll bar rather than stretching out the whole screen
* I can rearrange column order by drag and drop
* the active row is highlighted
* the style just feels a bit more modern and slicker somehow (though dark theme is obviously not usable right now)

I'd guess some of these you can probably also achieve with Dash data table by using appropriate styling parameters or arguments though?

These are just my impressions from playing around for a few seconds though, so don't take them too seriously - it's the first time I've used both sorts of table, so I don't know what each is capable of or how easily we can get all the features we want out of each.

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
```diff
@@ -104,13 +104,16 @@ def _get_inputs_of_chart_interactions(
         page=_get_triggered_page(action_id=action_id),
         action_function=action_function,
     )
-    return [
-        State(
-            component_id=_get_triggered_model(action_id=ModelID(str(action.id))).id,
-            component_property="clickData",
+    inputs = []
+    for action in chart_interactions_on_page:
```
antonymilne, line 108: I think there's quite a bit of renaming we should do since lots of stuff refers explicitly to `graph` or `chart` or the like.

Not sure what the right word is for "component that is reactive to filters and parameters" - maybe just `component` if that's not too ambiguous.

vizro-core/src/vizro/actions/_actions_utils.py
petar-qb: Although, the `_update_theme` callback (that is moved to the Graph) is triggered when users change the theme on the toggle button, we still need to apply `update_layout` to graph when the filter/parameter/filter_interaction action is triggered in purpose to apply the currently selected theme (as template) to the graph that will be returned as a response. 

Try to comment out the following code:
```
if hasattr(outputs[target], "update_layout"):
    outputs[target].update_layout(template="vizro_dark" if ctd_theme["value"] else "vizro_light")
```
then open the app, select light theme (on toggle button), and apply the filter which has Graph target. The graph will be presented in a dark theme, in contrast to the rest of the page, which is in a light theme.

.pre-commit-config.yaml
```diff
@@ -41,7 +41,7 @@ repos:
         args: [--autofix]
 
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.261
+    rev: v0.0.261 #TODO: upgrade to latest version
```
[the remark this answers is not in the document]
maxschulz-COL: Ruff `0.1.0` was released three days ago, so thought we might do it for this one, but we can also do a sweep of all linters if you prefer

vizro-core/src/vizro/models/_components/form/checklist.py
```diff
@@ -27,6 +27,9 @@ class Checklist(VizroBaseModel):
     title: Optional[str] = Field(None, description="Title to be displayed")
     actions: List[Action] = []
 
+    # Component properties for actions and interactions
+    _input_property: str = PrivateAttr("value")
+
```
petar-qb, line 32: There's a big correlation between  `_input_property` and `_action_validator_factory`.  

Just to remind that:
- `_action_validator_factory  argument` ("value" in this scenario) represents a default property when the model object is the `trigger` of the actions chain. 
- `_input_property  argument` ("value" in this scenario) represents a default property when the model object is the `state` of some action.

vizro-core/src/vizro/charts/tables/__init__.py
```diff
@@ -0,0 +1,4 @@
+from vizro.charts.tables.dash_table import dash_data_table
```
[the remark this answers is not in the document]
maxschulz-COL: I was thinking more about what the end state may look like. Currently this is our first so-called "custom chart" and it happens to be the table, however there may be other custom charts (e.g. waterfall, MACC, network) in the future, that may be quite nice to be imported from `vizro.charts`.

Given that the table behaves in exactly the same way (its a decorated function that you insert into a model argument) I felt that it fitted best in the place where I think these other charts may sit in the future. And ultimately a table is also a chart in a wider sense, just not a very visual one, so I liked it best like that

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -37,9 +38,15 @@ class Graph(VizroBaseModel):
...
 
     type: Literal["graph"] = "graph"
-    figure: CapturedCallable = Field(..., import_path=px)
+    figure: CapturedCallable = Field(
+        ..., import_path=px
+    )  # TODO: needs to be refactored so plotly-independent or extendable - No, as within the boundaries of Graph model
     actions: List[Action] = []
 
+    # Component properties for actions and interactions
+    _input_property: str = PrivateAttr("clickData")
```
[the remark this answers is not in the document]
petar-qb: There's a big correlation between _input_property and _action_validator_factory.

Just to remind that:

_action_validator_factory  argument ("value" in this scenario) represents a default property when the model object is the trigger of the actions chain.
_input_property  argument ("value" in this scenario) represents a default property when the model object is the state of some action.

We definitely need to refine this implementation especially considering that there are cases where `_input_property` doesn't make sense as a single value. Explained in another PR [here](https://github.com/mckinsey/vizro/pull/119/files#r1365118530).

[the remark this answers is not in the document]
maxschulz-COL: Good question, I thought about this as well, and I think the answer is mostly, but not always. You could argue that there can be different triggers to inputs if I am not mistaken, but @petar-qb may have a clearer view on this!

vizro-core/pyproject.toml
```diff
@@ -64,3 +64,6 @@ skip_covered = true
 branch = true
 parallel = true
 source_pkgs = ["vizro"]
+
+[tool.hatch.metadata]
+allow-direct-references = true
```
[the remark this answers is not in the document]
maxschulz-COL: This is a setting I took over and wanted to discuss in the PR. @huong-li-nguyen @chiara-sophie : any reason this was added initially?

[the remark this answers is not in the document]
huong-li-nguyen: You can delete it, it was just to install local whl files to test custom js and d3.js components :)

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -109,3 +119,14 @@ def build(self):
...
+    def _update_graph_theme(self):
+        @callback(
+            Output(self.id, "figure", allow_duplicate=True),
+            Input("theme_selector", "on"),
+            prevent_initial_call="initial_duplicate",
+        )
+        def update_graph_theme(theme_selector_on: bool):
+            patched_figure = Patch()
+            patched_figure["layout"]["template"] = themes.dark if theme_selector_on else themes.light
+            return patched_figure
```
petar-qb, line 132: Just to remind you that this means that one http request will be sent to the server `per' Graph object on the page (so it may happen that the response is not rendered at the same time for all graphs). 
In the callback implementation we had in the Page model, there was only one request sent to the server, no matter how many Graphs were on the page.

vizro-core/src/vizro/charts/tables/dash_table.py
maxschulz-COL: Yes, do not have a very strong opinion, but of the top of my hat:
- `table` (very general, and may collide with any future other table additions; really makes the repetition in words stand out, almost feels like precluding any other objects to be inserted)
- `data_table` (similar reasoning as above + somewhat refers to dash (`dash_table.DataTable`), but only very lightly)
- `dash_data_table` or `dash_table` (explicitly refers to dash, which may be useful for guidance, but could feel unnecessarily long)

My gut feeling is probably `dash_table` actually

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -0,0 +1,83 @@
...
+    actions: List[Action] = []
+
+    # Component properties for actions and interactions
+    _input_property: str = PrivateAttr("active_cell")
```
petar-qb, line 31: I removed the `_input_property` in new [PR](https://github.com/mckinsey/vizro/pull/119) because we need to propagate two different table properties if the table is the source of the filter_interaction. Explained in that PR [here](https://github.com/mckinsey/vizro/pull/119/files#r1365118530).

.pre-commit-config.yaml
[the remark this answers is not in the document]
antonymilne: No strong feelings - I'm not too bothered about keeping up to date with linters.

maxschulz-COL:
> [178559e](https://github.com/mckinsey/vizro/pull/114/commits/178559eec1260e2161b66714ae047425f710635c) is just a very rough and hacky demonstration of something, not to be included in this PR (I reverted it in the next commit). image

Great idea and cool demo! It shows that we are slowly getting there!

> Two questions I'm interested in now:
> 
> 1. Does the same thing work already if I try to use `Card.text` as input? e.g. by setting `_input_property = "n_clicks"` and seeing if I can use the card as an input object just by clicking on it. Silly example obviously, but would again demonstrate whether the current solution is fully general or not
> 2. Without making these changes to the `Card`, how would I achieve the same behave as in my demo just using a custom action?

Would need to take some time to test, but I think if I understand custom actions properly, no.2  should be fairly straightforward creating a custom action taking the filter as input, doing the sum, and then adding the card to the `output` of the action model. @petar-qb can probably tell at a glance if its possible

> Edit: just realised there's actually another thing I had to do to get it working - specify the card's id in `targets`. I'm not sure why this already worked for the `Table` actually but not `Card`? Where is the code that works out all the components on the page if `targets` is not set?

I think this has to do with the fact that the filter only targets component that have registered data (see [here](https://github.com/mckinsey/vizro/blob/840d1a41c4c074d3dd626e36b431f71bc44d23b4/vizro-core/src/vizro/managers/_data_manager.py#L77)). Potentially something we can improve as well.

vizro-core/src/vizro/models/_components/graph.py
maxschulz-COL: @huong-li-nguyen This may be relevant for you in the tabs considerations. So it looks like moving this from page causes and increase in server requests

vizro-core/src/vizro/models/_components/table.py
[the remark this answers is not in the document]
maxschulz-COL: This was indeed weird, fixed it :)

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -0,0 +1,82 @@
...
+    def __getitem__(self, arg_name: str):
+        # pydantic discriminated union validation seems to try Table["type"], which throws an error unless we
+        # explicitly redirect it to the correct attribute.
+        if arg_name == "type":
+            return self.type
+        return self.table[arg_name]
+
+    @_log_call
+    def build(self):
+        return html.Div(dash_table.DataTable(pd.DataFrame().to_dict('records'), []), id=self.id)
```
petar-qb, line 82: This looks weird to me. How can we support the AgGrid now? Isn't that we need to call `self.table._function(...)` in the build method?

petar-qb: Ah I see now. This now follows the `figure=create_empty_fig("")` approach in the `vm.Graph.build` method and it really makes sense 👍

vizro-core/src/vizro/models/types.py
```diff
@@ -43,6 +43,11 @@ def __init__(self, function, /, *args, **kwargs):
         """
         self.__function = function
         self.__bound_arguments = inspect.signature(function).bind_partial(*args, **kwargs)
+        # Below is required as otherwise kwargs provided to captured functions will sit under the key "kwargs"
+        kwargs_to_unfold = self.__bound_arguments.arguments.get("kwargs")
+        if kwargs_to_unfold:
+            self.__bound_arguments.arguments.update(kwargs_to_unfold)
+            del self.__bound_arguments.arguments["kwargs"]
```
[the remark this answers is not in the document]
maxschulz-COL: Did some further investigation and got to the issue of it. Check out: https://github.com/mckinsey/vizro/pull/114/commits/2a5db2b3d9d320f14f8fdb3df726650a0d04b163, where it shows that the problem is having the **kwargs argument.

Unsure how to fix this in any other way though...

vizro-core/src/vizro/models/_components/graph.py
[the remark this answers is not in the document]
petar-qb: So as I said:
> `_action_validator_factory argument` represents a default property when the model object is the `trigger` of the actions chain.
And `_input_property` represents a default property when the model object is the `State` of some action.

Also, we can say that `_output_property` represents a default property when the model object is the `Output` of some action.

The problem with the `_input_property` and `_output_property` approach is that default property `state` and `output` sometimes is not a single value like [here](https://github.com/mckinsey/vizro/pull/119/files#r1365118530) and it also could potentially depend on the concrete predefined action implementation.

The `"not a single value"` (in "_input_property" / "_output_property") problem could be solved by moving dependency of State/Output creation from `_callback_mapping_utils` to additional model methods. 
So for example, instead of having the following code inside the `_get_action_callback_outputs`:

```
    return {
        target: Output(
            component_id=target,
            component_property=model_manager[target]._output_property,  # type: ignore[attr-defined]
            allow_duplicate=True,
        )
        for target in targets
```

We can define `_get_model_action_callback_output` method (and _get_model_action_callback_input) inside the model in the following way:

```
# IMPORTANT: Inside this method we can return multiple outputs if we need.
def _get_model_action_callback_output(self):
    return Output(
            component_id=self.id,
            component_property="value"
            allow_duplicate=True,
        )
``` 

and call it inside the `_callback_mapping_utils` function `_get_action_callback_outputs` in a following way:

```
    return {
        target: target._get_model_action_callback_output()
        for target in targets
```

This still doesn't solve the problem if the inputs/outputs depend on the concrete predefined action implementation which could happen potentially. 

--- 

So what do you think, what's the correct approach?

vizro-core/src/vizro/models/_components/table.py
maxschulz-COL: Yes although there is still a weird behaviour I couldn't immediately figure out. When there are no controls (unlike the `Graph`) it doesn't load the table right. I suppose it doesn't trigger the on page load action?

vizro-core/src/vizro/actions/_actions_utils.py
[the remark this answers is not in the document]
maxschulz-COL: Yes can do, let me try that

vizro-core/src/vizro/models/_components/graph.py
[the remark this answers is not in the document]
antonymilne: Thanks for the explanation @petar-qb, it makes sense 👍 I don't have an immediate answer to this - will need to spend some time familiarising myself with the actions code again and thinking about the new possibilities that tables bring. And then definitely let's discuss it.

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Component properties for actions and interactions
+    _input_property: str = PrivateAttr("active_cell")
+    _output_property: str = PrivateAttr("children")
+
+    # validator
+    set_actions = _action_validator_factory("active_cell")  # type: ignore[pydantic-field]
+
+    @validator("table")
+    def process_component_data_frame(cls, table, values):
```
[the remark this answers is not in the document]
antonymilne: Yeah, let's put this into a reusable validator in `_components_utils.py`

vizro-core/src/vizro/charts/tables/dash_table.py
antonymilne: The two relevant table implementations at the moment would be `dash.dash_table.DataTable` and
`dash_ag_grid.AgGrid` so my vote on function names would be `data_table` and `ag_grid`.

maxschulz-COL: I'll ask him

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
antonymilne: I largely agree with this but not sure about the use of the word `chart = component with data_frame = component that responds to filters = graph + table`. What you say is consistent with @maxschulz-COL's view [here](https://github.com/mckinsey/vizro/pull/114#discussion_r1365336478) and is growing on me as an idea.

The reason for my hesitation is that I'm not sure why we shouldn't have all `components` reactive to filters. Like in my demo here, what happened if at some point wanted to make `Card` reactive to filters. Would it become a sort of `chart`? Maybe that would make sense if it is thought of as a chart if it behaves like that and depends on a dataframe. Or maybe this doesn't make sense as an idea so is irrelevant anyway...

vizro-core/src/vizro/actions/_actions_utils.py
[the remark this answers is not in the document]
maxschulz-COL: Done

vizro-core/src/vizro/models/_components/table.py
petar-qb: `_on_page_load` targets are all components returned from the `_get_components_with_data` function (so tables should also be included). 

But, the _on_page_load action will be created in the `Page.pre_build` only:
`if any(isinstance(component, Graph) for component in self.components):`

So we probably need to adjust this if statement so that it passes even if there are only table components on the page.

vizro-core/src/vizro/models/_components/graph.py
huong-li-nguyen: Ah, good point! Is there any time lag visible right now @maxschulz-COL? Then let's move it back, but the logic of the for-loop might have to change then because as soon as we start introducing sub-pages and tiles it will have to look for any Graph objects on page. But don't change the logic now, this should be done in a separate PR when we start introducing other container models

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
antonymilne: FYI from a model point of view the difference between `Card` compared to `Table`/`Graph` should I think be that the latter expose a `__call__` method which delegates to a field that has a `CapturedCallable`. So the terminology you're suggesting is basically that `chart` is something that is a callable component, which is probably ok.

vizro-core/src/vizro/models/_components/graph.py
petar-qb: It's quite hard to see any lag while the app is hosted locally, so I also recommend returning the callback back to the Page model.

antonymilne:
> > Edit: just realised there's actually another thing I had to do to get it working - specify the card's id in `targets`. I'm not sure why this already worked for the `Table` actually but not `Card`? Where is the code that works out all the components on the page if `targets` is not set?
> 
> I think this has to do with the fact that the filter only targets component that have registered data (see [here](https://github.com/mckinsey/vizro/blob/840d1a41c4c074d3dd626e36b431f71bc44d23b4/vizro-core/src/vizro/managers/_data_manager.py#L77)). Potentially something we can improve as well.

Ah I was wondering what that that method was for and whether we really needed it 😀 Would it make sense to do this check as `callable(model)` instead? Following my idea in https://github.com/mckinsey/vizro/pull/114#discussion_r1366728543, basically we would be saying model with registered data = model that can be reactive to filter = model with `__call__` method. And we might call all such models "charts".

Probably not worth making the change in this PR just to keep scope limited but let's bear it in mind for the discussions on how to name things are where the table functions should live.

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -120,11 +120,7 @@ def _get_parametrized_config(
     for target in targets:
         # TODO - avoid calling _captured_callable. Once we have done this we can remove _arguments from
         #  CapturedCallable entirely.
-        graph_config = (
-            deepcopy(model_manager[target].figure._arguments)  # type: ignore[attr-defined]
-            if hasattr(model_manager[target], "figure")
-            else deepcopy(model_manager[target].table._arguments)  # type: ignore[attr-defined]
-        )
+        graph_config = deepcopy(model_manager[target]._callable_component._arguments)  # type: ignore[attr-defined]
```
antonymilne, line 123: Sorry I wasn't clear here but what I meant was the `_callable_component: str = PrivateAttr("figure")` and then here it would be `getattr`. The `_callable_component` is just telling us what the relevant field is, not a replacement or copy of that field.

vizro-core/src/vizro/models/_components/table.py
antonymilne: This sounds a case where we might want to replace `isinstance(component, Graph)` with `callable(component)` if we like the idea https://github.com/mckinsey/vizro/pull/114#issuecomment-1772454425?

vizro-core/src/vizro/actions/_actions_utils.py
antonymilne: Thanks for the explanation! I'd like to have a think about whether there's a nicer way to do this but it seems fine for now 👍

vizro-core/src/vizro/actions/_actions_utils.py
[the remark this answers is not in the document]
antonymilne: Sorry I wasn't clear here but what I meant was `_callable_component: str = PrivateAttr("figure")` and then here it would use `getattr` to get the relevant field. The `_callable_component` is just telling us what the relevant field is, not a replacement or copy of that field.

vizro-core/src/vizro/models/_components/table.py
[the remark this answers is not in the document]
maxschulz-COL: Done

vizro-core/src/vizro/models/_components/graph.py
maxschulz-COL: Moved this back

vizro-core/src/vizro/models/_components/graph.py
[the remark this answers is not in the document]
maxschulz-COL: Scheduled a meeting to discuss this

vizro-core/src/vizro/models/_components/table.py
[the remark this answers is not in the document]
maxschulz-COL: Adressed through introduction of `_callable_component`

vizro-core/src/vizro/models/_components/table.py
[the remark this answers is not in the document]
maxschulz-COL: Done

vizro-core/src/vizro/actions/_actions_utils.py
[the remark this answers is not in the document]
maxschulz-COL: Ok gotcha, implemented this new version, and used it in the `_page` model to check for attribute

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
maxschulz-COL: Let's discuss this as well in our meeting on Monday

vizro-core/src/vizro/models/_components/table.py
maxschulz-COL: I already removed it in the PR, I left the `_output_property` as it makes more sense. We may also change that one though in the future

vizro-core/src/vizro/models/_components/table.py
maxschulz-COL: Implemented a slight alternative that was quick to do, which follows the idea that we now have `_callable_component`. Can refactor further in the future

vizro-core/src/vizro/charts/tables/dash_table.py
maxschulz-COL: Joe's vote goes for `dash_data_table` and if we have it in the future `dash_ag_grid`

vizro-core/src/vizro/charts/tables/__init__.py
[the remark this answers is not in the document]
maxschulz-COL: Decided on `vizro.graphs` and `vizro.tables`

vizro-core/src/vizro/models/_components/graph.py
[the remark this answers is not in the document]
maxschulz-COL: Outcome of discussion is that:
- we keep things as they are in this PR
- abstraction logic should lie with functions, as we do not want a multitude of models
- refactor later, as this is not public API

vizro-core/src/vizro/models/_components/table.py
maxschulz-COL: Done

vizro-core/src/vizro/models/_components/table.py
maxschulz-COL: Changed this to `hasattr(...,"figure")` given that we want to unify Graph and Table to carry this attritbute now

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
maxschulz-COL: For completeness, adding:

For now then we say that Figure = something that reacts to filters/parameters = model with figure field. In future we can extend this to further components if needed.

vizro-core/docs/pages/user_guides/graph.md
maxschulz-COL, line 1: This is split out as otherwise the guides were getting to long

vizro-core/examples/default/app.py
maxschulz-COL, line 1: This needs to be reverted

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class Table(VizroBaseModel):
+    """Wrapper for table components to visualize in dashboard.
+
+    Args:
+        type (Literal["table"]): Defaults to `"table"`.
+        figure (CapturedCallable): Table like object to be displayed. Current choices include:
+            [`dash_table.DataTable`](https://dash.plotly.com/datatable).
+        actions (List[Action]): See [`Action`][vizro.models.Action]. Defaults to `[]`.
```
huong-li-nguyen, line 26: We need to expose a `title` argument - distinct to the Graph model the dash datatable doesn't have a native title argument embedded. So we would probably have to expose it here and in the return it should be something like below: 

`return html.Div([html.H3(self.title) if self.title else None, dash_table.DataTable(pd.DataFrame().to_dict("records")], []), className="table-container", id=self.id)`

I think it's H3, but double-check, it should be of the same size as the chart title in Graph

vizro-core/src/vizro/static/css/table.css
```diff
@@ -0,0 +1,32 @@
...
+}
+
+div.right_side
+  .dash-table-container
+  .dash-spreadsheet-container
+  .dash-spreadsheet-inner
+  td,
+.dash-table-container .dash-spreadsheet-container .dash-spreadsheet-inner th {
+  padding: 0px 4px;
+}
```
huong-li-nguyen, line 32: I just did a quick refinement, but it would be great if you can create a proper ticket to style the table properly as we're currently deviating from the initial design and I think the design eventually also has to be updated because the proposed spacing look quite large on the dashboard (I didn't implement the proposed spacing). 

```suggestion
div.right_side .dash-table-container .dash-spreadsheet-container .dash-spreadsheet-inner td .dash-cell-value.unfocused {
    color: var(--text-primary);
    font-size: 14px;
}

div.right_side .dash-table-container .dash-spreadsheet-container .dash-spreadsheet-inner .column-header-name {
    color: var(--text-primary); /* Should be text-secondary when hover color is implemented */
    font-size: 14px;
}

div.right_side .table-container {
 padding: 16px;
}
```

vizro-core/src/vizro/tables/dash_table.py
```diff
@@ -0,0 +1,12 @@
...
+from dash import dash_table
+
+from vizro.models.types import capture
+
+
+@capture("table")
+def dash_data_table(data_frame: pd.DataFrame, **kwargs):
+    """Standard `dash_table.DataTable`."""
+    kwargs.setdefault("columns", [{"name": i, "id": i} for i in data_frame.columns])
+    return dash_table.DataTable(data=data_frame.to_dict("records"), **kwargs)
```
huong-li-nguyen, line 12: @antonymilne  @maxschulz-COL - preferably I would have kept our CSS styling only in the CSS files, but unfortunately that's not possible. The dash datatable comes with inline styling, so it always has precedence over anything I could eventually target via CSS selectors in the assets folder. The only way how to overwrite the native inline styling is to add an` ! important` tag to the CSS properties which is not feasible, as otherwise the users cannot overwrite it with `style_header` and `style_data `anymore. This is currently the only way I think we can ship the dash data table with our default styling. 

Unfortunately, the default styling would disappear as soon as the users use `style_header` and `style_data`, so if they want to keep the styling they would have to copy/paste dict entries from here. @maxschulz-COL shall we add that to the docs?

@maxschulz-COL - can you also double-check with Joe if he's fine with the default styling and functionality of the table right now? Or if we should turn on more properties by default? Compared to our internal version e.g. it doesn't have sorting and other properties natively turned on, but we could do that e.g. via 

```
kwargs.setdefault("sort_action", "native")
kwargs.setdefault("sort_mode", "single")
```

The `kwargs.setdefault("style_as_list_view", True)` removes the vertical grid lines, which makes it look a bit cleaner. It will then also be closer the proposed design.

```suggestion
    kwargs.setdefault("columns", [{"name": i, "id": i} for i in data_frame.columns])
    kwargs.setdefault("style_as_list_view", True)
    kwargs.setdefault("style_data", {"border_bottom": "1px solid var(--border-subtle-alpha-01)", "height": "40px"})
    kwargs.setdefault(
        "style_header",
        {
            "border_bottom": "1px solid var(--state-overlays-selected-hover)",
            "border_top": "1px solid var(--main-container-bg-color)",
            "height": "32px",
        },
    )
    return dash_table.DataTable(data=data_frame.to_dict("records"), **kwargs)
```

vizro-core/tests/unit/vizro/models/test_page.py
```diff
@@ -101,7 +101,7 @@ def test_valid_component_types(self, standard_px_chart):
         [vm.Checklist(), vm.Dropdown(), vm.RadioItems(), vm.RangeSlider(), vm.Slider()],
     )
     def test_invalid_component_types(self, test_component):
-        with pytest.raises(ValidationError, match=re.escape("(allowed values: 'button', 'card', 'graph')")):
+        with pytest.raises(ValidationError, match=re.escape("allowed values: ")):
```
huong-li-nguyen, line 104: Do we not want to test for the actual components?

huong-li-nguyen (changes requested):
Thanks for pushing through this! Looks like great progress 🚀 

I didn't review the code in detail as @antonymilne and @petar-qb have done a great job on this. I was mostly looking at visual refinements and defaults. Have some quick proposal to get closer to the actual proposed design (see screenshot below), but it would be great if you can create a proper ticket because currently we deviate from that.

Screenshot 2023-10-24 at 10 32 38

Could you also double-check w/ J if he's okay with the current default functionality/design or if certain functionalities should be turned on by default? Compared to our internal version e.g. it doesn't have sorting/column merging turned on by default. For me that's fine if this should be configured by the user.

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -0,0 +1,55 @@
...
+    def __getitem__(self, arg_name: str):
+        # pydantic discriminated union validation seems to try Table["type"], which throws an error unless we
+        # explicitly redirect it to the correct attribute.
+        if arg_name == "type":
+            return self.type
+        return self.figure[arg_name]
+
+    @_log_call
+    def build(self):
+        return html.Div(dash_table.DataTable(pd.DataFrame().to_dict("records"), []), id=self.id)
```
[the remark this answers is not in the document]
petar-qb: > You probably already discussed this, but I assume there is no way we can insert the self.id into the id of the actual dash datatable right? Currently if we want to use any interactions, we don't reference this self.id, but the id provided in the datatable.

You're right. 

>  Is there any use case where we would reference the self.id instead of the datatable id?

Since `self.id` is used here as the ID of the outer html.Div object, we reference it while calculating Output of the some predefined actions with "children" property.

> We need the className later 

Like it!

vizro-core/src/vizro/actions/_parameter_action.py
```diff
@@ -23,10 +24,10 @@ def _parameter(targets: List[str], **inputs: Dict[str, Any]) -> Dict[str, Any]:
     Returns:
         Dict mapping target component ids to modified charts/components e.g. {'my_scatter': Figure({})}
     """
-    targets = [target.split(".")[0] for target in targets]
+    target_ids: List[ModelID] = [target.split(".")[0] for target in targets]  # type: ignore[misc]
 
     return _get_modified_page_charts(
-        targets=targets,
+        targets=target_ids,
```
petar-qb, line 30: Should we revert these changes here or refactor them in other predefined actions too?

vizro-core/tests/unit/vizro/models/test_page.py
petar-qb, line 104: Do we also want to test vm.Table inside the `test_valid_component_types` test?

vizro-core/src/vizro/models/_components/table.py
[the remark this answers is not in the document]
petar-qb: Do we need to wrap an empty `dash_data.Datatable` into the `dcc.Loading`, similarly as we do with the `vm.Graph` object inside the `Graph.build` method?

vizro-core/src/vizro/models/types.py
petar-qb, line 50: Will the changes from https://github.com/mckinsey/vizro/pull/121 fix this?

petar-qb (approved):
Looks good 🤖! Really great job @maxschulz-COL.

petar-qb:
One more question. Do we need to test vm.Table as filter/parameter/export_data/on_page_load/filter_interaction `target` as a part of this PR?

vizro-core/examples/default/app.py
antonymilne: @maxschulz-COL this probably got lost in all the comments. Maybe worth making a new issue out of it?

vizro-core/src/vizro/models/_components/table.py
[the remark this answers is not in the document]
huong-li-nguyen: Hmm.. it depends - does it require a lot of time to load a table ? I would assume that this is actually quite fast as it's just displaying data, rather than compiling a trace. 

I also remember that in the old repository, we had a bug with the datatable flickering, so we had to remove the dcc.Loading wrapper which fixed it. I don't know if this is still the case. 

You can try it out and see if it works and if not, just leave it out

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
```diff
@@ -145,6 +145,7 @@ def _get_action_callback_inputs(action_id: ModelID) -> Dict[str, List[State]]:
 def _get_action_callback_outputs(action_id: ModelID) -> Dict[str, Output]:
     """Creates mapping of target names and their Output."""
     action_function = model_manager[action_id].function._function  # type: ignore[attr-defined]
+    # TODO: here and above, could we fix mypy (and ideally have better code), by defining these attributes in the base?
```
antonymilne, line 148: The right solution here is to not define new attributes on the base but instead to get mypy to recognise that `model_manager[action_id]` is of type `Action` and hence has the `function` attribute.

Ideally `model_manager.__getitem__` would handle this itself, possibly with suitable use of a `cast`. If not then we can do the `cast` to `Action` at the point of consumption here to avoid needing mypy ignores.

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -12,15 +12,15 @@
 
 rd = random.Random(0)
 
+# Type used for model_manager IDs
```
antonymilne, line 15: This seems kind of redundant as a comment given how it's directly used below?
```suggestion
```

vizro-core/src/vizro/models/_components/_components_utils.py
```diff
@@ -0,0 +1,34 @@
+import logging
+
+from vizro.managers import data_manager
+
+logger = logging.getLogger(__name__)
+
+
+# Validator for reuse in other models to validate pages
```
antonymilne, line 8:
```suggestion
```

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -37,39 +38,17 @@ class Graph(VizroBaseModel):
     """
 
     type: Literal["graph"] = "graph"
-    figure: CapturedCallable = Field(..., import_path=px)
+    figure: CapturedCallable = Field(
+        ..., import_path=px
+    )  # TODO: needs to be refactored so plotly-independent or extendable - No, as within the boundaries of Graph model
```
antonymilne, line 43:
```suggestion
    figure: CapturedCallable = Field(..., import_path=px)
```

I believe this is resolved?

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -37,39 +38,17 @@ class Graph(VizroBaseModel):
...
-            data_manager[dataset_name] = data_frame
-        except ValueError as exc:
-            logger.debug(exc)
-
-        data_manager._add_component(values["id"], dataset_name)
-
-        # No need to keep the data in the captured function any more so remove it to save memory.
-        del figure["data_frame"]
-        return figure
+    _validate_callable = validator("figure", allow_reuse=True, always=True)(_process_callable_data_frame)
```
antonymilne, line 51:
```suggestion
    _validate_callable = validator("figure", allow_reuse=True)(_process_callable_data_frame)
```

This shouldn't be needed since `figure` is a mandatory field.

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -0,0 +1,55 @@
...
+    type: Literal["table"] = "table"
+    figure: CapturedCallable = Field(..., import_path=vt, description="Table to be visualized on dashboard")
+    actions: List[Action] = []
+
+    # Component properties for actions and interactions
+    _output_property: str = PrivateAttr("children")
+
+    # validator
+    set_actions = _action_validator_factory("active_cell")  # type: ignore[pydantic-field]
+    _validate_callable = validator("figure", allow_reuse=True, always=True)(_process_callable_data_frame)
```
antonymilne, line 38:
```suggestion
    _validate_callable = validator("figure", allow_reuse=True)(_process_callable_data_frame)
```

vizro-core/src/vizro/models/_components/table.py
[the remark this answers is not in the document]
antonymilne: 1. If the point here is to just generate an empty table then please could we do it in the simplest way possible e.g. `dash_table.DataTable()` or whatever the minimal call here is? No need for `pd.DataFrame()` etc.
2. Actually what is the effect of this existing at all? When do you see the empty table? If it's for page loading purposes do we even need the `dash_table.DataTable()` there at all rather than an empty div? 🤔

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -57,14 +58,14 @@ class Filter(VizroBaseModel):
...
-            the page that use `column`.
+        targets (List[ModelID]): Target component to be affected by filter. If none are given then target all components
+            on the page that use `column`.
         selector (Optional[SelectorType]): See [SelectorType][vizro.models.types.SelectorType]. Defaults to `None`.
     """
 
     type: Literal["filter"] = "filter"
     column: str = Field(..., description="Column of DataFrame to filter.")
-    targets: List[str] = Field(
+    targets: List[ModelID] = Field(
```
antonymilne, line 68: We discussed this before but not sure if you tried fixing the circular imports - I think it should be possible to do so in this case though. Either way, if `ModelID` appears in model docs then it should definitely live in `vizro.models.types` and not hidden away. Currently a user will see from the docs that the field requires a `ModelID` but not know what that is, which is actually worse than how things currently are where they know it's a string.

Let's do that as a separate PR though.

vizro-core/src/vizro/models/types.py
antonymilne: Yes, we should be able to revert this change here 🙂

vizro-core/src/vizro/models/types.py
```diff
@@ -213,8 +218,25 @@ def wrapped(*args, **kwargs):
                 return CapturedCallable(func, *args, **kwargs)
 
             return wrapped
+        elif self._mode == "table":
+            # The table component
```
antonymilne, line 222:
```suggestion
```

Crusade against redundant comments 😀

vizro-core/src/vizro/models/types.py
```diff
@@ -213,8 +218,25 @@ def wrapped(*args, **kwargs):
...
+            @functools.wraps(func)
+            def wrapped(*args, **kwargs):
+                if "data_frame" not in inspect.signature(func).parameters:
+                    raise ValueError(f"{func.__name__} must have data_frame argument to use capture('table').")
```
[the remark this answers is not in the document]
antonymilne: Yes - if you do `@capture("graph")` then it will raise the same error.

vizro-core/src/vizro/models/types.py
```diff
@@ -161,7 +166,7 @@ class capture:
     [custom charts](../user_guides/custom_charts.md).
     """
 
-    def __init__(self, mode: Literal["graph", "action"]):
+    def __init__(self, mode: Literal["graph", "action", "table"]):
```
antonymilne, line 169: This function needs quite a few other updates now that `table` is an option. If you just search the file for `"graph"` then you should see where docstrings etc. need to be updated.

vizro-core/src/vizro/tables/dash_table.py
antonymilne: Would it work if we put this in CSS with `!important` and then require a user who wants to override that to put their own CSS in an assets folder also with `!important`?  Which `!important` would win out? 😬 This would mean that users can't use the `style_header/data` arguments but would keep CSS out of this file. Overall it's probably a worse solution than the above?

We should be able to fiddle with the current solution so that users can modify subkeys without needing to copy and paste though, just by rethinking how to handle `kwargs` a bit.

antonymilne (commented):
Basically an approval in terms of the code, with a few minor comments 🙂 

Haven't yet reviewed docs or tests but will do so in a separate review.

vizro-core/src/vizro/models/types.py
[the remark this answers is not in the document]
maxschulz-COL: Hm not sure if I understand, but it is already in the `graph` version of it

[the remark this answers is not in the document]
huong-li-nguyen: Ah got it, just saw that this only runs when `self._mode=table`. I was wondering why this only mentions the table in the ValueError if the same actually accounts for the graph as well. But I see that we repeat the ValueError

vizro-core/src/vizro/models/_components/table.py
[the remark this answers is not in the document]
huong-li-nguyen: The dcc.Loading and the empty figure population are two different concepts - the dcc.Loading is just for providing feedback to the user that something is loading 😄  So it's this waiting visual that often is not visible because it's loading fast. So e.g. the dcc.Loading stops when a figure (empty or with data is loaded). 

The empty figure population is for optimizing the performance of actions. As far as I understood, the actions wait for a figure to be loaded to trigger an update of the figure with the default filters/parameters applied. So instead of having to wait for a chart with all data, we just render an empty chart and then the filters/parameters with default selection get applied to render the full chart. Normally you wouldn't see the empty table, unless the actions require lots of time to apply filters etc.

vizro-core/src/vizro/tables/dash_table.py
huong-li-nguyen: @antonymilne that would work, the users `! important` tag would have precedence as their files get read in last.

So see it very nicely here, table.css is coming from our static assets folder, images.css is from the user assets folder. 
Screenshot 2023-10-24 at 19 00 31

But yeah, this would mean that users can't use the `style_header/data` arguments to specify the attributes defined here. But maybe that's okay? Personally, I am fine with the solution of users having to specify these attributes in a CSS file with the `! important ` tag then. What do you guys think?

antonymilne: Thanks for checking @huong-li-nguyen! My current thinking is that this is actually a worse solution than setting inline styles because:
* `style_*` parameters are well documented on Dash and so would be good if they worked
* I'm not sure whether setting through CSS would even work correctly with [style priorities](https://dash.plotly.com/datatable/style#styles-priority) which I guess may require the `style_*` arguments
* it looks like AG Grid [has similar arguments](https://dash.plotly.com/dash-ag-grid/styling-rows#conditional-row-styles) which do things that can't be achieved that definitely can't be achieved through pure CSS (the alternating row colouring could be with `:nth-child` but I'm pretty sure the whole `styleConditions` thing can't be

Hence the only real drawbacks I think are:
1. it doesn't feel very elegant
2. "Unfortunately, the default styling would disappear as soon as the users use style_header and style_data, so if they want to keep the styling they would have to copy/paste dict entries from here"

The first of these is ok since there doesn't seem to be a better way. The second of these we can fix by settings defaults less destructively:

```
from collections import defaultdict
from collections.abc import Mapping

def _set_defaults_nested(supplied: Mapping, defaults: Mapping) -> dict:
    supplied = defaultdict(dict, supplied)
    for default_key, default_value in defaults.items():
        if isinstance(default_value, Mapping):
            supplied[default_key] = _set_defaults_nested(supplied[default_key], default_value)
        else:
            supplied.setdefault(default_key, default_value)
    return dict(supplied)

# use in table function as something like this
defaults = {
    "columns": [{"name": i, "id": i} for i in data_frame.columns],
    "style_as_list_view": True,
    "style_data": {"border_bottom": "1px solid var(--border-subtle-alpha-01)", "height": "40px"},
    "style_header": {
        "border_bottom": "1px solid var(--state-overlays-selected-hover)",
        "border_top": "1px solid var(--main-container-bg-color)",
        "height": "32px",
    },
}
kwargs = _set_defaults_nested(kwargs, defaults)
```

In theory this should work for arbitrarily deep defaults dictionaries, e.g.
```
defaults = {"a": {"b": {"c": 1, "d": 2}}}
_set_defaults_nested({"a": {"b": {"c": 5}}}, defaults)
# gives {'a': {'b': {'c': 5, 'd': 2}}}
```
... but I haven't tested it very thoroughly so do write some unit tests for it!

vizro-core/src/vizro/models/_components/table.py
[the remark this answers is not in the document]
antonymilne: Thank you for the explanation @huong-li-nguyen! 🙏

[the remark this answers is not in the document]
maxschulz-COL: 1. @antonymilne removed the `pd.DataFrame`, for some reason this didn't work last time I tried, good spot.
2. Thanks for answering this already @huong-li-nguyen !

vizro-core/src/vizro/tables/dash_table.py
huong-li-nguyen: Ahh, I like this a lot! And if they can't do want to overwrite these defaults in style_XXX, they can still use the important tag and CSS files. @maxschulz-COL - could you test that out? 🙏

vizro-core/src/vizro/models/_components/table.py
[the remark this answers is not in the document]
huong-li-nguyen: @maxschulz-COL , could you try out the dcc.Loading around the table wrapper then and see if it still works properly? If it doesn't, then just leave it out.

vizro-core/src/vizro/models/_components/graph.py
maxschulz-COL: Yes

vizro-core/tests/unit/vizro/models/test_page.py
maxschulz-COL: I had to change the test just because I added a component so found that too specific. Ideally it checks for the type of error message, but does not need to be adjusted just because you added a new component

vizro-core/src/vizro/actions/_parameter_action.py
maxschulz-COL: Specifically the change on line 30?

vizro-core/src/vizro/models/types.py
maxschulz-COL: Done

vizro-core/src/vizro/tables/dash_table.py
maxschulz-COL: Tested this with a mix of new parameters and an overwrite of an existing parameter:

```py
style_header={"backgroundColor": "rgb(30, 30, 30)", "fontWeight": "bold", "color": "white","height": "100px"},
```

The below is the result, which keeps the default formatting (borders), changes the overwritten (height) and add some new stylings. I think this is the most desirable result.

image

vizro-core/src/vizro/actions/_parameter_action.py
petar-qb: I thought you should leave the name "targets" instead of this name "target_ids", but I don't mind this either, so it doesn't matter and leave it like this.

vizro-core/src/vizro/static/css/table.css
```diff
@@ -0,0 +1,56 @@
...
+
+.table-title {
+  padding-bottom: 14px;
+}
+
+.table-container {
+  height: 100%;
+  overflow: auto;
+  padding: 12px;
+  padding-left: 18px;
```
huong-li-nguyen, line 54: Can you check with 20px? Sorry, I realised this is not written down somewhere, but all our margins, paddings, gaps, font sizes etc. should be multiples of 4 :) 

```suggestion
  padding-left: 20px;
```

vizro-core/tests/unit/vizro/models/test_page.py
maxschulz-COL: Tbh I think this test is not going to stop anyone adding another component in the allowed types. If the reviewer and contributor know about this restriction, they will forbid it even without this test, if not, they will change the test.

I added it back in, but I think it's not really our place to test automatically generated error messages as a reminder. If we really care about this, why not add a note behind the component type, which is much clearer and less annoying?

vizro-core/src/vizro/static/css/table.css
```diff
@@ -0,0 +1,56 @@
...
+  padding-bottom: 14px;
+}
+
+.table-container {
+  height: 100%;
+  width: 100%;
+  overflow: auto;
+  padding: 12px;
+  padding-left: 20px;
+}
```
huong-li-nguyen, line 56: Sorry, I know this is a bit tedious. Seems like we don't need the title `padding` after all (the gap between the graph title and the graph seem to be of the same gap as the table title and the table now). I also think we need to increase the `padding-left`, such that the table does align with the y-axis title:

Screenshot 2023-10-25 at 21 05 50

I kind of don't like that the graph and table title don't align, but I think this is our initial design on the charts. I'll create an issue for that such that we re-evaluate the padding of the graph title, but feel free to ignore this one right now.

```suggestion
.table-container {
  height: 100%;
  width: 100%;
  overflow: auto;
  padding: 12px;
  padding-left: 32px;
}
```

huong-li-nguyen (approved):
Looks great 👍 Thanks for fixing!

vizro-core/tests/unit/vizro/models/test_page.py
huong-li-nguyen: But this would require that the reviewer and contributor remember this restriction. 

It just requires adding the str type in there, which I think is okay? We can also leave this thread open - I am not strictly against it if the team prefers that. The comment next to the `ControlType` and `ComponentType` would also be okay for me, though I still prefer the tests as I think they serve as documentation as well. And there I prefer to be more explicit in showing what's allowed and not allowed.

vizro-core/docs/pages/user_guides/table.md
```diff
@@ -0,0 +1,213 @@
...
+
+        ```
+    === "Result"
+        [![Table2]][Table2]
+
+    [Table2]: ../../assets/user_guides/table/styled_table.png
+
+#### Custom Table
+
+Lorem Ispum
```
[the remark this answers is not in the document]
maxschulz-COL: Yes, I will add another examples today - sorry, mentioned in the PR description, but not very prominently

vizro-core/tests/unit/vizro/tables/test_dash_table.py
```diff
@@ -0,0 +1,22 @@
...
+    [
+        ({}, {"a": {"b": {"c": 1, "d": 2}}, "e": 3}),  # nothing supplied
+        ({"e": 10}, {"a": {"b": {"c": 1, "d": 2}}, "e": 10}),  # flat main key
+        ({"a": {"b": {"c": 11, "d": 12}}}, {"a": {"b": {"c": 11, "d": 12}}, "e": 3}),  # updated multiple nested keys
+        ({"a": {"b": {"c": 1, "d": {"f": 42}}}}, {"a": {"b": {"c": 1, "d": {"f": 42}}}, "e": 3}),  # add new dict
+        ({"a": {"b": {"c": 5}}}, {"a": {"b": {"c": 5, "d": 2}}, "e": 3}),  # arbitrary nesting
+    ],
+)
+def test_set_defaults_nested(default_dictionary, input, expected_output):
+    assert _set_defaults_nested(input, default_dictionary) == expected_output
```
petar-qb, line 22: What do you think, do we also need to add and support next combinations?
`({"a": 10}, {"a": 10, "e": 3})` or 
`({"a": {"b": 10}}, }, {"a": {"b": 10}},  "e": 3})`

petar-qb (approved):
This is fantastic work Max 🚀. I just left one question below. 

After reverting `app.py`, `dashboard.yaml`, `pyproject.toml` changes and after adding changelog file, I think the PR will be ready for merging.

vizro-core/src/vizro/models/types.py
petar-qb, line 1: Do we also need to adjust the following ValueError message?
Lines: 140-144

```
elif not isinstance(callable_config, dict):
    raise ValueError(
        "You must provide a valid CapturedCallable object. If you are using a plotly express figure, ensure "
        "that you are using `import vizro.plotly.express as px`. If you are using a custom figure or action, "
        "that your function uses the @capture decorator."
    )
```

vizro-core/src/vizro/tables/dash_table.py
```diff
@@ -0,0 +1,36 @@
...
+            supplied[default_key] = _set_defaults_nested(supplied[default_key], default_value)
+        else:
+            supplied.setdefault(default_key, default_value)
+    return dict(supplied)
+
+
+@capture("table")
+def dash_data_table(data_frame: pd.DataFrame, **kwargs):
+    """Standard `dash_table.DataTable`."""
+    defaults = {
```
antonymilne, line 25: Possibly we'll want to make this `defaults` publicly accessible also so that people can easily use it in their custom table functions, but no rush to do that yet.

vizro-core/tests/unit/vizro/tables/test_dash_table.py
```diff
@@ -0,0 +1,22 @@
...
+from vizro.tables.dash_table import _set_defaults_nested
+
+
+@pytest.fixture
+def default_dictionary():
+    return {"a": {"b": {"c": 1, "d": 2}}, "e": 3}
+
+
+@pytest.mark.parametrize(
+    "input, expected_output",
```
antonymilne, line 12:
```suggestion
    "input, expected",
```
I think this is our convention still?

vizro-core/tests/unit/vizro/tables/test_dash_table.py
antonymilne: Not very important I think but now that you've typed them out we might as well add them 😀

antonymilne (approved):
Very happy to approve this subject to some minor suggestions and reverting the changes to the demo app or whatever your intention is. I haven't tried out the latest incarnation of the code myself but looks like others have done so and checked the styling etc. 👍 

Really amazing work overall - I know this wasn't easy and had lots of iteration but I think it turned out great! Definitely one for the new Highlights section in our changelog ✨ Very nice work on the docs and tests also 👍

In future let's try to avoid such big PRs with 100s of comments, which I know is mainly my fault for leaving so many comments and digressions and asking things to be reworked 😅 Sometimes it's hard to avoid these things or they just happen by accident I know, and I'm as guilty as anyone here, but let's be mindful of it next time and maybe work through it in a different way that's more manageable for everyone 🙂

vizro-core/src/vizro/models/types.py
maxschulz-COL: What needs to be changed?

vizro-core/docs/pages/user_guides/table.md
```diff
@@ -0,0 +1,286 @@
...
+        import vizro.plotly.express as px
+        from vizro import Vizro
+        from vizro.tables import dash_data_table
+
+        df = px.data.gapminder().query("year == 2007")
+
+        page = vm.Page(
+            title="Example of a Dash DataTable",
+            components=[
+                vm.Table(id="table", title="Dash DataTable", figure=dash_data_table(data_frame=df)),
```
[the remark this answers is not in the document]
maxschulz-COL: Hm, I sometimes things it can be nice to be a bit explicit in examples, as this is often the starting point for modification. Someone might like to refer to the table id for Parameters eventually. I'll leave it for now

vizro-core/docs/pages/user_guides/table.md
```diff
@@ -0,0 +1,286 @@
...
+        from vizro import Vizro
+        from vizro.models.types import capture
+
+        df = px.data.gapminder().query("year == 2007")
+
+
+        @capture("table")
+        def my_custom_table(data_frame=None, chosen_columns: List[str] = None):
+            """Custom table."""
+            columns = [{"name": i, "id": i} for i in chosen_columns]
```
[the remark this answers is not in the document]
maxschulz-COL: It is not the most compelling example, but it's a simple one! Yes you can pass a columns argument, but you cannot pass an argument that needs to be computed and target it with a parameter unless I am mistaken?  Ie `columns = [{"name": i, "id": i} for i in chosen_columns]` - how do get `chosen_columns` affected by a parameter.

vizro-core/docs/pages/user_guides/table.md
```diff
@@ -0,0 +1,286 @@
...
+            defaults = {
+                "style_as_list_view": True,
+                "style_data": {"border_bottom": "1px solid var(--border-subtle-alpha-01)", "height": "40px"},
+                "style_header": {
+                    "border_bottom": "1px solid var(--state-overlays-selected-hover)",
+                    "border_top": "1px solid var(--main-container-bg-color)",
+                    "height": "32px",
+                },
+            }
```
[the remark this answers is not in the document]
maxschulz-COL: Agree! I was hesitant to do this, but otherwise the table would have looked out of sync with the other examples, so I decided to had it. The focus here is more on the custom argument, less on "generality"

vizro-core/src/vizro/models/types.py
petar-qb: Sorry for the lack of context. I thought it might be something like:
"If you are using dash Datatable figures, make sure you are using `from vizro.tables import dash_data_table`."
before the last sentence.

l0uden:
Dashboards in my tests can't start with this error:

```
    from ._components import Card, Graph, Table
  File "/home/circleci/.venv/vizro-qa/lib/python3.8/site-packages/vizro/models/_components/__init__.py", line 5, in <module>
TypeError: 'ABCMeta' object is not subscriptable
```

vizro-core/examples/default/app.py
maxschulz-COL: Ah I just saw you made one. Left this open until getting to it, but you beat me to it :)

vizro-core/src/vizro/models/_components/table.py
maxschulz-COL: I think it is really not, also in many other places. Let me create a ticket for that as it's not something you wanna rush

vizro-core/src/vizro/models/_controls/filter.py
maxschulz-COL: Will create an issue for it

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -0,0 +1,63 @@
...
+    def __getitem__(self, arg_name: str):
+        # pydantic discriminated union validation seems to try Table["type"], which throws an error unless we
+        # explicitly redirect it to the correct attribute.
+        if arg_name == "type":
+            return self.type
+        return self.figure[arg_name]
+
+    @_log_call
+    def build(self):
+        return html.Div(
```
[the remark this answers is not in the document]
maxschulz-COL: Implemented now!

PR #115, opened 2023-10-16
Enable turning off marks when step is define in slider components

nadijagraca:
## Description
Alterations of slider and range_slider code. 
- Enabling users to turn off marks when step argument is provided. 
- Setting marks argument to default to `{}` instead of `None`

Users can now set step, but turn off visible marks on their sliders by providing `marks=None`

## Checklist

## Types of changes

## Notice


vizro-core/src/vizro/models/_components/form/_form_utils.py
```diff
@@ -94,5 +94,8 @@ def validate_step(cls, step, values):
     return step
 
 
-def set_default_marks(cls, v, values):
-    return v if values.get("step") is None else {}
+def set_default_marks(cls, marks, values):
+    if marks is None or (not marks and values.get("step") is None):
+        return None
+
+    return marks if values.get("step") is None else {}
```
huong-li-nguyen, line 101: Can you check if this simplified version works fully:

```suggestion
def set_default_marks(cls, marks, values):
    if values.get("step") is None:
        marks = None
    return marks
```

vizro-core/tests/unit/vizro/models/_components/form/test_slider.py
```diff
@@ -63,10 +63,10 @@ def test_create_slider_mandatory(self):
 
         assert hasattr(slider, "id")
         assert slider.type == "slider"
+        assert slider.marks is None
         assert slider.min is None
         assert slider.max is None
         assert slider.step is None
-        assert slider.marks is None
```
huong-li-nguyen, line 69: Can you add unit tests if these are not in here yet:
- marks == {} if step is set
- marks is None if step is not set
- marks is None if step is set and marks=None

huong-li-nguyen (changes requested):
Thanks for fixing the bug! 🚀

vizro-core/src/vizro/models/_components/form/_form_utils.py
lingyielia: - what's the expected behavior when step is None and marks is not None?
- what's the expected behavior when step and marks are not None, but they are conflicting? Shall we suggest in the user guide which one takes precedence?

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -226,20 +226,6 @@ def test_validate_step_invalid(self):
         ):
             vm.RangeSlider(min=0, max=10, step=11)
 
-    @pytest.mark.parametrize(
```
huong-li-nguyen, line 229: Instead of deleting these tests, shall we just re-write them?

Or is there any other unit test that shows what happens when you define steps / marks and all cross-combinations of these?

huong-li-nguyen (approved):
Thanks for this!! 🚀 

Could you answer one of the users who asked that this is now natively supported after merging? :)

huong-li-nguyen (commented):
@nadijagraca I just saw that you've changed the default of the step size to 1 - what I meant was to keep the default untouched and just to configure step=1 and marks=None in the app.py if possible?

I don't think we should change the default to 1 - e.g. what happens if min and max are 0 and 1? The defaults should be similar to the underlying component.

vizro-core/src/vizro/models/_components/form/_form_utils.py
```diff
@@ -94,5 +94,7 @@ def validate_step(cls, step, values):
     return step
 
 
-def set_default_marks(cls, v, values):
-    return v if values.get("step") is None else {}
+def set_default_marks(cls, marks, values):
+    if not marks and values.get("step") == 1:
```
huong-li-nguyen, line 98: I think this change also has to be reverted

nadijagraca: Good catch. 👀

lingyielia (approved):

Issue #111, opened 2023-10-12
Work out what to do with registering pages and model manager


antonymilne:
Rough notes for now... A few somewhat related problems:
* #56 
* `page.build` call is required to generate slider callbacks - could be solved by moving callback definition to `pre_build` but then it's dash-dependent
* so is https://github.com/mckinsey/vizro/pull/74/files#diff-624948dfc409dad5986a2413b74f744ec8910022352b840aa4b3aa85f3dfa14c
* where should `register_page` call live? See https://github.com/mckinsey/vizro/pull/74#discussion_r1345394402
* should we use `dash.page_registry` or model manager as the source of truth? How do they relate?
* how much of the model manager can we completely remove?

maxschulz-COL:
Adding https://github.com/mckinsey/vizro/issues/109 for visinility

petar-qb:
Here's a short recap of the conversation I had with @antonymilne last week on the following topic: "Reorganising models `pre_build` method invoking system and models `callbacks creation` so we can get rid of eager `page.build` inside the `vm.Dashboard.build()` method":

Here is a list of tests and code changes I made:
1. I defined vm.Filter(column="year", selector=vm.RangeSlider(title="Select timeframe")) inside the example.
2. I commented out the `page.build()` code from `vm.Dashboard.build()`,
3. `TESTED` - and it `did not` work.
4. Then I moved the creation of the RangeSlider callback inside its pre_build method.
5. `TESTED` - and it `did` work.
6. I redefined the filer without propagating a selector manually, so vm.Filter(column="year"),
7. `TESTED` - and it  `did not` work.
8. I've set a logic inside the Filter that calls self.selector.pre_build() if the selector is artificially added.
8. `TESTED` - and it `did` work.


I came up with two suggestions for the general purpose of solving the problem described above:
1. Should we relocate the internal callback creation within the model's pre_build method?
2. Do we wish to establish a convention where the object initiating the creation of a new object (during its pre_build phase) is also responsible for invoking the pre_build method of the newly created object?

And here is @antonymilne's opinion on the topic (also summarised):

"""
Relocating the callback creation to `pre_build` works nicely but it breaks the rule that `pre_build` should not have any dash dependency. It also feels a bit wrong because it’s tightly coupled to the components generated in the `build` method. I wonder whether we need some other mechanism for eagerly registering these callbacks like `build_callbacks` or something. I can imagine there might be other cases in future also so it’s not specific to the sliders?

[...]

PR #117, opened 2023-10-17
Refactor `Navigation` and `Accordion`

huong-li-nguyen:
## Description
- Refactor `Navigation` and `Accordion` and update unit tests accordingly

User relevant changes:
- Remove user warning if there are missing pages (not all registered pages added to navigation)
- Auto-populate `navigation.pages` if None

## Checklist

## Types of changes

## Notice


vizro-core/src/vizro/models/_navigation/_navigation_utils.py
```diff
@@ -0,0 +1,21 @@
...
+    from vizro.models import Page
+
+    pages_as_list = list(itertools.chain(*pages.values())) if isinstance(pages, dict) else pages
+
+    if not pages_as_list:
+        raise ValueError("Ensure this value has at least 1 item.")
+
+    # Ideally we would use dash.page_registry or maybe dashboard.pages here, but we only register pages in
+    # dashboard.pre_build and model manager cannot find a Dashboard at validation time.
+    # page[0] gives the page model ID.
```
huong-li-nguyen, line 17: @antonymilne - I've checked to pass on the `dashboard.pages` here again, but as I correctly remembered, this doesn't seem to work as, at the point of validation, a model with the type Dashboard cannot be found in the model manager.

So you'll get something like below - any ideas?
Screenshot 2023-10-17 at 18 14 39

vizro-core/src/vizro/models/_navigation/_navigation_utils.py
```diff
@@ -0,0 +1,21 @@
+import itertools
+
+from vizro.managers import model_manager
+
+
+def _validate_pages(pages):
+    """Validator for reuse in `Navigation` and `Accordion` to validate pages."""
+    from vizro.models import Page
+
+    pages_as_list = list(itertools.chain(*pages.values())) if isinstance(pages, dict) else pages
```
antonymilne, line 10: Do we actually need this conversion? Since this validator always runs after the one that coerces to dictionary. The pydantic type hint `Dict[str, List[str]]` should be guaranteed by this point?

vizro-core/src/vizro/models/_navigation/_navigation_utils.py
antonymilne: Ahhh yes I remember... Thanks for checking!

It's not easy to fix here, so let's leave as is for now. And eventually the fix to #111 will improve this.

vizro-core/tests/unit/vizro/models/_navigation/test_accordion.py
```diff
@@ -13,41 +14,39 @@
...
+            Accordion()
 
-    def test_field_invalid_pages_empty_list(self):
+    @pytest.mark.parametrize("pages", [{"SELECT PAGE": []}, []])
+    def test_invalid_field_pages_no_ids_provided(self, pages):
         with pytest.raises(ValidationError, match="Ensure this value has at least 1 item."):
-            Accordion(pages=[])
+            Accordion(pages=pages)
+
+    def test_invalid_field_pages_wrong_input_type(self):
```
antonymilne, line 41: Random comment - let's not change anything here, but worth bearing in mind.

To stop people from needing to copy and paste page IDs we should note this pattern is a good one:
```
pages = [Page(...), Page(...)]
dashboard = Dashboard(pages=pages, navigation=Navigation(pages=[page.id for page in pages])
```

If people find this confusing and make the mistake of inputting the `Page` object itself a lot then potentially we could also enable all these `pages` fields to accept that and immediately extract the `id` from that, so the above would be even simpler:
```
pages = [Page(...), Page(...)]
dashboard = Dashboard(pages=pages, navigation=Navigation(pages=pages))
# somewhere in the validator we do navigation.pages = [page.id for page in pages]
```
...but no need to do that unless it's a proven pain point.

antonymilne (approved):
Looks great, thank you very much for finishing this off! ⭐ 

Just one tip for PR hygiene: the changes to `_validate_pages` are quite hard to review because the code was changed and it moved file. Where possible (tbh often it's not unless you know what you're doing in advance) it's nice to do these commits in a fixed order:
1. Make all changes to `_validate_pages`
2. Then move it to a different file
The opposite order is also fine, but doing 1 then 2 then 1 again is not so good. The idea is that the reviewer should be able to see what are all the actual functional changes by looking at the diff between two commits that have the code in the same file rather than needing to compare between two different files.

Doesn't make any difference here since it's a pretty small function and I know what the changes are anyway, but just a general tip 🙁

maxschulz-COL:
Before taking a look, can you add a little bit of background to the PR why the refactoring was necessary? What didn't work before, and what is better now?

huong-li-nguyen:
> Before taking a look, can you add a little bit of background to the PR why the refactoring was necessary? What didn't work before, and what is better now?

There was room for clean-up and improvement and we had the capacity to do so, so there wasn't a necessity in terms of having to fix something. The new code works as before, so functionally nothing has changed except for the two minor changes I've added to the changelog.

Trying to remember everything - feel free to add @antonymilne :
- Extended validators to auto-populate `dashboard.navigation` for the case if `dashboard.navigation = Navigation()`
- Initiated type conversion (from list to dict) early on (in validator instead of Accordion.build) so we don't have to deal with the different types later on
- Removed shallow functions and just directly copied in the code in `Accordion.build `
- Aligned that build methods should always return a dash object, so instead of `None` we return a hidden `html.Div `in the `Accordion.build`

huong-li-nguyen:
> Looks great, thank you very much for finishing this off! ⭐
> 
> Just one tip for PR hygiene: the changes to `_validate_pages` are quite hard to review because the code was changed and it moved file. Where possible (tbh often it's not unless you know what you're doing in advance or are happy rewriting git history) it's nice to do these commits in a fixed order:
> 
> 1. Make all changes to `_validate_pages`
> 2. Then move it to a different file
> 
> The opposite order is also fine, but doing 1 then 2 then 1 again is not so good. The idea is that the reviewer should be able to see what are all the actual functional changes by looking at the diff between two commits that have the code in the same file rather than needing to compare between two different files.
> 
> Doesn't make any difference here since it's a pretty small function and I know what the changes are anyway, but just a general tip 🙁

Makes a lot of sense, will take into account in the future 👍

vizro-core/src/vizro/models/_navigation/_accordion.py
```diff
@@ -1,67 +1,52 @@
...
 
-    # validators
-    _validate_pages = validator("pages", allow_reuse=True, always=True)(_validate_pages)
+    _validate_pages = validator("pages", allow_reuse=True)(_validate_pages)
+
+    @validator("pages", pre=True)
+    def validate_pages_format(cls, pages):
+        if isinstance(pages, list):
+            return {ACCORDION_DEFAULT_TITLE: pages}
+        return pages
```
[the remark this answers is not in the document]
huong-li-nguyen: Makes sense! 👍  Yes, you are right - `pages = {"page_1"} ` would now automatically be converted to `pages = ["page_1"]`

vizro-core/src/vizro/models/_navigation/_navigation_utils.py
```diff
@@ -0,0 +1,21 @@
+import itertools
+
+from vizro.managers import model_manager
+
+
+def _validate_pages(pages):
+    """Validator for reuse in `Navigation` and `Accordion` to validate pages."""
```
maxschulz-COL, line 7: If we're actually writing a docstring here, shall we quickly say what this validates?

vizro-core/src/vizro/models/_navigation/_navigation_utils.py
huong-li-nguyen: Ah yes, you are right 👍 - I've removed the list conversion as we can loop through the` itertools.chain `object directly

vizro-core/tests/unit/vizro/models/_navigation/test_accordion.py
maxschulz-COL: I remember there was a warning for "forgotten" pages before, why did we decide to get rid of it? Is it because you mentioned above that somethins is now autmatically added?

huong-li-nguyen: Yes, agree! I added the test in here because I think at some point it got confusing with the other `Navigation` selectors and all of them having a `pages` argument. As long as it's clear that dashboard.pages accept Page object and all Navigation related models Page IDs, I think it's fine!

vizro-core/tests/unit/vizro/models/_navigation/test_accordion.py
```diff
@@ -13,41 +14,39 @@
...
-    def test_create_accordion_pages_as_dict(self, pages_as_dict):
+    def test_accordion_valid_pages_as_dict(self, pages_as_dict):
         accordion = Accordion(pages=pages_as_dict, id="accordion_id")
         assert accordion.id == "accordion_id"
         assert accordion.pages == pages_as_dict
 
-    def test_field_invalid_pages_input_type(self):
-        with pytest.raises(ValidationError, match="2 validation errors for Accordion"):
-            Accordion(pages=[vm.Button()])
+    def test_navigation_valid_pages_not_all_included(self):
```
maxschulz-COL, line 27: What does "not all included" mean?

vizro-core/tests/unit/vizro/models/_navigation/test_navigation.py
```diff
@@ -11,51 +12,59 @@
...
-    def test_navigation_pages_as_dict(self, pages_as_dict):
+    def test_navigation_valid_pages_as_dict(self, pages_as_dict):
         navigation = vm.Navigation(pages=pages_as_dict, id="navigation")
         assert navigation.id == "navigation"
         assert navigation.pages == pages_as_dict
 
-    def test_navigation_not_all_pages_included(self):
-        with pytest.warns(UserWarning):
-            vm.Navigation(pages=["Page 1"])
+    def test_navigation_valid_pages_not_all_included(self):
```
maxschulz-COL, line 34: Similar, what does not all included mean here?

vizro-core/src/vizro/models/_navigation/_navigation_utils.py
huong-li-nguyen: EDIT: We actually do need it - saw one of the unit tests failing just now 😄 We need the list conversion so the next check properly catches the case if `navigation.pages = {"SELECT PAGE": []}`

So we need it for this check after:

```
if not pages_as_list:
```

Otherwise this unit tests fails:
Screenshot 2023-10-18 at 10 10 56

vizro-core/tests/unit/vizro/models/_navigation/test_navigation.py
```diff
@@ -11,51 +12,59 @@
...
+        navigation = vm.Navigation(pages=["Page 1"], id="navigation")
+        assert navigation.id == "navigation"
+        assert navigation.pages == ["Page 1"]
+
+    def test_navigation_invalid_pages_empty_list(self):
+        with pytest.raises(ValidationError, match="Ensure this value has at least 1 item."):
+            vm.Navigation(pages=[], id="navigation")
```
maxschulz-COL, line 41: If `[None, vm.Navigation()]` works, why should this not work?

vizro-core/tests/unit/vizro/models/_navigation/test_navigation.py
```diff
@@ -11,51 +12,59 @@
...
-        assert result == expected
+@pytest.mark.parametrize("pages", [["Page 1", "Page 2"], {"Page 1": ["Page 1"], "Page 2": ["Page 2"]}])
+def test_navigation_build(pages):
+    navigation = vm.Navigation(pages=pages)
+    navigation.pre_build()  # Required such that an Accordion is assigned as selector
+    accordion = Accordion(pages=pages)
+    navigation._selector.id = accordion.id
+
+    result = json.loads(json.dumps(navigation.build(), cls=plotly.utils.PlotlyJSONEncoder))
+    expected = json.loads(json.dumps(accordion.build(), cls=plotly.utils.PlotlyJSONEncoder))
```
maxschulz-COL, line 69: I know this was already previously there, but are finding it ok to skip the direct object comparison, because the build of the `Accordion` has already been explicityl tested before?

maxschulz-COL (approved):
LGTM! Left a few comments, mainly for my understanding. Forgive if there are silly questions that I should have understood!

vizro-core/changelog.d/20231017_212005_huong_li_nguyen_navigation.md
```diff
@@ -0,0 +1,39 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category.
+
+-->
+
+### Changed
+
+- Autopopulate `navigation.pages` with registered pages during `Dashboard` validation if None ([#117](https://github.com/mckinsey/vizro/pull/117))
```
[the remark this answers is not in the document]
huong-li-nguyen: So in general you don't have to configure any navigation, you only do so if you want to customize things (e.g. change page group titles etc.), so

```
Dashboard(pages=[Page(title="Page 1", ...), Page(title="Page 2", ...)]], navigation=None)
Dashboard(pages=[Page(title="Page 1", ...), Page(title="Page 2", ...)]], navigation=Navigation())
```

will automatically give you:
`Dashboard(pages=[Page(title="Page 1", ...), Page(title="Page 2", ...)]], navigation=Navigation(pages=["Page 1", "Page 2"]))`

vizro-core/tests/unit/vizro/models/_navigation/test_accordion.py
huong-li-nguyen: If you don't include all registered pages into the Navigation, e.g. let's say you have a code as below:

`Dashboard(pages=[Page(title="Page 1", ...), Page(title="Page 2", ...)]], navigation=Navigation(pages=["Page 1"]))`

You have two registered pages, but only include one in the navigation. What will be visible in the navigation is the one page you included into the navigation

vizro-core/tests/unit/vizro/models/_navigation/test_navigation.py
huong-li-nguyen: Duplicate, will answer above

vizro-core/tests/unit/vizro/models/_navigation/test_navigation.py
huong-li-nguyen: Yes, we shouldn't have to test it again - what is important here is that the navigation.build() returns the build of it's selector. It's very similar to our tests for Filters, where in the build we also only that it is the same as the selector.build

vizro-core/tests/unit/vizro/models/_navigation/test_navigation.py
huong-li-nguyen: `[None, vm.Navigation()]` are the cases we take care of, so they don't have to specify a Navigation model at all. It's the two cases covered by the auto-population of `navigation.pages` in the Dashboard validator.

However, in this case, the user explicitly configured `dashboard.navigation = vm.Navigation(pages=[])`. It's similar to our checks in other models, e.g., `Page.components `or `Dashboard.pages` where we check that the lists they provide are not empty.

vizro-core/src/vizro/models/_navigation/_navigation_utils.py
huong-li-nguyen: Changed to: `Reusable validator in `Navigation` and `Accordion` to check if provided Page IDs exist as registered pages.`

huong-li-nguyen:
> LGTM! Left a few comments, mainly for my understanding. Forgive if there are silly questions that I should have understood!

No worries ❤️  I'll let you resolve the open threads yourself if it's clear to you, otherwise let's discuss :)

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -75,7 +75,8 @@ def validate_pages(cls, pages):
     def validate_navigation(cls, navigation, values):
         if "pages" not in values:
             return navigation
-        if navigation is None:
+
+        if navigation is None or not navigation.pages:
```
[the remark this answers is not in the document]
antonymilne: Ah yes thank you! 👍 All good.

FYI in case I forget again... the reason the second of those cases might be useful is so you could do something like this in the future:
```
dashboard.navigation = Navigation(selector=NavBar())  # change selector type but otherwise leave navigation the same
```
and then it would work without the user needing to explicitly list all the `pages` themselves. It's nice if a user can incrementally build and alter their navigation like this, as they can for a `Filter`.

vizro-core/src/vizro/models/_navigation/_navigation_utils.py
antonymilne: Sorry, I wasn't at all clear in my original comment. What I meant is not whether we need to do the coercion to list but whether we need to do the `isinstance` check? Can't we just do this because we're guaranteed that `pages` is `Dict[str, List[str]]` by this point?

```
pages = list(itertools.chain(*pages.values()))
```

huong-li-nguyen: Ah got it, it works for the Accordion, but not for the Navigation. In the Accordion pages will always be in a Dict format when it hits this validator, but this is not the case for the Navigation e.g. you can provide something like this: Navigation(pages=["Page 1", "Page 2", "Page 3"]).

It relates to the question yesterday when I was wondering whether we should move the dict conversion to the Navigation as well, but I think we agreed that we would only have that in the Accordion as as soon as the NavBar comes in we would require a different dict format than what the Accordion requires.

antonymilne: Ah yes, makes complete sense, thanks for the explanation 👍 I forgot that the same validator is used by `Navigation` where we can still have `List[str]`.

It's great that the tests are catching all of these cases so we don't make a mistake based on my bad suggestions 😀  💯

vizro-core/tests/unit/vizro/models/_navigation/test_accordion.py
huong-li-nguyen: @maxschulz-COL - we decided that the effort to maintain the warning wasn't worth the added benefit. As far as I understood @nadijagraca it became increasingly difficult to do that check with the Icons/Accordions. 

The user will still get a visual indication whether they included all pages or not, as only the included Pages in the Navigation will be visible

vizro-core/changelog.d/20231017_212005_huong_li_nguyen_navigation.md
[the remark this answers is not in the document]
maxschulz-COL: Got it! So registered means registered with the dashboard and not the model_manager right?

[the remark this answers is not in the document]
huong-li-nguyen: Yes, exactly :)

vizro-core/tests/unit/vizro/models/_navigation/test_accordion.py
maxschulz-COL: Ok, I was a fan of the warning, but if its too complicated to maintain, not too fussed about it 👍

vizro-core/tests/unit/vizro/models/_navigation/test_navigation.py
maxschulz-COL: Hm I see, but I think the case is slightly different, as here we would still know what to do, whereas in the cases you described, the configuration does not make any sense. If it was easy, I'd also handle this case, but happy to skip if this is a bother

PR #119, opened 2023-10-19
Table filter interactions

petar-qb:
## Description
- Filter interaction enabled from the `vm.Table` model when its `figure` argument represents the `dash.Datatable` object.

## Potential issues/inconsistencies:
vm.Table as output (of filter, parameter, on_page_load, filter_interactions) works fine because we always modify (as the action's result) outer Div component which ID we know. There are some problems when vm.Table is the input (trigger or state) of the action:
- Defining `actions` on the vm.Table object actually means defining a callback where the trigger is underlying table component fully created by user. It means that we can't set underlying table ID and we need to "force" dashboard users to add the ID if the table is action's trigger.
- Our vm.Table will behave differently depending on is the underlying component `dash.Datatable` or `AgGrid`. e.g. different filter_interaction input properties are considered and different way of calculating filter_interaction's result.
- Persistency of dash.Datatable "action_cell" and chart's "clickData" properties behave differently and we need to consider aligning them. e.g. you can unmark "active_cell" in different ways, and the only way to unmark chart's "clickData" is to refresh the page. 

## TODOs:
- [x] Implement unit tests.
- [x] Add table filter interaction example in docs
- [x] Introduce Figure in docs
- [x] Add table filter interaction example in app.py
- [x] Remove any AgGrid reference from project

## Checklist

## Types of changes

## Notice


vizro-core/src/vizro/models/_components/table.py
```diff
@@ -27,8 +27,9 @@ class Table(VizroBaseModel):
     table: CapturedCallable = Field(..., description="Table to be visualized on dashboard")
     actions: List[Action] = []
 
+    _datatable_id: str = PrivateAttr()
+
     # Component properties for actions and interactions
-    _input_property: str = PrivateAttr("active_cell")
```
petar-qb, line 31: Removed since there are two input properties ("active_cell" and "derived_viewport_data") for `dash_data_table` objects. And for the `AgGrid` implementation there are probably some other properties.

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
```diff
@@ -137,7 +154,7 @@ def _get_action_callback_inputs(action_id: ModelID) -> Dict[str, List[State]]:
         "filter_interaction": (
             _get_inputs_of_chart_interactions(action_id=action_id, action_function=filter_interaction.__wrapped__)
             if "filter_interaction" in include_inputs
-            else []
+            else {}
```
petar-qb, line 157: I probably need to adjust all inputs formats to the dictionary format filter_interaction use.

vizro-core/examples/default/app.py
petar-qb, line 1: Just a reminder to revert `app.py` changes.

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
```diff
@@ -98,22 +98,39 @@ def _get_inputs_of_controls(action_id: ModelID, control_type: ControlType) -> Li
...
+                }
+            )
+        else:
+            inputs.append(
+                {
+                    "clickData": State(component_id=triggered_model.id, component_property="clickData"),
+                }
+            )
+
+    return inputs
```
petar-qb, line 130: Consider do we want to move the following logic into Model implementations. Already explained in https://github.com/mckinsey/vizro/pull/114#discussion_r1364517261

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -31,10 +31,13 @@ class Table(VizroBaseModel):
     title: Optional[str] = Field(None, description="Title of the table")
     actions: List[Action] = []
 
+    _underlying_table_id: str = PrivateAttr()
+
```
petar-qb, line 35: Once we support AgGrid and others, we'll need to handle the `"active_cell"` argument in the following line:

`set_actions = _action_validator_factory("active_cell")`

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -51,12 +54,36 @@ def __getitem__(self, arg_name: str):
...
     @_log_call
     def build(self):
+        # TODO: We also need to take case what empty object we need to create here once we support AgGrid and others.
         return html.Div(
             [
                 html.H3(self.title, className="table-title") if self.title else None,
-                html.Div(dash_table.DataTable(), id=self.id),
+                html.Div(
+                    dash_table.DataTable(**({"id": self._underlying_table_id} if self.actions else {})), id=self.id
+                ),
```
petar-qb, line 86: After we support AgGrid and others, we will also need to take care of the empty object we create here.

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -51,12 +53,35 @@ def __getitem__(self, arg_name: str):
...
+
     @_log_call
     def build(self):
         return html.Div(
             [
                 html.H3(self.title, className="table-title") if self.title else None,
-                html.Div(dash_table.DataTable(), id=self.id),
+                html.Div(
+                    dash_table.DataTable(**({"id": self._underlying_table_id} if self.actions else {})), id=self.id
+                ),
```
petar-qb, line 84: After we support AgGrid and others, we will also need to take care of the empty object we create here.

vizro-core/docs/pages/user_guides/actions.md
```diff
@@ -103,17 +103,17 @@ a result, when a dashboard user now clicks the button, all data on the page will
...
-To configure this chart interaction follow the steps below:
+To configure this figure interaction follow the steps below:
 
-1. Add the action function to the source [`Graph`][vizro.models.Graph] and a list of IDs of the target charts into `targets`
+1. Add the action function to the source [`Graph`][vizro.models.Graph] or [`Table`][vizro.models.Table] component and a list of IDs of the target charts into `targets`.
 ```py
 actions=[vm.Action(function=filter_interaction(targets=["scatter_relation_2007"]))]
 ```
-2. Enter the filter columns in the `custom_data` argument of the underlying source chart `function`
+2. If the source figure is [`Graph`][vizro.models.Graph], enter the filter columns in the `custom_data` argument of the underlying source chart `function`.
```
petar-qb, line 116: Do I also need to add an example where the `vm.Table` is source of the filter interactions?

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -66,34 +63,97 @@ def _apply_filters(
     return data_frame
 
 
-def _apply_filter_interaction(
-    data_frame: pd.DataFrame,
-    ctds_filter_interaction: List[CallbackTriggerDict],
-    target: str,
+def _apply_chart_filter_interaction(
```
maxschulz-COL, line 66:
```suggestion
def _apply_graph_filter_interaction(
```
and in subsequent places

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -66,34 +63,97 @@ def _apply_filters(
...
+    data_frame: pd.DataFrame, target: str, ctd_filter_interaction: Dict[str, CallbackTriggerDict]
```
maxschulz-COL, line 67: Curious, what made the typing change here? This originates from `ctx` which is not in our code-base right? So why has this changed its type?

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -51,12 +53,35 @@ def __getitem__(self, arg_name: str):
...
+        data = data_manager._get_component_data(self.id)  # type: ignore
+        kwargs = self.figure._arguments.copy()
+        kwargs.pop("data_frame", None)
+        return self.figure._function(data_frame=data, **kwargs)
+
+    @_log_call
+    def pre_build(self):
+        if self.actions:
+            # The underlying table object is pre-built, so we can fetch its ID.
+            underlying_table_object = self._build_underlying_table_object()
```
maxschulz-COL, line 66: Do we need to build the actual table for this? Could we not just build an empty table? Then we can get rid of this whole arguments fetching stuff?

vizro-core/src/vizro/models/_components/table.py
maxschulz-COL: Do we really need the if clause? In the end why not always attach ID of the pre-built table?

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -66,34 +63,97 @@ def _apply_filters(
...
+            continue
+        column = ctd_active_cell["value"]["column_id"]
+        derived_viewport_data_row = ctd_active_cell["value"]["row"]
+        clicked_data = ctd_derived_viewport_data["value"][derived_viewport_data_row][column]
+        data_frame = data_frame[data_frame[column].isin([clicked_data])]
+
+    return data_frame
+
+
+def _apply_filter_interaction(
```
maxschulz-COL, line 126: Starting this thread to summarize the previous discussions. We agreed that for now, since this is not public API, we agree that:
- having the implementation logic in these help functions is fine
- in the future we want to abstract these methods to be attached to the captured callables

Is my understanding here correct?

vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
```diff
@@ -22,12 +25,22 @@ def _build_action_loop_callbacks() -> None:
...
     for actions_chain in actions_chains:
+        actions_chain_trigger_component_id = actions_chain.trigger.component_id
+        try:
+            actions_chain_trigger_component = model_manager[ModelID(str(actions_chain_trigger_component_id))]
+            # Use underlying table object as a trigger component.
+            if isinstance(actions_chain_trigger_component, Table):
+                actions_chain_trigger_component_id = actions_chain_trigger_component._underlying_table_id
+        # Not all action_chain_trigger_components are included in model_manager e.g. on_page_load_action_trigger
+        except KeyError:
+            pass
```
maxschulz-COL, line 36: How would we abstract this logic in the future? The Table and a future react model essentially need a way to provide the `underlying` id right?

Should we in this case not work already with something more abstract in the table model such as:

`_callable_object_id` and then work here with `hasattr("_callable_object_id)`?

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -66,34 +63,97 @@ def _apply_filters(
...
+    ctd_derived_viewport_data = ctd_filter_interaction["derived_viewport_data"]
+    if not ctd_active_cell["value"] or not ctd_derived_viewport_data["value"]:
+        return data_frame
+
+    source_table_id = ctd_active_cell["id"]
+    try:
+        source_table_actions = _get_component_actions(model_manager[source_table_id])
+    except KeyError:
+        # source_table_id is underlying table id, so actions from parent Vizro.Table component should be considered.
+        source_table_actions = _get_component_actions(_get_parent_vizro_table(source_table_id))
```
maxschulz-COL, line 113: I think I don't follow here. Can you elaborate a bit more?

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -66,34 +63,97 @@ def _apply_filters(
...
+    raise KeyError(f"No parent Vizro.Table component found for underlying table with id {_underlying_table_id}.")
+
+
+def _apply_table_filter_interaction(
+    data_frame: pd.DataFrame, target: str, ctd_filter_interaction: Dict[str, CallbackTriggerDict]
+) -> pd.DataFrame:
+    ctd_active_cell = ctd_filter_interaction["active_cell"]
+    ctd_derived_viewport_data = ctd_filter_interaction["derived_viewport_data"]
+    if not ctd_active_cell["value"] or not ctd_derived_viewport_data["value"]:
+        return data_frame
```
maxschulz-COL, line 106: I see now what data is filtered, and it's not just the state. Understood. So then every targets dataframe that share the same column id with the source is filtered:
- by the target column
- for the value that is clicked

Is this sufficient? Have users requested to customize this?

maxschulz-COL (commented):
First round of comments, thanks a lot, must not have been easy :)

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -66,34 +63,97 @@ def _apply_filters(
...
+def _get_parent_vizro_table(_underlying_table_id: str) -> Table:
+    from vizro.models import Table
 
-        for action in source_chart_actions:
-            if target not in action.function["targets"]:
-                continue
+    for _, table in model_manager._items_with_type(Table):
+        if table._underlying_table_id == _underlying_table_id:
+            return table
+    raise KeyError(f"No parent Vizro.Table component found for underlying table with id {_underlying_table_id}.")
```
huong-li-nguyen, line 97: Again, could we make this more generic? So we don't reference the table specifically but look for models with an underlying `_figure_id`

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -51,12 +53,35 @@ def __getitem__(self, arg_name: str):
...
+
+            # Underlying table object has to have "id" defined if it triggers actions chain.
+            if not hasattr(underlying_table_object, "id"):
+                raise ValueError(
+                    "Underlying table object has no attribute 'id'. To perform specified actions, a valid 'id' to the"
+                    " Underlying table object has to be provided."
```
huong-li-nguyen, line 72:
```suggestion
                    "Underlying `Table` callable has no attribute 'id'. To enable actions triggered by the `Table` a valid 'id' has to be provided to the `Table`callable.
```

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -51,12 +53,35 @@ def __getitem__(self, arg_name: str):
...
+        kwargs.pop("data_frame", None)
+        return self.figure._function(data_frame=data, **kwargs)
+
+    @_log_call
+    def pre_build(self):
+        if self.actions:
+            # The underlying table object is pre-built, so we can fetch its ID.
+            underlying_table_object = self._build_underlying_table_object()
+
+            # Underlying table object has to have "id" defined if it triggers actions chain.
```
huong-li-nguyen, line 68: Would remove the comments if it's clear from the error message and code
```suggestion
```

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -66,34 +63,97 @@ def _apply_filters(
...
+
+
+def _apply_table_filter_interaction(
+    data_frame: pd.DataFrame, target: str, ctd_filter_interaction: Dict[str, CallbackTriggerDict]
+) -> pd.DataFrame:
+    ctd_active_cell = ctd_filter_interaction["active_cell"]
+    ctd_derived_viewport_data = ctd_filter_interaction["derived_viewport_data"]
+    if not ctd_active_cell["value"] or not ctd_derived_viewport_data["value"]:
```
huong-li-nguyen, line 105: Could you remind me again, are `active_cell` and `derived_viewport_data` the only possible arguments? Or is it working like in the dcc.Graph and there are a few more? What other arguments exist?

vizro-core/src/vizro/models/_components/table.py
huong-li-nguyen: Depending on what's slicker in code, I would probably just return an empty datatable even for the ag-grid.

Visually we will aim to style them very similar, so the empty figure shouldn't look different when you use the AgGrid or the DataTable

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -51,12 +53,35 @@ def __getitem__(self, arg_name: str):
             return self.type
         return self.figure[arg_name]
 
+    def _build_underlying_table_object(self):
+        data = data_manager._get_component_data(self.id)  # type: ignore
+        kwargs = self.figure._arguments.copy()
+        kwargs.pop("data_frame", None)
```
huong-li-nguyen, line 59: Do we still have to do the popping of data_frame? I slightly remember we removed this from Max PR? Or am I remembering this incorrectly?

huong-li-nguyen (commented):
Great job, Petar!! 🚀  I quickly scanned through the code, would take a proper look on tests and docs in the second iteration.

Left a few comments, but I think we need to align on namings. 
- Any function only related to graph, I would rename from XXX_chart_XXXX to XXXX_graph_XXXX.
- Any function related to both Graph and Table, I think we said we would rename `chart` to `figure`

vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
huong-li-nguyen: Ah, just saw we were reviewing at the same time haha I suggested the same and was proposing `_figure_id`.
`_callable_object_id` would also be fine for me though

vizro-core/src/vizro/models/_components/table.py
petar-qb: > Do we really need the if clause? In the end why not always attach ID of the pre-built table?

Underlying table ID isn't always provided and it isn't always necessary. 
Underlying table ID is necessary only if `vm.Table` has `actions` defined (which means that underlying table triggers action/-s).

The reason why we need `id` defined in the empty `datatable` when the build method is invoked is because this `datatable` is input into `_on_page_load` action (so it has to exist in UI elements). We can potentially change this behaviour and say that `filter_interaction` doesn't need to be the input of the `_on_page_load` action, but we would probably need to discuss it. 

> I would probably just return an empty datatable even for the ag-grid

Since `filter_interaction` input will be different ("active_cell" and "derived_viewport_data" for `datatable` and "cellClicked" for `AgGrid`) it means that probably we would need to take care of what empty component is created as well. I guess it will happen that the Dash error will be raised because somewhere this `datatable` uses `"cellClicked"` property which is not available. I'm not sure so let's discuss it once we add AgGrid and others if you agree.

Since the `filter_interaction` input will be different ("active_cell" and "derived_viewport_data" for `datatable` and "cellClicked" for `AgGrid`), this means we should probably also care about which empty component is created. I suspect that the Dash error will occur because in `on_page_load` action this `datatable` is using a `"cellClicked"` property that is not available. I'm not sure so let's discuss when we add AgGrid and others if you agree.

vizro-core/examples/default/app.py
```diff
@@ -62,32 +53,52 @@ def AgGrid(data_frame=None):
...
-        vm.Table(id="table2", figure=dash_data_table(data_frame=data, style_header={"border": "1px solid green"})),
+        vm.Table(
+            id="table_id",
+            figure=dash_data_table(
+                id="dash_datatable_id",
+                data_frame="table_data",
+                style_cell={"border": "5px solid green"},
+            ),
+            actions=[
+                vm.Action(id="filter_interaction", function=filter_interaction(targets=["scatter_chart", "table_2_id"]))
```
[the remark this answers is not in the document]
petar-qb: We always filter by the column which cell you clicked (it's not  'State' column specific). 
image

vizro-core/src/vizro/actions/_actions_utils.py
petar-qb: It's caused by changes made in the `_callback_mapping_utils.py  _get_inputs_of_figure_interactions function`. Its `return value` is changed since some actions required multiple component properties to be propagated. So, I "packed" them into Dict[str, CallbackTriggerDict] format.

vizro-core/src/vizro/models/_components/table.py
petar-qb: I refactored and avoided to create a temporary datatable with the original data_frame. Does it look good now?

vizro-core/src/vizro/models/_components/table.py
huong-li-nguyen: Yes, that's fine @petar-qb 👍  I wouldn't worry too much of making everything compatible with the AgGrid in this PR. This can be done in a separate PR. The public API should just not change, which I think we already ensured in Max previous PR. So happy to move all considerations regarding the AgGrid which are more complex to implement right now in a separate ticket

vizro-core/src/vizro/actions/_actions_utils.py
petar-qb: I agreed with J about this behaviour because we don't need to handle artificially added `"custom_data"` dash.datatable parameter (as we do in vizx). Once we enable targeting filter interaction with the following format (you proposed):
`"source_column.target_id.target_column"` (or something similar), we will adjust calculations in this util function.

vizro-core/changelog.d/20231026_084022_petar_pejovic_table_filter_interaction.md
```diff
@@ -0,0 +1,41 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+
+### Added
+
+- Enable `filter_interaction` action when the source figure is a `vm.Table` component ([#119](https://github.com/mckinsey/vizro/pull/119))
```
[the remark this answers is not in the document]
petar-qb: Correct. There are examples for all the above actions in the `app.py` + table as a filter interaction source.

vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
```diff
@@ -22,12 +25,22 @@ def _build_action_loop_callbacks() -> None:
 
     gateway_inputs: List[Input] = []
     for actions_chain in actions_chains:
+        actions_chain_trigger_component_id = actions_chain.trigger.component_id
```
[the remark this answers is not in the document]
petar-qb: > Is there any reason we have to assign this to a variable instead of using actions_chain.trigger.component_id directly? Does the dot notation cause issues?

The reason is that we modify or reuse this variable in three places below and it's nice to extract it once and then adjust and reuse it.

---

> Also, what do you think about renaming to actions_chain_trigger_model_id to be more generic?

The word "component" is used everywhere in the project when it refers to the trigger component_id (dash syntax). Also, this code is the best example of how action source (trigger) doesn't need to be Vizro model (but some pure dash underlying component).

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -31,6 +31,8 @@ class Table(VizroBaseModel):
     title: Optional[str] = Field(None, description="Title of the table")
     actions: List[Action] = []
 
+    _underlying_table_id: str = PrivateAttr()
```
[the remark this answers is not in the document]
petar-qb: `_underlying_table_id` is replaced with `_callable_object_id` (the selection based on https://github.com/mckinsey/vizro/pull/119#discussion_r1374262606).

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -66,34 +63,97 @@ def _apply_filters(
...
+                target=target,
+                ctd_filter_interaction=ctd_filter_interaction,
+            )
 
-            for custom_data_idx, column in enumerate(custom_data_columns):
-                data_frame = data_frame[data_frame[column].isin([customdata[custom_data_idx]])]
     return data_frame
 
 
+def _validate_selector_value_NONE(value: Union[SingleValueType, MultiValueType]) -> ValidatedNoneValueType:
```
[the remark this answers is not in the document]
petar-qb: I agree but this code is written by @maxschulz-COL, so what do you think Max?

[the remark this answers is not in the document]
maxschulz-COL: Happy to change it, put the `NONE` as it was referring to the string, but probably better to keep lowercase

vizro-core/src/vizro/actions/_actions_utils.py
petar-qb: If the source of the filter interaction is `Graph` then `"clickData"` is the callback input.
If the source of the filter interaction is `Table` then `"active_cell"` and `"derived_viewport_data"` are the callback inputs.

huong-li-nguyen: I meant in general e.g. in the Graph model I know the default callback input is `clickData` but in the future we also want to enable `hoverData`, `selectedData`, etc.

Do several potential options also exist for the data table? Or is it just going to be `active_cell `and `derived_viewport_data`?

vizro-core/src/vizro/actions/_action_loop/_build_action_loop_callbacks.py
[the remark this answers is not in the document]
huong-li-nguyen: On your first point, got it - makes sense 👍 

Hmm.. yeah, I see where the confusion comes from. The term `components` is probably not the best - for me all models inside the projects `components` folder I consider as `components`, which is why I thought it would be better to rename to `actions_chain.trigger.model_id`, as we also have models outside of the `components` folder that can be triggers e.g. the `Page` model can be a trigger (at least it has an action argument attached), but is not a component for me in that sense.

I think your definition of components is more broad than mine as you consider all dash objects as components, whether they

vizro-core/src/vizro/actions/_actions_utils.py
petar-qb: Ah, I see, sorry.
There are always multiple potential properies for any dash object so there is for a dash datatable too. 
e.g. If we want to enable something like the old model interactions, we should enable recognition of which rows in datatable are selected, so we should also support the `"derived_virtual_selected_rows"` property.

vizro-core/src/vizro/models/_components/table.py
petar-qb: I change it now.

huong-li-nguyen (approved):
Looks good from my side - thanks for incorporating all the comments 👍 🚀  

Would be great if you could add the additional table example and remove any reference to the AG grid in the requirements and code. Other than that, all good from my side 👍

petar-qb:
There are few more things I'm currently working on (and are the part of the PR Description - ToDos section):
- Add table filter interaction example in docs.
- Introduce "Figure" in docs.
- Add table filter interaction example in app.py.
- Remove any reference to AgGrid in project.

In the meantime other parts of the PR are ready for review. 

FYI: @maxschulz-COL, @huong-li-nguyen 😄

maxschulz-COL (commented):
Thanks a lot! Have a couple more suggestions, but I think we are getting there 💪 

Need to be careful so that tests don't become unreadable! :)

maxschulz-COL:
> There are few more things I'm currently working on (and are the part of the PR Description - ToDos section):
> 
> * Add table filter interaction example in docs.
> * Introduce "Figure" in docs.
> * Add table filter interaction example in app.py.
> * Remove any reference to AgGrid in project.
> 
> In the meantime other parts of the PR are ready for review.
> 
> FYI: @maxschulz-COL, @huong-li-nguyen 😄

You beat me to my comment :) Reviewed the test, but happy to take another look once the above is done!

vizro-core/src/vizro/actions/_actions_utils.py
petar-qb: Yes exactly. It has the same format for all of currently predefined actions. Dictionary with keys ("filters", "parameters", "filter_interaction", "theme_selector") 

You can find more on  [Dash flexible callback signatures](https://dash.plotly.com/flexible-callback-signatures#dictionary-grouping)

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
```diff
@@ -96,24 +96,40 @@ def _get_inputs_of_controls(action_id: ModelID, control_type: ControlType) -> Li
...
-        )
-        for action in chart_interactions_on_page
-    ]
-
-
-def _get_action_callback_inputs(action_id: ModelID) -> Dict[str, List[State]]:
+    inputs = []
+    for action in figure_interactions_on_page:
+        triggered_model = _get_triggered_model(action_id=ModelID(str(action.id)))
+        if isinstance(triggered_model, Table):
```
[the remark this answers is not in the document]
petar-qb: We probably want to move this logic to a Figure Model method called `_get_filter_interaction_inputs` or something similar. This is one way to avoid using statements like `if isinstance(triggered_model, Table)`.

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -52,12 +55,32 @@ def __getitem__(self, arg_name: str):
         return self.figure[arg_name]
 
     @_log_call
+    def pre_build(self):
+        if self.actions:
+            kwargs = self.figure._arguments.copy()
```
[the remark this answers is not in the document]
petar-qb: I decided to propagate them because they can be used in a `custom table function` in some way (eg `kwargs["abc"]`) and I don't want to compromise the proper operation of the function.

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -52,12 +55,32 @@ def __getitem__(self, arg_name: str):
...
+            self._callable_object_id = underlying_table_object.id
+
     def build(self):
         return dcc.Loading(
             html.Div(
                 [
                     html.H3(self.title, className="table-title") if self.title else None,
-                    html.Div(dash_table.DataTable(), id=self.id),
+                    html.Div(
+                        dash_table.DataTable(**({"id": self._callable_object_id} if self.actions else {})), id=self.id
```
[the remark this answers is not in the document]
petar-qb: I answered here below https://github.com/mckinsey/vizro/pull/119#discussion_r1374425714. Do you think it's resolved?

vizro-core/tests/unit/vizro/actions/_callback_mapping/test_get_action_callback_mapping.py
```diff
@@ -29,7 +29,7 @@ def export_data():
 
 
 @pytest.fixture
-def managers_one_page_four_controls_two_graphs_filter_interaction(request):
+def managers_one_page_four_controls_two_figures_filter_interaction(request, dash_data_table_with_id):
```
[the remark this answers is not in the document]
petar-qb: Actually there are two figures, and one button (which is not a figure). If you still think we need to rename it, I'm happy to do that.

vizro-core/tests/unit/vizro/actions/_callback_mapping/test_get_action_callback_mapping.py
```diff
@@ -98,7 +108,13 @@ def action_callback_inputs_expected():
             dash.State("parameter_y_selector", "value"),
         ],
         "filter_interaction": [
-            dash.State("scatter_chart", "clickData"),
+            {
```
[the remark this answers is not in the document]
petar-qb: I just made a type change and wrap the `State` to `Dict[str, State]`, so I can enable table interactions when there are may properties. 
`dash.State("scatter_chart", "clickData")` -> `{"clickData": dash.State("scatter_chart", "clickData")}`

Search for return type changes made in `_get_inputs_of_figure_interactions` in `_callback_mapping_utils.py`

vizro-core/tests/unit/vizro/actions/_callback_mapping/test_get_action_callback_mapping.py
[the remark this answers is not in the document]
maxschulz-COL: Huh, I count: Graph, Graph, Table, Button

vizro-core/tests/unit/vizro/actions/_callback_mapping/test_get_action_callback_mapping.py
```diff
@@ -63,7 +73,7 @@ def managers_one_page_four_controls_two_graphs_filter_interaction(request):
             vm.Filter(id="filter_country", column="country", selector=vm.Dropdown(id="filter_country_selector")),
             vm.Parameter(
                 id="parameter_x",
-                targets=["scatter_chart.x", "scatter_chart_2.x"],
+                targets=["scatter_chart.x", "scatter_chart_2.x", "vizro_table.columns"],
```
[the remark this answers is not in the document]
maxschulz-COL: It's hard to see the diff now, have to changed it to something that works? Or have you just taken out the table parameter?

vizro-core/examples/default/app.py
maxschulz-COL, line 1: If we're deciding to update those changes, we should also transfer them to yaml, json and dict.

As long as we don't have an alternative to present every feature, we should keep them in sync

maxschulz-COL (approved):
LGTM now, but please still adress the last open issues

vizro-core/tests/unit/vizro/actions/_callback_mapping/test_get_action_callback_mapping.py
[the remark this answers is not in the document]
petar-qb: I added a new Parameter into `managers_one_page_four_controls_three_figures_filter_interaction`:
```
vm.Parameter(
                id="vizro_table_row_selectable",
                targets=["vizro_table.row_selectable"],
                selector=vm.Dropdown(
                    id="parameter_table_row_selectable",
                    options=["multi", "single"],
                    multi=False,
                    value="single",
                ),
            ),
```

which is tested with `def test_action_callback_mapping_outputs` and the following combination of inputs:
```
             "action_id, action_callback_outputs_expected",
            (
                "parameter_action_vizro_table_row_selectable",
                [
                    {"component_id": "vizro_table", "component_property": "children"},
                ],
            )
```

vizro-core/changelog.d/20231026_084022_petar_pejovic_table_filter_interaction.md
```diff
@@ -0,0 +1,41 @@
...
+-->
+
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+
+### Added
```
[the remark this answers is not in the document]
petar-qb: I think so. I added a new empty changelog file.

PR #120, opened 2023-10-19
Enable `_clear_state` and ensure test independence

maxschulz-COL:
## Description
- implement the private `_clear_state` function as suggested by @ned2
- apply to all tests to ensure test independence
- adjust preliminary warnings that guide towards this solution if cells are re-evaluated in Jupyter
- Replace `tech_logos.png`

## Checklist

## Types of changes

## Notice


vizro-core/src/vizro/_vizro.py
```diff
@@ -13,6 +13,18 @@
 logger = logging.getLogger(__name__)
 
 
+def _clear_state():
+    """Private method that clears all state in the vizro package."""
+    data_manager._reset()
+    model_manager._reset()
```
huong-li-nguyen, line 19: I think @antonymilne suggested to rename these methods to `_clear` as well

vizro-core/tests/unit/vizro/conftest.py
```diff
@@ -44,5 +43,3 @@ def dashboard(page1, page2):
 @pytest.fixture()
 def dashboard_prebuild(dashboard):
     yield dashboard.pre_build()
```
huong-li-nguyen, line 45: If we don't clean up after, let's change to a return instead of yield? Maybe double-check in other places too

vizro-core/changelog.d/20231019_130629_maximilian_schulz_implement_improved_clearing.md
```diff
@@ -0,0 +1,41 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category.
+
+-->
+
+### Changed
+
+- Warning for duplicated ID's in `data_manager` and `model_manager` now recommends `vizro._vizro._clear_state()` as a potential fix when working in a Jupyter notebook ([#120](https://github.com/mckinsey/vizro/pull/120))
```
huong-li-nguyen, line 22:
```suggestion
- Update warning for duplicated IDs  in `data_manager` and `model_manager` to now recommend `vizro._vizro._clear_state()` as a potential fix when working in a Jupyter notebook ([#120](https://github.com/mckinsey/vizro/pull/120))
```

vizro-core/src/vizro/_vizro.py
```diff
@@ -13,6 +13,18 @@
 logger = logging.getLogger(__name__)
 
 
+def _clear_state():
```
huong-li-nguyen, line 16: Shall we actually add this as a private method into the Vizro class? Then people can just call `Vizro._clear_state()` and don't have to import the private method via the private file etc.?

```suggestion
def _clear_state():
```

huong-li-nguyen (commented):
Thanks you two for fixing @maxschulz-COL  @ned2! 🚀 

@maxschulz-COL - could you add @ned2 to our list of contributors in our authors.md file to give him credits for the proposed solution?

vizro-core/src/vizro/_vizro.py
maxschulz-COL: Hm I think that may ultimately be the better option, let me implement

vizro-core/src/vizro/_vizro.py
```diff
@@ -76,6 +76,18 @@ def _pre_build():
             if hasattr(model, "pre_build"):
                 model.pre_build()
 
+    @staticmethod
+    def _clear_state():
+        """Private method that clears all state in the vizro package."""
```
huong-li-nguyen, line 81:
```suggestion
        """Private method that clears all states in the vizro app."""
```

vizro-core/tests/conftest.py
```diff
@@ -1,10 +1,19 @@
...
-from vizro.managers import data_manager, model_manager
+from vizro import Vizro
 
 
 @pytest.fixture(autouse=True)
 def reset_managers():
+    # this ensures that the managers are reset before and after each test
+    # in principle clearing after every test should suffice, but this serves as a safety measure
+    # there are two tests in `test_get_action_loop_components.py` that fail without this
+    Vizro._clear_state()
```
huong-li-nguyen, line 11: Do we know why we need the clearing twice?

vizro-core/tests/unit/vizro/conftest.py
```diff
@@ -43,6 +42,4 @@ def dashboard(page1, page2):
 
 @pytest.fixture()
 def dashboard_prebuild(dashboard):
```
huong-li-nguyen, line 44: Is this fixture even used now? I think it was replaced by the Vizro.pre_build?

huong-li-nguyen (approved):
Thanks for cleaning up! 🧹 🔥

vizro-core/src/vizro/_vizro.py
```diff
@@ -76,6 +76,18 @@ def _pre_build():
             if hasattr(model, "pre_build"):
                 model.pre_build()
 
+    @staticmethod
+    def _clear_state():
```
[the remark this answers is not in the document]
maxschulz-COL: Hm, if we only use `clear` it may be too linked to the `Vizro` class no? Ultimately I think it's ok to be attached to it, because it thematically belongs to it, but if people then type `Vizro._clear` they could think that the `Vizro` class is just some data structure. This clear method somewhat transcends that. But honestly not sure...

vizro-core/tests/conftest.py
maxschulz-COL: This is the strange issue I was telling you about - for some reason without the buildup clear there is some remnant left that affects the above two mentioned tests. So I am not entirely sure... Clearing in the build-up and teardown is also generally a more safe way to handle, so I think this is in principle good.

vizro-core/src/vizro/_vizro.py
maxschulz-COL: Do yo think it's confusing because we also include the `data_manager` and `model_manager` here? Not sure if they really count as being part of the app 🤔 ?

vizro-core/tests/unit/vizro/conftest.py
maxschulz-COL: It is used in the `dashboard` and `navigation` tests

huong-li-nguyen: Ah yeah, I see. I think in the dashboard it might be better to actually test the function so change to:

In the dashboard tests, shall we actually explicitly call the `dashboard.pre_build` given that it's testing that? Feel free to ignore if you think it doesn't matter much

```
@pytest.mark.usefixtures("vizro_app", "dashboard")
    def test_dashboard_page_registry(self, mock_page_registry, dashboard):
        dashboard.pre_build()
        result = dash.page_registry
        expected = mock_page_registry
```

vizro-core/tests/conftest.py
petar-qb: I just tried with `hatch run test-unit` and the `reset_managers` implementation: 
1. Vizro._clear_state(); yield
2. yield; Vizro._clear_state()
and in both cases I get all tests passed. Did I overlook something?

vizro-core/src/vizro/_vizro.py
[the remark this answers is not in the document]
huong-li-nguyen: I was thinking mostly about consistency but I definitely see your point. I don't have a strong preference on this one to be honest, so either way would be fine for me :)

vizro-core/tests/unit/vizro/conftest.py
```diff
@@ -43,6 +42,4 @@ def dashboard(page1, page2):
...
-    yield dashboard.pre_build()
-    del dash.page_registry["Page 1"]
-    del dash.page_registry["Page 2"]
+    dashboard.pre_build()
```
petar-qb, line 45: Can we just add `vizro_app` as input of this fixture and then replace:
`@pytest.mark.usefixtures("vizro_app", "dashboard_prebuild")` 
with:
`@pytest.mark.usefixtures("dashboard_prebuild")`
everywhere?

vizro-core/src/vizro/_vizro.py
huong-li-nguyen: For me they are part of the app as otherwise the app wouldn't run. I was a bit confused about the package term because I wouldn't know what clearing the state for a python package means 🤔

vizro-core/src/vizro/_vizro.py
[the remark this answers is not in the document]
maxschulz-COL: Let's leave for now unless someone else has a stronger opinion

vizro-core/tests/unit/vizro/conftest.py
maxschulz-COL: Yes I think that would be cleaner!

vizro-core/tests/conftest.py
maxschulz-COL: No this may work, as it executes the code before all tests. I suppose it is then for added safety to also execute after very test.

vizro-core/tests/unit/vizro/conftest.py
maxschulz-COL: I think this may not work in all cases, as there is other places, e.g. `TestGetActionLoopComponents`, where there is a different dashboard built. In order to acheive the above, there is a bigger rework needed

antonymilne (approved):
Thank you for the fix @maxschulz-COL and @ned2! 

I just have a couple of small comments and one concern about the test fixture but otherwise looks great 💯 

In the near future I think we might want to move some of this resetting behaviour so it always runs automatically in our `Vizro.__init__` or similar, and also make the rest that's left in `Vizro._reset` into public API. This can be handled as part of #111 because it's all related.

petar-qb (approved):
It looks good!

vizro-core/tests/conftest.py
```diff
@@ -1,10 +1,19 @@
...
     yield
-    model_manager._reset()
-    data_manager._reset()
+    Vizro._clear_state()
+
+
+@pytest.fixture()
+def vizro_app():
+    """Fixture to instantiate Vizro/Dash app. Required when needing to register pages."""
+    yield Vizro()
```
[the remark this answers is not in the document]
maxschulz-COL: Yes

vizro-core/tests/conftest.py
maxschulz-COL: It is even more complicated than that...

When running all tests the above is fine, just not when running this test alone. Going back to this I discovered that this runs fine if you `_reset` BEFORE the test, but not after.

It seems that when running this with other tests, the teardown of a previous test fixes the issue, taking the role of `_reset` before manually.

This feel like that there is something that it interferes with that needs to be explicitly be cleared. The error is quite strange, it starts to complain all of a sudden for the `Accordion` validation:

```
E   pydantic.error_wrappers.ValidationError: 1 validation error for Accordion
E   pages -> SELECT PAGE
E     none is not an allowed value (type=type_error.none.not_allowed)
```

maxschulz-COL: This was ressolved on Friday - the culprit was/is instatiating models outside of fixtures, in this case the harmful model was `vm.Navigation` in the navigation tests.

However since clearing the managers before every test is valid solution to this, and also because it can be quite convenient to define models outside fixtures (e.g. in parametrization) we decided that this was the correct course of action

PR #121, opened 2023-10-23
Fix CapturedCallable to work better with kwargs

antonymilne:
## Description

#114 uncovered a bug with `CapturedCallable`. In short, this would work ok:
```
function = lambda **kwargs: kwargs["a"]

CapturedCallable(function)(a=2) # a=2 only provided at runtime
```

And this would not:
```
CapturedCallable(function, a=2)() # a=2 is provided upfront
```

Now they should both work ok.

We have already had a couple of very small bugs that I've known about for a while, where the following sorts of functions do not work with `CapturedCallable`:
1. a function with variadic positional args like `def function(*args)`
2. a function with positional-only arguments like `def function(a, /)`

Fixing the new `**kwargs` bug, which is an important case, makes it much harder to fix these much less important cases, and so I've decided to just not support them at all. There's now an explicit check to make sure no one tries to use `CapturedCallable` with those sorts of functions (very unlikely they'd want to do so anyway).

Tests have been updated to be much more thorough so that we now test every case on all 3 sorts of valid function `positional_or_keyword_function, keyword_only_function, var_keyword_function`.

Thanks to @maxschulz-COL for spotting the bug and for his initial fix which is the basis for this!

## Checklist

## Types of changes

## Notice

### DIFF
--- a/vizro-core/changelog.d/20231024_101744_antony.milne_captured_callable.md
+++ b/vizro-core/changelog.d/20231024_101744_antony.milne_captured_callable.md
@@ -31,7 +31,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 ### Fixed
 
-- `CapturedCallable` now handles variadic keywords arguments (`**kwargs`) correctly ([#120](https://github.com/mckinsey/vizro/pull/120))
+- `CapturedCallable` now handles variadic keywords arguments (`**kwargs`) correctly ([#121](https://github.com/mckinsey/vizro/pull/121))
 
 <!--
 ### Security
--- a/vizro-core/changelog.d/20231024_101744_antony.milne_captured_callable.md
+++ b/vizro-core/changelog.d/20231024_101744_antony.milne_captured_callable.md
@@ -1,0 +1,41 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category.
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category.
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+
+### Fixed
+
+- `CapturedCallable` now handles variadic keywords arguments (`**kwargs`) correctly ([#120](https://github.com/mckinsey/vizro/pull/120))
+
+<!--
+### Security
+
+- A bullet item for the Security category.
+
+-->
--- a/vizro-core/src/vizro/models/types.py
+++ b/vizro-core/src/vizro/models/types.py
@@ -39,6 +39,9 @@ class CapturedCallable:
         """Creates a new CapturedCallable object that will be able to re-run `function`.
 
         Partially binds *args and **kwargs to the function call.
+
+        Raises:
+            ValueError if `function` contains positional-only or variadic positional parameters (*args).
         """
         # It is difficult to get positional-only and variadic positional arguments working at the same time as
         # variadic keyword arguments. Ideally we would do the __call__ as
@@ -201,6 +204,8 @@ class capture:
             # The more difficult case, where we need to still have a valid plotly figure that renders in a notebook.
             # Hence we attach the CapturedCallable as a property instead of returning it directly.
             # TODO: move point of checking that data_frame argument exists earlier on.
+            # TODO: also would be nice to raise errors in CapturedCallable.__init__ at point of function definition
+            #  rather than point of calling if possible.
             @functools.wraps(func)
             def wrapped(*args, **kwargs) -> _DashboardReadyFigure:
                 if "data_frame" not in inspect.signature(func).parameters:
--- a/vizro-core/src/vizro/models/types.py
+++ b/vizro-core/src/vizro/models/types.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import functools
 import inspect
-from copy import deepcopy
 from typing import Any, Dict, List, Literal, Protocol, Union, runtime_checkable
 
 from pydantic import Field, StrictBool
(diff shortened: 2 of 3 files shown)


vizro-core/tests/unit/vizro/models/test_types.py
```diff
@@ -8,113 +8,111 @@
 from vizro.models.types import CapturedCallable, capture
 
 
-@pytest.fixture
```
antonymilne, line 11: `test_varargs` and `test_positional_only` are now done in `test_invalid_parameter_kind` with `xfail` removed.

vizro-core/tests/unit/vizro/models/test_types.py
```diff
@@ -8,113 +8,111 @@
...
-def keyword_only_function():
-    def function(a, *, b):
-        return a + b
 
-    return CapturedCallable(function, 1)
+def keyword_only_function(a, *, b, c):
+    return a + b + c
 
 
-def test_keyword_only(keyword_only_function):
```
antonymilne, line 51: `keyword_only_function` is now tested on the whole suite rather than just one test here.

vizro-core/tests/unit/vizro/models/test_types.py
```diff
@@ -8,113 +8,111 @@
...
 
 @pytest.fixture
-def varkwargs_function():
-    def function(a, b=2, **kwargs):
-        return a + b + kwargs["c"]
-
-    return CapturedCallable(function, 1)
-
-
-def test_varkwargs(varkwargs_function):
```
antonymilne, line 63: `varkways_function` (now called `var_keyword_function`) is now tested on the whole suite rather than just one test here.

vizro-core/tests/unit/vizro/models/test_types.py
```diff
@@ -8,113 +8,111 @@
...
-    varkwargs_function(c=3, d=4) == 1 + 2 + 3
-
-
-@pytest.fixture
-def simple_function():
```
antonymilne, line 68: `simple_function` (now called `positional_or_keyword_function`) is tested on the same suite as before, just with order changed a bit to facilitate parametrisation of tests.

vizro-core/tests/unit/vizro/models/test_types.py
```diff
@@ -8,113 +8,111 @@
...
-        return a + b
 
-    return CapturedCallable(function, 1)
+def keyword_only_function(a, *, b, c):
+    return a + b + c
 
 
-def test_keyword_only(keyword_only_function):
-    assert keyword_only_function(b=2) == 1 + 2
+def var_keyword_function(a, **kwargs):
```
petar-qb, line 36: Do we need tests for the following cases:
`def keyword_only_function(*, a, b, c)`
 and 
`def var_keyword_function(**kwargs)`
?

antonymilne: I don't think so - they're actually simpler cases than testing unctions which have these cases mixed with positional-or-keyword arguments, and those are well tested already 🙂

vizro-core/tests/unit/vizro/models/test_types.py
```diff
@@ -8,113 +8,111 @@
...
+def positional_or_keyword_function(a, b, c):
+    return a + b + c
 
-@pytest.fixture
-def keyword_only_function():
-    def function(a, *, b):
-        return a + b
 
-    return CapturedCallable(function, 1)
+def keyword_only_function(a, *, b, c):
```
maxschulz-COL, line 32: Why is this called keyword only? Can't `a` be given as a positional argument?

vizro-core/tests/unit/vizro/models/test_types.py
```diff
@@ -8,113 +8,111 @@
...
 
         with pytest.raises(KeyError):
-            simple_function["a"]
+            captured_callable["a"]
+
+
+@pytest.mark.parametrize(
+    "captured_callable, expectation",
+    [
+        (positional_or_keyword_function, pytest.raises(TypeError, match="missing 1 required positional argument: 'c'")),
```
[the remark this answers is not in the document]
maxschulz-COL: I found it easier to read than the indirect parametrization above :P

maxschulz-COL (approved):
Really great work! I just have one concern which I may have agreed to last night not grasping the full consequence, but maybe I am wrong. Let's see!

Thanks also for the extensive tests, really a joy to read actually!

vizro-core/src/vizro/models/types.py
```diff
@@ -41,36 +40,68 @@ def __init__(self, function, /, *args, **kwargs):
...
+        parameters = inspect.signature(function).parameters
+        invalid_params = {
+            param.name
+            for param in parameters.values()
+            if param.kind in [inspect.Parameter.POSITIONAL_ONLY, inspect.Parameter.VAR_POSITIONAL]
+        }
+
+        if invalid_params:
+            raise ValueError(
+                f"Invalid parameter {', '.join(invalid_params)}. CapturedCallable does not accept functions with "
```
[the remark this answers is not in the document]
antonymilne: That is definitely still possible. The only things that aren't possible are things that wouldn't have worked well before anyway, namely positional-only arguments (things preceding `/` in the function signature) and variable positional arguments (`*args`).

vizro-core/src/vizro/models/types.py
```diff
@@ -41,36 +40,68 @@ def __init__(self, function, /, *args, **kwargs):
...
+                f"positional-only or variadic positional parameters (*args)."
+            )
+
         self.__function = function
-        self.__bound_arguments = inspect.signature(function).bind_partial(*args, **kwargs)
+        self.__arguments = inspect.signature(function).bind_partial(*args, **kwargs).arguments
+
+        # A function can only ever have one variadic keyword parameter. {""} is just here so that var_keyword_param
+        # is always unpacking a one element set.
+        (var_keyword_param,) = {
```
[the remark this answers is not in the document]
antonymilne: Yes exactly 👍 It doesn't have to be called `**kwargs` for this to work.

vizro-core/tests/unit/vizro/models/test_types.py
antonymilne: Yes, it can. The keyword-only arguments here are `b` and `c`. Maybe clearer to call this `function_including_keyword_only` I guess? Though that is a bit long. The same pattern is followed for other function names, e.g. `def var_keyword_function(a, **kwargs)` has an argument that is `**kwargs` but it has another argument also and could potentially be called `function_with_var_keyword`.

vizro-core/tests/unit/vizro/models/test_types.py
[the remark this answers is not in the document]
antonymilne: I also don't advocate that much either but again unfortunately there wasn't a good alternative here 😬

vizro-core/changelog.d/20231024_101744_antony.milne_captured_callable.md
```diff
@@ -0,0 +1,41 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+
+### Fixed
+
+- `CapturedCallable` now handles variadic keywords arguments (`**kwargs`) correctly ([#120](https://github.com/mckinsey/vizro/pull/120))
```
huong-li-nguyen, line 34:
```suggestion
- `CapturedCallable` now handles variadic keywords arguments (`**kwargs`) correctly ([#121](https://github.com/mckinsey/vizro/pull/121))
```

huong-li-nguyen (approved):
Thanks for this, looks great! 🚀  Was mostly reading through this out of interest, the tests really helped understanding it. Is it worth adding a note to the docs for the custom charts/actions or do you think the error messages suffice to provide guidance? Either way is fine for me 👍

antonymilne:
> Thanks for this, looks great! 🚀 Was mostly reading through this out of interest, the tests really helped understanding it. Is it worth adding a note to the docs for the custom charts/actions or do you think the error messages suffice to provide guidance? Either way is fine for me 👍

I think no need for anything in the narrative docs, since a user is so unlikely to ever try to do `CapturedCallable` on a function where it doesn't work that having any note on this is more likely to just confuse things. I will make a note of it in the docstring though!

vizro-core/src/vizro/models/types.py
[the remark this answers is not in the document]
maxschulz-COL: Ah got it, that's great

vizro-core/tests/unit/vizro/models/test_types.py
maxschulz-COL: As you like! I stumbled a bit though

PR #136, opened 2023-10-31
Tidy schema linting

antonymilne:
## Description

Just a small tidy following @maxschulz-COL's work in #119.
* `hatch run schema` will now include linting so no need to run `hatch run lint` manually after that.
* `hatch run lint` with no other arguments still lints all files (not just staged files)
* you can change this by overriding arguments e.g. `hatch run lint --file=...` to just lint some files
* Create a detached linting environment called `lint`

## Checklist

## Types of changes

## Notice


huong-li-nguyen (approved):
In general, looks good to me 👍  

Have we been testing this out? I remember Max previously merging 2-3 PRs on fixing the linting behaviour during the release process, because somehow it didn't behave the way we wanted it to 😄  I remember that double linting call was somehow required? @maxschulz-COL

maxschulz-COL (approved):
Thanks for implementing those 💪 Should work as intended, else we have to adjust again :)

maxschulz-COL:
> In general, looks good to me 👍
> 
> Have we been testing this out? I remember Max previously merging 2-3 PRs on fixing the linting behaviour during the release process, because somehow it didn't behave the way we wanted it to 😄 I remember that double linting call was somehow required? @maxschulz-COL

In principle with this change the linting should not be required at all, so also no double linting should be required. Given that we can always lint manually I would just test this out, and adjust in case this fails for some reason..

PR #139, opened 2023-10-31
Adjust CI scripts for package specific tasks

maxschulz-COL:
## Description
This PR adjusts the CI such that:
- linting is always done for all PRs
- checks for schema and changelog fragments are only applied if there are changes in the relevant folder of the package
- unit and integration tests are only applied if there are changes in the `src` folder of the relevant package

## Checklist

## Types of changes

## Notice


.github/workflows/checks-vizro-ai.yml
```diff
@@ -42,9 +46,6 @@ jobs:
       - name: List dependencies
         run: hatch run all.py${{ matrix.python-version }}:pip freeze
 
-      - name: Lint
```
huong-li-nguyen, line 45: Some parts here are still outdated, see this change + comment: https://github.com/mckinsey/vizro/pull/142/files#r1378559456

'hatch run docs:changelog' doesn't exist and has been updated - could you double-check?

huong-li-nguyen (approved):
The checks-vizro-ai.yml needs to be updated for the changelog command - we changed it in checks-vizro-core.yml a few days ago

.github/workflows/test-integration-vizro-core.yml
```diff
@@ -7,9 +7,13 @@ defaults:
 on:
   push:
     branches: [main]
+    paths:
+      - "vizro-core/src/**"
```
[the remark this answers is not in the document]
maxschulz-COL: Given that github is deprecting support for required workflows I will take out the requiredness: 
image

antonymilne:
> * linting is always done for all PRs

Why do we want this? Isn't it better to run linting just on the relevant vizro-core vs. vizro-ai directory?

maxschulz-COL:
> > * linting is always done for all PRs
> 
> Why do we want this? Isn't it better to run linting just on the relevant vizro-core vs. vizro-ai directory?

In principle yes, but there are a few complications:
- it is actually fairly annoying to code up (but manageable)
- what about changes outside either of those folders

Given that linting comes cheaply, and I have seen it cause bugs in mono-repos before (when only certain things are considered), I would prefer just to keep this general

antonymilne:
> In principle yes, but there are a few complications:
> 
> * it is actually fairly annoying to code up (but manageable)
> * what about changes outside either of those folders
> 
> Given that linting comes cheaply, and I have seen it cause bugs in mono-repos before (when only certain things are considered), I would prefer just to keep this general

OK, fair enough. The situations I want to avoid are:
* my PR in vizro-core fails checks due to some random linting error in vizro-ai which I didn't even touch
* my PR in vizro-core takes significantly longer to be checked due to linting running on vizro-ai

(plus the equivalent situations with vizro-core and vizro-ai swapped round, obviously)

But so long as our linting is fast and we keep the pre-commit versions pinned (so that random errors don't start popping up due to linter version changes) then the above situations shouldn't arise, so fine how you have it.

.github/workflows/lint-vizro-all.yml
```diff
@@ -0,0 +1,47 @@
...
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Hatch
+        run: pip install --upgrade hatch
+
+      - name: List dependencies
+        run: hatch run all.py${{ matrix.python-version }}:pip freeze
+
+      - name: Lint
+        run: hatch run all.py${{ matrix.python-version }}:lint
```
antonymilne, line 47: Since linting just requires `pre-commit` actually we should probably make a new `deatched=True` Hatch environment `lint` for it that doesn't include all of the dev dependencies. This way we could just do `hatch run lint:lint` on CI and it would be significantly faster. At the moment we install all the dependencies for `all.py3.8` and then run `pre-commit` that uses only one of those dependencies which is a bit silly.

Not a big deal though, so feel free to do in a separate PR another time if you prefer.

antonymilne (approved):

.github/workflows/lint-vizro-all.yml
maxschulz-COL: Hm good idea, let me see if I can incorporate that quickly still

maxschulz-COL:
> > In principle yes, but there are a few complications:
> > 
> > * it is actually fairly annoying to code up (but manageable)
> > * what about changes outside either of those folders
> > 
> > Given that linting comes cheaply, and I have seen it cause bugs in mono-repos before (when only certain things are considered), I would prefer just to keep this general
> 
> OK, fair enough. The situations I want to avoid are:
> 
> * my PR in vizro-core fails checks due to some random linting error in vizro-ai which I didn't even touch
> * my PR in vizro-core takes significantly longer to be checked due to linting running on vizro-ai
> 
> (plus the equivalent situations with vizro-core and vizro-ai swapped round, obviously)
> 
> But so long as our linting is fast and we keep the pre-commit hook versions stable (so that random errors don't start popping up due to linter version changes) then the above situations shouldn't arise, so fine how you have it.

Agreed, but since we most importantly do share the same pre-commit config (and hence linter version) the first should not arise. And the second should not really be a problem, we're still talking order of couple of seconds

.github/workflows/lint-vizro-all.yml
maxschulz-COL: On the other hand, `hatch run lint` is really used a lot, so would prefer not to have `hatch run lint:lint`. Or do you mean to keep `pre-commit` in the default environment as well?

maxschulz-COL: Let's revisit in another PR if we still want to do changes

antonymilne: @maxschulz-COL yes, you're right - I originally put `lint` in the `default` environment precisely so that we don't have to write `hatch run lint:lint`. But probably the right way to do it is to make the `hatch run lint:lint` the "correct" command with a detached `lint` environment and then put a script `"lint" = ["hatch run lint:lint"]` in the default environment to just act as a shortcut so you can still do `hatch run lint`.

PR #142, opened 2023-10-31
Move layout creation to `Dashboard`

huong-li-nguyen:
## Description
- Move layout creation to `Dashboard`
- For more context see: https://github.com/McK-Internal/vizro-internal/issues/20

## Checklist

## Types of changes

## Notice


.github/workflows/lint-vizro-ai.yml
```diff
@@ -64,9 +63,9 @@ jobs:
         run: |
           if [ -z "${{ steps.added-files.outputs.added_files }}" ];
           then
-            echo "No changelog fragment .md file within changelog.d was detected.  Run 'hatch run docs:changelog' to create such a fragment.";
+            echo "No changelog fragment .md file within changelog.d was detected.  Run 'hatch run changelog:add' to create such a fragment.";
```
antonymilne, line 66: This stuff shouldn't be there - I think you need to pull in `main`.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -94,8 +96,10 @@ def pre_build(self):
         # Note redirect_from=["/"] doesn't work and so the / route must be defined separately.
         for order, page in enumerate(self.pages):
             path = page.path if order else "/"
-            dash.register_page(module=page.id, name=page.title, path=path, order=order, layout=page.build)
-        self._create_error_page_404()
+            dash.register_page(
+                module=page.id, name=page.title, path=path, order=order, layout=partial(self._make_page_layout, page)
+            )
+        dash.register_page(module=MODULE_PAGE_404, layout=create_layout_page_404())
```
antonymilne, line 102: Putting this directly inline rather than having the separate `_create_error_page_404` is a small change but a big improvement to readability I think 👍 Very nice idea!

Just for consistency and ease of subclassing, let's rename `create_layout_page_404` to `_make_page_404_layout` and make it a `@staticmethod` inside `Dashboard` also I think.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -114,14 +118,41 @@ def build(self):
             fluid=True,
         )
 
+    def _make_page_layout(self, page):
```
antonymilne, line 121:
```suggestion
    def _make_page_layout(self, page: Page):
```
would probably make things a bit easier to understand and type check.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -114,14 +118,41 @@ def build(self):
...
+        )
+
+        # Shared across pages but slightly differ in content
+        page_title = html.H2(children=page.title, id="page_title")
+        navigation = cast(Navigation, self.navigation).build(active_page_id=page.id)
+
+        # Different across pages
+        page_content = page.build()
+        control_panel = page_content["control_panel_outer"]
+        component_container = page_content["component_container_outer"]
```
antonymilne, line 139: I think this is a very good solution 👍 It keeps the rule that `build` returns one Dash component and is indexable like a dictionary so easy to split apart.

Please could you just double check that in Dash/React there's no component other than `Div` that's specifically designed to be a "fake container" just to group things together. I don't think there is but just to be sure that `Div` is the best thing to use for this.

Note this "return one component" rule is not just our rule but [also in Dash/React](https://dash.plotly.com/react-for-python-developers):
> Notice that we're not allowed to return multiple elements from render(), but an element with children is totally fine.
They seem to use `div` for this purpose in their example also, but maybe there's more up to date React guidance somewhere that does is slightly differently.

vizro-core/src/vizro/models/_page.py
```diff
@@ -118,6 +115,11 @@ def pre_build(self):
     def build(self):
         self._update_graph_theme()
         controls_content = [control.build() for control in self.controls]
+        control_panel = (
+            html.Div(children=[*controls_content, html.Hr()], className="control_panel", id="control_panel_outer")
+            if controls_content
+            else html.Div(className="hidden", id="control_panel_outer")
```
antonymilne, line 121: I see you've changed some of the `None` returns to hidden Divs instead - just curious why is this?

antonymilne (approved):
This is a huge improvement I think! 💯 I'm much happier with this now, thank you for making the change!

Just some very small comments but it looks great.

Also while I remember - we should discuss hidden `Div`s briefly.

I think we should close the original "Take title out of dashboard" ticket and create a new one for the `grid-template-area` which has all the relevant information there just to be clear where things now stand 🙂

.github/workflows/lint-vizro-ai.yml
huong-li-nguyen: This got added yesterday when we added the vizro-ai package. It currently has an outdated linting script - that's why I just brought over the changes that you already added to `lint-vizro-core.yml` into `lint-vizro-ai.yml`, otherwise it will fail as it's trying to also look for a changelog file in vizro-ai. I already notified @maxschulz-COL and @Anna-Xiong , so they will clean up the CI workflows today :)

vizro-core/src/vizro/models/_page.py
huong-li-nguyen: It is not always required, but in this case, it is - given that we unpack the page content div later in the `Dashboard` model, the key "control_panel_outer" needs to exist in `page_content`. If we were to return None here, for example,  when there are no controls, they key would not be found and the code wouldn't run. So in this case, you need to create a hidden div that has the id. 

So we currently have the rule of returning hidden divs in case:
- You need to unpack and access the id later
- It's the return object of a build method

I thought it would be confusing to remember this, so I would prefer to always return hidden divs instead of None. WDYT?

vizro-core/src/vizro/models/_dashboard.py
huong-li-nguyen: Interesting point! I've just checked and using html.Divs to group things is correct and commonly used: https://www.shecodes.io/athena/1967-why-should-we-wrap-react-components-in-a-div

Right below that example where they argue for `html.Divs` though, they mention `react.fragments `which seem to work very similar to `html.Divs` but have the added advantage that they do not add any new element to the DOM.

Just looked through the dcc and html library and there don't seem to be a react.fragment element that we could use. Found this issue that has been closed: https://github.com/plotly/dash-html-components/issues/125 but it doesn't seem like they have added it into their library for users to use. 

So I think currently html.Divs should work fine 👍

.github/workflows/lint-vizro-ai.yml
huong-li-nguyen: Will revert changes here in favor for this PR to be merged first: https://github.com/mckinsey/vizro/pull/139

vizro-core/pyproject.toml
```diff
@@ -56,7 +56,7 @@ exclude_lines = [
   "if __name__ == .__main__.:",
   "if TYPE_CHECKING:"
 ]
-fail_under = 92
+fail_under = 91
```
huong-li-nguyen, line 59: Had to reduce as strangely this fails on the unit tests for 3.11 only: https://github.com/mckinsey/vizro/actions/runs/6718518053/job/18258559080?pr=142

We have a proper ticket to investigate further here:
https://github.com/McK-Internal/vizro-internal/issues/43

maxschulz-COL: For transparency:
- the reason this fails for python 3.11 and not others is not related to this PR: the coverage for different version is different also in previous checks (the reason seems to be the `kedro` integrations code, which skips some tests depending on the version)
- this means that this PR decreased the overall coverage just enough to push some tests below the threshold
- the culprit seems to lines 101-128 in `_dashboard.py`

Let's investigate this further in another PR, just wanted to clarify for the person that looks at this

vizro-core/src/vizro/models/_dashboard.py
antonymilne: Ah yes, a fragment is just the sort of thing I was thinking of a would have been perfect for this sort of thing! Pity it never made it into `dcc`. Thanks for checking.

In this case there's no extra element cluttering the DOM because we just take the children out of the `html.Div` straight away so not a big deal. In other cases it would have been nice to have a `dcc.Fragment` to avoid the extra element but good to know that `html.Div` is still the best way to do it 👍

vizro-core/src/vizro/models/_page.py
antonymilne: Ah ok - makes sense in this case, but do we also have a rule for the ones where we don't need to access the `id` later on. Like for `dashboard_title` - we used to leave this as `None` I think but now make a hidden div instead?

I actually think there will be a better way to do all this throughout Vizro which is the "we should discuss hidden `Div`s briefly" thing I mentioned in my previous review, but let's leave it for a separate PR. Happy to go with whatever works here for now.

vizro-core/pyproject.toml
antonymilne: Kedro 0.18.13 onwards support Python 3.11 so we should be able to remove our special skipping of kedro in that environment now anyway.

maxschulz-COL (approved):
Managed to look at it all now! I really like it, and it's indeed a huge improvement.

I have been thinking: given that our coverage decreased with this PR, e.g. for lines 101-121 of `_dashboard.py`, could this be that it is linked to the fact that we are using `partial` there? And hence the `_make_page_layout` function is actually never called in any unit tests!

If that's the case, then at least we know where to fix it - but given that time is of the essence I would not focus on it more here

vizro-core/pyproject.toml
huong-li-nguyen: Created a new issue for this here: https://github.com/McK-Internal/vizro-internal/issues/363

huong-li-nguyen:
> Managed to look at it all now! I really like it, and it's indeed a huge improvement.
> 
> I have been thinking: given that our coverage decreased with this PR, e.g. for lines 101-121 of `_dashboard.py`, could this be that it is linked to the fact that we are using `partial` there? And hence the `_make_page_layout` function is actually never called in any unit tests!
> 
> If that's the case, then at least we know where to fix it - but given that time is of the essence I would not focus on it more here

That could be it! We've linked to this PR in the relevant issue already, so this can be taken as reference

PR #149, opened 2023-11-06
Update docs on kedro integration and deployment

antonymilne:
## Description

A few corrections, clarification and additions to the docs.

## Checklist

## Types of changes

## Notice

### DIFF
--- a/vizro-core/docs/pages/user_guides/run.md
+++ b/vizro-core/docs/pages/user_guides/run.md
@@ -2,9 +2,9 @@
 
 This guide shows you how to launch your dashboard in different ways. By default, your dashboard apps run on localhost.
 
-## Default built-in web server in flask
+## Default built-in Flask web server
 
-!!! example "Default flask server"
+!!! example "Default built-in Flask web server"
     === "app.py"
         ```py
         from vizro import Vizro
@@ -34,8 +34,8 @@ Dash is running on http://127.0.0.1:8050/
  * Debug mode: on
 ```
 
-## Launch it in jupyter environment
-The dashboard application can be launched in jupyter environment in `inline`, `external`, and `jupyterlab` mode.
+## Jupyter
+The dashboard application can be launched in a Jupyter environment in `inline`, `external`, and `jupyterlab` mode.
 !!! example "Run in jupyter notebook in inline mode"
     === "app.ipynb"
         ```py linenums="1"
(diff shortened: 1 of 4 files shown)


vizro-core/docs/pages/user_guides/run.md
```diff
@@ -59,9 +59,17 @@ The dashboard application can be launched in jupyter environment in `inline`, `e
 - you can specify `jupyter_mode="external"` and a link will be displayed to direct you to the localhost where the dashboard is running.
 - you can use tab mode by `jupyter_mode="tab"` to automatically open the app in a new browser
 
-## Launch it via Gunicorn
+## Deployment of Vizro app
```
lingyielia, line 62: To be very accurate, this instruction is still more focus on "running the app with gunicorn" than how to deploy the app 🤔 

Shall we phrase it as something like, for deployment, it's recommended to run the app with gunicorn, and here is how

lingyielia (approved):
LGTM!

vizro-core/docs/pages/user_guides/run.md
maxschulz-COL: Agreed, in the current form, this point breaks the reading flow. This entire guide has the overarching title of "How to launch the dashboard".

I was going to suggest to write a separate section on how to deploy the dashboard, when I realised this is exactly what we had 😅 .

I would suggest keeping the old structure, but to maybe enhance it with the additional links and so on.

Overall then I think the title of this section should change back to "Gunicorn", and ideally the one above to simply "Jupyter Environment". Then the sections alrady contain the answer to the user guide question, which I think is an improvement.

maxschulz-COL (changes requested):
Lgtm! Thanks for improving the guide on kedro integrations! Let's discuss the changes on `run.md` where I think we need to consider the overall reading flow a little

vizro-core/docs/pages/user_guides/run.md
antonymilne: See what you think of the new version 🙂 I even did the special popup thing:
image

vizro-core/docs/pages/user_guides/run.md
```diff
@@ -80,24 +81,22 @@ The dashboard application can be launched in jupyter environment in `inline`, `e
 
         dashboard = vm.Dashboard(pages=[page])
         app = Vizro().build(dashboard)
-        server = app.dash.server
+        server = app.dash.server # (1)!
         ```
-You need to expose the server via `app.dash.server` in order to use gunicorn as your wsgi here.
-Run it via
+
+        1.  Expose the underlying Flask app through `app.dash.server`.
```
maxschulz-COL, line 87: ❤️

maxschulz-COL (approved):
Like it! Thanks for applying the changes :)

PR #151, opened 2023-11-08
Fix assets to work when hosted at subdirectory

antonymilne:
## Description

A Vizro dashboard might be accessible through 3 different sorts of route:
1. root level, e.g. http://www.example.com/
2. subdomain, e.g. http://subdomain.example.com/ (as with Heroku for example)
3. subdirectory, e.g. http://www.example.com/path/

**Before this change: 1 and 2 already worked correctly but 3 did not.
Now: all should work correctly.**

I've tested 1, will test 2 soon. @maxschulz-COL please can you test 3? I've bumped the version here to `0.1.6.dev2` so you can check you're using the right version when you deploy.

> [!IMPORTANT]  
> We now have a "proper" way to alter `assets_folder` from its default rather than going through `Vizro._user_assets_folder`, since I've exposed `**kwargs` in `Vizro()`.

Here's the rules for how we should write code so that paths are always correctly formed:
- always use `dash.get_relative_path` to link to pages with `href` (see `_make_page_404_layout` example link)
- always use `dash.get_relative_path(f"/{STATIC_URL_PREFIX}/..")` to refer to built-in assets in the `static` folder (see `_make_page_404_layout` example `html.Img`)
- always use `dash.get_asset_url` to refer to things in the user `assets` folder (we don't have any examples of this at the moment outside docs)

Still to do:
- [x] test case 2
- [x] test case 3 @maxschulz-COL 
- [x] update changelog
- [x] update deployment docs
- [x] update docs with the above rules
- [x] update docs about how to specify assets folder and how we can now do `**kwargs` in `Vizro.__init__`

## Checklist

## Types of changes

## Notice


vizro-core/docs/pages/user_guides/actions.md
```diff
@@ -63,8 +63,7 @@ a result, when a dashboard user now clicks the button, all data on the page will
 
         dashboard = vm.Dashboard(pages=[page])
 
-        if __name__ == "__main__":
-            Vizro().build(dashboard).run()
+        Vizro().build(dashboard).run()
```
antonymilne, line 66: This was unnecessarily complicated before so I simplified everywhere.

vizro-core/src/vizro/_vizro.py
```diff
@@ -88,19 +98,10 @@ def _reset():
...
-        for f in sorted(files):
-            path = os.path.join("/" + url_prefix, base, f) if base else os.path.join("/" + url_prefix, f)
-            extension = os.path.splitext(f)[1]
-            if extension == ".js":
-                _vizro_js.append({"src": path, "type": "module"})
-            elif extension == ".css":
-                _vizro_css.append(path)
-    return _vizro_js, _vizro_css
+    @staticmethod
+    def _get_external_assets(new_path: str, folder: Path, extension: str) -> List[str]:
```
antonymilne, line 102: The output of this is identical to `_append_styles` but it uses pathlib, which we should generally prefer to `os.path`.

vizro-core/src/vizro/_vizro.py
```diff
@@ -16,25 +16,35 @@
...
+    def __init__(self, **kwargs):
         """Initializes Dash."""
-        _js, _css = _append_styles(self._lib_assets_folder, STATIC_URL_PREFIX)
-        self.dash = dash.Dash(
-            use_pages=True,
-            pages_folder="",
-            external_scripts=_js,
-            external_stylesheets=_css,
-            assets_folder=self._user_assets_folder,
+        self.dash = dash.Dash(**kwargs, use_pages=True, pages_folder="")
```
antonymilne, line 21: Note this takes `**kwargs` now!

maxschulz-COL, line 21: Are we allowing `**kwargs` now so that users can set things like `url_base_pathname`? And what if for some reason they want to provide one of the two other arguments `use_pages` and `pages_folder`. Should we not intercept this with a custom error message?

vizro-core/src/vizro/_vizro.py
```diff
@@ -88,19 +98,10 @@ def _reset():
...
-                _vizro_js.append({"src": path, "type": "module"})
-            elif extension == ".css":
-                _vizro_css.append(path)
-    return _vizro_js, _vizro_css
+    @staticmethod
+    def _get_external_assets(new_path: str, folder: Path, extension: str) -> List[str]:
+        """Returns a list of paths to assets with given extension in folder, prefixed with new_path.
+        e.g. with new_path="/vizro", extension="css", folder="/path/to/vizro/vizro-core/src/vizro/static",
+        we will get ["/vizro/css/accordion.css", "/vizro/css/button.css", ...]
+        """
```
maxschulz-COL, line 106:
```suggestion
        """Returns a list of paths to assets with given extension in folder, prefixed with new_path.
        
        E.g. with `new_path="/vizro"`, `extension="css"` and `folder="/path/to/vizro/vizro-core/src/vizro/static"`,
        we will get `["/vizro/css/accordion.css", "/vizro/css/button.css", ...]`.
        """
```

vizro-core/src/vizro/_vizro.py
```diff
@@ -16,25 +16,35 @@
...
+        # Serve all assets (including files other than css and js) that live in vizro_assets_folder at the
+        # route /vizro. Based on code in Dash.init_app that serves assets_folder. This respects the case that the
+        # dashboard is not hosted at the root of the server, e.g. http://www.example.com/dashboard/vizro.
+        blueprint_prefix = self.dash.config.routes_pathname_prefix.replace("/", "_").replace(".", "_")
+        self.dash.server.register_blueprint(
+            flask.Blueprint(
+                f"{blueprint_prefix}vizro_assets",
+                self.dash.config.name,
+                static_folder=vizro_assets_folder,
+                static_url_path=static_url_path,
```
maxschulz-COL, line 44: Just to understand, this now works, as it dynamically adjusts with the `self.dash.config.requests_pathname_prefix`, something that it did not do before right?

maxschulz-COL (commented):
Great stuff! Couple of QQ

vizro-core/src/vizro/_vizro.py
antonymilne: Yes exactly!

vizro-core/src/vizro/_vizro.py
maxschulz-COL: Yh that was the case I was thinking off. It reminds me of the old decorator days, and struck me as an unintuitive error message because they did not provide the same keyword twice. Agree though that it is very unlikely, but maybe let's settle once and for all the policy on this case.

And alternative would be `setdefault` no? In fact, should this not be the path forward, as we do that similarly in table?

```py
kwargs = kwargs.setdefault("use_pages",True)
kwargs = kwargs.setdefault("pages_folder","")
```

antonymilne: This wouldn't make sense here because Vizro relies very heavily on Dash pages e.g. if you set `use_pages = False` then nothing will work at all. Overriding these options should not be possible.

If you think it's helpful then I can put in something which checks if the user provided `use_pages` or `pages_folder` and emits a different exception that explains it's invalid, but honestly I think this case is so unlikely to occur that it's not worth it.

maxschulz-COL: Hm ok, then ignore this one. I suppose someone every attempting to change those should actually be clever enough to read the source code

vizro-core/changelog.d/20231108_230816_antony.milne_fix_assets_pathname.md
```diff
@@ -0,0 +1,48 @@
...
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Fixed
```
maxschulz-COL, line 38: No fix?

antonymilne: Added now :)

vizro-core/docs/pages/user_guides/assets.md
```diff
@@ -19,11 +19,7 @@ The user-provided `assets` folder thus always takes precedence.
 ```
 
 ## Adding static images
-We leverage Dash's underlying functionalities to embed images into the app.
-For more information, see [here](https://dash.plotly.com/dash-enterprise/static-assets?de-version=5.1#embedding-images-in-your-dash-apps).
-
-For example, you can leverage the `dash.get_asset_url()` function in your custom components, such that any provided path does not require `assets` as a prefix in the relative path anymore.
-
+We use [Dash's underlying functionalities](https://dash.plotly.com/dash-enterprise/static-assets#embedding-images-in-your-dash-apps) to embed images into the app). You should always use `dash.get_asset_url` to refer to files inside your assets folder to ensure that the correct URL is generated in both development and production environments. For example, the file `collections.svg` in the above example would be referenced as `dash.get_asset_url("images/icons/collections.csv")`.
```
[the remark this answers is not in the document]
maxschulz-COL: One thing that I have been wondering about is what user would actually use `dash.get_asset_url("images/icons/collections.csv")`. 

As far as I can tell, the most use of images is inside a text card, where the guide currently says:

> Use markdown to render your image by using one of the following syntax:
>- Relative Image URL: !\[Image ALT text](/path/to/image.png)
>- Absolute Image URL: !\[Image ALT text](https://XXXXXX)

Do we need to change that guide? Or when do I apply the above, and when do I use `dash.get_asset_url`?

maxschulz-COL (approved):
LG from my side!

vizro-core/docs/pages/user_guides/assets.md
```diff
@@ -164,17 +155,14 @@ To achieve this, do the following:
 
     **Order of CSS being served to app**
 
-    1. Dash styling sheets
-    2. Vizro external styling sheets
-    3. User assets folder
-        - CSS/JS Files
-        - Folders
-           - CSS/JS Files
+    1. Dash built-in stylesheets
```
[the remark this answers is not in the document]
antonymilne: Done!

vizro-core/docs/pages/user_guides/assets.md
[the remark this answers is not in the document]
maxschulz-COL: Great! If you add it to an issue/keep a note of this then perfect!

Joseph-Perkins (approved):

huong-li-nguyen:
This looks great - thanks for fixing and solving the mystery! 🚀 🔥 

Looks very clean now - I also like that you don't have to provide an absolute path to the `assets_folder` anymore. This has been confusing for some users in the past!

PR #152, opened 2023-11-08
Flask caching per dataset

lingyielia:
## Description
Build on top of https://github.com/mckinsey/vizro/pull/133/commits/d6c22150bd5d58981c828deecf5993d9f231abc2

Options for configuring per-dataset arguments:
# 1.
```
data_manager["gapminder"] = VizroDataSet(
    retrieve_gapminder,
    timeout=600,
    # unless=lambda: True
)
data_manager["gapminder2"] = VizroDataSet(
    retrieve_gapminder,
    timeout=600
)
```

# 2.
```
data_manager["gapminder"] = retrieve_gapminder
print("to update cache config")
data_manager["gapminder"].set_cache_config(timeout=6000)

data_manager["gapminder2"] = retrieve_gapminder
print("to update cache config")
data_manager["gapminder2"].set_cache_config(unless=lambda: True)
```

## Checklist

## Types of changes

## Notice

### DIFF
--- a/vizro-core/src/vizro/managers/_data_manager.py
+++ b/vizro-core/src/vizro/managers/_data_manager.py
@@ -16,12 +16,12 @@ DatasetName = str
 pd_LazyDataFrame = Callable[[], pd.DataFrame]
 
 
-class VizroDataSet:
+class _Dataset:
     """A dataset with flask-caching config attached.
 
     Examples:
         >>> import plotly.express as px
-        >>> data_manager["iris"] = VizroDataSet(lambda: pd.DataFrame(), timeout=300)
+        >>> data_manager["iris"] = _Dataset(lambda: pd.DataFrame(), timeout=300)
         >>> data_manager.__lazy_data["iris"]._cache_arguments == {"timeout": 300}
     """
     def __init__(self, data: pd_LazyDataFrame, timeout: Optional[int] = None, unless: Optional[Callable] = None):
@@ -48,14 +48,14 @@ class DataManager:
     _cache = Cache(config={"CACHE_TYPE": "SimpleCache"})
 
     def __init__(self):
-        self.__lazy_data: Dict[DatasetName, VizroDataSet] = {}
+        self.__lazy_data: Dict[DatasetName, _Dataset] = {}
         self.__original_data: Dict[DatasetName, pd.DataFrame] = {}
         self.__component_to_original: Dict[ComponentID, DatasetName] = {}
         self._frozen_state = False
 
     # happens before dashboard build
     @_state_modifier
-    def __setitem__(self, dataset_name: DatasetName, data: Union[pd.DataFrame, pd_LazyDataFrame, VizroDataSet]):
+    def __setitem__(self, dataset_name: DatasetName, data: Union[pd.DataFrame, pd_LazyDataFrame]):
         """Adds `data` to the `DataManager` with key `dataset_name`.
 
         This is the only user-facing function when configuring a simple dashboard. Others are only used internally
@@ -65,22 +65,16 @@ class DataManager:
             raise ValueError(f"Dataset {dataset_name} already exists.")
 
         if callable(data):
-            data = VizroDataSet(data)
+            self.__lazy_data[dataset_name] = _Dataset(data)
         elif isinstance(data, pd.DataFrame):
             self.__original_data[dataset_name] = data
-        elif isinstance(data, VizroDataSet):
-            pass
         else:
             raise TypeError(
                 f"Dataset {dataset_name} must be a pandas DataFrame or callable that returns pandas DataFrame."
-                f"Additionally, it can be a VizroDataSet object."
             )
 
-        if isinstance(data, VizroDataSet):
-            self.__lazy_data[dataset_name] = data
-
-    def __getitem__(self, dataset_name: DatasetName) -> VizroDataSet:
-        """Returns the `VizroDataSet` object associated with `dataset_name`."""
+    def __getitem__(self, dataset_name: DatasetName) -> _Dataset:
+        """Returns the `_Dataset` object associated with `dataset_name`."""
         if dataset_name not in self.__original_data and dataset_name not in self.__lazy_data:
             raise KeyError(f"Dataset {dataset_name} does not exist.")
         if dataset_name in self.__original_data:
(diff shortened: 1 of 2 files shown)


vizro-core/src/vizro/managers/_data_manager.py
```diff
@@ -16,6 +16,26 @@
 pd_LazyDataFrame = Callable[[], pd.DataFrame]
 
 
+class VizroDataSet:
```
antonymilne, line 19:
```suggestion
class _Dataset:
```
Let's leave private for now (and maybe forever depending on how the API evolves) 🙂

vizro-core/src/vizro/managers/_data_manager.py
```diff
@@ -28,14 +48,14 @@ class DataManager:
...
-        self.__lazy_data: Dict[DatasetName, pd_LazyDataFrame] = {}
+        self.__lazy_data: Dict[DatasetName, VizroDataSet] = {}
         self.__original_data: Dict[DatasetName, pd.DataFrame] = {}
         self.__component_to_original: Dict[ComponentID, DatasetName] = {}
         self._frozen_state = False
 
     # happens before dashboard build
     @_state_modifier
-    def __setitem__(self, dataset_name: DatasetName, data: Union[pd.DataFrame, pd_LazyDataFrame]):
+    def __setitem__(self, dataset_name: DatasetName, data: Union[pd.DataFrame, pd_LazyDataFrame, VizroDataSet]):
```
antonymilne, line 58:
```suggestion
    def __setitem__(self, dataset_name: DatasetName, data: Union[pd.DataFrame, pd_LazyDataFrame]):
```

antonymilne: Since it'll be private, let's not expect the user to ever type `data_manager["iris"] = _Dataset` for now at least.

vizro-core/src/vizro/managers/_data_manager.py
```diff
@@ -45,14 +65,30 @@ def __setitem__(self, dataset_name: DatasetName, data: Union[pd.DataFrame, pd_La
...
+        if isinstance(data, VizroDataSet):
+            self.__lazy_data[dataset_name] = data
+
+    def __getitem__(self, dataset_name: DatasetName) -> VizroDataSet:
+        """Returns the `VizroDataSet` object associated with `dataset_name`."""
+        if dataset_name not in self.__original_data and dataset_name not in self.__lazy_data:
+            raise KeyError(f"Dataset {dataset_name} does not exist.")
+        if dataset_name in self.__original_data:
+            # no cache available
+            raise ValueError(f"Dataset {dataset_name} is not lazy.")
```
antonymilne, line 88: I'm not sure I like these switches around original data vs. lazy data but let's leave like this for now and then we can possibly change to Idea 2 after we merge this into #133. That will simplify things quite a bit I think.

vizro-core/src/vizro/managers/_data_manager.py
```diff
@@ -69,19 +105,16 @@ def _add_component(self, component_id: ComponentID, dataset_name: DatasetName):
         self.__component_to_original[component_id] = dataset_name
 
     def _load_lazy_data(self, dataset_name: DatasetName) -> pd.DataFrame:
-        self._cache_dataset_arguments = {"iris": {"timeout": 50}, "other_dataset": {"timeout": 100}}
+        logger.debug("reloading lazy data: %s", dataset_name)
 
-        @self._cache.memoize(**self._cache_dataset_arguments.get(dataset_name, {}))
+        @self._cache.memoize(**self.__lazy_data[dataset_name]._cache_arguments)
```
antonymilne, line 110:
```suggestion
        @self._cache.memoize(**self[dataset_name]._cache_arguments)
```

vizro-core/src/vizro/managers/_data_manager.py
```diff
@@ -69,19 +105,16 @@ def _add_component(self, component_id: ComponentID, dataset_name: DatasetName):
...
-        def inner():
-            self.__lazy_data[dataset_name]()
-
-        """Returns the original data for `dataset_name`."""
-        logger.debug("reloading lazy data: %s", dataset_name)
-        time.sleep(2.0)
-        return inner()
+        def inner(dataset):
+            time.sleep(2.0)
+            return self.__lazy_data[dataset].data()
```
antonymilne, line 116:
```suggestion
            return self[dataset].data()
```

vizro-core/src/vizro/managers/_data_manager.py
```diff
@@ -69,19 +105,16 @@ def _add_component(self, component_id: ComponentID, dataset_name: DatasetName):
...
         # timeout=0.000001) and doesn't update cache.
         # timeout and unless need to depend on dataset_name
-        def inner():
-            self.__lazy_data[dataset_name]()
-
-        """Returns the original data for `dataset_name`."""
-        logger.debug("reloading lazy data: %s", dataset_name)
-        time.sleep(2.0)
-        return inner()
+        def inner(dataset):
```
antonymilne, line 114: Why is this argument needed? I guess there was some error without it but not sure what? 🤔

vizro-core/src/vizro/managers/_data_manager.py
```diff
@@ -16,6 +16,26 @@
...
+        self.data = data
+        self._cache_arguments: Dict[str, int] = {}
+        self.set_cache_config(timeout, unless)
+
+    def set_cache_config(self, timeout: int = None, unless: Callable = None):
+        """Sets the cache configuration for the dataset."""
+        self._cache_arguments["timeout"] = timeout
+        self._cache_arguments["unless"] = unless
+        logger.debug(f"set_cache_config: {self._cache_arguments}")
+
```
antonymilne, line 37: Let's change this so that a user can directly do `data_manager["iris"]._cache_arguments = {"timeout": 10}` rather than having a `set_cache_config` function. That will be much nicer and more Pythonic.

vizro-core/src/vizro/managers/_data_manager.py
```diff
@@ -28,14 +48,14 @@ class DataManager:
     _cache = Cache(config={"CACHE_TYPE": "SimpleCache"})
```
antonymilne, line 48: Now that we are using `self` in the memoize decorator, we should be able to move this into the `__init__` method so it's an instance rather than class property, which is really where it should always have gone.

antonymilne (approved):
This looks great, thank you for fleshing it out! 

I've thought through some other options and think the solution we have here is the right one, but just to explain: actually options 1 and 2 are basically the same, it's just a question of whether we expose `_Dataset` or not.

As it stands, after my proposed changes, the user would do this:
```
data_manager["iris"] = lambda: pd.DataFrame()
data_manager["iris"]._cache = {"timeout": 10}
data_manager["iris"]._something_else_in_future = ...
```

Under the hood this is creating the `_Dataset` instance in `data_manager.__setitem__` and fetching it in `data_manager.__getitem__` but the user doesn't need to know anything about it.

In the future, maybe `_Dataset` would become public in which case you could set configuration at the same time as when you set the data, like this:
```
data_manager["iris"] = Dataset(lambda: pd.DataFrame, cache={"timeout": 10}, something_else_in_future=...)
```

So if we keep `_Dataset` private for now then it's hidden from the user and the public interface is that you set `data_manager["iris"]` and then alter properties of that, but in future maybe you'll be able to do it in one step.

There's some more changes I think we'll make here, but if you implement the stuff in my suggestions then feel free to merge and then we can continue work in the `dev/flask_caching-AM` branch 🙂

vizro-core/src/vizro/managers/_data_manager.py
lingyielia: Because `@self._cache.memoize` is decorating inner. `dataset` needs to be there so memoize knows to cache the function + argument combination

antonymilne: Got it, thank you! 👍

PR #166, opened 2023-11-13
Fix flickering graphs

antonymilne:
## Description

There's currently a smaller flicker that's detectable when you change theme and then screens. I _think_ I've been able to remove this by changing `create_empty_fig` to use a transparent background.

**Before**

**Now**
Hopefully no flicker.

~@petar-qb and/or @nadijagraca please could you give this a careful test to see if (a) you can see the flickering currently and (b) that this has fixed it? I had several failed attempts at fixing this so would be great to double check this one works!~

@petar-qb checked and all fixed now.

Also in this PR:
- remove some dead code
- added some more comments just documenting my understanding of the theme update stuff because I found it confusing before

## Checklist

## Types of changes

## Notice


vizro-core/src/vizro/actions/__init__.py
```diff
@@ -9,12 +9,3 @@
 
 # Please keep alphabetically ordered
 __all__ = ["export_data", "filter_interaction"]
-
```
antonymilne, line 12: This was not used anywhere

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -23,10 +23,6 @@
 logger = logging.getLogger(__name__)
 
 
-def update_theme(on: bool):
```
antonymilne, line 26: This was not used anywhere

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -60,13 +56,6 @@ def validate_navigation(cls, navigation, values):
             return Navigation(pages=[page.id for page in values["pages"]])
         return navigation
 
-    def __init__(self, *args, **kwargs):
```
antonymilne, line 63: @petar-qb I'm pretty sure we don't want this any more?

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -267,7 +267,4 @@ def _get_modified_page_figures(
         outputs[target] = model_manager[target](  # type: ignore[operator]
             data_frame=filtered_data[target], **parameterized_config[target]
         )
-        if hasattr(outputs[target], "update_layout"):
```
antonymilne, line 270: I've moved this to `Graph.__call__` but let's discuss @petar-qb

vizro-core/src/vizro/models/_page.py
```diff
@@ -136,18 +136,25 @@ def build(self):
...
+        # not good.
+        # Putting graphs as inputs here would be a nice way to trigger the theme change automatically so that we don't
+        # need the update_layout call in Graph.__call__, but this results in an extra callback and the graph
+        # flickering.
+        # TODO: consider making this clientside callback and then possibly we can remove the update_layout in
+        #  Graph.__call__ without any flickering.
+        # TODO: consider putting this as dashboard level like with _update_theme (though possibly pointless because you
+        #  don't see graphs on other pages and then when you change page it always redraws graphs anyway).
+        # TODO: consider putting the Graph-specific logic here in the Graph model itself (whether clientside or
+        #  serverside) to keep the code here abstract.
```
petar-qb, line 149: I think only the first `TODO` looks like a candidate for improvement.

For the second one, you've already described why it's not a good idea.

I can't see how to implement the last TODO you wrote without applying all the filters, parameters again (which is slow).

petar-qb (approved):
These changes look really good!
I guess there's still a lot of work in purpose to adjust tests here.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -56,6 +45,16 @@ def __call__(self, **kwargs):
         # Remove top margin if title is provided
         if fig.layout.title.text is None:
             fig.update_layout(margin_t=24)
+
+        try:
+            # if "theme_selector" in ctx.states # Probably want this here given code should work outside Dashboard
+            # for now
+            fig.update_layout(template="vizro_dark" if ctx.states["theme_selector.on"] else "vizro_light")
```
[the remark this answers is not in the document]
antonymilne: Thanks for the review @petar-qb!

You're ok with the approach, but do you think it's actually an improvement on the current solution so worth making the change? My current thinking is "probably yes" but I would do it in a separate PR.

vizro-core/src/vizro/models/_page.py
antonymilne: On the last TODO, my thinking was that the graph-specific logic like `patched_figure["layout"]["template"] = themes.dark if theme_selector_on else themes.light` should ideally live in the `Graph` model, and we would somehow try to simplify the `ifinstance(component, Graph)`. The callback itself would still live here like it does now.

It would just be a bit of refactoring, not changing the number of callbacks or anything like that.

vizro-core/src/vizro/models/_components/graph.py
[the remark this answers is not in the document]
petar-qb: I agree, let's do this change in a separate PR.

lingyielia (approved):
LGTM! Also tested on my local and server. Works very well 👍

PR #172, opened 2023-11-13
vizro_ai support for whl file check

l0uden:
## Description

whl file check fix for vizro_ai

## Checklist

## Types of changes

## Notice


.github/workflows/check-release.yml
```diff
@@ -81,11 +81,15 @@ jobs:
           password: ${{ env.PYPI_TOKEN }}
       - name: Check correct package uploaded to PyPI
         run: |
+          if [ "${{ needs.check-version.outputs.package_name }}" == "vizro-ai" ]; then
+            export package_name=vizro_ai
+          else export package_name=${{ needs.check-version.outputs.package_name }}
```
lingyielia, line 86: Thanks for quickly fix this 🥇 
1. Do you think it make sense to move this renaming to the beginning of this script? (If there is a possibility that other jobs might need this variable "vizro_ai" too)
2. Maybe we can name it as something like `package_name_normalized` instead of `package_name` to reduce confusion with `needs.check-version.outputs.package_name` when reading the code

lingyielia (approved):

antonymilne (approved):
Agree with @lingyielia's comments that we should make this clearer but otherwise looks good, thank you for the fix!

.github/workflows/check-release.yml
antonymilne: We should call this `dist_filename` or similar rather than `package_name_normalized` since the normalised package name on PyPI is actually the version with `-` (see https://packaging.python.org/en/latest/specifications/name-normalization/#normalization).

l0uden: Thanks @lingyielia and @antonymilne for the comments.
Agreed and did as you've proposed.

PR #174, opened 2023-11-14
Move update of graph theme to `Graph.__call__`

antonymilne:
## Description

As prototyped in #166, a bit of further refactoring with no change in functionality.

**Before**: the code to update graph theme lived in `_callback_mapping_utils.py`.
**Now**: it lives in `Graph`, plus a couple of other improvements.

Why is this (probably) better?
- callbacks no longer need to all carry around `ctd_theme`
- callbacks are not directly tied to `Graph` implementation
- the duplicated logic between the `update_graph_theme` callback and the `update_layout` call is centralised to a single `Graph._update_theme` method
- the `update_theme` callback is generic so in theory works for components other than `Graph` so long as they define a `_update_theme` method (probably not very useful in practice)

Possible drawbacks:
- `Graph.__call__` is coupled to the Dash callback context. This was actually aways my original intention, but it was never made explicit before, and the current actions implementation works somewhat differently. To be seen whether it causes problems

Future work:
- (small/easy) see if we can get rid of the `_update_theme` call in `Graph.__call__` altogether by making `update_theme` a clientside callback. This would be even nicer, but the current solution is perfectly performant so it's not a big problem
- (bigger/difficult) consider whether more functionality should move from callbacks into `__call__`

## Checklist

## Types of changes

## Notice


vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
```diff
@@ -154,7 +154,7 @@ def _get_action_callback_inputs(action_id: ModelID) -> Dict[str, Any]:
             if "filter_interaction" in include_inputs
             else []
         ),
-        "theme_selector": (State("theme_selector", "on") if "theme_selector" in include_inputs else []),
+        "theme_selector": State("theme_selector", "on") if "theme_selector" in include_inputs else [],
```
antonymilne, line 157: Debatable whether we really want to keep this key here at all or just not set it at all `if "theme_selector" not in include_input`, but I think current solution is ok.

vizro-core/tests/unit/vizro/actions/test_on_page_load_action.py
antonymilne, line 1: I actually didn't need to alter any tests across the other files, and the tests in this file were working great in verifying that the refactoring didn't break anything. Great job @petar-qb for writing these originally 🙏

vizro-core/src/vizro/actions/_on_page_load_action.py
antonymilne, line 1: I actually didn't need to alter any tests across the other files, and the tests in this file were working great in verifying that the refactoring didn't break anything. Great job @petar-qb for writing these originally 🙏

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -44,6 +46,16 @@ def __call__(self, **kwargs):
...
+
+        # Possibly we should enforce that __call__ can only be used within the context of a callback, but it's easy
+        # to just swallow up the error here as it doesn't cause any problems.
+        try:
+            # At the moment theme_selector is always present so this if statement is redundant, but possibly in
+            # future we'll have callbacks that do Graph.__call__() without theme_selector set.
+            if "theme_selector" in ctx.args_grouping:
+                fig = self._update_theme(fig, ctx.args_grouping["theme_selector"]["value"])
+        except MissingCallbackContextException:
+            logger.info("fig.update_layout called outside of callback context.")
```
petar-qb, line 58: I agree we your comments. Let's leave this logic here.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -81,3 +93,10 @@ def build(self):
             color="grey",
             parent_className="loading-container",
         )
+
+    @staticmethod
+    def _update_theme(fig: go.Figure, theme_selector: bool):
+        # Basically the same as doing fig.update_layout(template="vizro_light/dark") but works for both the call in
+        # self.__call__ and in the update_graph_theme callback.
+        fig["layout"]["template"] = themes.dark if theme_selector else themes.light
+        return fig
```
petar-qb, line 102: I hope you're right that this part of the code is the same as doing: `fig.update_layout(template="vizro_light/dark")` even from the performances point of view (at least almost the same).

It's true assuming that the `pio` object is the `server` global variable that probably just change the key such as `"vizro_light"` with the value `themes.light` in some point of moment (don't exactly know when).

If the `pio` object makes these changes somehow in the `frontend` side (by storing the `pio` mapper on the client side and making `"vizro_light"` -> `themes.light` change on the frontend), then I guess we should go with this approach (`fig.update_layout(template="vizro_light/dark")`) because of performances.

petar-qb (approved):
Great tidying up! 🚀

vizro-core/src/vizro/models/_components/graph.py
antonymilne: Yeah, this is something I already checked but you are absolutely right to think of it!

These things are exactly the same:
```
px.scatter([], template=vizro._themes.dark)
px.scatter([], template="vizro_dark")
px.scatter([]).update_layout(template=vizro._themes.dark)
px.scatter([]).update_layout(template="vizro_dark")
```

If you supply `template` as a string `"vizro_dark"`, all plotly does it look for key `"vizro_dark"` in the `pio.defaults` dictionary and return the corresponding `vizro._themes.dark` object (type `Template`, basically a dict). This is all done serverside, no frontend magic involved. So actually the new solution is very slightly more performant because we don't need to do the dictionary lookup (if possible I would have done it using the strings though, since they are "more public" than the internal `Template` objects).

What it means is that if we did want to make this a clientside callback, we basically have to store the JSON content of `vizro_dark` and `vizro_light` themes clientside in `dcc.Store` to switch between them.

Joseph-Perkins (approved):

vizro-core/src/vizro/models/_components/graph.py
petar-qb: Yeah, I see. I just tested it and everything is super clear to me now. Thanks for the explanation 😄.

> What it means is that if we did want to make this a clientside callback, we basically have to store the JSON content of vizro_dark and vizro_light themes clientside in dcc.Store to switch between them.

Yes, that's the idea I have. Do we want to load client machine memory for a little (~50KB) and to optimise time execution when the theme is changed is the question we need to consider.

PR #186, opened 2023-11-29
Enable tooltips in `NavLink` and add minor CSS fixes for `Navigation`

nadijagraca:
## Description
Minor css fixes on navlink and accordion components:
- added lined between each accordion item (the line is not new design, but rather old that got lost in the refactoring)
- adjusted highlighting of active icon
- implemented and styled tooltip component
## Screenshot
Screenshot 2023-11-29 at 19 32 55

## Checklist

## Types of changes

- [] New feature (non-breaking change which adds functionality)

## Notice


vizro-core/examples/default/app.py
```diff
@@ -526,7 +526,19 @@ def create_home_page():
...
             "Analysis": ["Homepage", "Variable Analysis", "Relationship Analysis", "Country Analysis"],
             "Summary": ["Continent Summary"],
-        }
+        },
+        nav_selector=vm.NavBar(
+            items=[
+                vm.NavLink(
+                    label="Label 1",
+                    pages={"Analysis": ["Variable Analysis", "Homepage"], "Summary": ["Continent Summary"]},
+                ),
```
petar-qb, line 535: Can we keep the same attributes order like below? So:
```
                vm.NavLink(
                    pages={"Analysis": ["Variable Analysis", "Homepage"], "Summary": ["Continent Summary"]},
                    label="Label 1",
                ),
```

vizro-core/examples/default/app.py
petar-qb, line 1: Can we also adjust other examples (dict, json, yaml)?

vizro-core/src/vizro/models/_navigation/nav_link.py
```diff
@@ -58,10 +59,14 @@ def build(self, *, active_page_id=None):
...
-                # try dbc.Popover or Dash mantine components tooltip?).
-                # dbc.Tooltip(html.P(self.label), target=self.id, placement="bottom", className="custom-tooltip"),
+                dmc.Tooltip(
+                    label=self.label,
+                    offset=6,
+                    withArrow=True,
+                    zIndex=1,
+                    children=[html.Span(self.icon, className="material-symbols-outlined")],
+                    position="bottom-start",
+                )
```
petar-qb, line 69: Just curious: Is the Tooltip always presented, without any option to remove it?

petar-qb (commented):
I'm not fully familiar with this feature, but still was curious in it. So, I left few comments 😄.

vizro-core/examples/default/app.py
antonymilne: I think these changes were probably just to test it out rather than meant to remain as changes to the default app.py?

antonymilne (approved):
Great stuff, thank you for all the fixes! 🙏 

The tests look good. I just made a small change to make them a bit less verbose: [6500d8f](https://github.com/mckinsey/vizro/pull/186/commits/6500d8f4c9e1469870d90a058b12304eb6e7d6b2). It's the same tests as you wrote, just putting some stuff into a variable to make it clearer.

vizro-core/changelog.d/20231201_152948_nadija_ratkusic_graca_navigation_css_refinements.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Enable tooltip functionality for `NavLink` icons. ([#186](https://github.com/mckinsey/vizro/pull/186))
```
huong-li-nguyen, line 22:
```suggestion
- Enable tooltips for `NavLink` icons ([#186](https://github.com/mckinsey/vizro/pull/186))
```

vizro-core/src/vizro/models/_navigation/nav_link.py
huong-li-nguyen: This is actually a very good question - people should be able to remove it if they want. Currently, if you provide an empty string, the tooltip is still visible. Shouldn't the `label` be Optional @antonymilne @nadijagraca ? I understand that for the horizontal alignment it should be a mandatory field, but for the vertical alignment it should be Optional?

vizro-core/src/vizro/models/_navigation/nav_link.py
```diff
@@ -57,10 +58,14 @@ def build(self, *, active_page_id=None):
 
         button = dbc.Button(
             [
-                html.Span(self.icon, className="material-symbols-outlined"),
-                # TODO: commented out until we insert styling for the tooltip or find a better way to display it (e.g.
-                # try dbc.Popover or Dash mantine components tooltip?).
-                # dbc.Tooltip(html.P(self.label), target=self.id, placement="bottom", className="custom-tooltip"),
+                dmc.Tooltip(
+                    label=self.label,
+                    offset=6,
```
huong-li-nguyen, line 63: Could you double-check the CSS applied to the tooltip? It seems they are not aligned with the designs that are in Figma? Or did D. give you other designs? 

In Figma/Storybook, it would be good if you double-check the offset, zIndex and the colors. On dark theme e.g. the tooltip is white, while on light theme the tooltip is dark. Currently it's the other way around.

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -138,25 +138,16 @@ div.dashboard_container .custom-tooltip {
...
-.icon-button.btn.btn-primary.active {
-  background-color: var(--state-overlays-selected);
-}
-
 .loading-container {
   height: 100%;
   width: 100%;
 }
+
+.mantine-Tooltip-tooltip {
```
huong-li-nguyen, line 149: I think if you give the relevant mantine component a className, you can reference it here, instead of relying on the default mantine className. I would prefer that :)

vizro-core/tests/unit/vizro/models/_navigation/test_nav_bar.py
```diff
@@ -73,11 +74,21 @@ def test_items_with_with_pages_icons(self, pages_as_dict):
 class TestNavBarBuildMethod:
     """Tests NavBar model build method."""
 
+    common_args = {"offset": 6, "withArrow": True, "zIndex": 1, "position": "bottom-start"}
```
huong-li-nguyen, line 77: Can we actually move all of these defaults to CSS instead of having it inside the arguments of the dmc component?

vizro-core/src/vizro/models/_navigation/nav_link.py
```diff
@@ -57,10 +58,14 @@ def build(self, *, active_page_id=None):
...
+                    withArrow=True,
+                    zIndex=1,
+                    children=[html.Span(self.icon, className="material-symbols-outlined")],
```
huong-li-nguyen, line 66: In the designs there is actually a max. width for the tooltip defined, so in case it exceeds it, text wrapping will be activated automatically. We should have this as our default as well, but optimally users should be able to overwrite it via CSS if they want.

See Figma designs for details:
Screenshot 2023-12-05 at 13 30 43

huong-li-nguyen (changes requested):
Hey @nadijagraca , great that you found a way to style the tooltips now! 🚀  Have some change requests mostly related to design and some default options.

It would be great if you could apply the Figma designs we have for tooltips here. I noticed some discrepancies. Ignore the comment, if the Figma designs are outdated and these are indeed our new designs?

According to the Figma designs, we would also need a fixed width for the tooltip, such that text wrapping is activated if the label is very long. I think that makes sense and should also be our default.

Could you double-check with @antonymilne whether the label indeed should be mandatory? For me it makes sense, that users can turn off the tooltips if they want. Currently that's not possible.

Could you update the `accordion_inside_nav_bar.png` such that it also shows the last page in the Accordion? In the new screenshot it's currently cut off, but I think it makes sense to keep it in, as we're explaining the grouping in the docs section

vizro-core/tests/unit/vizro/models/_navigation/test_nav_bar.py
antonymilne: I don't think these are CSS styling arguments though? They are [component properties](https://www.dash-mantine-components.com/components/tooltip) and not part of `style`.

vizro-core/src/vizro/models/_navigation/nav_link.py
antonymilne: I think we should leave this compulsory:
* it will be easy (non-breaking) in future to make it optional, but the reverse direction is not true
* as you say, when it comes to a horizontal navbar, it would be compulsory, and having something optional in one layout and not the other could be tricky and confusing
* always requiring a tooltip is arguably important for accessibility and good design principles (?)
* in reality, much of the time this label is automatically populated anyway (from a higher level `pages` field) and so the user does not need to explicitly specify it anyway

The intention here is to keep things simple and non-committal since we don't actually know much about how people will use the components in reality (or even whether they will use them at all). We can always change in future after we've seen real world usage and identified what common requirements/complaints/etc. are. e.g. maybe we'll want a whole new field `tooltip` that's distinct from `label`, maybe not; maybe we will make it so that an empty string or other `False`y value for `label` doesn't render a tooltip. But unless you think these are important requirements to enable upfront, I would not do so yet.

vizro-core/tests/unit/vizro/models/_navigation/test_nav_bar.py
huong-li-nguyen: Interesting, I am wondering why these are actually not part of style, because offset/position can be CSS properties: https://developer.mozilla.org/en-US/docs/Web/CSS/offset-position

If this is not possible, I guess users should create custom components of the `NavLink` when they want to style their tooltip differently in terms of offset/position. 

@nadijagraca , if you could just ensure that these are aligned with the designs, that would be great! E.g. we should stick with the design rule that paddings/margins should always be multiples of 4.

vizro-core/src/vizro/models/_navigation/nav_link.py
huong-li-nguyen: Got it! No, all good - I am fine with not enabling the possibility of disabling the tooltip until people request it

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -131,32 +131,28 @@ div.dashboard_container .tooltip-inner {
...
+  font-family: Inter;
+  font-size: var(--text-size-01);
+  font-weight: var(--text-weight-light);
+  line-height: 16px;
+  letter-spacing: var(--letter-spacing-help-text);
```
huong-li-nguyen, line 148: Do you actually have to set all of these? I would have thought that some of them such as font-family etc. are not required anymore because they are defined in our global CSS variables

huong-li-nguyen (approved):
The new tooltips look great now, love the box-shadowing on the light theme as well! 🚀  Thanks for incorporating all the changes :) Left one small comment and could you replace the screenshot as mentioned here:

`Could you update the accordion_inside_nav_bar.png such that it also shows the last page in the Accordion? In the new screenshot it's currently cut off, but I think it makes sense to keep it in, as we're explaining the grouping in the docs section`

The rest looks great! After updating the screenshot and just checking whether we need all the CSS properties defined in `.mantine-Tooltip-tooltip `, feel free to merge after all the tests pass.

PR #188, opened 2023-11-30
Tidy hidden `Div`s and default values

antonymilne:
## Description

No change to functionality; just a small tidy. Seems a bit pointless, but it will make it easier to clean up some tests in a following PR...

* prefer to use `None` over `html.Div(hidden=True)` in the case that we don't need to refer to the element in any way (basically whenever you don't set an `id`). e.g. `html.P(self.title) if self.title else None`
* prefer to use `html.Div(hidden=True)` over `None` in the case that we do need to refer to the element (basically when you do set an `id`). e.g. `html.Div(hidden=True, id="nav_panel_outer")`. Generally these can be identified by the fact that `build` return values have a type like `_NavBuildType`
* prefer to use `""` as default value for optional fields which are `str`. These fields do not need to accept `None` values at all

## Checklist

## Types of changes

## Notice


lingyielia (approved):
Thanks for the house-keeping! Is there a way for us to record this type of coding standards? So for future components to come, we can be consistent in terms of best practices

nadijagraca (approved):
Thanks for tidying it up. 🚀

antonymilne:
> Thanks for the house-keeping! Is there a way for us to record this type of coding standards? So for future components to come, we can be consistent in terms of best practices

Great question. I used to have various bits of documentation on this but they're a bit scattered and out of date. Let me make a ticket to consolidate and update them because definitely it would be useful 👍

PR #190, opened 2023-12-04
Minor update to `Navigation` docs

huong-li-nguyen:
## Description
- Fix typing in docstrings
- Remove remaining `className="hidden"` and replace with `hidden=True` (there was one case that got back in)
- Consolidate formatting for default values in docstrings
- Add validator for provided icon names (transforms to lower-case and replaces spaces with underscores)

## Checklist

## Types of changes

## Notice


vizro-core/src/vizro/models/_navigation/navigation.py
```diff
@@ -14,14 +16,13 @@ class Navigation(VizroBaseModel):
...
-        pages (Optional[NavPagesType]): See [`NavPagesType`][vizro.models.types.NavPagesType].
-            Defaults to [].
+        pages (NavPagesType): See [`NavPagesType`][vizro.models.types.NavPagesType]. Defaults to `[]`.
         nav_selector (Optional[NavSelectorType]): See [`NavSelectorType`][vizro.models.types.NavSelectorType].
             Defaults to `None`.
     """
 
     pages: NavPagesType = []
-    nav_selector: NavSelectorType = None
+    nav_selector: Optional[NavSelectorType] = None
```
huong-li-nguyen, line 25: @antonymilne - here I wasn't sure whether the `Optional` was removed intentionally?

vizro-core/docs/pages/user_guides/navigation.md
```diff
@@ -174,11 +174,11 @@ Another way to group together pages in the navigation is to use a [`NavBar`][viz
     [NavBar]: ../../assets/user_guides/navigation/nav_bar.png
 
 
-Here, the first level of the navigation hierarchy ("Group A" and "Group B") is represented by an icon in a navigation bar, and the second level of the navigation (the pages) is represented by an accordion. By default, the icons are the [`filter` icons from the Google Material icons library](https://fonts.google.com/icons?icon.query=filter).
+Here, the first level of the navigation hierarchy ("Group A" and "Group B") is represented by an icon in a navigation bar, and the second level of the navigation (the pages) is represented by an accordion. By default, the icons are from [the Google Material `filter` icons library](https://fonts.google.com/icons?icon.query=filter).
```
antonymilne, line 177: This made more sense to me before because the Google Material icons library is the library as a whole, and the `filter` icons are just some icons inside that library?

Maybe this would be clearer than the original?
```
By default, the set of icons used are the [`filter` icons from the Google Material icons library](https://fonts.google.com/icons?icon.query=filter).
```

vizro-core/docs/pages/user_guides/navigation.md
```diff
@@ -228,6 +228,8 @@ The same configuration for [grouping pages](#grouping-your-pages) applies inside
 
 You can alter the icons used by specifying the name of the icon in the [Google Material icons library](https://fonts.google.com/icons):
 
+*Note: Spaces in the icon's name are replaced with underscores*
```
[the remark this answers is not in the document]
antonymilne: Even better, why don't we just do the conversion for them? We can easily add something to `NavLink` that does `str.lower().replace(" ", "_")` and then it will work straight away.

vizro-core/src/vizro/models/_navigation/navigation.py
```diff
@@ -37,11 +38,12 @@ def pre_build(self):
 
     @_log_call
     def build(self, *, active_page_id=None) -> _NavBuildType:
-        nav_selector = self.nav_selector.build(active_page_id=active_page_id)
+        nav_selector: NavSelectorType = cast(NavSelectorType, self.nav_selector).build(active_page_id=active_page_id)
```
antonymilne, line 41: This doesn't look right: `nav_selector` here is the _built_ version, i.e. a dash component and not itself a `NavSelectorType`. Is this what you meant?
```suggestion
        nav_selector = cast(NavSelectorType, self.nav_selector).build(active_page_id=active_page_id)
```

antonymilne: P.S. I think maybe one of my reasons for removing the `Optional` was so that we didn't need to do this `cast` here at all.

vizro-core/tests/unit/vizro/models/_navigation/test_navigation.py
```diff
@@ -82,7 +82,7 @@ def test_default_nav_selector(self, pages, request):
         built_navigation = navigation.build(active_page_id="Page 1")
         assert_component_equal(
             built_navigation["nav_bar_outer"],
-            html.Div(className="hidden", id="nav_bar_outer"),
+            html.Div(hidden=True, id="nav_bar_outer"),
             keys_to_strip={"children"},
```
antonymilne, line 86: Does this still pass with `keys_to_strip={}`?

vizro-core/src/vizro/models/_navigation/navigation.py
antonymilne: To be confirmed...

vizro-core/docs/pages/user_guides/navigation.md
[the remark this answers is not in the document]
antonymilne: The more I think about this, the more I think that the preferred way for a user to enter `icon` should actually be the human-readable icon name like `Double Arrow` rather than the computer-readable snakecase version `double_arrow`.

Please could you change our docs and tests to reflect this? I think just a find and replace of `icon=` should do it. e.g. in docs instead of saying `icon="bar_chart"` we should do `icon="Bar Chart"`.

If a user enters `bar_chart` or `bar chart` etc. it will still work fine, but `Bar Chart` should be the first thing we suggest they try rather than `bar_chart`.

vizro-core/tests/unit/vizro/models/_navigation/test_navigation.py
huong-li-nguyen: Yes, it still passes with `keys_to_strip={}` and `keys_to_strip` not being specified at all.

vizro-core/src/vizro/models/_navigation/navigation.py
huong-li-nguyen: Ah yes! We do something very similar in `_dashboard.py`:

`navigation: _NavBuildType = cast(Navigation, self.navigation).build(active_page_id=page.id)`

I am also fine with removing the `Optional`, but then we should do it through the entire code basis because in the Dashboard model we currently also have this:

`navigation: Optional[Navigation] = None`

So either we remove it everywhere or we keep it?

vizro-core/docs/pages/user_guides/navigation.md
huong-li-nguyen: Yeah, that makes more sense! I think the filter notation is just very confusing if one hasn't understood that this is just a category name for a set of Google icons. Will change to above :)

vizro-core/docs/pages/user_guides/navigation.md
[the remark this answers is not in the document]
huong-li-nguyen: Ah, yes, I like that! Will implement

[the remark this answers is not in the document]
huong-li-nguyen: Implemented in this commit: https://github.com/mckinsey/vizro/pull/190/commits/afe2968ec316b25a64d91126f6340d7a410e8086

vizro-core/docs/pages/user_guides/navigation.md
```diff
@@ -239,7 +240,7 @@ You can alter the icons used by specifying the name of the icon in the [Google M
                     items=[
                         vm.NavLink(
                             label="Section 1",
-                            icon="bar_chart",
+                            icon="Bar Chart",
```
nadijagraca, line 243: Can we revert to "bar_chart" as "Bar chart" is not a valid google icons name, and leads to error in loading the icon from the library.

vizro-core/docs/pages/user_guides/navigation.md
```diff
@@ -259,7 +260,7 @@ You can alter the icons used by specifying the name of the icon in the [Google M
             type: nav_bar
             items:
               - label: Section 1
-                icon: bar_chart
+                icon: Bar Chart
```
nadijagraca, line 263: Needs to be reverted to "bar_chart" in this example as well.

vizro-core/tests/unit/vizro/models/_navigation/test_nav_item.py
```diff
@@ -31,6 +31,14 @@ def test_nav_link_mandatory_and_optional(self, pages_as_list):
         assert nav_link.icon == "home"
         assert nav_link.pages == pages_as_list
 
+    @pytest.mark.parametrize("icon", ["Bar Chart", "bar chart", "bar_chart", "Bar_Chart"])
```
nadijagraca, line 34: When tested locally "Bar Chart" and "bar chart" are not valid icon names as they do not produce an icon.

nadijagraca (changes requested):
Thanks for tidying it up. I left some small comments regarding the icon names. But otherwise, looks good.

vizro-core/docs/pages/user_guides/navigation.md
huong-li-nguyen: Hey Nadija, the newly added validator converts "Bar Chart" to "bar_chart", so both of them produce the same results.

vizro-core/tests/unit/vizro/models/_navigation/test_nav_item.py
huong-li-nguyen: Can you double-check if the latest commits have been in your local branch? I've just double-checked again and it does reproduce the icon. Given that this unit test passes, the validator seems to be producing the right results as well?

Try with these configurations for example:

```
dashboard = vm.Dashboard(
    pages=[
        create_home_page(),
        create_variable_analysis(),
        create_relation_analysis(),
        create_continent_summary(),
        create_country_analysis(),
    ],
    navigation=vm.Navigation(
        nav_selector=vm.NavBar(
            items=[
                vm.NavLink(
                    label="Link 1",
                    icon="Bar Chart",
                    pages={
                    },
                ),
                vm.NavLink(
                    label="Link 2",
                    icon="Arrow Drop Down",
                    pages={
                    },
                ),
            ]
        )
    ),
)
```

vizro-core/docs/pages/user_guides/navigation.md
nadijagraca: Got it. I tested on my local branch. Sorry.

nadijagraca (approved):
Good job with the validator. 🚀

vizro-core/tests/unit/vizro/models/_navigation/test_navigation.py
antonymilne: In this case please can you change to `keys_to_strip={}`? Will explain why in a near-future PR

vizro-core/docs/pages/user_guides/navigation.md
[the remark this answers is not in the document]
antonymilne: I just left a couple of comments on that commit (not sure if you can see the conversation anywhere else 🤔) but looks good!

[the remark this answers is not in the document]
huong-li-nguyen: I've incorporated them in :)

vizro-core/src/vizro/models/_navigation/navigation.py
huong-li-nguyen: After reading through this section of [the migration guide for pydantic V2](https://docs.pydantic.dev/2.0/migration/#required-optional-and-nullable-fields) and this [forum post](https://stackoverflow.com/questions/76466468/pydantic-model-fields-with-typing-optional-vs-typing-optional-none) it seems we should stick with `Optional[XXX] = None`.

To sum up the relevant parts:

**Runtime implications**
In practice this means that for the following model all fields behave identically in the sense that neither is required during initialization because all of them will receive None as their value, when no other value is provided, and the type of each of them is set to be Optional[str].

```

from typing import Optional, Union
from pydantic import BaseModel, Field

class Model(BaseModel):
    a: str | None
    b: Union[str, None]
    c: Optional[str]
    d: Optional[str] = None
    e: Optional[str] = Field(default=None)
    f: Optional[str] = Field(default_factory=lambda: None)
    g: str = None
```

It seems that the g case, the one we are talking about here is under the hood converted to Optional[str] = None:

```
Note that even g is implicitly handled in a way that sets its type to be Optional[str], even though we annotate it with just str. You can verify this by doing the following:

for field in Model.__fields__.values():
    print(repr(field))
```

What will change in pydantic V2 is the way how Optional[X] are handled if no defaults are provided as far as I undertsand e.g. `Optional[str]` and `Optional[str] = None` were both considered as non-required and a default was automatically set in V1. This is not the case anymore with V2. Optional[str] without a default will be considered as a required field, while `Optional[str] = None` is not required, can be None, and is None by default.

So I think to be more explicit, we should stick with `Optional[XXX] = None` unless I am missing something :)

vizro-core/src/vizro/models/_navigation/navigation.py
huong-li-nguyen: Followed up here: https://github.com/mckinsey/vizro/pull/190#discussion_r1416628062

vizro-core/src/vizro/actions/export_data_action.py
```diff
@@ -23,7 +23,7 @@ def export_data(
     """Exports visible data of target charts/components on page after being triggered.
 
     Args:
-        targets: List of target component ids to download data from. Defaults to None.
+        targets: List of target component ids to download data from. Defaults to `None`.
```
maxschulz-COL, line 26: Just reminded myself of how mkdocstring handles defaults. It is unfortunately a little confusing...

For pydantic models: we need to mention the default (e..g defaults to `None`), because the default is not recognized

For python functions (such as actions), this is actually automatically recognized (i.e. it will know from the type hint)

While being (for the time being) two different approaches, I would suggest that we remove the "defaults to XY" from all actions.

Wdyt?

vizro-core/src/vizro/actions/filter_interaction_action.py
```diff
@@ -25,7 +25,7 @@ def filter_interaction(
 
     Args:
         targets: List of target component ids to filter by chart interaction. If missing, will target all valid
-            components on page. Defaults to None.
+            components on page. Defaults to `None`.
```
maxschulz-COL, line 28: Hence here as well

vizro-core/src/vizro/models/_components/form/_user_input.py
```diff
@@ -22,14 +22,14 @@ class UserInput(VizroBaseModel):
         actions (Optional[List[Action]]): Defaults to `[]`.
         title (str): Title to be displayed. Defaults to `""`.
         input_type (Literal["text", "number", "password", "email", "search", "tel", "url", "range", "hidden"]):
-            Type of value to validate user input against. Defaults to 'text'.
+            Type of value to validate user input against. Defaults to `"text"`.
```
maxschulz-COL, line 25: but not here! :)

vizro-core/changelog.d/20231204_201354_huong_li_nguyen_update_nav_docs.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
```
maxschulz-COL, line 40: Maybe worth mentioning the CSS fix. Actually I couldn't find it, which code change was that?

maxschulz-COL (approved):
Lgtm! Thanks for the PR. Have a couple of small suggestions, but generally looks good

vizro-core/changelog.d/20231204_201354_huong_li_nguyen_update_nav_docs.md
huong-li-nguyen: It got merged in the meantime in another PR - haven't updated this one yet as @antonymilne still wanted to double-check the Optional/None open point :) 

So nothing to add here anymore

vizro-core/src/vizro/actions/export_data_action.py
huong-li-nguyen: Given that it seems quite confusing and people would then have to remember when to add it or not, I would actually just leave it as is and always provide it? Currently it's consistently added to the docstrings everywhere and this seems like a simpler rule to remember until we have consolidated everything?

maxschulz-COL: Hm ok, then let's leave it

vizro-core/src/vizro/models/_navigation/navigation.py
huong-li-nguyen: As discussed w/ @antonymilne : Changes will be reverted here and handled properly in separate PR

antonymilne (approved):

vizro-core/src/vizro/models/_navigation/navigation.py
antonymilne: See https://github.com/mckinsey/vizro/pull/210

PR #192, opened 2023-12-05
Fix integration tests to work in a single pytest session

### LINKED ISSUE #125: Optimise windows integration tests
> [!IMPORTANT]
> All the BELOW is based on some very quick glancing through of recently run Actions. Before making any changes we should look properly at how long these jobs take and which jobs seem redundant.

I want our CI to remain super fast and slick. At the moment it's not too bad but there are several things I think could be optimised.

Let's start with the most blatant offender and probably easiest to fix, which seems to be windows integration tests. Very roughly it looks like these can take 6-8 mins when all other jobs take a maximum of 3 mins.

Currently we run integration tests on the following matrix: https://github.com/mckinsey/vizro/blob/1cfa2470556397c73fa7c3e0244f6e122b6db2df/.github/workflows/test-integration-vizro-core.yml#L29-L30

Things to ponder:
* can we speed these Windows tests up significantly?
* if not, do we actually want to run it on every commit in every PR? (probably not I think)

Probably it's not really worthwhile covering all versions of Python on Windows but since they seem to be run in parallel, the actual number of jobs doesn't make a big difference I think, just the speed of each one.

antonymilne:
## Description

Main change: we can now do `pytest tests/integration` in vizro-core in a single pytest session rather than needing to run each test separately. This is good because it means that any new integration tests we write (like `test_navigation` I added recently) will automatically run.

@l0uden thought that https://github.com/plotly/dash/issues/1933 was the issue here, but after some careful investigation (using `keep_open=True` is very useful) I decided that the failures were actually correct and not just a quirk of `dash_duo`. The root of the problem was something to do with the way we were handling imports, which I've changed to use `runpy` now.

Other minor changes:
* closes #125 by removing all Windows integration tests. I looked through our history of Github actions and decided that these don't add anything useful and are really just there for historic reasons
* tided some of our configurations e.g. ones which didn't need to use `matrix` to use a single version of Python
* tided some of our hatch scripts for running tests, because these always confuse me and we actually only use a couple of the commands
* made a couple more divs into `hidden`
## Checklist

## Types of changes

## Notice

### DIFF
--- a/.github/workflows/secret-scan.yml
+++ b/.github/workflows/secret-scan.yml
@@ -5,6 +5,10 @@ on:
   workflow_dispatch:
   schedule:
     - cron: "0 4 * * *" # run once a day at 4 AM
+
+env:
+  PYTHON_VERSION: "3.9"
+
 jobs:
   scan:
     name: gitleaks
@@ -14,10 +18,10 @@ jobs:
 
       - uses: actions/setup-go@v4
 
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python $PYTHON_VERSION
         uses: actions/setup-python@v4
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: $PYTHON_VERSION
 
       - name: Install pre-commit
         run: pip install pre-commit
--- a/.github/workflows/secret-scan.yml
+++ b/.github/workflows/secret-scan.yml
@@ -18,10 +18,10 @@ jobs:
 
       - uses: actions/setup-go@v4
 
-      - name: Set up Python $PYTHON_VERSION
+      - name: Set up Python ${{ env.PYTHON_VERSION }}
         uses: actions/setup-python@v4
         with:
-          python-version: $PYTHON_VERSION
+          python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Install pre-commit
         run: pip install pre-commit
--- a/.github/workflows/test-integration-vizro-core.yml
+++ b/.github/workflows/test-integration-vizro-core.yml
@@ -20,45 +20,27 @@ env:
 
 jobs:
   run:
-    name: Python ${{ matrix.python-version }} on ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
+    name: Python ${{ matrix.python-version }}
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
         python-version: ["3.8", "3.9", "3.10", "3.11"]
 
     steps:
       - uses: actions/checkout@v4
-      - name: Get branch name
-        id: branch-name
-        uses: tj-actions/branch-names@v7
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
 
-      - uses: actions/cache@v3
-        with:
-          path: ${{ env.pythonLocation }}
-          key: ${{ matrix.os }}-${{ matrix.python-version }}-${{ steps.branch-name.outputs.current_branch }}-pip-${{ hashFiles('hatch.toml') }}-${{ hashFiles('pyproject.toml') }}
-          restore-keys: |
-            ${{ matrix.os }}-${{ matrix.python-version }}-${{ steps.branch-name.outputs.current_branch }}-pip-
+      - name: Install Hatch
+        run: pip install --upgrade hatch
 
-      - name: Run ubuntu integration tests
-        if: ${{ matrix.os == 'ubuntu-latest' }}
-        run: |
-          pip install --upgrade hatch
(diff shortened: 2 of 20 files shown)


.github/workflows/secret-scan.yml
```diff
@@ -14,10 +18,10 @@ jobs:
 
       - uses: actions/setup-go@v4
 
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python ${{ env.PYTHON_VERSION }}
         uses: actions/setup-python@v4
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: ${{ env.PYTHON_VERSION }}
```
antonymilne, line 24: This actually wasn't working before because `matrix` wasn't defined here.

.github/workflows/test-integration-vizro-core.yml
```diff
@@ -20,45 +20,27 @@ env:
...
         with:
           python-version: ${{ matrix.python-version }}
 
-      - uses: actions/cache@v3
-        with:
-          path: ${{ env.pythonLocation }}
-          key: ${{ matrix.os }}-${{ matrix.python-version }}-${{ steps.branch-name.outputs.current_branch }}-pip-${{ hashFiles('hatch.toml') }}-${{ hashFiles('pyproject.toml') }}
-          restore-keys: |
-            ${{ matrix.os }}-${{ matrix.python-version }}-${{ steps.branch-name.outputs.current_branch }}-pip-
+      - name: Install Hatch
```
antonymilne, line 38: @l0uden I wasn't sure whether to remove this cache or not. Is it only useful on Windows or in general? We don't have it on our unit tests or any other github workflows so not sure whether it's actually speeding up anything or not? I feel like either it should be on all workflows or it should be on none of them?

vizro-ai/hatch.toml
antonymilne, line 1: This is just tidying, no functional changes.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -83,7 +83,7 @@ def build(self):
         return dbc.Container(
             id="dashboard_container_outer",
             children=[
-                html.Div(id=f"vizro_version_{vizro.__version__}"),
+                html.Div(vizro.__version__, id="vizro_version", hidden=True),
```
antonymilne, line 86: @huong-li-nguyen this is a very small and insignificant change but feels like a much tidier way to do it since it separates the label `vizro_version` from the data itself: `<div id="vizro_version" hidden>0.1.6</div>`.

vizro-core/tests/integration/test_examples.py
antonymilne, line 1: This file is the major change. Now we can have a single parametrised test rather than 4 separate tests.

vizro-core/hatch.toml
```diff
@@ -57,16 +49,14 @@ test = [
...
 ]
-test-cov = "coverage run -m pytest tests/unit {args}"
-test-integration = [
-  "- pytest tests/integration/test_examples.py::test_default_dashboard --headless",
-  "- pytest tests/integration/test_examples.py::test_dict_dashboard --headless",
-  "- pytest tests/integration/test_examples.py::test_json_dashboard --headless",
-  "- pytest tests/integration/test_examples.py::test_yaml_dashboard --headless",
-  "pytest tests/integration/test_navigation.py"
-]
+test-integration = "pytest tests/integration --headless {args}"
```
antonymilne, line 52: This is the key improvement.

vizro-ai/hatch.toml
```diff
@@ -22,14 +22,6 @@ dependencies = [
 VIZRO_AI_LOG_LEVEL = "DEBUG"
 
 [envs.default.scripts]
```
antonymilne, line 24: This is just tidying, no functional changes

vizro-core/tests/integration/test_examples.py
```diff
@@ -1,5 +1,6 @@
 # ruff: noqa: F403, F405
```
antonymilne, line 1: This file is the major change. Now we can have a single parametrised test rather than 4 separate tests.

vizro-core/src/vizro/models/_dashboard.py
huong-li-nguyen: Makes sense!

huong-li-nguyen (approved):
Looks like a great improvement to me - thanks for figuring it all out! 🚀

.github/workflows/test-integration-vizro-core.yml
antonymilne: Cool, I'll leave it deleted then.

Can I also remove `.github/requirements.txt` now? I can't remember what that's needed for.

l0uden: Yes, it was used only for windows tests

antonymilne: Perfect, I'll delete that too then.

PR #70, opened 2023-09-26
Enable `NavBar` as selector for `Navigation`

nadijagraca:
## Description

@antonymilne comment: Some of the below description is out of date. The best place to look for what this PR does is the [user guide on navigation](https://vizro.readthedocs.io/en/stable/pages/user_guides/navigation/).

As well as navigation, this PR adds an important new testing functionality with `assert_component_equals`. I will write about this separately in another PR.

---

The main idea is that the `Navigation` model is responsible for delivering navigation components, nav_bar and nav_panel and Page model is responsible for  arranging components. 

Models added:
### 1. NavBar model
NavBar model is responsible for building nav_bar_div. It takes `items` and `pages` as arguments. `items` are list of NavItem models, and  `pages` are list of page ids or page titles. 
- `pages`: list or dict mapping of page ids or page titles. 
- `items`: list of `NavItem` models

### 2. NavItem model
NavItem model is responsible for building individual icons inside the nav_bar div. Click on the icon leads you to the first page defined inside NavItem.pages.
- `pages`: list or dict mapping of page ids or page titles. 
- `icon`: If None provide default image (home icon from Google icons)(but only on vertical), if a user sets icon = ""  they can turn off our defaults
- `text`: If text goes above our max_text_length, it will be truncated to length defined in `max_text_length` which defaults to 8. If not already provided, tooltip is automatically populated with the full text length
- `tooltip`: No special logic except for automatic population based on text length
- `max_text_length`: Defaults to 8 characters. If a user wants to exceed the max text length, they can set it here in which case the text should wrap automatically (vertical alignment). For the horizontal alignment, we might just set a different max_text_length as default or None?
- `selector`: exposed, but can only take `Accordion` model

### Possible configuration options for dashboard navigation:

- Option 1
**Configuration**: `navigation=Navigation()` -> gives -> `navigation=Navigation(selector=Accordion(pages=[])` where `pages` is list of all registered pages in the dashboard. 
**Outcome**: navigation panel with accordion and 'SELECT PAGE' default accordion title

- Option 2
**Configuration**: `navigation=Navigation(pages=['Page_1', 'Page_2', 'Page_3'])`  -> gives -> `navigation=Navigation(selector=Accordion(pages=['Page_1', 'Page_2', 'Page_3'])` the same is true when `pages` is provided as a `dict`
**Outcome**: navigation panel with accordion and 'SELECT PAGE' default accordion title if pages provided as list or custom title with pages provided as dicc

- Option 3
[...]

PR #195, opened 2023-12-06
Introduce `assert_component_equal` for tests

antonymilne:
## Description

I made a new testing utility function `assert_component_equal` as part of rewriting of tests in #70 but didn't advertise it widely before. I've just tweaked how it works so I'm happy with it now. I'll do a learning session on this on Weds 20th December to explain all.

Just in terms of reviewing, there's not really anything to check here because most of the commits are just tidying up my implementation of `assert_component_equal`.

The things that are of interest are:
* https://github.com/mckinsey/vizro/pull/195/commits/360e4bab3a30e2dffd6649448d4f7f09347b5511 shows an example of basic usage and how we should update our tests
* https://github.com/mckinsey/vizro/blob/tests/assert-components-equal/vizro-core/tests/tests_utils/asserts.py is the code behind `assert_component_equal`
* https://github.com/mckinsey/vizro/blob/tests/assert-components-equal/vizro-core/tests/tests_utils/demo_asserts.py shows some example advanced usage

## Basic usage

When we compare the output of a `build` method to a Dash component tree, previously we were doing something like this:

```
result = json.loads(json.dumps(x, cls=plotly.utils.PlotlyJSONEncoder))
expected = json.loads(json.dumps(y, cls=plotly.utils.PlotlyJSONEncoder))
assert result == expected
```

Now we do:
```
assert_component_equal(x, y)
```

For example:
```
from dash import html
assert_component_equal(html.Div([html.P("blah")]), html.Div([html.P("blah")]))
```

So it's just a nicer way of writing the same check. Think of it a bit like [`pandas.testing.assert_frame_equal`](https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_frame_equal.html).

## Advanced usage: `keys_to_strip`

`assert_component_equal` also has a `keys_to_strip` argument that sets properties that will be ignored in the component comparison. Sometimes it's useful to set `key_to_strip={"id"}` since our `id`s are randomly generated (they're model manager ids) and we don't care about actually comparing them.

The other use for this is in testing "high-level containers"  vs. "low-level contents". This is used a lot in the navigation tests and demonstrated with a toy example in https://github.com/mckinsey/vizro/blob/tests/assert-components-equal/vizro-core/tests/tests_utils/demo_asserts.py.

In general any test where we have `keys_to_strip={"children"}` would be followed by a test on `result.children` using `STRIP_ALL`.

The idea is:
1. `test_X_build` tests stuff that's only in `X`
2. `test_Y_build` tests stuff that's only in `Y` 
3. `test_Y_build` _also_  tests that `Y.build()` is correctly calling `X.build()` but not the details of what `X.build()` returns

The typical way you'd do case 3 is by providing a mock `X` and then doing something like `assert X.call_count == 4`. The current solution is arguably slightly "wrong" in that some of the implementation of `X.build` is leaking into the `test_Y_build`, because we are checking against `html.Div`. This is fine by me though, because it means we test the interface a bit more strictly, and it's much clearer, easier and less error-prone than writing mocks.

## Next steps

* Convert all our tests that do `json.loads(json.dump(...))` into `assert_component_equal`. Most of these will just be simple find and replace as per the basic usage
* Write tests that are currently missing e.g. for `Page.build` using the "advanced usage"

## Checklist

## Types of changes

## Notice


vizro-core/tests/unit/vizro/models/_components/test_button.py
antonymilne, line 1: This is the one file you should look at to understand how `assert_component_equal` should be used.

vizro-core/tests/tests_utils/demo_asserts.py
```diff
@@ -0,0 +1,46 @@
...
+
+def test_Y_build():
+    # Test for higher-level container.
+    many_x = [X(text="Hello world") for _ in range(4)]
+    result = Y(id="y", children=many_x).build()
+    # We don't want to compare the whole component tree here. Instead compare the bit that's specifically in Y and
+    # ignore the children:
+    assert_component_equal(result, html.Div(id="y", className="container"), keys_to_strip={"children"})
+    # And also compare the "interface" between X and Y:
+    assert all(isinstance(child, html.Div) for child in result.children)
```
huong-li-nguyen, line 46: Just for me to when we apply this everywhere: In this case where you want to test for the higher-level component and therefore ignore its children, would you always add this additional check then:

`assert all(isinstance(child, html.Div) for child in result.children)`

or should we actually not care at all as the children should be tested properly in the tests for the children?

vizro-core/tests/unit/vizro/models/_components/test_button.py
```diff
@@ -53,9 +37,13 @@ def test_set_action_via_validator(self):
...
-        assert result == expected
+    def test_button_build(self):
+        button = vm.Button(id="button_id", text="My text").build()
+        assert_component_equal(
+            button,
+            html.Div(
+                dbc.Button(id="button_id", children="My text", className="button_primary"),
+                className="button_container",
+                id="button_id_outer",
+            ),
```
huong-li-nguyen, line 48: Any preference whether to input the object directly or store it in a variable called `expected` before?

huong-li-nguyen (approved):
Thanks for this super helpful PR description! 🚀  Everything is clear 👍

vizro-core/tests/unit/vizro/models/_components/test_button.py
maxschulz-COL: Hm, I think it would look easier to read if we actually stored it in `expected`.

vizro-core/tests/tests_utils/asserts.py
```diff
@@ -4,27 +4,31 @@
...
 
-def component_to_dict(component: dash.development.base_component.Component) -> dict:
+def _component_to_dict(component: dash.development.base_component.Component) -> dict:
+    """Prepares a Dash component for comparison by conversion to JSON object."""
     return json.loads(json.dumps(component, cls=plotly.utils.PlotlyJSONEncoder))
 
 
-# TODO: implement some sort of depth limit to comparison so can use in high level tests, roll out more widely across
-# tests
 def assert_component_equal(left, right, keys_to_strip=None):
```
maxschulz-COL, line 21: Given that this labelled as a public function, shall we also add it to the docs?

vizro-core/tests/tests_utils/demo_asserts.py
maxschulz-COL: Or rather in the test for `X` in this case no? Are we not sure due to pydantic and the tests for `X` that this is correct?

maxschulz-COL (commented):
Lgtm! Have one suggestion, but it's a really cool feature, and really saves a lot of confusion and headache :)

vizro-core/tests/tests_utils/asserts.py
```diff
@@ -4,27 +4,31 @@
...
+    Examples:
+        >>> from dash import html
+        >>> assert_component_equal(html.Div(), html.Div())
+        >>> assert_component_equal(html.Div(id="a"), html.Div(), keys_to_strip={"id"})
+        >>> assert_component_equal(html.Div(html.Div()), html.Div(), keys_to_strip={"children"})
+    """
+    keys_to_strip = keys_to_strip or {}
+
+    left = _strip_keys(_component_to_dict(left), keys_to_strip)
+    right = _strip_keys(_component_to_dict(right), keys_to_strip)
```
petar-qb, line 33: The tests would be faster if we first `strip (remove) the object attributes` (similar to `_strip_keys` in dict) and then `translate the "truncated" object into a dictionary`. If you find it easy to implement, I'd love to see it in a PR, otherwise ignore this comment 😄.

petar-qb (approved):
I really like the "assert_component_equal".  
P.S. Thanks for the awesome description! 🚀

vizro-core/tests/unit/vizro/models/_components/test_button.py
antonymilne: I would prefer not to be too prescriptive about this in general. I don't think we should insist on always having a variable called `expected` if the assert is very simple like `assert_component_equal(button, html.Div())`. Once the `expected` gets beyond a certain complexity, I think it's more readable to extract into a variable instead of having it inline, but the boundary of what that "certain complexity" is just a matter of taste really.

In this case I think I agree with @maxschulz-COL so I will extract it into a variable, but I don't think we need to require very simple cases to do that.

Worth noting that PyCharm handles the actual and expected in exactly the same way regardless of whether you extract it into a variable or not (left is expected, right is actual):
image

vizro-core/tests/tests_utils/asserts.py
antonymilne: This is not a public function because it's not part of the public `vizro` package. It's "public" only to developers who write tests for vizro.

Possibly in future we would have some public API `from vizro.testing import assert_component_equal` (like Dash and pandas) if vizro _users_ would find it useful, e.g. to write tests for their custom components. This PR doesn't do anything like that though - all the code here lives in `tests`, not in `src`.

vizro-core/tests/tests_utils/asserts.py
antonymilne: Please could you explain a bit more what you mean here? This is what a component looks like before conversion to dictionary:

```
component = Div(children=Button(children='My text', id='button_id', className='button_primary'), id='button_id_outer', className='button_container')
```

Each Dash component has an attribute `available_properties` and `children`, so I guess something like:

```
for key in keys_to_strip:
    remove component.available_properties[key]
    # recurse through component.children and do the same
```

Is that what you mean?

vizro-core/tests/tests_utils/asserts.py
maxschulz-COL: Ooops, I overlooked that its in the `tests` folder. Nevermind then 👍

vizro-core/tests/tests_utils/asserts.py
petar-qb: Yes, that's what I meant.

So my assumptions are that this line:
`_component_to_dict(_strip_component_properties(dash_component, properties_to_strip))`

is faster for bulky and nested dash components than this one:
`_strip_keys(_component_to_dict(dash_component), keys_to_strip)`

I can't guess how much time we can save for the `hatch run test-unit', but it might be worth a try.

vizro-core/tests/tests_utils/demo_asserts.py
antonymilne: I'm glad you asked because I actually meant to rewrite this and then forgot... Now done in [85dae66](https://github.com/mckinsey/vizro/pull/195/commits/85dae66647274a914ebe5d3d4273ac4b6068368c) by introducing a new `STRIP_ALL` option.

In general any test where we have `keys_to_strip={"children"}` would be followed by a test on `result.children` using `STRIP_ALL`.

The idea is:
1. `test_X_build` tests stuff that's only in `X`
2. `test_Y_build` tests stuff that's only in `Y` 
3. `test_Y_build` _also_  tests that `Y.build()` is correctly calling `X.build()` but not the details of what `X.build()` returns

The typical way you'd do case 3 is by providing a mock `X` and then doing something like `assert X.call_count == 4`. The current solution is arguably slightly "wrong" in that some of the implementation of `X.build` is leaking into the `test_Y_build`, because we are checking against `html.Div`. This is fine by me though, because it means we test the interface a bit more strictly, and it's much clearer, easier and less error-prone than writing mocks.

vizro-core/tests/tests_utils/asserts.py
antonymilne: I think you're _probably_ right, although it's not obvious that would be faster actually because it would still need to be recursive code either way.

I played around with this for a while and couldn't quickly find a simple way to do it though. I did check performance of the current solution, and given that our nested dictionaries are never that big the time optimisation here is going to be extremely minimal even if `assert_component_equal` is used 10000s of times across the test suite. There are much bigger overheads in running the test suite than this at the moment, so even if we could speed the check up by a factor of 100 it would have minimal impact on overall running time.

vizro-core/tests/tests_utils/demo_asserts.py
huong-li-nguyen: Got it! That makes sense - could you add it to the PR description? 🙏  Just in case this conversation gets resolved and we try to find it when applying these changes elsewhere 👍

vizro-core/tests/tests_utils/asserts.py
antonymilne: So definitely a good suggestion, but I'm going to leave it how it is 🙂

PR #202, opened 2023-12-08
Fix custom actions multiple return types

antonymilne:
## Description

Fix for the problem discovered with custom actions where `namedtuple` cannot have a field with name that's not a valid Python identifier (e.g. `a-b`).

This turned out to be a much bigger job than expected since we reworked how we handle both inputs and outputs, but the result is much clearer now I think.

* custom actions always handle `inputs` and `outputs` using position only. So type hint in the model is `List[str]`
* the name of parameters in the function is no longer relevant
* no more munging of return values from the callback function; instead we just rely on [Dash's flexible callback signatures](https://dash.plotly.com/flexible-callback-signatures)
* separation of `internal` and `external` `intputs/outputs` to make it clearer what's "behind the scenes"
* pre-defined functions use `Dict[str, str]` for `inputs` and `outputs`

In future:
* we can allow `dict[str, str]` for `inputs` and `outputs`. Only at this point would it become possible to return a dictionary from a custom action function and have it mapped onto multiple outputs. This might be nice to have but is much less important than it used to be because the main motivator for it previously was reusability of a custom action function with fixed parameter names, which we've already achieved here just by relying on ordering. So the main advantage of this would just be for people who like to be explicit or who return lots of different things in their action function
* (related) maybe handle pre-defined actions exactly the same way as custom ones. I don't like that we switch between two different types (dict vs. list) using `if self.inputs` and `if self.outputs`.

## Checklist

## Types of changes

## Notice


vizro-core/docs/pages/user_guides/custom_actions.md
```diff
@@ -133,6 +133,10 @@ The following example shows how to create a custom action that shows the clicked
 
         Vizro().build(dashboard).run()
         ```
+
+        1. Just as for a normal Python function, the names of the arguments `show_species` and `points_data` are arbitrary and do not need to match on to the names of `inputs` in any particular way.
```
petar-qb, line 137: Do we need to emphasise that action function arguments are propagated in the following order:
1. static variables e.g. (`True` from the `my_custom_action(show_species=True)`),
2. dynamic variables e.g. `inputs`

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -74,70 +74,79 @@ def _get_callback_mapping(self):
...
-        elif hasattr(return_value, "_asdict") and hasattr(return_value, "_fields"):
-            # return_value is a namedtuple.
-            if set(return_value._fields) != set(outputs):
+        if isinstance(inputs, Mapping):
+            return_value = self.function(**inputs)
+        else:
+            return_value = self.function(*inputs)
+
+        # Delegate all handling of the return_value and mapping it onto appropriate outputs to Dash - we don't modify
+        # return_value to reshape it in any way. All we do is do some error checking to raise clearer error messages.
```
huong-li-nguyen, line 121:
```suggestion
        # Delegate all handling of the return_value and mapping to appropriate outputs to Dash - we don't modify
        # return_value to reshape it in any way. All we do is do some error checking to raise clearer error messages.
```

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -146,25 +155,40 @@ def build(self):
...
-        for name, object in callback_inputs.items():
-            logger.debug(f"--> {name}: {object}")
+        logger.debug("---------- INPUTS ----------")  # AM: fix this
+        # for name, object in callback_inputs["external"].items():
+        #     logger.debug(f"--> {name}: {object}")
         logger.debug("---------- OUTPUTS ---------")
-        for name, object in callback_outputs.items():
-            logger.debug(f"--> {name}: {object}")
+        # for name, object in callback_outputs["external"].items():
+        #     logger.debug(f"--> {name}: {object}")
```
huong-li-nguyen, line 184: Remove commented out code?

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -34,12 +34,12 @@ class Action(VizroBaseModel):
...
         [],
         description="Inputs in the form `<component_id>.<property>` passed to the action function.",
-        regex="^[a-zA-Z0-9_]+[.][a-zA-Z_]+$",
+        regex="^[^.]+[.][^.]+$",
     )
     outputs: List[str] = Field(
         [],
         description="Outputs in the form `<component_id>.<property>` changed by the action function.",
-        regex="^[a-zA-Z0-9_]+[.][a-zA-Z_]+$",
+        regex="^[^.]+[.][^.]+$",
```
huong-li-nguyen, line 42: I don't think this is what we want? I think how we've had it before was safer because it only allowed for alphanumeric characters. In the new regex, something like this would now pass, which didn't pass before: `inputs=["#1.eo"]` 

Or does it not matter at all now?

huong-li-nguyen (commented):
Tried to understand as much as I could - generally looks good to me 👍  Have a few comments/questions to understand the changes. Also shall we add a test somewhere that could have initially prevented the first merge?

vizro-core/src/vizro/models/_action/_action.py
antonymilne: This shouldn't matter now that we always use position. The only limitation on a valid `input` or `output` is that it must contain a single `.` - otherwise anything is allowed.

I don't think there are any restrictions on Dash component IDs (e.g. "1 blah" would work fine) so it should be fine for us also now.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -146,25 +152,40 @@ def build(self):
...
+            "internal": {"trigger": Input({"type": "action_trigger", "action_name": self.id}, "data")},
+        }
+        callback_outputs = {
+            "internal": {"action_finished": Output("action_finished", "data", allow_duplicate=True)},
+        }
+
+        # If there are no outputs then we don't want the external part of callback_outputs to exist at all.
+        # This allows the action function to return None and match correctly on to the callback_outputs dictionary
+        # The (probably better) alternative to this would be just to define a dummy output for all such functions
+        # so that the external key always exists.
```
antonymilne, line 167: @petar-qb any thoughts on this idea. I think options are:
1. you think it's a huge improvement _and_ it's quick and easy to do ➡️ let's do it here
2. you think it's a small improvement or it's not quick and easy to do ➡️ let's do it but not before release
3. you think it's not an improvement at all ➡️ let's not do it at all and remove the comment

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -282,67 +271,104 @@ def test_action_callback_function_return_value_valid(
         result = action._action_callback_function(inputs={}, outputs=callback_outputs)
         assert result == expected_function_return_value
```
antonymilne, line 272: @petar-qb Are there any cases here where we don't have `expected_function_return_value == custom_action_function_mock_return()`? I don't think so, because all `_action_callback_function` does now is validate stuff.

If I'm right then let's remove the  `expected_function_return_value` and `assert` altogether and just do `action._action_callback_function(inputs={}, outputs=callback_outputs)`. If that incorrectly raises an exception then the test will fail.

I think this `test_action_callback_function_return_value_valid` has become much less valuable but is still worth keeping. Not sure we want to try all these different cases any more because some of them are quite self-evident now though.

Assuming I'm right then

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -282,67 +271,104 @@ def test_action_callback_function_return_value_valid(
...
 
+    @pytest.mark.parametrize("callback_outputs", [[], {}, None])
+    @pytest.mark.parametrize(
+        "custom_action_function_mock_return",
+        [False, 0, "", [], (), {}],
+        indirect=True,
+    )
+    def test_action_callback_function_no_outputs_return_value_exists(
```
antonymilne, line 280: Very nice test 👏

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -282,67 +271,104 @@ def test_action_callback_function_return_value_valid(
...
-            ({"component_1_property": "new_value"}, ["component_1_property", "component_2_property"]),
-            ({"component_1_property": "new_value", "component_2_property": "new_value_2"}, []),
-            (
-                {"component_1_property": "new_value", "component_2_property": "new_value_2"},
-                ["component_1_property", "component_2_property", "component_3_property"],
-            ),
+            (None, ["component_1_property"]),
+            (False, ["component_1_property"]),
+            (0, ["component_1_property"]),
+            (123, ["component_1_property"]),
```
antonymilne, line 296: Please could we add a case with `custom_action_function_mock_return` a dictionary here?

vizro-core/docs/pages/user_guides/custom_actions.md
```diff
@@ -66,7 +66,7 @@ The following example shows how to create a custom action that postpones executi
 ### Interacting with dashboard inputs and outputs
 When a custom action needs to interact with the dashboard, it is possible to define `inputs` and `outputs` for the custom action.
```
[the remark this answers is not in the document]
petar-qb: I think we have a perfect match when users search for the keyword `callbacks`. There is only one matching result that leads to the beginning of `/user_guides/actions` documentation. 

image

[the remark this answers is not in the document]
huong-li-nguyen: Thanks for checking! Somehow I get different results than you, for me the first result are custom components and then these action docs. This means if someone would look for callbacks or how to write their own callbacks, they currently won't find the right docs by keyword search. I think it would be good to mention it in the docs for the custom actions, such that it pops up as one of the first results when looking for "writing own callback", "callbacks" etc.

vizro-core/src/vizro/actions/_filter_action.py
```diff
@@ -33,7 +33,7 @@ def _filter(
     """
     return _get_modified_page_figures(
         targets=targets,
-        ctds_filter=ctx.args_grouping["filters"],
-        ctds_filter_interaction=ctx.args_grouping["filter_interaction"],
-        ctds_parameters=ctx.args_grouping["parameters"],
+        ctds_filter=ctx.args_grouping["external"]["filters"],
```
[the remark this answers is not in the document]
petar-qb: "external" dictionary key is used for handling both pre-defined and custom actions. 
"internal" dictionary key is used for handling internal inputs and outputs like:
- `{"action_finished": Output("action_finished", "data", allow_duplicate=True)},` or
- ` {"trigger": Input({"type": "action_trigger", "action_name": self.id}, "data")}`

This structure helps us to avoid "juggling" (e.g. removing action_finished as `outputs.remove("action_finished")`) in the callback_wrapper. 

This change doesn't affect users (so it can be changed any time) and only helps us not to worry about `"hardcoded"` inputs and outputs that exist for every action (pre-defined or custom).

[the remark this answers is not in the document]
huong-li-nguyen: Thanks for explaining! That makes sense 👍

vizro-core/src/vizro/models/_action/_action.py
maxschulz-COL: So in other words, the components and properties mentioned in `inputs` and `outputs` are still meaningful right? I'm stuck stuck on how this matches with things being positional.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -74,70 +74,79 @@ def _get_callback_mapping(self):
...
+        self,
+        inputs: Union[Dict[str, Any], List[Any]],
+        outputs: Union[Dict[str, Output], List[Output], Output, None],  # AM: check type hints
+    ) -> Any:
         logger.debug("=============== ACTION ===============")
         logger.debug(f'Action ID: "{self.id}"')
         logger.debug(f'Action name: "{self.function._function.__name__}"')
+        # PP: Trim logs when longer than N characters
         logger.debug(f"Action inputs: {inputs}")
+        logger.debug(f"Action outputs: {outputs}")  # AM: check what these look like
```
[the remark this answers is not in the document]
petar-qb: I added basic logging logic for the actions, and the following TODO in places where it looks like we can do some more elegant solution:
`# TODO: solve custom logging format using logging.Handler and custom logging.Filter objects`.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -74,70 +74,79 @@ def _get_callback_mapping(self):
...
 
         action_components = _get_action_callback_mapping(action_id=ModelID(str(self.id)), argument="components")
 
         return callback_inputs, callback_outputs, action_components
 
-    def _action_callback_function(self, inputs: Dict[str, Any], outputs: List[str]) -> Dict[str, Any]:
+    def _action_callback_function(
+        self,
+        inputs: Union[Dict[str, Any], List[Any]],
+        outputs: Union[Dict[str, Output], List[Output], Output, None],  # AM: check type hints
```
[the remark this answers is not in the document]
petar-qb: I double checked it and it looks they are right.

vizro-core/src/vizro/models/types.py
```diff
@@ -81,32 +84,61 @@ def __init__(self, function, /, *args, **kwargs):
...
-        """Run the `function` with the initial arguments overridden by **kwargs.
-
-        Note *args are not possible here, but you can still override positional arguments using argument name.
+        # For a function with parameter **kwargs this converts self.__bound_arguments = {"kwargs": {"a": 1}} into
+        # self.__bound_arguments = {"a": 1}.
+        if var_keyword_param in self.__bound_arguments:
+            self.__bound_arguments.update(self.__bound_arguments[var_keyword_param])
+            del self.__bound_arguments[var_keyword_param]
+
+    def __call__(self, *args, **kwargs):
```
maxschulz-COL, line 93: Given that we have worked on this before, why was there another change needed? It doesn't seem to appear in the PR description

vizro-core/src/vizro/models/_action/_action.py
petar-qb: Exactly. I tested the graph with the id: `"1scatter #- chart"`, and it works with any predefined or custom action.

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -282,67 +299,104 @@ def test_action_callback_function_return_value_valid(
...
+        ],
+        indirect=["custom_action_function_mock_return"],
+    )
+    def test_action_callback_function_outputs_mapping_return_value_not_mapping(
+        self, custom_action_function_mock_return, callback_outputs
+    ):
+        action = Action(function=custom_action_function_mock_return())
+        with pytest.raises(
+            ValueError,
+            match="Action function has not returned a dictionary or similar "
```
maxschulz-COL, line 357: What do we mean by similar? I assume this is a custom error function right? Should we not be more precise, e.g. say we expect `abc.mapping`?

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -282,67 +299,104 @@ def test_action_callback_function_return_value_valid(
...
+        ],
+        indirect=["custom_action_function_mock_return"],
+    )
+    def test_action_callback_function_outputs_list_return_value_length_not_match(
+        self, custom_action_function_mock_return, callback_outputs
+    ):
+        action = Action(function=custom_action_function_mock_return())
+        with pytest.raises(
+            ValueError,
+            match="Number of action's returned elements .+ does not match the number of action's defined outputs .+",
```
maxschulz-COL, line 382: Just out of curiosity, whast does the `.+` do?

maxschulz-COL (approved):
Also took a look just for my understanding, great stuff. Have a few small questions, but would be happy to merge

vizro-core/tests/unit/vizro/models/_action/test_action.py
petar-qb: It really makes sense. I decreased the number of test cases and just leaving a few valid cases for "no outputs", "single output", "multiple list outputs" and "multiple dict outputs".

vizro-core/src/vizro/models/_action/_action.py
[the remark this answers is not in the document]
antonymilne: Improved again in [47f75ba](https://github.com/mckinsey/vizro/pull/202/commits/47f75bafc544df7263027b5312380533fc28f250).

vizro-core/tests/unit/vizro/models/_action/test_action.py
petar-qb: What do you mean? The following case will pass: `({'a': 1}, ["component_1_property"])` (and it should pass). Remember that we transform outputs to outputs[0] before `_action_callback_function` happens

vizro-core/src/vizro/models/_action/_action.py
antonymilne: Excellent, let's leave it for a future change then. Please will you make sure it's captured somewhere? (Maybe that long list of planned improvements to the action implementation?)

vizro-core/tests/unit/vizro/models/_action/test_action.py
petar-qb: It matches 1 or more characters and we put it here because the original exception message is dynamically created like this:
`f"Number of action's returned elements ({len(return_value)}) does not match the number of action's defined outputs ({len(outputs)})."`

vizro-core/tests/unit/vizro/models/_action/test_action.py
antonymilne: Technically the error message should indeed be "Action function has not return a `Mapping`" but I decided that was too technical/confusing because most users know what a dictionary is but not necessarily a Mapping. So the "or similar" case is a hint towards the fact that really you don't have to return a dictionary, just something that's dictionary-like.

If these seems to add more confusion then I would just remove the "or similar" rather than use the terminology "Mapping" I think.

vizro-core/docs/pages/user_guides/custom_actions.md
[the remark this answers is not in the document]
petar-qb: We got different results because I searched for `"calbacks"` and I assume you searched for `"callback"`. 😄 
Even we add a `"callback"` key word in the custom action `.md` file, the "How to create custom component" result will appear as the first suggested answer.

vizro-core/src/vizro/models/_action/_action.py
maxschulz-COL: But this is just for the name of the component targeted right? If this is in first position, it will be passed to the first argument of the function.

We did this, so that we can allow non-pythonic names in components, and pass them positionally to the action functions (same for outputs)?

vizro-core/tests/unit/vizro/models/_action/test_action.py
maxschulz-COL: I think dictionary-like might be best! It has the most used word in it, but it also yields mapping to people that know what that is. With "or similar" I think it's a bit confusing whther it could also be say a list or something

vizro-core/src/vizro/models/_action/_action.py
petar-qb: Yes, Max. With the new implementation we can only propagate UI component values as positional so component with any ID can be and input and output for the custom action 😄.

vizro-core/tests/unit/vizro/models/_action/test_action.py
antonymilne: Good plan! Done in [3052306](https://github.com/mckinsey/vizro/pull/202/commits/30523061ecd43b78d008785ff76967c81800dc49).

vizro-core/docs/pages/user_guides/custom_actions.md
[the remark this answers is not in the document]
petar-qb: It's added to the `custom_actions.md` the `"callbacks"` search now looks like this:
image

petar-qb (approved):
With these changes, the custom actions feature takes a whole new vibe. 🔝

PR #184, opened 2023-11-28
[Snyk] Security upgrade aiohttp from 3.8.6 to 3.9.0

leonnallamuthu:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to fix one or more vulnerable packages in the `pip` dependencies of this project.</h3>

#### Changes included in this PR

- Changes to the following files to upgrade the vulnerable dependencies to a fixed version:
    - vizro-ai/snyk/requirements.txt

<details>
<summary>⚠️ <b>Warning</b></summary>

```
openai 0.28.1 requires aiohttp, which is not installed.

```
</details>

##### By pinning:
Severity                   | Priority Score (*)                   | Issue                   | Upgrade                   | Breaking Change                   | Exploit Maturity
:-------------------------:|-------------------------|:-------------------------|:-------------------------|:-------------------------|:-------------------------
medium severity  |  **663/1000**  <br/> **Why?** Proof of Concept exploit, Recently disclosed, Has a fix available, CVSS 5.4  | Improper Input Validation <br/>[SNYK-PYTHON-AIOHTTP-6091621](https://snyk.io/vuln/SNYK-PYTHON-AIOHTTP-6091621) |  `aiohttp:` <br> `3.8.6 -> 3.9.0` <br>  |  No  | Proof of Concept 
medium severity  |  **663/1000**  <br/> **Why?** Proof of Concept exploit, Recently disclosed, Has a fix available, CVSS 5.4  | Improper Input Validation <br/>[SNYK-PYTHON-AIOHTTP-6091622](https://snyk.io/vuln/SNYK-PYTHON-AIOHTTP-6091622) |  `aiohttp:` <br> `3.8.6 -> 3.9.0` <br>  |  No  | Proof of Concept 

(*) Note that the real score may have changed since the PR was raised.

Some vulnerabilities couldn't be fully fixed and so Snyk will still find them when the project is tested again. This may be because the vulnerability existed within more than one direct dependency, but not all of the affected dependencies could be upgraded.

Check the changes in this PR to ensure they won't cause issues with your project.

------------

**Note:** *You are seeing this because you or someone else with access to this repository has authorized Snyk to open fix PRs.*

For more information:  
🧐 [View latest project report](https://app.snyk.io/org/vizro/project/c5b16d47-7abb-426e-8e60-468beb0e8aba?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr)

[...]

PR #204, opened 2023-12-08
Resolve snyk warnings and expand model options

lingyielia:
## Description
- Address snyk issue suggested by local snyk scan in python 3.9.
    - https://github.com/mckinsey/vizro/pull/184 is a false positive warning. As `aiohttp` dropped support for python 3.7 since version 3.9.0 (https://github.com/aio-libs/aiohttp/releases/tag/v3.9.0). And vizro-ai support python 3.9 and above. Warning raised due to snyk's python 3.7 environment.
- Add two more model options for user testing

## Checklist

## Types of changes

## Notice

### DIFF
--- a/vizro-ai/docs/pages/user_guides/model_config.md
+++ b/vizro-ai/docs/pages/user_guides/model_config.md
@@ -41,9 +41,8 @@ Vizro-AI currently supports the following LLMs:
 - gpt-4-1106-preview (under testing, not suitable for production use)
 
 These models provide different levels of performance and
-cost. In general, `gpt-3.5-turbo` collection provides the most cost-effective models,
-which would be a good starting point for most users. `gpt-4` is more powerful than the other, like
-it allows for more tokens per request. You can refer to these models' [capabilities](https://platform.openai.com/docs/models/overview)
+cost. In general, the `gpt-3.5-turbo` collection provides the most cost-effective models,
+which would be a good starting point for most users. `gpt-4` models are more powerful than `gpt-3` models, e.g. they allow for more tokens per request. You can refer to these models' [capabilities](https://platform.openai.com/docs/models/overview)
 and [pricing](https://openai.com/pricing) for more information.
 
 We are working on supporting more models and more vendors. Please stay tuned!
--- a/vizro-ai/docs/pages/user_guides/model_config.md
+++ b/vizro-ai/docs/pages/user_guides/model_config.md
@@ -41,7 +41,7 @@ Vizro-AI currently supports the following LLMs:
 - gpt-4-1106-preview (under testing, not suitable for production use)
 
 These models provide different levels of performance and
-cost. In general, `gpt-3.5-turbo` collection provides the most cost-effective models, 
+cost. In general, `gpt-3.5-turbo` collection provides the most cost-effective models,
 which would be a good starting point for most users. `gpt-4` is more powerful than the other, like
 it allows for more tokens per request. You can refer to these models' [capabilities](https://platform.openai.com/docs/models/overview)
 and [pricing](https://openai.com/pricing) for more information.
--- a/vizro-ai/docs/pages/user_guides/model_config.md
+++ b/vizro-ai/docs/pages/user_guides/model_config.md
@@ -35,14 +35,15 @@ Vizro-AI currently supports the following LLMs:
 
 ### OpenAI models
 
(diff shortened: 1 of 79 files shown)


vizro-ai/pyproject.toml
```diff
@@ -16,7 +16,7 @@ dependencies = [
   "pandas",
   "tabulate",
   "openai>=0.27.8,<1.0.0",  # TODO add support for openai>=1.0.0
-  "langchain==0.0.325",
+  "langchain==0.0.329",
```
huong-li-nguyen, line 19: Can't we just do:

```suggestion
  "langchain>=0.0.329",
```

This has been raised by Snyk several times now always requiring the latest version. Is there a reason why we would want to pin to a specific version?

lingyielia: I think there are 2 reasons on why we ended up with pin a specific langchain version.
1. Due the langchain is a extremely active (release many times in a week) and thus less stable library, we would always expect to see Snyk found an vulnerability from an older version and ask us to bump to the latest. So even if we do `langchain>=0.0.329`, few weeks later we might need to do something like `langchain>=0.0.345`
2. Also due to langchain is not stable, we want to be sure the version we allow won't break current Vizro-AI. Other than unit tests and integration tests we have, there is another step of user testing (manually run the example and check whether the final output make sense)

lingyielia: > 1. In this case I would actually expect that Snyk would not complain anymore, because `langchain>=0.0.329` would automatically install the latest version and thus have the newest security upgrades included?

Snyk still complains in this case. Say if the latest version is already 0.0.400 and it found an issue in version 0.0.330, which might be fixed in 0.0.399. It will suggest a version bump to langchain>=0.0.399 still. Basically to enforce users to avoid installing versions that are vulnerable.

 

> 2\. I understand that langchain is not stable and then we don't want to run into the risk of breaking the current vizro-ai version, but at the same time we won't be able to identify non-working future versions if we pin to a specific older version. I think we should do something like this `langchain>=0.0.329, <0.1.0` in case we consciously want to upgrade, but pinning to a specific version seems too limiting. WDYT?

I'm quite open to this argument actually. As long as we expect there might be unexpected changes, but the benefit is providing a wider range of versions. @maxschulz-COL , are you aware of any other reasons why we pin to a specific version?

vizro-ai/docs/pages/user_guides/model_config.md
```diff
@@ -35,14 +35,15 @@ Vizro-AI currently supports the following LLMs:
...
+- gpt-4-1106-preview (under testing, not suitable for production use)
 
 These models provide different levels of performance and
-cost. In general, `gpt-3.5-turbo-0613` is the most cost-effective model, which would be a good
-starting point for most users. `gpt-4-0613` is more powerful than the other, like
-it allows for more tokens per request.
-You can refer to these models' [capabilities](https://platform.openai.com/docs/models/overview)
+cost. In general, `gpt-3.5-turbo` collection provides the most cost-effective models,
+which would be a good starting point for most users. `gpt-4` is more powerful than the other, like
+it allows for more tokens per request. You can refer to these models' [capabilities](https://platform.openai.com/docs/models/overview)
```
maxschulz-COL, line 46:
```suggestion
cost. In general, the `gpt-3.5-turbo` collection provides the most cost-effective models,
which would be a good starting point for most users. `gpt-4` models are more powerful than `gpt-3` models, e.g. they allow for more tokens per request. You can refer to these models' [capabilities](https://platform.openai.com/docs/models/overview)
```

vizro-ai/pyproject.toml
maxschulz-COL: I think @Anna-Xiong had quite some extensive difficulties with langchain changing in fundamental ways even between these patch versions. So this is the main reason I can remember why we are pinning the version. At times it was as bad as things breaking several times a week.

Given that we currently have somewhat lower scope to maintain/keep up with breaking changes in langchain I would be cautious in opening up the upper bound. If we ever get the request, it is easy to simply bump to the latest version, run automatic and manual tests, and then bump the dependency (or not). What we gain is a more stable package that does not require constant maintenance and further releases.

maxschulz-COL (approved):

huong-li-nguyen (approved):

Issue #176, opened 2023-11-17
Updating Meta Tags


AnnMarieW:
### What's the problem this feature will solve?

One of the cool features of Dash Pages is that it’s easy to update the meta tags for each page.  This is  good for SEO and  creating nice looking preview cards when sharing a link  on social media.  

Currently with the Vizro demo app,  only the title is included in the meta tags.  Here's what the link preview looks like:
![vizro_on_forum](https://github.com/mckinsey/vizro/assets/72614349/a791971c-59e3-4ced-9222-dc410e90e493)

However,  Dash Pages will update the meta tags with an image and description if supplied.  For example,  see  the [Dash Example Index](https://dash-example-index.herokuapp.com).  It  has over 100 pages, each with it's own title, description and image:

![vizro_vs_dash_example_index](https://github.com/mckinsey/vizro/assets/72614349/392bab7d-a09d-4bd1-be3a-dea237486508)

### Describe the solution you'd like

It's easy to enable this feature in Vizro.  Simple include (an optional) `description`, `image`, and `image_url` parameter in the `Page`. 

I could do a pull request if you would like this feature enabled.   I got it working locally.  You can't see what the link looks like until the app is deployed , but when inspecting the browser, you can see that the meta tags now include an image and description.  

![vizro_updated_meta](https://github.com/mckinsey/vizro/assets/72614349/4dae7a10-dfdd-4d8c-b2c2-c0a5ab822c66)

### Alternative Solutions

No

### Additional context

None at this time.

### Code of Conduct

- [X] I agree to follow the  [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

antonymilne:
Hi @AnnMarieW, thanks very much for the suggestion and the offer to raise a PR 🙏 It's been on my list for a while to check what other things we could set on `dash.register_page` in case we were missing any nice tricks like this, so I love the idea! 

I do have a few hesitations about adding this feature right away though:
* we haven't yet enabled functionality to insert a dashboard-level logo (it's on our backlog though), and I'd like to make sure that however we handle that would be consistent with page-level image information. Related, #70 is our first foray into using navbar icons and we have opted to use Google material icons for that, so I'm not sure what the current thinking is on logos
* a page-level `description` makes a lot of sense to me outside of meta tags anyway - I wonder if we could also use the information as part of the page content and/or homepage navigation cards themselves. I'm not sure whether this has been considered before or if we have designs for it though
[...]

PR #205, opened 2023-12-11
Rearrange and rename main page containers

huong-li-nguyen:
## Description

- Split up `_make_page_layout ` into 2 sub-functions for simplified layout overwrite: 
     - `_get_page_divs`: Creates all the relevant containers that we might want to re-arrange
     - `_arrange_page_divs`: Arranges all 8 containers
- Change `TypedDict` classes that store `html.Divs` to [alternative syntax](https://peps.python.org/pep-0589/#alternative-syntax) such that we can use dashes inside the keys instead of underscores (required for CSS IDs)
- Refactor hideable container behaviour into check `_is_hidden`
- Remove `className` and use `id` only for unique containers, includes renaming:
    - nav_panel_outer -> nav-panel
    - nav_bar_outer -> nav-bar
    - component_container -> components
    - control_panel_outer -> control-panel
- Remove underscores from CSS IDs and replace with dashes for main containers only as usage of dashes are better practice for CSS names (a separate PR will be created to apply the same for the rest of the code basis)

## Screenshot
**This graphic helps understand the current layout arrangements - please take a look:**

In general, there are theoretically 8 main containers that we can re-arrange across the page:
- dashboard-title
- page-title
- settings (currently only includes theme switch)
- nav-panel
- nav-bar
- control-panel
- components
- (logo) (not available yet)

These 8 containers are allocated to these outer containers currently and these outer containers make up our page layout:

**Left Side**
- left-header
- left-sidebar
- left-main

**Right Side**
- right-header
- right-main

Screenshot 2023-12-12 at 09 46 04

## Notice

  - I have not referenced individuals, products or companies in any commits, directly or indirectly
  - I have not added data or restricted code in any commits, directly or indirectly


vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -95,12 +108,19 @@ def build(self):
             fluid=True,
         )
 
-    def _make_page_layout(self, page: Page):
+    def _get_page_divs(self, page: Page) -> PageDivs:
```
huong-li-nguyen, line 111: Splitting this function out should enable the user to create any kind of layout more easily by overwriting the `_make_page_layout` function and re-arranging the elements from the TypedDict returned by `self._get_page_divs`. Some custom CSS might be required as well, but at least the re-arrangement should be easier now.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -95,12 +108,19 @@ def build(self):
...
         # Identical across pages
+        # TODO: Implement proper way of automatically pulling file called logo in assets folder (should support svg, png and it shouldn't matter where it's placed in the assets folder)
+        logo = (
+            html.Div([html.Img(src=get_asset_url("logo.svg"), className="logo-img")], className="logo", id="logo")
+            # TODO: Implement condition check if image can be found/not found
+            if True
+            else html.Div(id="logo", hidden=True)
+        )
```
huong-li-nguyen, line 119: @nadijagraca - this part is what I am leaving up to you. It's hard-coded here just so I could test a few things, but in general that True/False condition would have to be detected automatically. See the relevant TODO :)

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -109,26 +129,62 @@ def _make_page_layout(self, page: Page):
...
-        nav_control_panel = (
-            html.Div(nav_control_elements, className="nav_control_panel")
-            if any(not getattr(element, "hidden", False) for element in nav_control_elements)
+        left_header_divs = [page_divs["dashboard_title"]]
+        left_sidebar_divs = [page_divs["nav_bar"]]
+
+        if getattr(page_divs["nav_bar"], "hidden", False) is False:
+            left_sidebar_divs.insert(0, page_divs["logo"])
+        else:
+            left_header_divs.insert(0, page_divs["logo"])
```
huong-li-nguyen, line 159: Not really pretty, so I am up for improvement suggestions if there are some :) But in general I realised that we could enable all the common 13 layouts as long as we can move the logo to different containers and considering two alternative design proposals I have

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -109,26 +130,63 @@ def _make_page_layout(self, page: Page):
...
-
-        # Arrangement
-        header = html.Div(children=[page_title, theme_switch], className="header", id="header_outer")
-        nav_control_elements = [dashboard_title, nav_panel, control_panel]
-        nav_control_panel = (
-            html.Div(nav_control_elements, className="nav_control_panel")
-            if any(not getattr(element, "hidden", False) for element in nav_control_elements)
-            else None
+        components = page_content["component_container_outer"]
+        return {
```
antonymilne, line 141: Unless there's a good reason not to do so, I'd suggest doing this as an `html.Div` directly rather than a dictionary. Indexing would then be by the relevant `id` like `["dashboard-title"]`.

vizro-core/src/vizro/models/_dashboard.py
antonymilne: This is worth discussing first @nadijagraca because @AnnMarieW will need something very similar for her work on meta tags - see #176. We should end up with something similar to [Dash's `_infer_images`](https://github.com/plotly/dash/blob/dev/dash/_pages.py#L24) but implemented in Pathlib, which makes it very easy to search recursively (see our own `Vizro.__get_external_assets`).

vizro-core/src/vizro/models/_dashboard.py
antonymilne: In general I think we will find this easier if instead of manipulating lists/dictionaries we are manipulating the actual Dash component tree.

If this `getattr(element, "hidden", False)` thing becomes used a lot then let's put it in a helper function also.

vizro-core/src/vizro/models/_dashboard.py
AnnMarieW: Just FYI - I''m making progress on retrieving the images for the logo, meta tags, and the favicon from the app's assets folder. :slightly_smiling_face:

huong-li-nguyen: Amazing - thanks @AnnMarieW ! Let me know if you need any help or we should review :) The icon part will be removed from this PR in the mean time

vizro-core/src/vizro/models/_dashboard.py
huong-li-nguyen: Currently implemented as `_get_hideable_parent_div` function

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -28,6 +28,29 @@
...
+def _is_hidden(components: List[html.Div]):
+    """Returns True if all components are either None and/or have hidden=True."""
+    return all(div is None or getattr(div, "hidden", False) for div in components)
+
+
+# This is just used for type checking. Ideally it would inherit from some dash.development.base_component.Component
+# (e.g. html.Div) as well as TypedDict, but that's not possible, and Dash does not have typing support anyway. When
+# this type is used, the object is actually still a dash.development.base_component.Component, but this makes it easier
+# to see what contract the component fulfills by making the expected keys explicit.
+_PageDivsType = TypedDict(
```
huong-li-nguyen, line 40: Notice the change for the alternative syntax of a TypedDict - it was required because we've decided to use hyphens instead of underscores for CSS classes and IDs. I'll create a separate PR where I'll consolidate it for the rest of the code basis, currently it's only implemented for the main page containers.

vizro-core/src/vizro/models/_dashboard.py
huong-li-nguyen: I went with your proposal from today and implemented a `_is_hidden` check [here](https://github.com/mckinsey/vizro/pull/205/commits/9210ca302411694fdad867267b1abdf0d4bcfac1) - there were definitely pros/cons in this new check compared to the previous function. 

- A pro was definitely that it simplified the code and the tests
- A minor con was that the children were indeed visible in the DOM but also as a hidden attribute (didn't bother me that much)
- A major con was that the CSS stylings with `XXXX:not(:empty)` didn't work properly anymore as hidden children do not count as empty (therefore the containers didn't collapse properly and the dividers were visible even though there was no content)

There is an important CSS fix I've added which I'll explain later on, but this basically got rid of the major con here + fixed some other ambiguous behaviour which we had because we sometimes return `None` or `html.Div(hidden=True)`. So I sticked with this approach 👍

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -153,6 +164,18 @@ div.dashboard_container .custom-tooltip {
...
+  flex-direction: column;
+  width: 64px;
+  background-color: var(--surfaces-bg-02);
+  border-right: 1px solid var(--border-subtle-alpha-01);
+  gap: 40px;
+}
+
+#page-container div[hidden] {
+  display: none;
+}
```
huong-li-nguyen, line 181: This is the important CSS fix @antonymilne 

We've been changing this a few times now but I think the current approach is the least ambiguous one now. Initially we had `className=hidden` that applied `display:none` to the relevant containers, but later decided to return `html.Div(id=XXX, hidden=True)` for containers with IDs while returning `None` for containers without IDs. This lead to different behaviour which became apparent in this PR, as I've provided IDs to the 8 main containers and therefore returned `hidden=True` instead of previously returning `None`.

The odd behaviour was that the containers didn't collapse properly + they took space and created gaps even though there was no content. This can e.g. happen if these hidden containers:
- have padding embedded (in which case the padding is visible even though it's hidden)
- are located in an outer flexbox container with flexbox with a default gap (hidden containers created gaps where there shouldn't be)


**As a reminder:**
- `display:none `: Will not appear on page, visible in the DOM, no space allocated to the components
- `visibility:hidden`: Will not appear on page, visible in the DOM, space is allocated to the components even if hidden
- `None`: Has the advantage of both - not visible in the DOM, not allocating space

So in short - this CSS fix basically applies `display:none` to all hidden Divs automatically. We can apply the same rule as before, but now the visual behaviour should be the same regardless whether we return hidden=True divs or None.

huong-li-nguyen:
@l0uden - could you help me figure out why the tests are failing? Locally I've just checked and it seems to work fine :) On CircleCI I see a couple of timeout exceptions? I've changed some of the CSS IDs, so if you use them in your tests they might have to get updated. If there are visual changes which are differing, then let me know! I should then fix it on my side.

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -105,7 +116,7 @@
   width: 100%;
 }
 
-.nav-bar {
+#nav-bar {
   align-items: center;
   background-color: var(--surfaces-bg-02);
```
nadijagraca, line 121: It's not change made in this PR, but maybe we should delete `background-color: var(--surfaces-bg-02);` from #nav-bar and icon-button classes, and let those components inherit background color from the #left-sidebar. 

Apart from being unnecessary property, it makes it hard for users to overwrite it, as currently if users want to overwrite the background color of entire left sidebar, they have to overwrite all three classes.

vizro-core/src/vizro/static/css/layout.css
antonymilne: Makes complete sense, thank you for the full explanation!

antonymilne (approved):
Really great work and an excellent description! ⭐ 🌟 ⭐ 

Just some small questions and suggestions, nothing big though.

I'm very much enjoying the clean commit history. If you want to take it to the next level then you could pull updates from `main` using rebase rather than merge ([also possible from Github UI if you press the little arrow next to the Update button)](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/keeping-your-pull-request-in-sync-with-the-base-branch#updating-your-pull-request-branch).

vizro-core/src/vizro/models/_dashboard.py
antonymilne: This seems like a great solution, thank you for working through it 👍

antonymilne: The whole thing looks much nicer than the original bit of code we were commenting on here with the `getattr` and the `insert` 💯

antonymilne:
Somewhat related to this PR so let me ask it here... Is there a reason we have this in `Dashboard.build`?
```
dbc.Container(
        id="dashboard_container_outer",
```

It's the only example of the `dbc.Container` we have - everything else is done with `html.Div`. Is it for the `fluid=True` or just historical accident?

vizro-core/src/vizro/static/css/layout.css
huong-li-nguyen: That's a great suggestion - thanks! ⭐ 

I've decided to get rid of the `background-color` from all the inner containers and just add it to the outermost container. The inner ones will automatically inherit. Implemented here: https://github.com/mckinsey/vizro/pull/205/commits/f2fb529489c8e6bab78494e72492c319eda39c7d

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -99,41 +122,55 @@ def build(self):
...
         dashboard_title = (
-            html.Div(children=[html.H2(self.title), html.Hr()], className="dashboard_title", id="dashboard_title_outer")
+            html.Div([html.H2(self.title)], id="dashboard-title")
             if self.title
-            else html.Div(hidden=True, id="dashboard_title_outer")
+            else html.Div(id="dashboard-title", hidden=True)
         )
-        theme_switch = daq.BooleanSwitch(
-            id="theme_selector", on=self.theme == "vizro_dark", persistence=True, persistence_type="session"
+        settings = html.Div(
```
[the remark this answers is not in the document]
huong-li-nguyen: Ah, good point! 

In general, we should remove all outer divs if they are not required (keeps the DOM clean and prevents weird sizing issues). That's why `page_title` isn't wrapped inside another div, same should apply for `dashboard_title` (I'll fix that now!).

The reason why the theme switch is embedded in an outer div is mostly premature optimization 😄 In the future we want to add other buttons inside that div currently called settings, so I already created that outer div here.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -99,41 +122,55 @@ def build(self):
...
-            else None
-        )
+        control_panel = page_content["control-panel"]
+        components = page_content["components"]
+        return html.Div([dashboard_title, settings, page_title, nav_bar, nav_panel, control_panel, components])
 
-        left_side = html.Div(children=[nav_bar, nav_control_panel], className="left_side", id="left_side_outer")
-        right_side = html.Div(children=[header, component_container], className="right_side", id="right_side_outer")
-        return html.Div([left_side, right_side], className="page_container", id="page_container_outer")
+    def _arrange_page_divs(self, page_divs: html.Div):
```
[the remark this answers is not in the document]
huong-li-nguyen: Ah yes, I missed this one! That makes sense! 👍

vizro-core/src/vizro/static/css/accordion.css
```diff
@@ -104,7 +104,7 @@
   width: 100%;
 }
 
-div.page_container .accordion-item-button {
+#page-container .accordion-item-button {
```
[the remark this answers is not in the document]
huong-li-nguyen: No, that's a good question! It has less to do with CSS, but the ability to overwrite some of Dash default CSS properties. In some cases doing `.accordion-item-button` isn't sufficient to overwrite the default CSS coming from Dash, so adding `#page-container` at the beginning gives the CSS selector higher specificity and therefore the ability to overwrite.

However, I've just checked this entire file and in all of the CSS classes the #page-container wasn't necessary to overwrite the properties, so I've removed it everywhere 👍 

@nadijagraca - we should probably double-check at some point for which CSS selectors we actually need it.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -28,6 +28,29 @@
 logger = logging.getLogger(__name__)
 
 
+def _is_hidden(components: List[html.Div]):
+    """Returns True if all components are either None and/or have hidden=True."""
+    return all(div is None or getattr(div, "hidden", False) for div in components)
```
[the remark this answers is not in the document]
huong-li-nguyen: I like it! `Component` would be of `Any` type?

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -153,6 +164,18 @@ div.dashboard_container .custom-tooltip {
...
+#left-sidebar {
+  display: flex;
+  flex-direction: column;
+  width: 64px;
+  background-color: var(--surfaces-bg-02);
+  border-right: 1px solid var(--border-subtle-alpha-01);
+  gap: 40px;
+}
+
+#page-container div[hidden] {
```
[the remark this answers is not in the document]
huong-li-nguyen: `div[hidden]` also includes `h2` :) 
Without the div, this actually wouldn't work for any divs, also not h2.

vizro-core/src/vizro/models/_dashboard.py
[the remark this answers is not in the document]
antonymilne: No, `Component` is specifically `dash.development.base_component.Component`.

vizro-core/src/vizro/models/_dashboard.py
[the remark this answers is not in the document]
antonymilne: Makes sense, thank you. Definitely agree with not having the outer divs in general - makes things much cleaner.

vizro-core/tests/unit/vizro/models/test_page.py
```diff
@@ -157,4 +157,4 @@ def test_page_build_left_side_removed(standard_px_chart):
     page = vm.Page(title="Single Page", components=[vm.Graph(id="scatter_chart", figure=standard_px_chart)])
     dashboard = vm.Dashboard(pages=[page])
     dashboard.pre_build()
-    assert "className='nav_control_panel'" not in str(page.build())
+    assert "id='left-main'" not in str(page.build())
```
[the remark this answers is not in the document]
huong-li-nguyen: Ah, I think this is an outdated test. I've just checked and it doesn't make any sense anymore since we introduced the hidden attribute and moved the layout creation to the dashboard. 

I would kill this test completely, because this test was intended to check whether the left-main side collapses properly if there is only one page with no controls and no navigation. Even if I would re-write this to the dashboard.build(), "left-main" will actually be present in the component tree, but its attribute is hidden. So I think testing whether it then actually collapses need to be captured by the screenshot tests.

vizro-core/src/vizro/models/_dashboard.py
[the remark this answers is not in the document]
huong-li-nguyen: Got it!

vizro-core/tests/unit/vizro/models/_navigation/test_nav_bar.py
```diff
@@ -93,13 +93,9 @@ def test_nav_bar_active_pages_as_dict(self, pages_as_dict):
...
             ]
         )
-        assert_component_equal(built_nav_bar["nav_bar_outer"], expected_button, keys_to_strip={"id", "className"})
-        assert_component_equal(
-            built_nav_bar["nav_panel_outer"],
-            html.Div(id="nav_panel_outer", className="nav_panel"),
-            keys_to_strip={"children"},
-        )
-        assert_component_equal(built_nav_bar["nav_panel_outer"].children, [dbc.Accordion()], keys_to_strip=STRIP_ALL)
+        assert_component_equal(built_nav_bar["nav-bar"], expected_button, keys_to_strip={"id", "className"})
```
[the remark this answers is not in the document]
huong-li-nguyen: That's a good point - I've double-checked and they were required everywhere as most of them were ignoring the `classNames` in the icons, which I didn't change 👍

huong-li-nguyen:
> Details

Very good question! I've seen this for years now, so I didn't every question it 😅 I think it's historical left-over. I've just replaced it with an `html.Div` given that the `fluid` argument has no effect. The `fluid` argument adds 15px padding to the screen (which we don't want). However, this was never visible because we overwrite it in our global CSS settings.

See this for reference: https://stackoverflow.com/questions/68823379/dash-dbc-container-fluid-true-does-not-cover-all-available-space

huong-li-nguyen:
> also possible from Github UI if you press the little arrow next to the Update button

Very good question! I think historical accident 😅 

The two behave pretty much the same. The `fluid=True` in theory adds 15px padding to the view, but this actually never worked as we globally overwrite it with our CSS. So there is no advantage in using the dbc.Container compared to a simple html.Div as far as I know. I've replaced it everywhere now.

Read the answer on this one for reference: https://stackoverflow.com/questions/68823379/dash-dbc-container-fluid-true-does-not-cover-all-available-space

vizro-core/tests/unit/vizro/models/test_page.py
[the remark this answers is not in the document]
antonymilne: In an ideal world I think we would test this as part of tests for `dashboard.build`. We'd make a page with no controls and no navigation and check for the hidden component in the build output using `assert_component_equal`. I wrote very similar tests for navigation to check that the nav panel is hidden when you have an accordion with only one item.

But writing those tests now is not needed, since it would take some time to do I know. I'd consider it part of the point I mention as future work writing more tests in #195.

antonymilne:
> Very good question! I think historical accident 😅
> 
> The two behave pretty much the same. The `fluid=True` in theory adds 15px padding to the view, but this actually never worked as we globally overwrite it with our CSS. So there is no advantage in using the dbc.Container compared to a simple html.Div as far as I know. I've replaced it everywhere now.
> 
> Read the answer on this one for reference: https://stackoverflow.com/questions/68823379/dash-dbc-container-fluid-true-does-not-cover-all-available-space

Thanks for checking this. It's always bugged me that we had this inconsistency 😀 Everything looks so much cleaner than it used to now 💯 

Some time next time I think we should discuss again the different options for how to handle the layout, e.g. the `grid-template-area` vs. using `dbc.Grid` vs. our current approach. @AnnMarieW had some good points here about performance on small screens here so would be good to talk it through with her I think. Also Ottis knows lots about this sort of thing so would be worth brainstorming with.

But for now I love the approach! ⭐

huong-li-nguyen:
> > Very good question! I think historical accident 😅
> > The two behave pretty much the same. The `fluid=True` in theory adds 15px padding to the view, but this actually never worked as we globally overwrite it with our CSS. So there is no advantage in using the dbc.Container compared to a simple html.Div as far as I know. I've replaced it everywhere now.
> > Read the answer on this one for reference: https://stackoverflow.com/questions/68823379/dash-dbc-container-fluid-true-does-not-cover-all-available-space
> 
> Thanks for checking this. It's always bugged me that we had this inconsistency 😀 Everything looks so much cleaner than it used to now 💯
> 
> Some time next time I think we should discuss again the different options for how to handle the layout, e.g. the `grid-template-area` vs. using `dbc.Grid` vs. our current approach. @AnnMarieW had some good points here about performance on small screens here so would be good to talk it through with her I think. Also Ottis knows lots about this sort of thing so would be worth brainstorming with.
> 
> But for now I love the approach! ⭐

Yes, I agree! I that these topics on top of my list and to also discuss w/ @AnnMarieW ! 🚀

AnnMarieW:
> It's the only example of the dbc.Container we have - everything else is done with html.Div. Is it for the fluid=True or just historical accident?

I think the bigger question is:  Why are you using Dash Bootstrap Components at all?  You aren't loading a stylesheet, correct?  This is a big problem since it's not uncommon to have a Bootstrap stylesheet in a user's assets folder.   Here's what the example app app looks like if someone has the Spacelab theme in /assets

image

huong-li-nguyen:
> > It's the only example of the dbc.Container we have - everything else is done with html.Div. Is it for the fluid=True or just historical accident?
> 
> I think the bigger question is: Why are you using Dash Bootstrap Components at all? You aren't loading a stylesheet, correct? This is a big problem since it's not uncommon to have a Bootstrap stylesheet in a user's assets folder. Here's what the example app app looks like if someone has the Spacelab theme in /assets
> 


Hey @AnnMarieW, that's a very good point! 

We have yet to figure out how to make templating as easy as possible. So far, we create our templates [here](https://github.com/mckinsey/vizro/blob/main/vizro-core/src/vizro/static/css/variables.css) and users can configure them in `dashboard.theme`. Currently, it only supports `vizro_dark` and `vizro_light`, but the future idea is to add more templates and facilitate the process for users to create their own.

We generally prefer dbc components over dcc components when dcc becomes challenging to style (some dash properties were tricky to overwrite) or when the dbc components enable additional functionality. Let me create another issue on this one so we can discuss it. :)

PR #210, opened 2023-12-13
Tidy/typing

antonymilne:
## Checklist

## Types of changes

## Notice

### DIFF
--- a/vizro-core/src/vizro/models/_components/form/_form_utils.py
+++ b/vizro-core/src/vizro/models/_components/form/_form_utils.py
@@ -1,8 +1,10 @@
 """Helper functions for models inside form folder."""
-from typing import Union
+from typing import Union, List
+
+from typing_extensions import reveal_type
 
 from vizro._constants import ALL_OPTION
-from vizro.models.types import MultiValueType, OptionsType, SingleValueType
+from vizro.models.types import MultiValueType, OptionsType, SingleValueType, OptionsDictType
 
 
 def get_options_and_default(options: OptionsType, multi: bool = False):
@@ -14,6 +16,7 @@ def get_options_and_default(options: OptionsType, multi: bool = False):
             options = [ALL_OPTION, *options]
 
     if all(isinstance(option, dict) for option in options):
+        # Each option is a OptionsDictTypo.
         default_value = options[0]["value"]  # type: ignore[index]
     else:
         default_value = options[0]
--- a/vizro-core/src/vizro/models/_dashboard.py
+++ b/vizro-core/src/vizro/models/_dashboard.py
@@ -43,7 +43,7 @@ class Dashboard(VizroBaseModel):
     theme: Literal["vizro_dark", "vizro_light"] = Field(
         "vizro_dark", description="Layout theme to be applied across dashboard. Defaults to `vizro_dark`"
     )
-    navigation: Optional[Navigation] = None
+    navigation: Navigation = None  # type: ignore[assignment]
     title: str = Field("", description="Dashboard title to appear on every page on top left-side.")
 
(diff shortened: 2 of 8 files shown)


vizro-core/src/vizro/models/_components/form/_form_utils.py
```diff
@@ -14,6 +14,7 @@ def get_options_and_default(options: OptionsType, multi: bool = False):
             options = [ALL_OPTION, *options]
 
     if all(isinstance(option, dict) for option in options):
+        # Each option is a OptionsDictTypo.
```
huong-li-nguyen, line 17:
```suggestion
        # Each option is a OptionsDictType
```

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -43,7 +43,7 @@ class Dashboard(VizroBaseModel):
     theme: Literal["vizro_dark", "vizro_light"] = Field(
         "vizro_dark", description="Layout theme to be applied across dashboard. Defaults to `vizro_dark`"
     )
-    navigation: Optional[Navigation] = None
+    navigation: Navigation = None  # type: ignore[assignment]
```
huong-li-nguyen, line 46: Won't repeat this comment, but this will apply everywhere you changed it: In theory we would also have to update the docstrings to reflect that change e.g. here it would be 

`navigation (Navigation): See [`Navigation`][vizro.models.Navigation]. Defaults to `None`.`

Is that a good or bad idea?

huong-li-nguyen (approved):
Thanks for double-checking again! 🚀 

What you wrote makes a lot of sense and I agree that this is probably still a bit naughty, but better than the other alternative 👍

vizro-core/src/vizro/models/_dashboard.py
maxschulz-COL: Good catch! And I am not sure what the answer is. We had previous discussions about the type hints, but I cannot find them quickly. The discussion evolved around whether we want to add as type what the user puts in, or what thing it is coerced into.

That being said, currently this type hint referred to what the user could put in (at least in this case): `None` | `Navigation`, which `Optional[Navigation]` described well. So as long as we agree with this, I think we should actually keep as is.

maxschulz-COL (approved):
I like this change, and I think the reasoning is sound although of course in a perfect world this would not be necessary

vizro-core/src/vizro/models/_dashboard.py
antonymilne: Hmmm, it's a good question. What @maxschulz-COL says is completely right, and currently `None` is a valid value. Maybe we should change it to keep consistency with the field description though because even though it's valid, it's not really an intended input. The fact that the field is optional (in a non-required sense) is clear from the fact that there's a default value listed. The fact that you could set `navigation = None` if you wanted isn't something we should expect anyone to do really (you would just not supply a value for `navigation` at all instead).

So I see pros and cons both ways and it's a very minor thing, but overall I'm tempted to change it.

huong-li-nguyen: This reminds me actually - I did change it already e.g. in the last PR for `nav_selector` in `vm.Navigation`. If we decide to not change it, you would then at least have to change the docstring back. 

But I agree that I am also tempted to change it because None is not a value we expect anyone to actually insert.

maxschulz-COL: Let's go with changing then, also makes the docstring shorter :)

antonymilne: Done!

PR #212, opened 2023-12-13
Update configuration for demo dashboard

huong-li-nguyen:
## Description
- Tidy up some of the data generation code snippets
- Improve demo dashboard
  - Move homepage to its own icon
  - Format numbers in data table for benchmark analysis page
  - Improve layout of benchmark analysis page (e.g. Add parameter, change layout, change title so it's clearer what the user can interact with)

Note: This is just a quick fix to improve the last screen of the demo dashboard. Hence, I didn't keep the yaml, dict, json examples in sync as we still need to align whether we want to keep these for this demo dashboard or not. This will be tackled in another PR after alignment on what we should do with the demo / feature compendium dashboard.

## Screenshot
Before:
Screenshot 2023-12-13 at 14 57 59

After:
Screenshot 2023-12-13 at 14 48 55

## Checklist

## Types of changes

## Notice

### DIFF
--- a/vizro-core/examples/default/app.py
+++ b/vizro-core/examples/default/app.py
@@ -8,7 +8,6 @@ from vizro import Vizro
 from vizro.actions import export_data, filter_interaction
 from vizro.tables import dash_data_table
 
-# DATA -----
 df = px.data.gapminder()
 df_mean = (
     df.groupby(by=["continent", "year"]).agg({"lifeExp": "mean", "pop": "mean", "gdpPercap": "mean"}).reset_index()
@@ -21,7 +20,6 @@ df_transformed["pop"] = df.groupby(by=["continent", "year"])["pop"].transform("s
 df_concat = pd.concat([df_transformed.assign(color="Continent Avg."), df.assign(color="Country")], ignore_index=True)
 
 
-# CONFIGURATIONS -----
 def create_variable_analysis():
     """Function returns a page with gapminder data to do variable analysis."""
     page_variable = vm.Page(
--- a/vizro-core/examples/default/app.py
+++ b/vizro-core/examples/default/app.py
@@ -453,14 +453,14 @@ def create_benchmark_analysis():
             ),
         ],
         controls=[
+            vm.Filter(column="continent", selector=vm.Dropdown(value="Europe", multi=False, title="Select continent")),
+            vm.Filter(column="year", selector=vm.RangeSlider(title="Select timeframe", step=1, marks=None)),
             vm.Parameter(
                 targets=["line_country.y"],
                 selector=vm.Dropdown(
                     options=["lifeExp", "gdpPercap", "pop"], multi=False, value="gdpPercap", title="Choose y-axis"
                 ),
             ),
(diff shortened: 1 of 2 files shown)


petar-qb (approved):
I like these changes.

Small suggestion: Maybe we should put the Parameter after Filters inside the benchmark page.

antonymilne (approved):
Looks good!

Do you think we should change the `NavLink` to have some nice icons here rather than the default `filter_1/2/3`? It would make our demo look a bit slicker maybe, but also adds complexity to the example app. Up to you!

vizro-core/examples/default/app.py
```diff
@@ -8,21 +8,20 @@
...
-    df = px.data.gapminder()
-    mean = (
-        df.groupby(by=["continent", "year"]).agg({"lifeExp": "mean", "pop": "mean", "gdpPercap": "mean"}).reset_index()
-    )
-    return mean
+df_transformed = df.copy()
+df_transformed["lifeExp"] = df.groupby(by=["continent", "year"])["lifeExp"].transform("mean")
+df_transformed["gdpPercap"] = df.groupby(by=["continent", "year"])["gdpPercap"].transform("mean")
+df_transformed["pop"] = df.groupby(by=["continent", "year"])["pop"].transform("sum")
+df_concat = pd.concat([df_transformed.assign(color="Continent Avg."), df.assign(color="Country")], ignore_index=True)
```
maxschulz-COL, line 20: I have only just read this code properly for the first time. I think in its current form it is actually largely confusing: We are associating the mean and sum across continents to individual countries again without changing the column names to reflect this, or without eliminating the countries so that we only look a continent and year.

I would probably change this to something like:

```py
df_transformed = df.groupby(by=["continent", "year"]).agg({"lifeExp": "mean", "pop": "sum", "gdpPercap": "mean"}).reset_index()
```

maxschulz-COL: I think this may be related to the reason why the average shows up weirdly (including being labeled with Zimbabwe here): 
image

Can we check this quickly?

maxschulz-COL (changes requested):
Generally really like it, but I found some off behviour in the transformed df display. Can you check quickly?

vizro-core/examples/default/app.py
huong-li-nguyen: I tried your suggestion above before actually, but it's not the same 😄 If you do above, it will only give you the country results - not the country next to the continent average result.

huong-li-nguyen: The results also seem coherent for me? 
Screenshot 2023-12-13 at 16 28 35

The chart on the right only makes sense if you filter on one country, so select a cell in the table

maxschulz-COL: Oh haha ok :)

Hm ok that makes more sense. I guess it's not quite ideal, but I can see why now!

huong-li-nguyen: Is it good to go for you then? 😄 Would merge after your approval

maxschulz-COL (approved):

Issue #194, opened 2023-12-06
Convert warnings into errors in tests


antonymilne:
Currently our tests show any deprecation warnings raised during running but these are easy to miss and I expect everyone ignores them (I do anyway). Let's get stricter about this so (a) we don't miss future deprecations in our dependencies that could break us and (b) our users aren't bothered by deprecation warnings.

This applies to both `vizro-ai` and `vizro-core`, though could be done in two separate PRs if easier.

* Add `filterwarnings = "error"` to `tool.pytest.ini_options` in `pyproject.toml`
* Run all tests and see what new errors are raised
* Go through these errors and work out whether we should fix them or ignore them
* For any that should be ignored, ignore them in the "least global" way possible.

See https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html for more guidance.

PR #214, opened 2023-12-14
Convert warnings into errors in tests

huong-li-nguyen:
## Description
Following up on https://github.com/mckinsey/vizro/issues/194 - only 2 warnings were converted to errors:
```

UserWarning: Importing LLMChain from langchain root module is no longer supported. Please use langchain.chains.LLMChain instead. 

HTTPResponse.getheader() is deprecated and will be removed in urllib3 v2.1.0. Instead use HTTPResponse.headers.get(name, default)
```

- The first one is fixed by simply changing the import statement in vizro-ai
- This second one is coming from dash[testing] as we use `dash_duo` for our tests only. The deprecation warning seems to be fixable by [upgrading selenium to 4.6](https://stackoverflow.com/questions/75382508/deprecationwarning-httpresponse-getheader-is-deprecated-and-will-be-removed-i). However, dash[testing] has an upper bound on selenium and [an issue](https://github.com/plotly/dash/issues/2590) has already been created on dash being aware of the deprecation warning (though I am not sure if they release the upper bound anytime soon). For now, I would just suggest to ignore it.

## Checklist

## Types of changes

## Notice


vizro-core/tests/integration/test_examples.py
```diff
@@ -33,6 +33,8 @@ def dashboard(request, monkeypatch):
     return app["dashboard"]
 
 
+# Ignore deprecation warning until this is solved: https://github.com/plotly/dash/issues/2590
+@pytest.mark.filterwarnings("ignore:HTTPResponse.getheader()")
```
huong-li-nguyen, line 37: Currently only ignoring it locally for this test as this is the only place where we use dash_duo. If at some point we use dash_duo in other places as well, we can decide to move this ignore one level up etc.

maxschulz-COL (approved):
Makes sense, thanks for taking care of this

petar-qb (approved):
Makes sense!

Issue #193, opened 2023-12-06
Add Python 3.12 to tests


antonymilne:
Quick and easy:
* In our `matrix` on Github actions, we currently test on `python-version` up to and including 3.11. Let's add 3.12 to the list
* Include 3.12 in `classifiers` in `pyproject.toml`
* Include 3.12 in our badge in the README

Could be difficult but I don't expect it will be necessary:
* Fix any problems that make our code incompatible with 3.12

PR #216, opened 2023-12-14
Add Python 3.12 to tests

huong-li-nguyen:
## Description
Following up on  [Add Python 3.12 to tests:
](https://github.com/mckinsey/vizro/issues/193)

**In vizro-core and vizro-ai:**
- [x]  Include 3.12 in classifiers in pyproject.toml
- [x]  Include 3.12 in our badge in the README
- [x]  Include 3.12 in [[envs.all.matrix]]
- [x]  Make sure the relevant 3.12 jobs are required for merging in Github branch protection settings
- [x]  Fix any problems that make our code incompatible with 3.12 (only got one new deprecation warning coming from pandas as pandas doesn't support python 3.12 just yet - decided to ignore that particular deprecation warning for now)

## Checklist

## Types of changes

## Notice


vizro-core/hatch.toml
```diff
@@ -1,11 +1,11 @@
 [envs.all]
 
 [[envs.all.matrix]]
-python = ["3.8", "3.9", "3.10", "3.11"]
+python = ["3.8", "3.9", "3.10", "3.11", "3.12"]
 
 [envs.all.overrides]
 matrix.python.features = [
-  {value = "kedro", if = ["3.8", "3.9", "3.10"]}
+  {value = "kedro", if = ["3.8", "3.9", "3.10", "3.11"]}
```
huong-li-nguyen, line 8: Kedro is now compatible with 3.11

antonymilne: 1. In order for the tests to run you'll also need to remove `kedro = pytest.importorskip("kedro")` also. Look at the GHA logs to see which tests are currently running in the different Python versions. This would close https://github.com/McK-Internal/vizro-internal/issues/362
2. Then ideally we would just make this `matrix.python.features = ["kedro"]` and it would run on 3.12 also
3. Just glancing at the kedro source though, I think this might raise some warning which we then convert to an error... So let me know what happens after step 2 and we can figure out what to do!

vizro-ai/pyproject.toml
```diff
@@ -48,4 +49,8 @@ parallel = true
 source_pkgs = ["vizro_ai"]
 
 [tool.pytest.ini_options]
-filterwarnings = "error"
+filterwarnings = [
+  "error",
+  # Ignore this warning here as it comes from pandas until pandas is made compatible with Python 3.12
+  'ignore:.*utcfromtimestamp.*:DeprecationWarning'
```
antonymilne, line 55: Which tests raise this warning or is it all over the place?

vizro-core/hatch.toml
huong-li-nguyen: Ah, got it!

So with Python 3.12 you do get a warning converted to an error:
Screenshot 2023-12-14 at 18 26 19

All the tests with Python 3.11 work fine 👍 

Shall we keep it as we have right now then?

antonymilne: Can we do an ignore with the `filterwarnings` thing in the right file and still run the tests on 3.12?

vizro-ai/pyproject.toml
huong-li-nguyen: Like the imports of the tests don't even work:

```
ImportError while loading conftest '/Users/huong_li_nguyen/Box Sync/GitRepos/vizro/vizro-core/tests/conftest.py'.
tests/conftest.py:3: in <module>
    from vizro import Vizro
src/vizro/__init__.py:4: in <module>
    from ._vizro import Vizro
src/vizro/_vizro.py:9: in <module>
    from vizro.managers import data_manager, model_manager
src/vizro/managers/__init__.py:2: in <module>
    from ._data_manager import data_manager
src/vizro/managers/_data_manager.py:4: in <module>
    import pandas as pd
```

vizro-core/hatch.toml
huong-li-nguyen: Somehow it doesn't work with the Kedro warning - I've tried different combinations now and also what worked already in other tests. But on the Kedro one, it just doesn't filter it. Maybe you see something obviously wrong that I can't see right now 😄 

```
@pytest.mark.filterwarnings("ignore:Kedro is not yet fully compatible with this Python version.")
def test_datasets_from_catalog(catalog_path):
    from kedro.io import DataCatalog  # noqa: E402
    from vizro.integrations.kedro import datasets_from_catalog  # noqa: E402
    ...
```

Tried these as well:
```
@pytest.mark.filterwarnings("ignore::UserWarning")
@pytest.mark.filterwarnings("ignore:Kedro")
```

or directly inside the pyproject.toml:
```
filterwarnings = [
  "error",
  'ignore:.*utcfromtimestamp.*:DeprecationWarning',  # this works
  'ignore:.*Kedro is not yet fully compatible with this Python version.*',
  'ignore:.*Kedro is not yet fully compatible.*:UserWarning',
  'ignore:.*Kedro.*:UserWarning'
]
```

However, somehow none of them work on the Kedro warning and it doesn't get filtered out during the tests 🤔

antonymilne:
Note in case we need to use the environment `all.py3.12` locally some time... Here's how I did it:

1. `brew install python@3.12`
2. Trying to do anything with the hatch `all.py3.12` environment fails with [this error](https://stackoverflow.com/questions/77364550/attributeerror-module-pkgutil-has-no-attribute-impimporter-did-you-mean)
3. Hatch uses virtualenv for its environments, but it's not possible to do any hatch commands in the environment to fix it because as soon as you activate `all.py3.12` it starts running `pip`, which is the cause of the failure. Hence to fix you need to find the environment (`hatch env find`) and then `source activate` to manually activate the virtualenv
4. After that doing `python -m ensurepip --upgrade` fixed the pip version for me, but maybe you'd also need to do some of the other commands there to clear the virtualenv cache

I also did `pipx upgrade hatch` but not sure if it's necessary.

vizro-core/hatch.toml
antonymilne: Yeah I think you're right here. I tried to fix it in [a1b5b88](https://github.com/mckinsey/vizro/pull/216/commits/a1b5b8807ab042c6236f3731633c9bfcd792bc87) but couldn't actually get things running locally so not sure if it will work.

If it doesn't work then please can you try adding the same `ignore` to `filterwarnings` in addition to where it is now?

And if that doesn't work and you can't figure out after spending ~20 mins on it then just put it back to how you had it in [b5a6a45](https://github.com/mckinsey/vizro/pull/216/commits/b5a6a4545a0aa0c39744a78af858b8c358197db8) where we don't do `kedro` tests in 3.12.

antonymilne: If you can get it working then remember to remove `kedro = pytest.importorskip("kedro")` also 🙂 Otherwise if we are skipping kedro in 3.12 then it needs to stay there.

.github/workflows/test-integration-vizro-ai.yml
```diff
@@ -30,7 +30,9 @@ jobs:
             hatch-env: all.py3.10
           - python-version: "3.11"
             hatch-env: all.py3.11
-          - python-version: "3.11"
+          - python-version: "3.12"
+            hatch-env: all.py3.12
+          - python-version: "3.12"
```
antonymilne, line 35:
```suggestion
          - python-version: "3.11"
```

Given Python 3.12 is pretty new and a bit problematic, let's keep 3.11 as our "standard" version for running stuff in CI, including as the version for `lower-bounds`.

Same change would apply across all the github workflows.

vizro-core/hatch.toml
huong-li-nguyen: Interesting that you couldn't run it locally! I actually didn't have to do any of the steps you mentioned below. I just installed Python 3.12 and then `run hatch run all.py3.12:test -vv` while `all.py3.09` is my hatch default environment. This way I was able to run all the tests for each python version locally.

Yes, I've tried what you suggested above for like 30 min yesterday but couldn't get it working :( 
The Kedro warning/error keeps popping even when adding the ignores locally + globally or only one of them. I tried all combinations yesterday 😓

antonymilne: So after many attempts to get my hatch.toml right I feel like the latest commit [962e4ea](https://github.com/mckinsey/vizro/pull/216/commits/962e4ea98a12537ed5d7ad7d009d749248d5338f) should work, but for some reason it doesn't even try to install kedro on our CI for Python 3.12. Does it work for you locally with that?

huong-li-nguyen: Locally I get a different error now than in CI, but it's still triggered by the Kedro warning 😅 

Screenshot 2023-12-15 at 18 03 12

antonymilne: Ah ok, this is a step forward! If you add the same `ignore` to `filterwarnings` in the `pytest.ini_options` does the test pass?

antonymilne (approved):
Approving subject to some resolution on the open conversation, even if that resolution is that we should give up 😀

vizro-core/hatch.toml
huong-li-nguyen: Ah yes!! Now it does 🥳

antonymilne: Nice! 🎉 

At a glance I have no idea why CI isn't installing kedro. Maybe something wrong with our hatch.toml environment configuration but no idea what. Next things I would try would be:
* remove hatch environments locally and run the same steps to see if you can reproduce the CI behaviuor
* make the `hatch pip freeze` step into `hatch -v pip freeze` to see the verbose output in CI
* login to the CI machine and try to run it on there since debugging through just repeatedly pushing to GHA to trigger the job is slow and tedious

vizro-ai/changelog.d/20231214_163129_huong_li_nguyen_add_python_312.md
maxschulz-COL, line 1: I would add this to the changelog

vizro-core/changelog.d/20231214_163119_huong_li_nguyen_add_python_312.md
maxschulz-COL, line 1: same here

maxschulz-COL (approved):
Lgtm, but I didn't follow the thread in detail. Would update the changelog though

vizro-core/hatch.toml
huong-li-nguyen: I was able to reproduce the CI behaviour locally but the verbose output didn't add any new insights. After trying for way too long, I've decided to revert the changes and not run the tests for Python 3.12 😞 

So we have it as before, but at least the kedro tests are additionally running for 3.11 now.

vizro-ai/changelog.d/20231214_163129_huong_li_nguyen_add_python_312.md
antonymilne: I would definitely not add this to the changelog because it's not actually a user facing change at all, just adding something to our tests.

vizro-core/changelog.d/20231214_163119_huong_li_nguyen_add_python_312.md
antonymilne: Same here from me too 😀

vizro-ai/changelog.d/20231214_163129_huong_li_nguyen_add_python_312.md
huong-li-nguyen: I thought about that, and was a bit indifferent. I see Max point as I did check whether the code was compatible with Python 3.12. In the end no actual code changes were made, but just a few warnings had to be ignored, so I am also fine with not adding it in.

vizro-core/pyproject.toml
```diff
@@ -71,6 +72,11 @@ addopts = [
   # Allow test files to have the same name in different directories.
   "--import-mode=importlib"
 ]
-filterwarnings = "error"
+filterwarnings = [
+  "error",
+  # Ignore this warning here as it comes from pandas until pandas is made compatible with Python 3.12
+  "ignore:.*utcfromtimestamp:DeprecationWarning",
+  "ignore:Kedro is not yet fully compatible with this Python version"
```
antonymilne, line 79: We might as well remove this given that we don't trigger the warning any more.
```suggestion
```

vizro-core/hatch.toml
antonymilne: This is fine by me, thank you for trying it out! Just to understand for future reference though:
1. how did you reproduce the CI behaviour locally? Did just removing the hatch environment and rerunning do the same thing?
2. please could you put a brief comment in the `hatch.toml` file about why kedro is excluded here with a link to this PR?

vizro-ai/changelog.d/20231214_163129_huong_li_nguyen_add_python_312.md
antonymilne: I think it's misleading to suggest that this version somehow adds support for Python 3.12, because there aren't actually any changes to the package at all. It's always worked fine with Python 3.12, just we weren't testing it before. So it's really just an internal testing detail, not something user-facing. A user shouldn't try and choose between version 0.1.7 and 0.1.8 on the basis of this change.

vizro-core/hatch.toml
huong-li-nguyen: 1) I run these commands in order - I don't think the conda environment creation was actually necessary, just did it to run the same pip install command as in CI:
```
hatch env prune
conda create -n hatch_test python=3.12
conda activate hatch_test       
pip install --upgrade hatch     
hatch run all.py3.12:pip freeze   
hatch run all.py3.12:test-unit-coverage
```

-> And then I got the same CI error that Kedro was not installed

2) Yes of course!

vizro-ai/changelog.d/20231214_163129_huong_li_nguyen_add_python_312.md
huong-li-nguyen: Yeah, that makes sense to me 👍 . Will remove it then

maxschulz-COL: Ok no you are right! In my head I was actually thinking that we are somewhat adding support, but I guess we just checked if it works. Let's not then! Apologies

PR #225, opened 2023-12-18
Enabled collapsible panel

nadijagraca:
## Description
Implemented collapsible navigation panel:
- added implicitly without possibility to turn off
- callback for opening closing implemented as clientside callback
- used Google material icons library for collapsable icon

## Screenshot
Opened
Screenshot 2024-01-17 at 12 54 51

Opened with tooltip
Screenshot 2024-01-17 at 12 56 43

Closed with tooltip
Screenshot 2024-01-17 at 12 57 04

## Checklist

## Types of changes

## Notice


huong-li-nguyen:
With the docs and tests we can wait, but the rotating collapsible icon you could already start implementing

vizro-core/src/vizro/static/js/models/dashboard.js
```diff
@@ -1,3 +1,9 @@
 export function _update_dashboard_theme(on) {
   return on ? "vizro_dark" : "vizro_light";
 }
+
+export function _collapse_nav_panel(n_clicks, is_open) {
+  if (n_clicks) return !is_open;
+
+  return is_open;
+}
```
antonymilne, line 9: Switching the collapse icon between `chevron_right` and `collapse_left` in this clientside callback makes sense to me as a possible solution (I think you mentioned it in standup?) but do you get the same nice animation that you get with [`dbc.Accordion`](https://dash-bootstrap-components.opensource.faculty.ai/docs/components/accordion/) when you click the arrows? I guess not.

huong-li-nguyen:
@Joseph-Perkins gave the approval for the first version to be per dashboard level and expanded by default (no possibility to configure expand/collapsed). So I think we can go forward with this PR @nadijagraca and finish it off this week :)

vizro-core/src/vizro/static/js/models/dashboard.js
huong-li-nguyen: I think we should be able to achieve that via CSS transitions: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_transitions/Using_CSS_transitions

@nadijagraca, if you can check that out - that would be great!

huong-li-nguyen (commented):
This already works and looks amazing! Great job 🔥 

I am nit picking right now when it comes to CSS, but could we:
- Reposition the icon such that the right-border of the left-side container goes exactly through the middle of the icon (supposedly between the two arrows)
- Change the cursor icon: https://developer.mozilla.org/en-US/docs/Web/CSS/cursor?retiredLocale=de - somehow it's showing this weird line which doesn't make much sense. It should either show the normal cursor or nothing.
- Change the position of the tooltip to be on the same line/height as the icon
- Can you quickly show how it looks like without and with this transparent background for the icon? Not sure if it looks with the transparency or not

vizro-core/docs/pages/user_guides/navigation.md
```diff
@@ -3,6 +3,7 @@
 This guide shows you how to use and customize the navigation that appears on the left of your dashboard.
 
 The [`Dashboard`][vizro.models.Dashboard] model accepts a `navigation` argument, where you can enter a [`Navigation`][vizro.models.Navigation] model. This enables you to group pages together and customize how they appear in your navigation.
+Navigation panel is designed with a collapsible feature, allowing users to minimize or expand it as needed. This functionality is achieved by clicking on the dedicated collapse button located in the top right corner of navigation panel.
```
huong-li-nguyen, line 6:
```suggestion
The dashboard includes a collapsible side panel that users can easily minimize or expand by a button click. The collapse button, located in the top right corner of the side panel, is visible by default for user convenience.
```

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -159,6 +165,28 @@ def _get_page_divs(self, page: Page) -> _PageDivsType:
...
+                            id="collapse_icon",
+                        )
+                    ],
+                    position="right-end",
+                    arrowOffset=10,
+                    style={"marginTop": "12px", "marginLeft": "8px"},
+                )
+            ],
+            className="collapsable-div",
+        )
```
huong-li-nguyen, line 188: I would suggest the following changes - some will make more sense when we add the relevant CSS:

- Remove the outer div, as the outer div is not necessary. Instead we should leverage [absolute/relative positioning](https://www.geeksforgeeks.org/how-to-set-position-absolute-but-relative-to-parent-in-css/) and the parent element should be the `left-side`
- Remove the brackets around the html.Span 
- Change position to right
- Remove inline-style (To achieve what you wanted to achieve, we don't need the inline style, but instead need to adapt the height of the target element. The dmc.Tooltip will by default be positioned correctly then) -> I'll explain later in the custom css changes
- Add `className="collapse-button-tooltip"` (we need this later to update the height or the target element) - optimally I would like to do that only via the ID, but somehow this doesn't work with the mantine component. So had to add the className here. But maybe try and see if we could also do this only via the `id=collapse_tooltip` instead of adding a className here

```suggestion
        collapsable_icon = dmc.Tooltip(
            html.Span("keyboard_double_arrow_right", className="material-symbols-outlined", id="collapse_icon"),
            id="collapse_tooltip",
            label="Hide Menu",
            offset=24,
            withArrow=True,
            position="right",
            arrowOffset=10,
            className="collapse-button-tooltip",
        )
```

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -175,7 +203,14 @@ def _arrange_page_divs(self, page_divs: _PageDivsType):
...
 
-        return html.Div([left_side, right_side], id="page-container")
+        collapsable_left_side = dbc.Collapse(
+            children=[left_side],
+            id="collapse",
+            is_open=True,
+            dimension="width",
+            navbar=True,
+        )
+        return html.Div([collapsable_left_side, collapsable_icon, right_side], id="page-container")
```
huong-li-nguyen, line 213: If we could get rid of the dbc.Collapse and instead do it directly via CSS/JS, I think that would be nice given the boostrap theme problem we currently have. If it takes too much time or the same functionality cannot be achieved, then don't worry about it.

I would also add these changes:
`left_side = html.Div([left_sidebar, left_main, collapsable_icon], id="left-side")`

and

`return html.Div([left_side, right_side], id="page-container")`

vizro-core/src/vizro/static/css/collapse.css
```diff
@@ -0,0 +1,33 @@
...
+.collapsing.collapse-horizontal {
+  width: 0;
+  height: auto;
+  transition: width 0.35s ease;
+}
+@media (prefers-reduced-motion: reduce) {
+  .collapsing.collapse-horizontal {
+    transition: none;
+  }
+}
```
huong-li-nguyen, line 33: If we get rid of the dbc.collapse, then double-check what of this CSS we then actually need

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -175,3 +176,24 @@ div.dashboard_container .custom-tooltip {
...
   display: none;
 }
+
+.collapsable-div {
+  background-color: var(--surfaces-bg-02);
+  height: 100%;
+  padding-top: 44px;
+  width: 12px;
+  padding-right: 12px;
+}
```
huong-li-nguyen, line 186: We can remove this after we've removed the outer div

```suggestion
```

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -175,3 +176,24 @@ div.dashboard_container .custom-tooltip {
...
+
+#collapse_icon.material-symbols-outlined {
+  z-index: 1;
+  position: absolute;
```
huong-li-nguyen, line 190: This also shouldn't be required anymore

```suggestion
```

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -175,3 +176,24 @@ div.dashboard_container .custom-tooltip {
...
+  background-color: var(--surfaces-bg-card);
```
huong-li-nguyen, line 191: I think we've discussed w/ Joe that we would update the background-color and color here?

The background-color is currently not optimal, because on dark / light it looks different e.g. in the dark theme it has a different color to the side panel, while on the light theme it looks like it has exactly the same color?

I think for color we wanted to use the disabled color instead of text-secondary

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -175,3 +176,24 @@ div.dashboard_container .custom-tooltip {
...
+  border-radius: 50%;
+  cursor: pointer;
+  color: var(--text-secondary);
+}
+
+#collapse_icon.material-symbols-outlined:hover {
+  color: var(--text-active);
+}
```
huong-li-nguyen, line 199: This is the change you would need to position the tooltip correctly without needing the inline style and an outer div

```suggestion
}

.collapse-button-tooltip {
  position: relative;
  top: 44px;
  left: -12px;
  height: 24px;
  width: 0 !important;
}
```

huong-li-nguyen (changes requested):
Hey @nadijagraca - visually it already looks good! 

Have some CSS improvement suggestions. I would recommend reading about this: https://www.geeksforgeeks.org/how-to-set-position-absolute-but-relative-to-parent-in-css/

This is the foundation of most of my proposals here :) 
It would also be great if we could already use the dash notation instead of underscore notation in the CSS classes and IDs

vizro-core/src/vizro/models/_dashboard.py
nadijagraca: All suggestions are included, but the last one as targeting by ID didn't work.

vizro-core/src/vizro/static/css/layout.css
nadijagraca: z-index is removed, but I need position property for positioning the icon when navigation panel is collapsed.

vizro-core/src/vizro/models/_dashboard.py
nadijagraca: As agreed, first version will be merged with dbc.Collapse component, and I will create ticket for it to be removed in the future.

vizro-core/changelog.d/20240112_163629_nadija_ratkusic_graca_enable_collapsible_panel.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Implemented the functionality allowing users to collapse or expand the side panel with a button click. ([#225](https://github.com/mckinsey/vizro/pull/225))
```
huong-li-nguyen, line 22:
```suggestion
-  Enable the side panel to collapse/expand via a button that is present by default ([#225](https://github.com/mckinsey/vizro/pull/225))
```

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -159,6 +170,21 @@ def _get_page_divs(self, page: Page) -> _PageDivsType:
...
+                className="material-symbols-outlined",
+                id="collapse_icon",
+            ),
+            id="collapse_tooltip",
+            label="Hide Menu",
+            offset=24,
+            withArrow=True,
+            position="right",
+            arrowOffset=10,
+            className="collapse-button-tooltip",
```
huong-li-nguyen, line 185: Let's use dash notation instead of underscore notation for all the newly added CSS IDs and classNames

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -175,7 +201,14 @@ def _arrange_page_divs(self, page_divs: _PageDivsType):
         right_main = page_divs["components"]
         right_side = html.Div([right_header, right_main], id="right-side")
 
-        return html.Div([left_side, right_side], id="page-container")
+        collapsable_left_side = dbc.Collapse(
+            left_side,
+            id="collapse",
```
huong-li-nguyen, line 206:
```suggestion
            id="collapsable-left-side",
```

vizro-core/src/vizro/models/_dashboard.py
huong-li-nguyen: Thanks for checking!

huong-li-nguyen (approved):
Thanks for incorporating all the changes - looks great to me 👍 

Let's already use the dashes instead of underscores for CSS IDs and classNames and could you double-check on some of these screenshots whether the left-side is actually required to display the relevant topic/feature? 

For example for the actions_table_filter_action.png, I actually don't see a reason why we have the left-side there as the focus is the chart interaction on the right-side. If you double-check the ones that you've updated, that would be great! 

Then in the future we don't have to update these anymore if we do changes on the left-side.

huong-li-nguyen:
Could you update the PR description as well? I think some of the screenshots/descriptions are outdated now :)

vizro-core/src/vizro/static/js/models/dashboard.js
```diff
@@ -1,3 +1,35 @@
 export function _update_dashboard_theme(checked) {
   return checked ? "vizro_light" : "vizro_dark";
 }
+
+export function _collapse_nav_panel(n_clicks, is_open) {
+  if (!n_clicks)
```
antonymilne, line 6: Is this the case that is relevant when you change page? Why does it need a separate `if` from the below one?

antonymilne (approved):
This looks amazing, really nice work 💯 

Now I've played around with it a bit more I think it does look a bit weird that the arrow spins around every time you change page. Not worth fixing right now unless it's easy, but I think we should put it on the backlog to fix that. 

I also observed a couple of small glitches trying it out. Most of the time everything works beautifully but when I click on the expand/collapse many times then about 10% of the time it ends up with an extra vertical line like here:
image
image

Not a big problem at all, so I wouldn't worry about it unless it's easy to reproduce and fix.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -175,7 +201,14 @@ def _arrange_page_divs(self, page_divs: _PageDivsType):
         right_main = page_divs["components"]
         right_side = html.Div([right_header, right_main], id="right-side")
 
-        return html.Div([left_side, right_side], id="page-container")
+        collapsable_left_side = dbc.Collapse(
+            left_side,
+            id="collapsable-left-side",
+            is_open=True,
+            dimension="width",
+            navbar=True,
```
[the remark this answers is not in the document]
nadijagraca: `navbar=True` is used inside a navbar, but I tried without and it does not make any difference so I removed it.

vizro-core/src/vizro/static/js/models/dashboard.js
nadijagraca: This was relevant when you change page, but I changed the icons and rotation which makes it redundant now. Good catch, I removed it now.

PR #185, opened 2023-11-28
Updating Meta Tags

AnnMarieW:
## Description
Closes #176

## Checklist

- [x] I have not referenced individuals, products or companies in any commits, directly or indirectly
- [x] I have not added data or restricted code in any commits, directly or indirectly
- [x] I have updated the docstring of any public function/class/model changed
- [ ] I have added tests to cover my changes (if applicable)

## Types of changes

- [ ] Docs/refactoring (non-breaking change which improves codebase)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

  - I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
  - I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
  - I certify that the use of this contribution as authorized by the Apache 2.0 license does not violate the intellectual property rights of anyone else.

PR #228, opened 2023-12-19
Improve dashboard title

antonymilne:
## Description

Small precursor to #185, mentioned as part of #176.

When `Dashboard.title` is specified we now include it in the page title in the dash registry. This makes no difference to anything on the screen other than the `<title>` tag and a meta tag. These are useful to get right because it means that the title of the page is much more meaningful in your browser window, when you try to bookmark a dashboard, and if you share a link of a Vizro dashboard that gets previewed e.g. on social media.

## Screenshot

image

## Checklist

## Types of changes

## Notice

### DIFF
--- a/vizro-core/changelog.d/20231218_204302_antony.milne_improve_dashboard_title.md
+++ b/vizro-core/changelog.d/20231218_204302_antony.milne_improve_dashboard_title.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Security
+
+- A bullet item for the Security category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
--- a/vizro-core/changelog.d/20231218_204302_antony.milne_improve_dashboard_title.md
+++ b/vizro-core/changelog.d/20231218_204302_antony.milne_improve_dashboard_title.md
@@ -16,16 +16,15 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 - A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
 
 -->
-<!--
+
 ### Added
 
-- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+- When set, the dashboard title appears alongside the individual page title as the text labelling a browser tab. ([#228](https://github.com/mckinsey/vizro/pull/228))
 
--->
 <!--
 ### Changed
 
-- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#228](https://github.com/mckinsey/vizro/pull/228))
 
 -->
 <!--
--- a/vizro-core/changelog.d/20231218_204302_antony.milne_improve_dashboard_title.md
+++ b/vizro-core/changelog.d/20231218_204302_antony.milne_improve_dashboard_title.md
@@ -19,7 +19,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 ### Added
 
-- When set, the dashboard title appears alongside the individual page title as the text labelling a browser tab. ([#228](https://github.com/mckinsey/vizro/pull/228))
+- When set, the dashboard title appears alongside the individual page title as the text labeling a browser tab. ([#228](https://github.com/mckinsey/vizro/pull/228))
 
 <!--
 ### Changed
--- a/vizro-core/examples/default/app.py
+++ b/vizro-core/examples/default/app.py
@@ -545,7 +545,6 @@ def create_home_page():
 
 
 dashboard = vm.Dashboard(
-    title="My dashboard",
     pages=[
         create_home_page(),
         create_variable_analysis(),
--- a/vizro-core/examples/default/app.py
+++ b/vizro-core/examples/default/app.py
@@ -545,6 +545,7 @@ def create_home_page():
 
 
 dashboard = vm.Dashboard(
+    title="My dashboard",
     pages=[
         create_home_page(),
         create_variable_analysis(),
--- a/vizro-core/examples/default/app.py
+++ b/vizro-core/examples/default/app.py
@@ -545,6 +545,7 @@ def create_home_page():
 
 
 dashboard = vm.Dashboard(
+    title="Vizro demo",
     pages=[
         create_home_page(),
         create_variable_analysis(),
--- a/vizro-core/examples/default/app.py
+++ b/vizro-core/examples/default/app.py
@@ -545,6 +545,7 @@ def create_home_page():
 
 
 dashboard = vm.Dashboard(
+    title="Blah blah blah",
     pages=[
         create_home_page(),
         create_variable_analysis(),
--- a/vizro-core/examples/default/app.py
+++ b/vizro-core/examples/default/app.py
@@ -545,7 +545,6 @@ def create_home_page():
 
 
 dashboard = vm.Dashboard(
-    title="Blah blah blah",
     pages=[
         create_home_page(),
         create_variable_analysis(),
--- a/vizro-core/examples/default/app.py
+++ b/vizro-core/examples/default/app.py
@@ -545,7 +545,6 @@ def create_home_page():
 
 
 dashboard = vm.Dashboard(
-    title="Vizro demo",
     pages=[
         create_home_page(),
         create_variable_analysis(),
--- a/vizro-core/src/vizro/models/_dashboard.py
+++ b/vizro-core/src/vizro/models/_dashboard.py
@@ -67,9 +67,13 @@ class Dashboard(VizroBaseModel):
         # For now the homepage (path /) corresponds to self.pages[0].
         # Note redirect_from=["/"] doesn't work and so the / route must be defined separately.
         for order, page in enumerate(self.pages):
-            path = page.path if order else "/"
             dash.register_page(
-                module=page.id, name=page.title, path=path, order=order, layout=partial(self._make_page_layout, page)
+                module=page.id,
+                name=page.title,
+                title=f"{self.title}: {page.title}" if self.title else page.title,
+                path=page.path if order else "/",
+                order=order,
+                layout=partial(self._make_page_layout, page),
             )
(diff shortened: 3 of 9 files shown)


vizro-core/changelog.d/20231218_204302_antony.milne_improve_dashboard_title.md
```diff
@@ -0,0 +1,47 @@
...
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- When set, the dashboard title appears alongside the individual page title as the text labelling a browser tab. ([#228](https://github.com/mckinsey/vizro/pull/228))
+
```
antonymilne, line 23: This is meant to be a "non-expert user understandable" way of saying `<title>` and is inspired by the description of Dash's `title` argument.

vizro-core/examples/default/app.py
```diff
@@ -545,6 +545,7 @@ def create_home_page():
 
 
 dashboard = vm.Dashboard(
+    title="My dashboard",
```
antonymilne, line 548:
```suggestion
```

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -67,9 +67,13 @@ def pre_build(self):
         # For now the homepage (path /) corresponds to self.pages[0].
         # Note redirect_from=["/"] doesn't work and so the / route must be defined separately.
         for order, page in enumerate(self.pages):
-            path = page.path if order else "/"
             dash.register_page(
-                module=page.id, name=page.title, path=path, order=order, layout=partial(self._make_page_layout, page)
+                module=page.id,
+                name=page.title,
+                title=f"{self.title}: {page.title}" if self.title else page.title,
```
antonymilne, line 73: This one line is the only functional change in the PR.

vizro-core/tests/unit/vizro/conftest.py
```diff
@@ -60,11 +60,3 @@ def vizro_app():
     app instantiation.pages.
     """
     return Vizro()
-
```
antonymilne, line 63: I've wanted to move this to the navigation conftest for a while, and now that I've removed all its other uses from test_dashboard.py it has been set free.

vizro-core/tests/unit/vizro/models/test_dashboard.py
```diff
@@ -149,27 +66,90 @@ def test_field_invalid_theme_input_type(self, page_1):
 class TestDashboardPreBuild:
     """Tests dashboard pre_build method."""
 
-    def test_dashboard_page_registry(self, prebuilt_two_page_dashboard, mock_page_registry):
-        result = dash.page_registry
-        expected = mock_page_registry
-        # Str conversion required as comparison of OrderedDict values result in False otherwise
-        assert str(result.items()) == str(expected.items())
```
antonymilne, line 156: This was not a good way of doing the comparison before: actually comparing two dictionaries works fine and does not need `str`. The actual problem was the comparison of the `layout` functions, which is very tricky to get right, hence why I ignore them here.

vizro-core/tests/unit/vizro/models/test_dashboard.py
```diff
@@ -149,27 +66,90 @@ def test_field_invalid_theme_input_type(self, page_1):
...
-        mocker.patch("vizro.models._dashboard.get_relative_path")
-        result = prebuilt_two_page_dashboard._make_page_404_layout()
-        result_image = result.children[0]
-        result_div = result.children[1]
-
-        assert isinstance(result, html.Div)
-        assert isinstance(result_image, html.Img)
-        assert isinstance(result_div, html.Div)
+    def test_page_registry(self, vizro_app, page_1, page_2, mocker):
+        mock_register_page = mocker.patch("dash.register_page", autospec=True)
```
antonymilne, line 70: Generally we should avoid mocking much, but this is a good case where it really helps the tests become more focussed.

antonymilne:
@AnnMarieW: I wanted to add you as a reviewer for this but for some reason can't do so - maybe some limitation of our github org settings I guess 🤔 

Are you able to add yourself as reviewer?

There's two reasons I wanted to add you anyway:
1. so you can check I haven't done anything stupid
2. I was thinking about the tests in your PR and decided our current tests weren't quite right so rewrote them

I think we should aim to get this merged ASAP before #185 so you can base your tests on the "new" style.

AnnMarieW:
@antonymilne I might have to be invited as a collaborator or a member before I can be added as a reviewer, but I think tagging me here works too :-) 

I like the new style of the tests and will wait until this is merged before writing the tests in my PR.  Thanks for the help!

I also like the new format for the title. Are you planning on updating the docs in a different PR?

vizro-core/changelog.d/20231218_204302_antony.milne_improve_dashboard_title.md
huong-li-nguyen: Just wondering whether we should align the naming - I think we refer to it as dashboard title while Dash refers to it as app title. Dashboard title makes more sense to me as it's an argument in the dashboard model, but under the hood we seem to overwrite the app title

This might also be just worth mentioning in the docs for people coming from Dash

vizro-core/examples/default/app.py
```diff
@@ -545,6 +545,7 @@ def create_home_page():
 
 
 dashboard = vm.Dashboard(
+    title="Vizro demo",
```
huong-li-nguyen, line 548: You probably just left this in for demo purposes, but I would remove it if this is close to being merged. It's not optimal in terms of design if there are controls/navigation on the page, as we still have this open discussion of where this might go. 

Looking at this, I prefer the upper header container we discussed the other day even more 😄

vizro-core/src/vizro/models/_dashboard.py
huong-li-nguyen: I guess if people do want to overwrite it, they would overwrite the pre-build of the model here?

I was actually wondering how this works compared to the `dash.app.title`. If you set the `dash.app.title`, I noticed that it appears for a second and then it gets overwritten by anything defined in `dash.register_page`.

This might be worth mentioning in the docs if people try to overwrite it via the `dash.app.title`. This wouldn't actually work, because in the end it gets overwritten by the code here again. But I think this is also how it works in a pure Dash app?

```
app = Vizro(assets_folder="../assets").build(dashboard)
app.title = "My title"
app.run()
```

vizro-core/tests/unit/vizro/models/test_dashboard.py
huong-li-nguyen: Yeees, the layout was the problem. Even better if we are fine ignoring it 👍

vizro-core/tests/unit/vizro/models/test_dashboard.py
```diff
@@ -149,27 +66,90 @@ def test_field_invalid_theme_input_type(self, page_1):
...
+        vm.Dashboard(pages=[page_1], title="My dashboard").pre_build()
+
+        mock_register_page.assert_any_call(
+            module=page_1.id,
+            name="Page 1",
+            title="My dashboard: Page 1",
+            path="/",
+            order=0,
+            layout=mocker.ANY,  # partial call is tricky to mock out so we ignore it.
+        )
```
huong-li-nguyen, line 109: I really like these new tests - looks much cleaner! 🚀

huong-li-nguyen (approved):
Looks great to me 👍 

If we can just mention in the docs how the `dash.app.title` and `dashboard.title/page.title ` overwriting works - I think that would help. I wasn't e.g. sure how it would actually work until I tested it out and saw that the `dash.app.title` is always overwritten by the title defined in the `dash.page_registry`. I guess that's the default behaviour coming from dash?

vizro-core/examples/default/app.py
antonymilne: Actually I left it here so that we would hopefully apply the same change to vizro.mckinsey.com next time we update the demo, but let me remove it if we're not yet happy about the visual design.

The reason I think it would be nice to include sooner rather than later is because every time someone shares a link to vizro.mckinsey.com the social media preview will have a slightly strange label (like just "Homepage") - it would be nicer if it instead said "Vizro demo: Homepage" or similar, to make it clear what the link being shared actually is.

If we don't work out the page layout/visual design soon then possible alternatives might be:
* we remove the title from the dashboard layout altogether from `_make_page_layout` for now
* we remove the title from the dashboard layout just by customising our demo app

So I'll revert the change here and leave it up to you to decide out what's the best solution here depending on how long the reworking of the page layout takes 🙂 It's not a huge thing but I don't think we should wait too long (say > 1 month) to get an improved shareable link to our demo app just because we're blocked on where to put the title on screen.

vizro-core/src/vizro/models/_dashboard.py
antonymilne: This is indeed how it works in a pure Dash app also:
1. `Dash(title="X")` corresponds to the property `dash.app.title` and sets the initial value of `<title>X</title>`
2. but this only appears very briefly when you're using Dash pages, because it gets [overwritten straight away](https://github.com/plotly/dash/blob/7527383050b6fe59a11833bbeac6f1a741436084/dash/dash.py#L2159) by  `dash.page_registry[page.id]["title"]`
3. after this, the original app `title=X` has no effect on anything

So for Vizro users where Dash pages is always used, the `title` argument basically has no effect. @AnnMarieW please do correct me if I've missed something here.

We have possibly added to the confusion here because we have our own `Dashboard.title` property, and it's not obvious to a user what effect that has on anything. I considered before whether we should do `dash.app.title = dashboard.title` in `Vizro.build()` but decided it was basically pointless because, like you say, it just gets overwritten straight away. I think it would be a very small improvement though, since that way you wouldn't see the "Dash" title at all, so let me add it in here.

vizro-core/changelog.d/20231218_204302_antony.milne_improve_dashboard_title.md
antonymilne: Yeah, I think "dashboard title" is best here because currently the only way a user might set their title is through `Dashboard.title` since we don't advertise the `Vizro(title=...)` method anywhere.

So I will leave this alone but add something to the docs that clarifies.

vizro-core/examples/default/app.py
huong-li-nguyen: **we remove the title from the dashboard layout altogether from _make_page_layout for now**
I think this would be a breaking change right? Or are we still fine saying that any visual changes are not considered breaking changes? 😄 I would be fine with this one to be honest - if people reach out, then we at least know people have been using a dashboard title and need it 😅 

**we remove the title from the dashboard layout just by customising our demo app**
As a quick-fix, this one seems safer. The demo app would be a bit out of sync for a while with the example app here, but I think it's fine as we manage it.

antonymilne: Reverted in [5cef34a](https://github.com/mckinsey/vizro/pull/228/commits/5cef34afdef26e2baf0a5ee4a4302beeb61fe5ea).

vizro-core/src/vizro/models/_dashboard.py
antonymilne: Change made in [c00fe4e](https://github.com/mckinsey/vizro/pull/228/commits/c00fe4ed0b58ab9b13645ab4aeb8484b08dde524).

antonymilne:
> @antonymilne I might have to be invited as a collaborator or a member before I can be added as a reviewer, but I think tagging me here works too :-)
> 
> I like the new style of the tests and will wait until this is merged before writing the tests in my PR. Thanks for the help!
> 
> I also like the new format for the title for the browser and the meta tags. Are you planning on updating the docs in a different PR?

I've just updated the docs in [7f981fb](https://github.com/mckinsey/vizro/pull/228/commits/7f981fb1c327ef11f9dfd295a61950124f184750).

When we write docs for your meta tag image and description pieces, I think it would belong in pages.md and then let's link to it from my text that mentions meta elements here in dashboard.md.

I've changed a github setting that should mean that all users can now review PRs, but I still don't seem to be able to add you here. Maybe it will work in future PRs though, let's see.

vizro-core/examples/default/app.py
antonymilne: I wouldn't consider removing the title from the layout a breaking change since it's just something visual and no one's code will break in any way. Admittedly it's kind of a drastic one rather than just tweaking styling, but I'm completely fine with it and, like you say, if people ask where their title went then it's a good signal they want it there.

Which solution we go for depends to me on how close we are to resolving the question of where the title goes and how upsetting people find the current placement:
1. if we're really not happy with how the dashboard title appears at the moment and it feels a long way off resolving let's go for option 1
2. if we think it's ok currently but still don't want it on our demo app then let's go for 2
3. if we think the current dashboard title is already completely fine then let's just leave it as it is and add a title to the example app

Personally I'm completely fine with the current visual appearance, which is why I just went for option 3 to begin with in this PR. But you have a better idea of how upsetting people with more design sensibilities find that and where future changes to layout might take us here, so completely up to you which option to choose 🙂

vizro-core/examples/default/app.py
```diff
@@ -545,6 +545,7 @@ def create_home_page():
 
 
 dashboard = vm.Dashboard(
+    title="Blah blah blah",
```
antonymilne, line 548:
```suggestion
```

vizro-core/examples/default/app.py
antonymilne: Cool, thank you! Sounds good to me.

I didn't actually realise how weird it looks like when there's no controls/navigation until I just looked again now. You're right that doesn't look good at all, so I take back my statement above that option 3 was ok by me. Just in case you thought I had no design sensibilities at all 😀

antonymilne: P.S. hopefully the demo update would be as simple as `dash.layout["dashboard_title_outer"].hidden = True` or similar rather than needing any customisation of `_make_page_layout` with a custom `Dashboard` model.

Issue #187, opened 2023-11-30
Use vizro.models.button to switch to another page


Tom-Sachs:
### Question

Hi there! 

As far as I know, the only way to create a clicktable container/button to switch to another Page is to use a Card, and use the `href` parameter.

I was wondering if this was possible to do the same thing using Button instead of Card ? The Button design seems a bit more user-friendly. At the moment, Button needs to take an action as an input, but the only two actions available are filter actions, or data exports, so I haven't found a way to do this.

Thank you!

### Code of Conduct

- [X] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

petar-qb:
Hi @Tom-Sachs, great question! Currently, we are working on enabling this feature natively through the `vm.Button` component. However, this behaviour is already possible to construct utilising the [custom components](https://vizro.readthedocs.io/en/stable/pages/user_guides/custom_components/#extend-an-existing-component) and here is an example on how to achieve it:

```
# Inherit the Vizro Button model in purpose to slightly modify its behaviour
class HrefButton(vm.Button):
    href: str = "/"

    def build(self):
        # Calling vm.Button build method that returns the html.Div object:
        # html.Div(
        #     [
        #         dbc.Button(
        #             id=self.id,
        #             children=self.text,
        #             className="button_primary",
        #         ),
        #     ],
        #     className="button_container",
        #     id=f"{self.id}_outer",
        # )
        button_build_obj = super().build()
        # button_build_obj[self.id] fetches the underlying dbc.Button object
        button_build_obj[self.id].href = self.href
        # Modified object with the "href" attributed is returned
        return button_build_obj 

dashboard = vm.Dashboard(
    pages=[
        vm.Page(
            title="Homepage",
            components=[
                HrefButton(
                    text="Go to analysis",
                    href="/analysis",
                ),
            ]
        ),
        vm.Page(
            title="Analysis",
            components=[
                vm.Card(text="Analysis")
            ]
        )
    ]
)

Vizro(assets_folder="../assets").build(dashboard).run()
```

---

Instead of calling `super().build()` inside the HrefButton `build` method, there is an option to completely copy-paste `vm.Button` `build` method implementation and then just adjust it by adding the `href` attribute as a `dbc.Button` argument.

```

class HrefButton(vm.Button):
    href: str = "/"

    def build(self):
        return html.Div(
            [
                dbc.Button(
                    id=self.id,
[...]

PR #229, opened 2023-12-19
Improvement of custom charts docs

maxschulz-COL:
## Description
- short PR to improve docs on custom charts - increased emphasis that they can be used with `FIlter` and `Parameter`
- especially the emphasis on Parameter hopefully guides people towards the great flexibility that this offers

Now added:
- change the docs on custom components to have a shorter example - they now make use of `super` instead of copying the entire build method

## Notice


petar-qb (approved):
Thanks for addressing this so immediately. 🚀 The flexibility of the custom charts feature  is much cleaner now.

---

Do you think we should change the order of custom charts with Filters and Parameters and list them in the order:
 - custom chart with Filter and then
 - custom chart with Parameter 
 
---

Also, do you think we should add following comments below `def scatter_with_line(...)` and `def waterfall(...)`:
 - \# "data_frame" is filtered.
 - \# "hline" is parametrised.
or something like that?

vizro-core/docs/pages/user_guides/custom_charts.md
```diff
@@ -65,13 +69,18 @@ Building on the above, there are several routes one can take. The following exam
...
+                    selector=vm.Slider(min=2, max=5, step=1, value=3, title="Horizontal line"),
+                ),
             ],
         )
         dashboard = vm.Dashboard(pages=[page_0])
 
         Vizro().build(dashboard).run()
         ```
+
+        1.  Note how we are able to target *any custom argument* of our custom chart. Since there is complete flexibility in terms of what can be derived from such custom arguments this allows for a wide range of customization options by the dashboard user.
```
huong-li-nguyen, line 83: Are there custom arguments? or are they just arguments?

```suggestion
        1.  Note how we can target *any argument* of our custom chart. Since there is complete flexibility regarding what can be derived from such arguments, the dashboard user has a wide range of customization options.
```

vizro-core/docs/pages/user_guides/custom_charts.md
```diff
@@ -29,11 +29,15 @@ def minimal_example(data_frame:pd.DataFrame=None):
     return go.Figure()
 ```
 
-Building on the above, there are several routes one can take. The following examples are guides on the most common custom requests, but also serve as an illustration of more general principles:
+Building on the above, there are several routes one can take. The following examples are guides on the most common custom requests, but also serve as an illustration of more general principles.
+
+!!! tip
+
+    Custom charts can be targeted by [Filters](filters.md) or [Parameters](parameters.md). We will showcase both possibilities in the following examples. In particular the `Parameters` in combination with custom charts can be extremely versatile in achieving custom functionality.
```
huong-li-nguyen, line 36:
```suggestion
    Custom charts can be targeted by [Filters](filters.md) or [Parameters](parameters.md) without any additional configuration. We will showcase both possibilities in the following examples. In particular the `Parameters` in combination with custom charts can be highly versatile in achieving custom functionality.
```

huong-li-nguyen (commented):
LGTM 👍 

I have one favor to ask given that this fits into this PR quite well - there is an item on our list to update the docs for the custom components as well. @antonymilne can give you more details, but in general it's also just updating the example and adding 1-2 sentences. 

In the current docs, the changes for the custom slider look quite complex, while it can actually be achieved quicker and easier - see @petar-qb answer on the user question here: https://github.com/mckinsey/vizro/issues/187

maxschulz-COL:
> Thanks for addressing this so immediately. 🚀 The flexibility of the custom charts feature is much cleaner now.
> 
> Do you think we should change the order of custom charts with Filters and Parameters and list them in the order:
> 
> * custom chart with Filter and then
> * custom chart with Parameter
> 
> Also, do you think we should add following comments below `def scatter_with_line(...)` and `def waterfall(...)`:
> 
> * # "data_frame" is filtered.
> * # "hline" is parametrised.
>   or something like that?

I implemented the above suggestions as clickable code annotations. Also improved the title of the examples, but I did not swap the order. Actually I prefer having the parameter on top, as its more important to get across.

vizro-core/docs/pages/user_guides/custom_charts.md
```diff
@@ -65,13 +70,19 @@ Building on the above, there are several routes one can take. The following exam
...
+                ),
             ],
         )
         dashboard = vm.Dashboard(pages=[page_0])
 
         Vizro().build(dashboard).run()
         ```
+
+        1.  Note that arguments of the custom chart can be parametrized. Here we choose to parametrize the `hline` argument (see below).
+        2.  Here we parametrize the `hline` argument, but any other argument can be parametrized as well. Since there is complete flexibility regarding what can be derived from such arguments, the dashboard user has a wide range of customization options.
```
huong-li-nguyen, line 85: Seems like repetition looking at the second point? Feel free to ignore though if you think this is more helpful
```suggestion
        1.  Note that arguments of the custom chart can be parametrized.
        2.  Here we parametrize the `hline` argument, but any other argument can be parametrized as well. Since there is complete flexibility regarding what can be derived from such arguments, the dashboard user has a wide range of customization options.
```

vizro-core/docs/pages/user_guides/custom_charts.md
```diff
@@ -136,13 +147,15 @@ The below examples shows a more involved use-case. We create and style a waterfa
...
-                vm.Filter(column="x", selector=vm.Dropdown(title="Financial categories", multi=True)),
+                vm.Filter(column="x", selector=vm.Dropdown(title="Financial categories", multi=True)),# (1)!
             ],
         )
         dashboard = vm.Dashboard(pages=[page_0])
 
         Vizro().build(dashboard).run()
         ```
+
+        1.  Note how we are able to apply a filter even to a custom chart
```
huong-li-nguyen, line 158: I would expect this as a basic functionality - using "even" kind of sounds like it's special? haha

```suggestion
        1.  Note how we are able to apply a filter to a custom chart
```

vizro-core/docs/pages/user_guides/custom_components.md
```diff
@@ -214,12 +127,13 @@ vm.Parameter.add_type("selector", TooltipNonCrossRangeSlider)
...
+        2.  Here we overload the `build` method by altering the output of `super().build()`. Alternatively one could copy the source code of the build method and alter it directly.
+        3.  `range_slider_build_obj[self.id]` then fetches the underlying `dcc.range_slider` object.
+        4.  This is the change that makes the `RangeSlider` not cross itself when moving the handle.
+        5.  This is the change that displays the tooltip below the handle.
+        6.  **Don't forget!** If part of a discriminated union, you must add the new component to the parent model where it will be inserted. In this case the new `TooltipNonCrossRangeSlider` will be inserted into the `selector` argument of the `Filter` model, and thus must be added as an allowed type.
+        7.  **Don't forget!** If part of a discriminated union, you must add the new component to the parent model where it will be inserted. In this case the new `TooltipNonCrossRangeSlider` will be inserted into the `selector` argument of the `Parameter` model, and thus must be added as an allowed type.
+        8.  The new component can now be inserted into a regular dashboard.
```
huong-li-nguyen, line 136: Mostly minor rephrasing suggestions:

```suggestion
        2.  Here, we overload the `build` method by altering the output of `super().build()`. Alternatively, one could copy the source code of the build method and change it directly.
        3.  `range_slider_build_obj[self.id]` then fetches the underlying `dcc.range_slider` object.
        4.  This change prevents the `RangeSlider` from crossing itself when moving the handle.
        5.  This is the change that displays the tooltip below the handle.
        6 **Remember!** If part of a discriminated union, you must add the new component to the parent model where it will be inserted. In this case, the new `TooltipNonCrossRangeSlider` will be inserted into the `selector` argument of the `Filter` model and thus must be added as an allowed type.
        7 **Remember!** If part of a discriminated union, you must add the new component to the parent model where it will be inserted. In this case, the new `TooltipNonCrossRangeSlider` will be inserted into the `selector` argument of the `Parameter` model and thus must be added as an allowed type.
        8.  The new component can now be inserted into a regular dashboard.
```

huong-li-nguyen (approved):
LGTM 👍  Have minor comments on rephrasings - feel free to add it or not :)

vizro-core/docs/pages/user_guides/custom_charts.md
```diff
@@ -65,13 +70,19 @@ Building on the above, there are several routes one can take. The following exam
                 ),
             ],
             controls=[
-                vm.Filter(column="petal_width"),
+                vm.Parameter( # (2)!
```
antonymilne, line 73: This is a really nice example. Should we put in an animated gif just to make it completely obvious what's possible with this example?

Or in future maybe we would have this in our "feature showcase" dashboard and just link to that.

vizro-core/docs/pages/user_guides/custom_components.md
```diff
@@ -35,12 +35,11 @@ at the same time, Vizro is easily extensible, so that you can tweak any componen
...
 
 
-You can extend an existing component by sub-classing the component you want to alter.
-The below example is a bit lengthy, but the annotations should guide you through the most important lines of that code. Remember that when sub-classing a component
+You can extend an existing component by sub-classing the component you want to alter. Remember that when sub-classing a component
 you have access to all fields of all parent models, but you can choose to overwrite any field or method, or define new ones.
 
 The aim for this example is to enhance the [`RangeSlider`][vizro.models.RangeSlider] model so that
-one slider handle cannot cross the other, and to have a permanent tooltip showing the current value. You will note that it is often easier to copy parts of the source-code when needing to change complex methods
+one slider handle cannot cross the other, and to have a permanent tooltip showing the current value. You will note that it is often easier to call `super()` when overloading a complex method
```
antonymilne, line 42:
```suggestion
one slider handle cannot cross the other, and to have a permanent tooltip showing the current value. You will note that it is often easier to call `super()` when overriding a complex method
```

See https://creasecode.com/python/overloading-vs-overriding-methods-functions/

vizro-core/docs/pages/user_guides/custom_components.md
```diff
@@ -214,12 +127,13 @@ vm.Parameter.add_type("selector", TooltipNonCrossRangeSlider)
...
-        3.  This is the change that makes the `RangeSlider` not cross itself when moving the handle.
-        4.  This is the change that displays the tooltip below the handle.
-        5.  **Don't forget!** If part of a discriminated union, you must add the new component to the parent model where it will be inserted. In this case the new `TooltipNonCrossRangeSlider` will be inserted into the `selector` argument of the `Filter` model, and thus must be added as an allowed type.
-        6.  **Don't forget!** If part of a discriminated union, you must add the new component to the parent model where it will be inserted. In this case the new `TooltipNonCrossRangeSlider` will be inserted into the `selector` argument of the `Parameter` model, and thus must be added as an allowed type.
-        7.  The new component can now be inserted into a regular dashboard.
+        2.  Here we overload the `build` method by altering the output of `super().build()`. Alternatively one could copy the source code of the build method and alter it directly.
```
antonymilne, line 130:
```suggestion
        2.  Here we override the `build` method by altering the output of `super().build()`. Alternatively one could copy the source code of the build method and alter it directly.
```

vizro-core/docs/pages/user_guides/custom_components.md
```diff
@@ -214,12 +127,13 @@ vm.Parameter.add_type("selector", TooltipNonCrossRangeSlider)
...
+        3.  `range_slider_build_obj[self.id]` then fetches the underlying `dcc.range_slider` object.
```
antonymilne, line 131:
```suggestion
        3.  `range_slider_build_obj[self.id]` then fetches the underlying [`dcc.RangeSlider`](https://dash.plotly.com/dash-core-components/rangeslider) object.
```

antonymilne (approved):
Thanks for putting the extra custom component change in there too!

Just while I remember, @Coding-with-Adam had some feedback that as a new user it was a little difficult to understand from the current docs the difference between filter and parameter and action. Probably he has lots of other valuable feedback too so we should quiz him in the new year 🙂

vizro-core/docs/pages/user_guides/custom_charts.md
maxschulz-COL: The thing is, since those things are clickable, they do not appear at the same time, so I think the repetition is fine, especially if someone just opens one of them

vizro-core/docs/pages/user_guides/custom_charts.md
maxschulz-COL: Good point. I added a section, and also a GIF. I do think that this actually belongs also to a feature showcase, as we do not use GIFs anywhere else, but this will do for now

Coding-with-Adam:
> Probably he has lots of other valuable feedback too so we should quiz him in the new year 🙂

Happy to help. adam@plot.ly

PR #237, opened 2023-12-22
Bug fix on update nested graph properties

nadijagraca:
## Description
Fixed bug on update nested graph properties. 

Issue: https://github.com/McK-Internal/vizro-internal/issues/270
## Notice

### DIFF
--- a/vizro-core/changelog.d/20231222_112619_nadija_ratkusic_graca_parameter_bug.md
+++ b/vizro-core/changelog.d/20231222_112619_nadija_ratkusic_graca_parameter_bug.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Security
+
+- A bullet item for the Security category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
--- a/vizro-core/changelog.d/20231222_112619_nadija_ratkusic_graca_parameter_bug.md
+++ b/vizro-core/changelog.d/20231222_112619_nadija_ratkusic_graca_parameter_bug.md
@@ -37,8 +37,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 ### Fixed
 
-- Fix bug while updating nested graph properties.  ([#273](https://github.com/mckinsey/vizro/pull/237))
-
+- Fix bug while updating nested graph properties. ([#273](https://github.com/mckinsey/vizro/pull/237))
 
 <!--
 ### Security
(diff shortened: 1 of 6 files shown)


vizro-core/changelog.d/20231222_112619_nadija_ratkusic_graca_parameter_bug.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Fix bug while updating nested graph properties. ([#273](https://github.com/mckinsey/vizro/pull/237))
```
huong-li-nguyen, line 40: Not sure if below correctly describes the issue as this is my current understanding. Please double-check :)

In general these descriptions here should be understandable from a user perspective, so sometimes it helps to be more verbose to explain the issue

```suggestion
- Fix a bug that prevented the proper update of graph properties through parameters when the graph property has not been previously defined. ([#273](https://github.com/mckinsey/vizro/pull/237))
```

huong-li-nguyen (approved):
LGTM 👍

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -177,8 +177,10 @@ def _create_target_arg_mapping(dot_separated_strings: List[str]) -> Dict[str, Li
 def _update_nested_graph_properties(graph_config: Dict[str, Any], dot_separated_string: str, value: Any):
```
maxschulz-COL, line 177:
```suggestion
def _update_nested_graph_properties(graph_config: Dict[str, Any], dot_separated_string: str, value: Any) -> Dict[str, Any]:
```
This should probably have been there before

vizro-core/tests/unit/vizro/actions/test_actions_utils.py
```diff
@@ -3,34 +3,52 @@
...
+            ({"node": {"label": "A", "color": "blue"}}, "node.color", {"node": {"label": "A", "color": "red"}}),
+            (
+                {"nodes": {"A": {"color": "blue"}, "B": {"color": "green"}}},
+                "nodes.A.color",
+                {"nodes": {"A": {"color": "red"}, "B": {"color": "green"}}},
+            ),
+        ],
+    )
+    def test_update_nested_graph_properties_multiple_levels(self, graph, dot_separated_strings, expected):
+        result = _update_nested_graph_properties(graph, dot_separated_strings, "red")
```
maxschulz-COL, line 25: Can we add `red` to the parametrization, makes it easier to read

vizro-core/tests/unit/vizro/actions/test_actions_utils.py
```diff
@@ -3,34 +3,52 @@
...
+                {"nodes": {"A": {"color": "blue"}, "B": {"color": "red", "value": "red"}}},
+            ),
+            (
+                {"nodes": {"A": {"color": "blue"}, "B": {"color": "green"}}},
+                "nodes.B",
+                {"nodes": {"A": {"color": "blue"}, "B": "red"}},
+            ),
+        ],
+    )
+    def test_update_nested_graph_properties_add_keys(self, graph, dot_separated_strings, expected):
```
maxschulz-COL, line 48: Here as well

vizro-core/tests/unit/vizro/actions/test_actions_utils.py
```diff
@@ -3,34 +3,52 @@
...
+        result = _update_nested_graph_properties(graph, dot_separated_strings, "red")
         assert result == expected
 
-    def test_update_nested_graph_properties_invalid_key(self, fake_graph_property_dict):
```
maxschulz-COL, line 30: Do we need to delete the fixture?

vizro-core/tests/unit/vizro/actions/test_actions_utils.py
```diff
@@ -3,34 +3,52 @@
...
-        with pytest.raises(KeyError, match="C"):
-            _update_nested_graph_properties(fake_graph_property_dict, "nodes.C.color", "red")
-
     def test_update_nested_graph_properties_invalid_type(self):
```
maxschulz-COL, line 52: I remember we still wanted to check the type change (ie from string to number), and why this test fails?

vizro-core/changelog.d/20231222_112619_nadija_ratkusic_graca_parameter_bug.md
maxschulz-COL, line 40:
```suggestion
- Fix a bug that prevented the update of nested graph properties through parameters when the graph property was not previously defined. ([#273](https://github.com/mckinsey/vizro/pull/237))
```
maybe we can be a little more specific, but bot versions are fine for me

vizro-core/tests/unit/vizro/actions/test_actions_utils.py
nadijagraca: The fixture is not used anymore and the test is no longer valid check as we are now adding missing keys to the graph property dict.

vizro-core/tests/unit/vizro/actions/test_actions_utils.py
nadijagraca: This test is very confusing because it mixes numbers and str but from what I understand is that the test fail has nothing to do with type change from str to number or viceversa, but to the fact you cannot change, in this example' {'color': 'blue'} -> {'color': {'value': 42}}. 
I'm still investigating.

maxschulz-COL: Ok let me know how it goes

vizro-core/tests/unit/vizro/actions/test_actions_utils.py
```diff
@@ -3,38 +3,60 @@
...
+            ),
+            (
+                {"nodes": {"A": {"color": "blue"}, "B": {"color": "green"}}},
+                "nodes.B",
+                "red",
+                {"nodes": {"A": {"color": "blue"}, "B": "red"}},
+            ),
+        ],
+    )
+    def test_update_nested_graph_properties_add_keys(self, graph, dot_separated_strings, value, expected):
```
maxschulz-COL, line 52:
```suggestion
    def test_update_nested_graph_properties_add_or_overwrite_keys(self, graph, dot_separated_strings, value, expected):
```

maxschulz-COL: And could we also add an empty `graph`, that I think is the last remaining case that might happen right?

maxschulz-COL (approved):
Lgtm! Thanks for adressing comments. Have two small more suggestions

PR #238, opened 2023-12-22
Introduce page-header for dashboard-title and settings

huong-li-nguyen:
## Description
**Context: There are currently some issues with the dashboard-title position:**
- Current positions takes a lot of horizontal space if there are no controls/navigations
- Hierarchy is not clear as title is on the same level as nav-bar and page-title
Screenshot 2024-01-02 at 22 05 16

**Introduction of `page-header`**:
- Solves for issues explained above 
- Creates space for yet-to-come horizontal nav-bar 
- Enables simple positioning of logo (squared and rectangular format)

Above solution required several CSS changes such that it looks good with the page-header as well, see code for more details

### Layout Screenshots
Take a look at the Miro board - there are several screenshots that show you how it looks now [here](https://miro.com/app/board/o9J_ldRDmCQ=/?moveToWidget=3458764572871119913&cot=14)

### Container Structure

https://miro.com/app/board/o9J_ldRDmCQ=/?moveToWidget=3458764572914015668&cot=14
Screenshot 2024-01-03 at 13 00 18

## Screenshot
In the Miro board there are more screenshots - adding a selected one here:

Screenshot 2024-01-02 at 22 06 02

## Notice


maxschulz-COL:
Starting having a look. I tried to understand these changes in the context of https://github.com/mckinsey/vizro/pull/205, but it was a little tricky to do as the graphic does not seem to match anymore. Do you think it makes sense to provide an update of that?

huong-li-nguyen:
> Starting having a look. I tried to understand these changes in the context of #205, but it was a little tricky to do as the graphic does not seem to match anymore. Do you think it makes sense to provide an update of that?

Sure, I've updated the PR description 👍  In general, it's best to check the attached Miro links, I've added screenshots of the layout everywhere so you can imagine how it will look like.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -155,23 +155,30 @@ def _get_page_divs(self, page: Page) -> _PageDivsType:
...
         left_sidebar_divs = [page_divs["nav-bar"]]
         left_main_divs = [
-            html.Div(left_header_divs, id="left-header", hidden=_all_hidden(left_header_divs)),
             page_divs["nav-panel"],
             page_divs["control-panel"],
         ]
+        right_header_divs = [page_divs["page-title"]]
+
+        # Apply different container position logic based on condition
+        page_header_divs.append(page_divs["settings"]) if self.title else right_header_divs.append(
```
maxschulz-COL, line 167: In the future this would be if there was a logo but no title?

maxschulz-COL (approved):
Lgtm! 🚀  The graphic really helped, I think its good to have it in a PR for public view 🙏

vizro-core/src/vizro/models/_dashboard.py
huong-li-nguyen: Yes, exactly. Here is actually a previous commit where I've prototyped with the logo: https://github.com/mckinsey/vizro/pull/238/commits/ac696cf4bad24e9c77324bd7fbaca44ad277c3c5

I've removed it from this PR as this will be part of another PR

vizro-core/changelog.d/20240103_210035_huong_li_nguyen_dashboard_title_position.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Move dashboard tite into top header container `page-header` ([#238](https://github.com/mckinsey/vizro/pull/238))
```
petar-qb, line 28:
```suggestion
- Move dashboard title into top header container `page-header` ([#238](https://github.com/mckinsey/vizro/pull/238))
```

huong-li-nguyen: Thanks! ⭐

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -120,14 +125,12 @@
...
   flex-direction: column;
-  padding-top: 26px;
-  width: 64px;
 }
 
 .icon-button {
   align-items: center;
   display: flex;
-  height: 64px;
+  height: 80px;
```
nadijagraca, line 133: Is the new height value according to figma designs? The icon looks a bit small compared to the empty space around it.

nadijagraca (approved):
I just had one small comment, but otherwise looks good. I like the new design with the dashboard title separated (makes much more sense and the hierarchy is clear). 🚀

vizro-core/src/vizro/static/css/layout.css
huong-li-nguyen: That's a good question! I've consciously deviated from the old Figma design because it didn't make much sense with the current paddings e.g. we have an outer padding of 32px everywhere except for the navigation bar. So the `dashboard-title`or logo always look a bit misaligned with the navigation bar, while they are aligned with the rest of the elements (e.g. when the navigation bar doesn't exist). I think the old figma designs did not consider that, and we really do need designs with all containers on page. 

However, until we get that, I took Kedro Viz as a reference: https://demo.kedro.org/ - Their navigation bar is also 80px wide. This also looks better aligned for us.

vizro-core/src/vizro/models/_dashboard.py
huong-li-nguyen: Separate PR on the logo: https://github.com/mckinsey/vizro/pull/248

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -155,23 +155,30 @@ def _get_page_divs(self, page: Page) -> _PageDivsType:
         return html.Div([dashboard_title, settings, page_title, nav_bar, nav_panel, control_panel, components])
 
     def _arrange_page_divs(self, page_divs: _PageDivsType):
```
antonymilne, line 157: Is `_PageDivsType` still correct or does it need updating? (Hopefully mypy would have caught if it needed updating, but good to check...)

vizro-core/src/vizro/models/_dashboard.py
antonymilne: It's not very good practice to use a ternary operator for something that doesn't have an explicit return value but instead relies on side-effects (i.e. `append` here), Better to spell out in full as:
```
if self.title:
    page_header_divs.append
else:
   ...
```

antonymilne (approved):
This is great I think, definitely feels like a good solution to me 👍 

As per a comment left on your other PR, I do wonder whether we should be changing some of these `div`s to more semantic elements, but it's really not important and definitely not worth doing now.

PR #245, opened 2024-01-03
Replace `toggle-switch` and fix CSS

huong-li-nguyen:
## Description
- Replaced `daq.BooleanSwitch` with `dmc.Switch` such that we can remove the entire `daq` dependency
- Added CSS for `dmc.Switch` and extended it such that it will be applied to all toggle-switches (e.g. if a user creates a custom component based on `dmc.Switch` and wants to add it anywhere on page, it will be styled correctly automatically)
- Turned the toggle-switch on for `vizro_light` instead of `vizro_dark` (as our default  is `vizro_dark` and should be turned off by default if a user didn't click on it)
- Added CSS for the toggle-switch when turned on

## Screenshot
Screenshot 2024-01-03 at 17 22 57
Screenshot 2024-01-03 at 17 23 04

## Notice

### DIFF
--- a/vizro-core/src/vizro/static/js/models/dashboard.js
+++ b/vizro-core/src/vizro/static/js/models/dashboard.js
@@ -1,3 +1,3 @@
 export function _update_dashboard_theme(on) {
-  return on ? "vizro_dark" : "vizro_light";
+  return on ? "vizro_light" : "vizro_dark";
 }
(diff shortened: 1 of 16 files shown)


vizro-core/src/vizro/static/js/models/dashboard.js
```diff
@@ -1,3 +1,3 @@
 export function _update_dashboard_theme(on) {
-  return on ? "vizro_dark" : "vizro_light";
+  return on ? "vizro_light" : "vizro_dark";
```
petar-qb, line 2: Maybe we should replace the "on" with the "checked" variable name even here.

huong-li-nguyen: Makes sense 👍 When you are back - let's schedule a session where you can explain me these js tests 😄

maxschulz-COL (approved):
Great stuff!

vizro-core/src/vizro/static/js/models/dashboard.js
maxschulz-COL: Would also like a refresher @petar-qb  If I recall correctly there are some js tests outstanding even no? for some clientside callback

nadijagraca (approved):
Looks good. 👍

petar-qb (approved):
Looks good!

PR #254, opened 2024-01-05
Introduce `Container` model

huong-li-nguyen:
## Description
- Enables `Container` model as a new page component with these fields: `title`, `components`, `layout`
- Renames selected CSS classes (other ideas welcome, it's getting trickier now with all the components, containers and their nested elements 😄)
- Refactor previously private methods of `Page` as reusable functions
- Transform duplicated validators into reusable validators and apply to `Page`, `Container` and `Form`
- Refactor actions code to incorporate for `Containers` and nested `Containers` 

Note: Tests and Docs will be done in a separate PR as this one is already quite big

- [x] Docs: https://github.com/mckinsey/vizro/pull/262
- [x] Tests: https://github.com/mckinsey/vizro/pull/265

## Screenshot
Screenshot 2024-01-05 at 16 51 49

## Notice


vizro-core/examples/default/app_dev.py
```diff
@@ -0,0 +1,337 @@
+"""Example to show dashboard configuration."""
```
huong-li-nguyen, line 1: This file is just for dev demo purposes - will be deleted later

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -24,3 +25,53 @@ def get_unique_grid_component_ids(grid: List[List[int]]):
...
+        grid = [[i] for i in range(len(values["components"]))]
+        return Layout(grid=grid)
+
+    unique_grid_idx = get_unique_grid_component_ids(layout.grid)
+    if len(unique_grid_idx) != len(values["components"]):
+        raise ValueError("Number of page and grid components need to be the same.")
+    return layout
+
+
+def _create_component_container(self, components_content):
```
huong-li-nguyen, line 53: I am wondering if this is the most efficient way to refactor these functions. These were previously private methods of the Page model, but they can now be re-used inside the Container and Form model. So it didn't make much sense to me to keep them as private methods to Page.

On the other hand, it's now more difficult to overwrite just these functions, as users now have to overwrite the entire build of the relevant model. Another option would be to wrap these functions inside private methods, but then we would have to duplicate them inside Page, Container, and Form.

vizro-core/src/vizro/models/_page.py
```diff
@@ -116,7 +131,8 @@ def pre_build(self):
                     ),
                     actions=[
                         Action(
-                            id=f"{ON_PAGE_LOAD_ACTION_PREFIX}_action_{self.id}", function=_on_page_load(page_id=self.id)
+                            id=f"{ON_PAGE_LOAD_ACTION_PREFIX}_action_{self.id}",
+                            function=_on_page_load(targets=targets),
```
petar-qb, line 135: Targets are calculated in advance for the `_on_page_load` action (aligned with eagerly calculating targets for the `_filter` action).

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -25,6 +25,7 @@ def __init__(self):
         self.__models: Dict[ModelID, VizroBaseModel] = {}
         self._frozen_state = False
 
+    # TODO: Consider storing "page_id" or "parent_model_id" and make searching helper methods easier?
```
petar-qb, line 28: Delete all "Consider" TODOs before merging.

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -46,12 +47,60 @@ def __iter__(self) -> Generator[ModelID, None, None]:
...
+        def __get_model_children_helper(model: VizroBaseModel) -> None:
+            model_children.append(ModelID(str(model.id)))
+            if hasattr(model, "components"):
+                for sub_model in model.components:
+                    __get_model_children_helper(model=sub_model)
+
+        __get_model_children_helper(model=self[model_id])
+        return model_children
+
+    # TODO: Consider moving this method in the Dashboard model or some other util file
```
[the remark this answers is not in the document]
maxschulz-COL: Similar for the `get_model_children`.

Or is that fact that we explicitly mention higher level models an argument against this?

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -46,12 +47,60 @@ def __iter__(self) -> Generator[ModelID, None, None]:
...
+                    page_model_ids.append(actions_chain.id)
+                    for action in actions_chain.actions:
+                        page_model_ids.append(action.id)
+
+            for control in page.controls:
+                page_model_ids.append(control.id)
+                if hasattr(control, "selector") and control.selector:
+                    page_model_ids.append(control.selector.id)
+
+            # TODO: Add navigation, accordions and other page objects
```
maxschulz-COL, line 99: I would simplify this to a recursive search using something like:

```py
[a for a in dir(page) if not (a.startswith('__') or a.startswith('_') or callable(getattr(page, a)))]
```

which in the case of page finds `['actions', 'components', 'controls', 'description', 'id', 'layout', 'path', 'title']` which we can the feed back again, and always attach if we find a model

Like that this is automatically extensive and we do not have to add things here when we add a new field e.g.

maxschulz-COL: Maybe this approach could be comined also with the children finding

vizro-core/src/vizro/models/_models_utils.py
maxschulz-COL: Would a potential solution be to have a 2 step process:
- `Container`, `Form` and `Page` have a method each to themselves
- that method just calls some commonly shared private fuction

This would keep the repetition to a minimum while allowing users to just overwrite their respective calling methods.

vizro-core/src/vizro/models/_page.py
```diff
@@ -103,10 +94,34 @@ def __init__(self, **data):
...
+
+        return page_actions_chains
+
+    def _get_page_model_ids_with_figure(self) -> List[ModelID]:
+        """Gets all components from the page that have a 'figure' registered."""
+        return [
+            model_id
+            for model_id in model_manager._get_model_children(model_id=ModelID(str(self.id)))
+            if hasattr(model_manager[model_id], "figure")
+        ]
```
maxschulz-COL, line 118: This follows our definition here: https://github.com/mckinsey/vizro/pull/114#discussion_r1369160777

but are we sure that this is stable enough, or should this be even more broad and check for registered data?

In any case I think we should explain this "figure" business in a comment

vizro-core/src/vizro/models/_page.py
```diff
@@ -103,10 +94,34 @@ def __init__(self, **data):
...
+        for model_id in model_manager._get_model_children(model_id=ModelID(str(self.id))):
+            model = model_manager[model_id]
+            if hasattr(model, "actions"):
+                page_actions_chains.extend(model.actions)
+
+        for control in self.controls:
+            if hasattr(control, "selector") and control.selector:
+                page_actions_chains.extend(control.selector.actions)
+
+        return page_actions_chains
```
maxschulz-COL, line 110: Maybe this can also make use of the recursive approach outlined before, as this looks again quite implementation specific

vizro-core/src/vizro/models/types.py
```diff
@@ -362,7 +362,7 @@ class OptionsDictType(TypedDict):
 [`Parameter`][vizro.models.Parameter]."""
 
 ComponentType = Annotated[
-    Union["Button", "Card", "Graph", "Table"],
+    Union["Button", "Card", "Graph", "Table", "Container"],
```
maxschulz-COL, line 365:
```suggestion
    Union["Button", "Card", "Container", "Graph", "Table"],
```

vizro-core/tests/unit/vizro/models/test_page.py
```diff
@@ -110,7 +110,10 @@ def test_valid_component_types(self, standard_px_chart, standard_dash_table):
         [vm.Checklist(), vm.Dropdown(), vm.RadioItems(), vm.RangeSlider(), vm.Slider()],
     )
     def test_invalid_component_types(self, test_component):
-        with pytest.raises(ValidationError, match=re.escape("(allowed values: 'button', 'card', 'graph', 'table')")):
+        with pytest.raises(
+            ValidationError,
+            match=re.escape("(allowed values: 'button', 'card', 'graph', 'table', 'container')"),
```
maxschulz-COL, line 115:
```suggestion
            match=re.escape("allowed values: "),
```

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
```diff
@@ -1,91 +1,30 @@
...
+        for actions_chain in page._get_page_actions_chains()
         for action in actions_chain.actions
         if action.function._function == action_function
     ]
 
 
 # CALLBACK STATES --------------
-def _get_inputs_of_controls(action_id: ModelID, control_type: ControlType) -> List[State]:
+def _get_inputs_of_controls(page: Page, control_type: ControlType) -> List[State]:
     """Gets list of States for selected control_type of triggered page."""
```
maxschulz-COL, line 27:
```suggestion
    """Gets list of States for selected `control_type` of triggered `page`."""
```

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
```diff
@@ -97,17 +36,16 @@ def _get_inputs_of_controls(action_id: ModelID, control_type: ControlType) -> Li
 
 
 def _get_inputs_of_figure_interactions(
-    action_id: ModelID, action_function: Callable[[Any], Dict[str, Any]]
+    page: Page, action_function: Callable[[Any], Dict[str, Any]]
 ) -> List[Dict[str, State]]:
     """Gets list of States for selected chart interaction `action_name` of triggered page."""
```
maxschulz-COL, line 41:
```suggestion
    """Gets list of States for selected chart interaction `action_function` of triggered `page`."""
```
If this is correct, then please check the other docstrings

maxschulz-COL (commented):
Looks really good already, and what a biest of a PR. Went through as much as I could and have a few suggestions. Thanks for taking this on you two!

vizro-core/src/vizro/models/_models_utils.py
huong-li-nguyen: Yeeees, that would be the last option I proposed. That would also be fine for me, but was wondering what people usually do here. Also wondering whether these functions are even something people would like to overwrite.

The whole assignment of grid areas could e.g. also be done via pure CSS so one wouldn't have to change these private methods.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -25,7 +25,7 @@
 from vizro.models.types import MultiValueType, SelectorType
 
 if TYPE_CHECKING:
-    from vizro.models import Page
+    pass
```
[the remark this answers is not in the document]
huong-li-nguyen: Ah yes, this can be removed now! We previously needed it for the circular import errors, but given that @petar-qb refactored it, we don't need the import of the Page model anywhere

vizro-core/examples/default/app_dev.py
```diff
@@ -0,0 +1,337 @@
...
+
+                #### Observation
+                A global trend of increasing life expectancy emerges, with some exceptions in specific African
+                countries. Additionally, despite similar population growth rates across continents, the overall
+                global population continues to expand, with India and China leading the way.  Meanwhile, GDP per
+                capita experiences growth in most regions.
+
+            """
+        ),
+        vm.Graph(
```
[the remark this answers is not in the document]
huong-li-nguyen: Can you show me the screenshot? Somehow I get re-directed to the PR when I click on the image that you are linking to! @petar-qb - do you see all the graphs?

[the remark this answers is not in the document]
huong-li-nguyen: For me it looks like this:
Screenshot 2024-01-11 at 14 59 47

It could be that your screen height is lower than mine, in which case the chart is then not visible. This is a general issue of the layout rather than the container though, as our grid-layout will always try to squeeze in all components on one page. So sometimes the charts might be squeezed too much and are then not visible. In these cases the user has to increase the row min height e.g. if you add the code below to the relevant Page model config, can you see the chart then?


`layout=vm.Layout(grid=[[i] for i in range(4)], row_min_height="450px")`

vizro-core/tests/unit/vizro/models/test_page.py
huong-li-nguyen: I think we've had this discussion one time and said that we want to explicitly write out the allowed values here. Even though this test has to be updated every time we add a new component, I still prefer to have this test more verbose and explicit about which values are allowed.

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
huong-li-nguyen: Should this even matter? Happy to follow a guideline here, but so far I don't think we have been consistently adding single quotation marks to model fields and Models in private methods. Shall we start doing that? Or shall we be less strict about it? I am fine either way

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
huong-li-nguyen: Should this even matter? Happy to follow a guideline here, but so far I don't think we have been consistently adding single quotation marks to model fields and Models in private functions. Shall we start doing that? Or shall we be less strict/pedantic about it? I am fine either way

vizro-core/tests/unit/vizro/models/test_page.py
huong-li-nguyen: Actually, never mind - in the test above we actually test for valid components and list them out explicitly already. So this one here is a bit redundant. I'll remove it!

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
maxschulz-COL: I proposed it here as it helped me read the docstring better/easier in other cases where it was in place. Here e.g. I read the docstring three times to ensure I was getting the meaning (e.g. maybe the triggered page is coming from somewhere else). 

I think we don't need to enforce it, but I think it's always good to make things more readable/improve. In the case below e.g. it was referring even to arguments not present which is not so good.

Given that it's super easy to accept a suggestion it could be a good addition?

vizro-core/examples/default/app_dev.py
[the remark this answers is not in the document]
maxschulz-COL: Ah weird why it never loads, but you are right, that will be it!

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
huong-li-nguyen: If it helps you read it, then let's be consistent about it :) 

We can change the relevant parts here in the PR only, but then I'll create a ticket where someone can quickly go through all docstrings and have function arguments in single quotation marks then 👍

vizro-core/tests/unit/vizro/models/test_page.py
maxschulz-COL: If I understand, you mean:
- we kill the `test_valid_component_types` test
- list the valid components in the `match`?

Clever idea! But I think we miss then the case in which a valid component doesn't work for some reason, which could happen if someone creates a new component, adds it to the union, but say forgets to adhere to something. Maybe not so likely, but the test is already written lol

huong-li-nguyen: But `test_valid_component_types` also wouldn't capture that 😄  And I think we do test these use cases actually in the types tests on custom components that Antony wrote?

maxschulz-COL: I thought we were testing the instatiation of the model with various valid and invalid components. In my eyes it would suffice even to just check if one invalid component fails, without a specific error message, as this really is testing pydantic.

But maybe I also don't get smth, happy either way

huong-li-nguyen: Yes, that's what I also just thought. I think we can:
- Kill `test_valid_component_types`
- Test only one invalid input e.g. `vm.Dropdown`
- Keep the error message though with the valid inputs

I'll push that change then 👍

vizro-core/src/vizro/models/_models_utils.py
huong-li-nguyen: Returning to this one, I quickly wanted to show what I meant. I am convinced that the users don't need to overwrite these two private methods, `_create_component_container` and `_assign_component_grid_area.` And in this case, I don't see a strong reason anymore to keep them as private methods in the models. 

### Why is that?
I can only imagine people wanting to overwrite these private methods when they want to change our default layout, which is a CSS grid, to a Flex-Box layout. This can be the case if they don't care about a fixed grid layout but, e.g., want their components to be stacked subsequently (remember the prototype @antonymilne once built with the custom project cards). So there are two approaches:

#### 1. Use our existing CSS grid layout
No need to overwrite the private methods, as our default is working for them

**Result**
Screenshot 2024-01-12 at 09 30 24


#### 2. Use a Flexbox by leveraging custom CSS

```
.grid-layout {
  display: flex;
  flex-direction: row;
  flex-wrap: wrap;
  justify-content: flex-start;
}

.card_container {
  min-width: 450px;
}
```

**Result**

Screenshot 2024-01-12 at 09 45 49

@antonymilne - I realized method 2) would have saved you some time creating that layout 😅 

I think the above is essential to understand because leveraging custom CSS here is much easier than overwriting these private methods.

maxschulz-COL: Hm very interesting and you are probably right that the only time things would be changed by the user can be served by CSS. Would be happy to go with that solution then.

The only thing going for the other solution would then be the "we don't know what we don't know", as in who knows what ppl come up with, but that might quite unlikely :)

vizro-core/src/vizro/managers/_model_manager.py
petar-qb: That sounds promising. I'll test it 👍.

vizro-core/src/vizro/models/_page.py
petar-qb: Do you think we should call `_has_registered_data` from the `DataManager` instead?
If so, then this is the only place where we make this kind of check (in that case I think we should keep `_has_registered_data` method inside the DataManager then).

maxschulz-COL: I am unsure, but wanted to discuss this quickly. So there are two options right?

1. Rely on our taxonomy, and trigger whenever a model has the attribute `figure`?

2. Go via the `DataManager` and check if the component has a registered dataset

On first thought, Option 2 seems more preferable because it is more general, but it may also lead to undesired side effect (e.g. if a registered data set should not have any consequences)

What do you think?

vizro-core/src/vizro/managers/_model_manager.py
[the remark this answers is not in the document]
maxschulz-COL: That is also a strong case of having them in one place. Curious what A think

l0uden:
@huong-li-nguyen 

I've found old bug from the previous branch of tabs an containers where `table` in one container overwrites title of the second container

image

l0uden:
@huong-li-nguyen , can we do smth with x-axis name conflicting with the legend items? it is present on the main screenshot of this PR with four charts in the container.

huong-li-nguyen:
> @huong-li-nguyen , can we do smth with x-axis name conflicting with the legend items? it is present on the main screenshot of this PR with four charts in the container.

This has to be manually set by the user by providing a custom Layout and providing a `row_min_height`. This then activates scrolling. The other option would be to use custom charts and just turn off the legend. But generally all solutions have to be done via the configuration. Are you using this example for your tests right now?

huong-li-nguyen:
> @huong-li-nguyen
> 
> I've found old bug from the previous branch of tabs an containers where `table` in one container overwrites title of the second container
> 
> image

Thanks! I'll take a look later

l0uden:
> > @huong-li-nguyen , can we do smth with x-axis name conflicting with the legend items? it is present on the main screenshot of this PR with four charts in the container.
> 
> This has to be manually set by the user by providing a custom Layout and providing a `row_min_height`. This then activates scrolling. The other option would be to use custom charts and just turn off the legend. But generally all solutions have to be done via the configuration. Are you using this example for your tests right now?

I'm using my own, but it has the same problem. I will try to use first option

vizro-core/src/vizro/models/_models_utils.py
antonymilne: Very good discussion 👍 You're absolutely right that CSS here would have been a much better way to do that layout with a variable number of cards.

Generally I agree with the approach here, but a few things I think we could improve further - let's discuss tomorrow:
1. the function signatures/separation of responsibility of the two helper functions don't feel quite right to me
2. `_models_utils` feels too generic as a home. These functions are specifically related to `Layout`, so , in order of smallest to biggest possible change:
  * maybe best to put them in `_layout` instead
  * in fact, why don't we even make them methods inside `Layout`? This makes it still possible to customise them by subclassing without any duplication
  * should we even have `Layout.build`? I did consider this in an early prototype and chose not to do it, but maybe it's a good idea now. wdyt? Let me see if I can remember why I didn't do it also.

Even if 99% of customisations that a user might ever want to do can be done through CSS, and we could consider the grid functionality as a fundamental part of Vizro that does not need to be user configurable, it would still be nice (but not very important) for this to be extensible through Python, which is most easily achieved by having the methods live in `Layout`.

Something else to consider is whether the same helper functions work already as they are in `Form`. Although maybe we ultimately won't need `Form` now that we have `Container` 🤔

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -0,0 +1,53 @@
...
+    set_layout,
+)
+from vizro.models.types import ComponentType
+
+if TYPE_CHECKING:
+    from vizro.models import Layout
+
+
+class Container(VizroBaseModel):
+    """Html.Div that can wrap other `Page.components`.
```
antonymilne, line 27:
```suggestion
    """Container to group together a set of components on a page.
```

`html.Div` is quite technical/not something that most people actually care about.

It's also arguably not what we should be using - probably we should favour [semantic elements like `<section>` instead](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section). The same could be said for a lot of our divs, like `<nav>`. It's not important or worth changing anywhere now but something for us to have in the back of our minds I think.

It does make me think that `Section` is possibly a better name than `Container` though 🤔 Even though we discussed naming of this to death already, actually I'm not sure we considered that name before...

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Args:
+        type (Literal["container"]): Defaults to `"container"`.
+        components (List[ComponentType]): See [ComponentType][vizro.models.types.ComponentType]. At least one component
+            has to be provided.
+        title (str): Title to be displayed.
+        layout (Layout): Layout to place components in. Defaults to `None`.
+
+    Raises:
+        ValueError: If number of page and grid components is not the same
```
antonymilne, line 37: @maxschulz-COL should we get rid of this across all models (not all functions)? There's any number of reasons a pydantic model could fail validation, and keeping this up to date seems like a hassle (it's already not up to date here or in `Page` since it doesn't mention the "at least one component per page" restriction).

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Re-used validators
+    _validate_components = validator("components", allow_reuse=True, always=True)(set_components)
+    _validate_layout = validator("layout", allow_reuse=True, always=True)(set_layout)
+
+    @_log_call
+    def build(self):
+        components_content = _assign_component_grid_area(self)
+        components_container = _create_component_container(self, components_content)
+        return html.Div(children=[html.H3(self.title), components_container], className="page-component-container")
```
antonymilne, line 53: Probably we should give something here `id=self.id`, but not sure whether it would be the top-level `html.Div` or `components_container`.

antonymilne: Same goes for `Page` actually, where we didn't have it before either. But not very important so feel free to ignore.

antonymilne (commented):
I've now reviewed the model manager bits too 🙂 @petar-qb generally looks great I think. I am not worried about getting it all perfect here since I have a big task as part of #111 to do further refactoring here anyway. The changes you've made are a definite improvement so will make future changes much easier, thank you! 🙏

antonymilne:
@maxschulz-COL @huong-li-nguyen @petar-qb this all gets a kind of pre-approval from me 💯 I left lots of comments but once those are resolved let me know and I'll re-review and chances are very high it will be an approval 😀  Just give me a shout if it's easier to discuss the comments in person or work through them together instead of getting too many long comment threads on github.

vizro-core/src/vizro/models/_components/container.py
huong-li-nguyen: Aligned with @maxschulz-COL that we'll remove it 👍

maxschulz-COL: Sorry forgot about this comment. Yes I agree, it seems in line with the  [google style guidelines](https://google.github.io/styleguide/pyguide.html#doc-function-raises):

> You should not document exceptions that get raised if the API specified in the docstring is violated

vizro-core/src/vizro/models/_components/container.py
huong-li-nguyen: I've implemented the docstring, but for the semantic elements point or even re-naming - let's discuss that on Monday 😄

vizro-core/src/vizro/models/_components/container.py
huong-li-nguyen: The components_container in the Page already has an ID attached, look for `components_container.id = "page-components"`. So if we want to provide the `self.id`, we should probably add it to the outermost container 👍

vizro-core/src/vizro/models/_models_utils.py
huong-li-nguyen: As discussed, we've moved the helper functions to the Layout model and introduced a build method there.

I went via 2 steps:
- Move everything to Layout and have `components` as an argument to the `Layout.build()`: https://github.com/mckinsey/vizro/pull/254/commits/a20859d6a2c18791c5f376a16bc206a094ff5776
- Remove components as an argument to the `Layout.build()` and call `component.build()` inside `Container, Page, Form`: https://github.com/mckinsey/vizro/pull/254/commits/591aae7ed930659511ee61775f47ebc3f3bfa4af


So the skeleton or the grid itself is created by the `Layout,` while the `components` are created and populated inside that skeleton within the relevant models 👍

huong-li-nguyen:
> I still need to review the actions stuff, but here's a partial review so you're not waiting too long...
> 
> Generally this looks great I think! Some general questions:
> 
> 1. `Page.controls` applies to all figures on the page, regardless of whether they're in a (possibly nested) container, right?
> 2. did you test the case where you have `Page.components = [vm.Container(), vm.Graph()]`, so a container and a non-container as siblings?
> 3. are we saving the `Container.controls` field for a separate PR?
> 4. for naming, I'd suggest maybe using `grid` more since ultimately that's what these classes and functions are for. They don't make sense without a `Layout.grid`.

@antonymilne - regarding your questions:
1. Yes, that's the case
2. Yes, I've also added a test example here: https://github.com/mckinsey/vizro/pull/254/commits/8ccf044ffe6c16cda937d6435fdb575b5545251f
3. `Container.controls` will come later
4. What does that refer to again? 😄 


@l0uden - in https://github.com/mckinsey/vizro/pull/254/commits/8ccf044ffe6c16cda937d6435fdb575b5545251f I've also added the example with a Table and a Container and there is no overlapping anymore. Can you double-check if that's the bug you've referred to? :)

l0uden:
> > I still need to review the actions stuff, but here's a partial review so you're not waiting too long...
> > Generally this looks great I think! Some general questions:
> > 
> > 1. `Page.controls` applies to all figures on the page, regardless of whether they're in a (possibly nested) container, right?
> > 2. did you test the case where you have `Page.components = [vm.Container(), vm.Graph()]`, so a container and a non-container as siblings?
> > 3. are we saving the `Container.controls` field for a separate PR?
> > 4. for naming, I'd suggest maybe using `grid` more since ultimately that's what these classes and functions are for. They don't make sense without a `Layout.grid`.
> 
> @antonymilne - regarding your questions:
> 
> 1. Yes, that's the case
> 2. Yes, I've also added a test example here: [8ccf044](https://github.com/mckinsey/vizro/commit/8ccf044ffe6c16cda937d6435fdb575b5545251f)
> 3. `Container.controls` will come later
> 4. What does that refer to again? 😄
> 
> @l0uden - in [8ccf044](https://github.com/mckinsey/vizro/commit/8ccf044ffe6c16cda937d6435fdb575b5545251f) I've also added the example with a Table and a Container and there is no overlapping anymore. Can you double-check if that's the bug you've referred to? :)

No, the bug is still reproducing. To do it just put table into the separate container. In your example it's just a graph.

huong-li-nguyen:
@l0uden  - got it! I've updated the example and added a fix. Can you check again? :)

I'll probably take that fix out and create a separate PR on main, such that @maxschulz-COL can pull it later for his PR on the ag-grid to see how it looks like 👍

vizro-core/src/vizro/managers/_model_manager.py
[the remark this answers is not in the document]
petar-qb: Thanks for comments and suggestions!
Page searching helper methods are now the part of the ModelManager implementation.

vizro-core/src/vizro/models/_page.py
petar-qb: For now, I just left the naive approach (`if hasattr(model_manager[model_id], "figure")`) as it works. Once we introduce something like `Figure` protocol then we can change this part of the code as well.

vizro-core/src/vizro/models/_page.py
petar-qb: I also agree, I will add TODOs where these improvements are mentioned so we can develop these when we decide which way ModelManager will go.

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
```diff
@@ -130,27 +68,25 @@ def _get_inputs_of_figure_interactions(
...
-            _get_inputs_of_controls(action_id=action_id, control_type=Filter) if "filters" in include_inputs else []
-        ),
+        "filters": (_get_inputs_of_controls(page=page, control_type=Filter) if "filters" in include_inputs else []),
         "parameters": (
-            _get_inputs_of_controls(action_id=action_id, control_type=Parameter)
-            if "parameters" in include_inputs
-            else []
+            _get_inputs_of_controls(page=page, control_type=Parameter) if "parameters" in include_inputs else []
         ),
         # TODO: Probably need to adjust other inputs to follow the same structure List[Dict[str, State]]
```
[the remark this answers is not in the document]
petar-qb: Yep, there is a "TODO" above that says: "# TODO: Probably need to adjust other inputs to follow the same structure List[Dict[str, State]]"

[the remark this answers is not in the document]
petar-qb: This part of the code is something that will be adjusted in the "actions v2" process or when we introduce a generic way how the Vizro components deals with the actions world.

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -46,12 +47,60 @@ def __iter__(self) -> Generator[ModelID, None, None]:
...
+    # TODO: Consider moving this method in the Dashboard model or some other util file
+    def _get_model_page(self, model_id: ModelID) -> Page:  # type: ignore[return]
+        """Gets the id of the page containing the model with "model_id"."""
+        from vizro.models import Page
+
+        for page_id, _ in model_manager._items_with_type(Page):
+            page_model_ids = [page_id, self._get_model_children(model_id=page_id)]
+            page: Page = cast(Page, self[page_id])
+
+            if hasattr(page, "actions"):
```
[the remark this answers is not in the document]
petar-qb: You are right, `page.actions` always exists, so I removed the following line `if hasattr(page, "actions")`.

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -46,12 +47,60 @@ def __iter__(self) -> Generator[ModelID, None, None]:
...
+        __get_model_children_helper(model=self[model_id])
+        return model_children
+
+    # TODO: Consider moving this method in the Dashboard model or some other util file
+    def _get_model_page(self, model_id: ModelID) -> Page:  # type: ignore[return]
+        """Gets the id of the page containing the model with "model_id"."""
+        from vizro.models import Page
+
+        for page_id, _ in model_manager._items_with_type(Page):
+            page_model_ids = [page_id, self._get_model_children(model_id=page_id)]
```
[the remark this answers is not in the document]
petar-qb: You're right, the order of these doesn't matter. Unfortunately, our unit tests are not written in too flexible way 😞. If we swap `list` with `set` here, many tests from different files start to fail nondeterministically. To make them work, we would  swap `list` with `set` in many places in the code and tests. 

I don't think it's urgent to "fix" this, so I've left another TODO above `_get_model_children` for now.

vizro-core/src/vizro/models/_page.py
```diff
@@ -103,10 +94,34 @@ def __init__(self, **data):
                 f"as the page title. If you have multiple pages with the same title then you must assign a unique id."
             ) from exc
 
+    def _get_page_actions_chains(self) -> List[ActionsChain]:
+        """Gets all ActionsChains present on the page."""
+        page_actions_chains = []
```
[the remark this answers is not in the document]
petar-qb: Answered in a https://github.com/mckinsey/vizro/pull/254#discussion_r1459035986.

vizro-core/src/vizro/managers/_model_manager.py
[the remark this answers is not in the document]
antonymilne: Cool, sounds good - thank you for checking 👍

vizro-core/src/vizro/models/_layout.py
```diff
@@ -25,6 +26,39 @@ class ColRowGridLines(NamedTuple):
...
+        grid = [[i] for i in range(len(values["components"]))]
+        return Layout(grid=grid)
+
+    unique_grid_idx = _get_unique_grid_component_ids(layout.grid)
+    if len(unique_grid_idx) != len(values["components"]):
+        raise ValueError("Number of page and grid components need to be the same.")
+    return layout
+
+
+# TODO: Insert layout only
```
antonymilne, line 59: Should this TODO still be here?

antonymilne (approved):
This is 99% there I think 👍 Top work to all involved!

@huong-li-nguyen I think we should do the `Layout.build` stuff slightly differently - let me push a commit to illustrate what I mean, since it will be easier than leaving in PR comments.

vizro-core/src/vizro/models/_layout.py
huong-li-nguyen: No! Could you remove it for me when you push the other change? :)

vizro-core/src/vizro/models/_models_utils.py
antonymilne: I think the second approach here is preferable overall. The only thing I didn't like is that relying on `children[idx].children` feels fragile, so I changed that to using a named key instead. See what you think: https://github.com/mckinsey/vizro/pull/273

Let me know if it makes sense. I haven't tested it at all so please do check before merging anything to `main` but hopefully it didn't break anything...

vizro-core/changelog.d/20240105_163349_huong_li_nguyen_enable_containers.md
```diff
@@ -4,23 +4,22 @@ A new scriv changelog fragment.
 Uncomment the section that is right (remove the HTML comment wrapper).
 -->
 
-<!--
 ### Highlights ✨
 
-- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+- Introduce `Container` model as a new `Page` component ([#254](https://github.com/mckinsey/vizro/pull/254))
```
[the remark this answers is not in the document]
maxschulz-COL: You could just update this file with the docs PR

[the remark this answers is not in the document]
huong-li-nguyen: Yes, that was my plan :) But would add it inside the docs PR

maxschulz-COL (approved):
Wohoo, looks good, thanks for getting this over the finish line!

PR #256, opened 2024-01-09
Add keyword `explain` to private vizro-ai method

maxschulz-COL:
## Description
Currently the docs example:

```py
from vizro_ai import VizroAI

vizro_ai = VizroAI()

df = px.data.gapminder()
code_string = vizro_ai._get_chart_code(df, "describe life expectancy per continent over time", explain=True)
```

fails due to the keyword `explain` not being present.

Given that users might want to extract the code, explanation and insights separately, it makes sense to allow this keyword AND to also allow the return of the entire dictionary containing the code string, explanation and insights (see updated docs example).

## Notice


vizro-ai/src/vizro_ai/_vizro_ai.py
```diff
@@ -91,17 +91,18 @@ def _run_plot_tasks(
...
         Can be used in integration with other application if only code snippet return is required.
 
         Args:
             df: The dataframe to be analyzed
             user_input: User questions or descriptions of the desired visual
+            explain: Flag to include explanation in response
         """
         # TODO refine and update error handling
-        return self._run_plot_tasks(df, user_input, explain=False).get("code_string")
+        return self._run_plot_tasks(df, user_input, explain=explain)
```
Anna-Xiong, line 105: this _get_chart_code is specifically for application integration, just for get code. 
- 'if only code snippet return is required.' is the only requirement. 
- if user wants to have explanation, this is equal to use  _run_plot_tasks call
- better to just keep one call if that is the case

Anna-Xiong (changes requested):

vizro-ai/src/vizro_ai/_vizro_ai.py
maxschulz-COL: Hm good point. In this case a user actually ran it with `explain` but I think this may be mainly because it was wrong in the example. I will revert then to code only return, but maybe refer to the other function in the docs.

vizro-ai/docs/pages/user_guides/run_vizro_ai.md
```diff
@@ -71,15 +71,12 @@ Vizro-AI's `_get_chart_code` method returns a dictionary containing the Python c
...
-        business_insights = output.get("business_insights")
+        code_string = vizro_ai._get_chart_code(df, "describe life expectancy per continent over time")
         ```
     === "code_string"
         [![ResultCode]][ResultCode]
 
     [ResultCode]: ../../assets/user_guides/code_string_app_integration.png
 
 The returned `code_string` can be used to dynamically render charts within your application. You may have the option to encapsulate the chart within a `fig` object or convert the figure into a JSON string for further integration.
+In case you would like to use the insights or code explanation, you can use `vizro_ai._run_plot_tasks(df, ..., explain=False)`, which returns a dictionary containing the code explanation and chart insights alongside the code.
```
Anna-Xiong, line 82:
```suggestion
In case you would like to use the insights or code explanation, you can use `vizro_ai._run_plot_tasks(df, ..., explain=True)`, which returns a dictionary containing the code explanation and chart insights alongside the code.
```

Anna-Xiong (approved):
approved already but we need to change it to True to get explanation otherwise it is none for explanation.

PR #268, opened 2024-01-16
[Feat] Apply custom CSS to AgGrid

maxschulz-COL:
## Description
This PR extends https://github.com/mckinsey/vizro/pull/260 to include styling. It is kept separate in order to keep PRs small.

The target designs can be found below:

image
image
image
image

### Open tasks
- [x] Left align left most column
- [x] Change the sorting icon similar to designs 
- [x] Design the potential sorting/filtering pop-ups
- [x] Base custom design on `alpine` theme in order to improve functionality (implemented at https://github.com/mckinsey/vizro/pull/268/commits/7e40f37c1c86488e634d141e6538617fc0086be9)
- [x] Right align numerical columns automatically
- [x] Create distinction between unselected, hover and selected header
- [x] (minor) Header height to 40px
- [x] (minor) Fix the slight mismatch of header and design header on hover
- [x] ~~(hard) Create the column gaps as per designs~~
- [x] ~~(minor) Change the sorting icon to match exactly the designs~~
- [x] Figure out the table sizing (see also https://github.com/mckinsey/vizro/pull/268#discussion_r1504288071 and **next section**):
- [x] Styling the icons of the filter menu (lens and calendar) + change the hover state of the filter menu itself (see @huong-li-nguyen [comment](https://github.com/mckinsey/vizro/pull/268#pullrequestreview-1895770131)
- [x] Style the Radio item to our Vizro specs

@nadijagraca Could you take a look at the last two items?

### On the grid extension and scrolling behaviour
@AnnMarieW  made a further observation about the scrolling behaviour. @huong-li-nguyen I summarized my current understanding below:

The requirements for the AG Grid are:

1. the AG Grid should extend to the full size of the container, horizontally and vertically
2. if the columns naturally extend to the horizontal width (ie there is enough of them, or the content is very long), then a horizontal scroll bar should appear
3. in the above case, the column width should be determined by the column content
4. if the columns do not extend to the horizontal width, they should be stretched to do so
5. if the columns naturally extend to the vertical height (ie there are more rows than fit on the screen), then a vertical scroll bar should appear
6. Everything of the above should happen, but when users resize the window, the table resizes as well AND the column width proportions are kept

The current code does:

1. yes
2. yes, but it appears “out of sight” ie only visible if scrolled to bottom
3. no
4. yes
5. yes, but the header disappears as we are scrolling the container
6. yes

Originally I was only concerned to make 3 happen on top of everything else. With your @AnnMarieW observation, the caveats on 2 and 5 are now also an issue.

Do we think we can solve all 6 points at the same time? Or at least all but 3?

I am starting to think that we may want to loose 6 (ie the proportions are not kept, and just use `responsizeSizetoFit`)?

## Notice


AnnMarieW:
Hi @maxschulz-COL 

Just a heads up that AG Grid supports creating an entirely new theme using Figma.  You can find more information here:  https://ag-grid.com/react-data-grid/ag-grid-design-system/

Alternately, you can get a long way by starting with the `ag-theme-alpine` and customizing the selectors needed to match the theme.

For example, here are the selectors I use to style the grid for light and dark mode in any Bootstra theme.  

```
.dbc-ag-grid .ag-theme-alpine {
  --ag-alpine-active-color: var(--bs-primary);
  --ag-selected-row-background-color: rgba(var(--bs-primary-rgb), 0.3);
  --ag-row-hover-color: rgba(var(--bs-primary-rgb), 0.1);
  --ag-column-hover-color: rgba(var(--bs-primary-rgb), 0.1);
  --ag-input-focus-border-color: rgba(var(--bs-primary-rgb), 0.4);
  --ag-range-selection-background-color: rgba(var(--bs-primary-rgb), 0.2);
  --ag-range-selection-background-color-2: rgba(var(--bs-primary-rgb), 0.36);
  --ag-range-selection-background-color-3: rgba(var(--bs-primary-rgb), 0.49);
  --ag-range-selection-background-color-4: rgba(var(--bs-primary-rgb), 0.59);
  --ag-background-color: var(--bs-body-bg);
  --ag-foreground-color: var(--bs-body-color);
  --ag-border-color:  rgba(173,181,189, 0.40);
  --ag-cell-horizontal-border:  rgba(173,181,189, 0.20);
  --ag-secondary-border-color:  rgba(173,181,189, 0.20);
  --ag-header-background-color:  rgba(173,181,189, 0.20);
  --ag-odd-row-background-color:  rgba(173,181,189, 0.05);
  --ag-control-panel-background-color: var(--bs-body-bg);
  --ag-subheader-background-color:  var(--bs-body-bg);
  --ag-invalid-color: var(--bs-form-invalid-color);
  --ag-font-family: var(--bs-font-family);
  --ag-tooltip-background-color: var(--bs-gray-400);
}
```

Here is a start with some Vizro variables to make it work in light and dark:

```

 .ag-theme-alpine.vizro {
  --ag-background-color: var(--main-container-bg-color);
  --ag-foreground-color: var(--text-primary);
  --ag-border-color:  var(--main-container-bg-color);
  --ag-cell-horizontal-border:  var(--border-subtle-alpha-01);
  --ag-secondary-border-color:  var(--border-subtle-alpha-01);
  --ag-header-background-color:  var(--main-container-bg-color);
  --ag-odd-row-background-color:  var(--main-container-bg-color);
  --ag-control-panel-background-color: var(--main-container-bg-color);
  --ag-subheader-background-color:  var(--main-container-bg-color);
  --ag-row-height: 40px;
  --ag-header-height: 30px;
}

.ag-theme-alpine .ag-header {
    border-bottom: 1px solid var(--border-subtle-alpha-01);
}
```

Note that the className on the  AG Grid component needs to include the base theme (Alpine) like this:

```python
def dash_ag_grid(data_frame=None, **kwargs):
    """Custom AgGrid."""
    return dag.AgGrid(
        className= "ag-theme-alpine vizro",
        rowData=data_frame.to_dict("records"),
       # columnDefs=[{"field": col} for col in data_frame.columns],    Needs to be set as a default prop like in DataTable
        **kwargs
    )


```

AnnMarieW:
Also, if you would like to change the demo app to use the grid instead of DataTable, here is how to add the conditional formatting and the number formatting:

```python
cellStyle = {
        "styleConditions": [
            {
                "condition": "params.value < 1045",
                "style": {"backgroundColor": "#ff9222"},
            },
            {
                "condition": "params.value >= 1045 && params.value <= 4095",
                "style": {"backgroundColor": "#de9e75"},
            },
            {
                "condition": "params.value > 4095 && params.value <= 12695",
                "style": {"backgroundColor": "#aaa9ba"},
            },
            {
                "condition": "params.value > 12695",
                "style": {"backgroundColor": "#00b4ff"},
            },
        ]
    }

    columnDefs = [
        {"field": "country"},
        {"field": "continent"},
        {"field": "year"},
        {
            "field": "lifeExp",
            "type": "rightAligned",
            "valueFormatter": {"function": "d3.format('.1f')(params.value)"},
        },
        {
            "field": "gdpPercap",
            "type": "rightAligned",
            "valueFormatter": {"function": "d3.format('$,.1f')(params.value)"},
            "cellStyle": cellStyle
        },
        {
            "field": "pop",
            "type": "rightAligned",
            "valueFormatter": {"function": "d3.format(',.0f')(params.value)"},
        },
    ]

    page_country = vm.Page(
        title="Benchmark Analysis",
        # description="Discovering how the metrics differ for each country and export data for further investigation",
        # layout=vm.Layout(grid=[[0, 1]] * 5 + [[2, -1]], col_gap="32px", row_gap="60px"),
        components=[
            vm.Table(
                id="table_country_new",
                title="Click on a cell in country column:",
                figure=dash_ag_grid(
                    id="dash_ag_grid_country",
                    data_frame=df,
                    columnDefs=columnDefs,
                    defaultColDef={"resizable": True, "sortable": True},
                    columnSize="sizeToFit",
                ),
                actions=[vm.Action(function=filter_interaction(targets=["line_country"]))],
            ),
    

```

AnnMarieW:
Open tasks -- here are some links that might be helpful:
- creating custom icons:  https://dash.plotly.com/dash-ag-grid/custom-icons-global
- right align number columns - this will be much easier in the next release (coming soon!) with the new feature of [Cell Data Types](https://ag-grid.com/react-data-grid/cell-data-types/)

maxschulz-COL:
> Hi @maxschulz-COL
> 
> Just a heads up that AG Grid supports creating an entirely new theme using Figma. You can find more information here: https://ag-grid.com/react-data-grid/ag-grid-design-system/

Amazing, this should be the approach for our designer once we have the development capacity for it. I will pass it on!

> Alternately, you can get a long way by starting with the `ag-theme-alpine` and customizing the selectors needed to match the theme.

Thanks for the approach, I will also look into this one. In the docs there was the mention

>  If your desired look and feel is very different from the provided theme, at some point it becomes easier to start from scratch

so we opted for a new theme. Do you think that is pushing it too far?

AnnMarieW:
> so we opted for a new theme. Do you think that is pushing it too far?

Haha - yes.  Creating an entirely new theme is certainly possible, and might be easy if you have phd in CSS :sweat_smile: .  But seriously, if that's what you want to do, then the Figma process might be the way to go.

There is a lot of CSS necessary to make the grid function correctly, and I think the design you posted would be achievable by customizing  the Alpine theme  (except for the column gap in the borders - I have no clue how you would do that).

maxschulz-COL:
> > so we opted for a new theme. Do you think that is pushing it too far?
> 
> Haha - yes. Creating an entirely new theme is certainly possible, and might be easy if you have phd in CSS 😅 . But seriously, if that's what you want to do, then the Figma process might be the way to go.
> 
> There is a lot of CSS necessary to make the grid function correctly, and I think the design you posted would be achievable by customizing the Alpine theme (except for the column gap in the borders - I have no clue how you would do that).

@huong-li-nguyen Shall we revert from the new theme then and try to adjust the existing `alpine` theme?

maxschulz-COL:
> There is a lot of CSS necessary to make the grid function correctly, and I think the design you posted would be achievable by customizing the Alpine theme (except for the column gap in the borders - I have no clue how you would do that).

@AnnMarieW That seems a convincing argument to not create our own theme. Didn't know that the functionality was also being compromised. Will definitely look into this

maxschulz-COL:
> > There is a lot of CSS necessary to make the grid function correctly, and I think the design you posted would be achievable by customizing the Alpine theme (except for the column gap in the borders - I have no clue how you would do that).
> 
> @AnnMarieW That seems a convincing argument to not create our own theme. Didn't know that the functionality was also being compromised. Will definitely look into this

@huong-li-nguyen @AnnMarieW I have based the design on the `alpine` theme now - works well and has the added benefit that previously not designed elements such as the filter fly out look better now 💪

vizro-core/src/vizro/static/css/aggrid.css
```diff
@@ -0,0 +1,46 @@
...
+  font-weight: normal;
+}
+
+#page-container .ag-theme-alpine .ag-row {
+  border-bottom: 1px solid var(--border-subtle-alpha-02);
+}
+#page-container .ag-theme-alpine .ag-cell {
+  display: flex;
+  align-items: center;
+  padding-left: 0px;
```
AnnMarieW, line 36: This makes it so that you can't align the column in the column defs

vizro-core/src/vizro/static/css/aggrid.css
```diff
@@ -0,0 +1,46 @@
+#page-container .ag-theme-alpine.vizro {
+  --ag-background-color: var(--main-container-bg-color);
+  --ag-foreground-color: var(--text-primary);
+  --ag-odd-row-background-color: var(--main-container-bg-color);
+  --ag-header-foreground-color: var(--text-primary);
+  --ag-header-background-color: var(--main-container-bg-color);
+  --ag-header-column-resize-handle-display: none;
```
AnnMarieW, line 7: Removing this makes it harder for the user to resize the columns. Is there some other solution that would be consistent with the design?  Maybe show on hover?

vizro-core/src/vizro/tables/dash_aggrid.py
```diff
@@ -4,10 +4,12 @@
...
 @capture("action")
-def dash_ag_grid(data_frame=None):
+def dash_ag_grid(data_frame=None, **kwargs):
     """Custom AgGrid."""
     return dag.AgGrid(
         id="get-started-example-basic",
         rowData=data_frame.to_dict("records"),
-        columnDefs=[{"field": col} for col in data_frame.columns],
+        columnDefs=[{"field": col, "filter": True} for col in data_frame.columns],
+        dashGridOptions={"rowHeight": 40},
```
AnnMarieW, line 13: The row height can be set in the .css file with  `--ag-row-height`

AnnMarieW (commented):
Glad you decided to go with modifying the Alpine theme :slightly_smiling_face: 
I like the header styling!

vizro-core/examples/default/app_dev.py
```diff
@@ -41,8 +40,12 @@ def create_benchmark_analysis():
                 title="Click on a cell in country column:",
                 figure=dash_ag_grid(
                     data_frame=df,
+                    className="ag-theme-alpine vizro",
```
huong-li-nguyen, line 43: The className I would move to the existing dash_ag_grid and define as a default there if the user hasn't specified anything

vizro-core/examples/default/app_dev.py
```diff
@@ -41,8 +40,12 @@ def create_benchmark_analysis():
...
+                    # className="ag-theme-custom-theme",
+                    defaultColDef={"resizable": True, "sortable": True},
+                    columnSize="sizeToFit",
```
huong-li-nguyen, line 46: I've noticed some strange behaviour but not sure if it's the configuration or some outer container we add on our side, but it seems like the ag-grid doesn't resize properly in the Vizro app?

E.g. if I start of with a smaller viewport and then just extend my screen, the app looks like this:

Screenshot 2024-01-17 at 16 27 51

After refreshing the page, it then resizes. Would be good to understand where this comes from and if we can fix it. Try to plug it into a pure dash app, if the issue still remains, then I think it's probably some outer container we add?

vizro-core/examples/default/app_dev.py
```diff
@@ -78,43 +81,43 @@ def create_benchmark_analysis():
...
-            vm.Filter(column="continent", selector=vm.Dropdown(value="Europe", multi=False, title="Select continent")),
-            vm.Filter(column="year", selector=vm.RangeSlider(title="Select timeframe", step=1, marks=None)),
-            vm.Parameter(
-                targets=["line_country.y"],
-                selector=vm.Dropdown(
-                    options=["lifeExp", "gdpPercap", "pop"], multi=False, value="gdpPercap", title="Choose y-axis"
-                ),
+                #     actions=[vm.Action(function=filter_interaction(targets=["line_country"]))],
             ),
+            # vm.Graph(
```
huong-li-nguyen, line 86: As soon as we merge the ag-grid, I agree with AnnMarie, that we should replace the data table with the ag-grid  in our demos :)

vizro-core/src/vizro/static/css/aggrid.css
```diff
@@ -0,0 +1,46 @@
...
+  --ag-font-size: 14px;
+  --ag-borders: none;
+  --ag-icon-font-family: agGridBalham;
+  --ag-icon-size: 14px;
+}
+
+#page-container .ag-theme-alpine .ag-header-cell {
+  border-bottom: 1px solid var(--border-subtle-alpha-02);
+  padding-left: 0px;
```
huong-li-nguyen, line 16: I wouldn't remove the padding from only one side if that's possible. Otherwise, it will be difficult to center content properly. 

A few things I've noticed that I think we can definitely fix:
- set header height to 40px
- set row height to 48px
- set font-size everywhere to 14px (currently I see 15px and 14px mixed)
- fix padding in the cells

Would be great if you could double-check the Figma specifications on dev mode. Let me know if we should connect on that one, but e.g. for the cells we have this:

```
display: flex;
height: 48px;
padding: 0px 12px;
align-items: center;
align-self: stretch;
```

vizro-core/src/vizro/tables/dash_aggrid.py
huong-li-nguyen, line 13: Shall we set the className here as well? Similar how we do it for the data table though - so only take our default if none is given

vizro-core/src/vizro/static/css/aggrid.css
```diff
@@ -0,0 +1,46 @@
+#page-container .ag-theme-alpine.vizro {
+  --ag-background-color: var(--main-container-bg-color);
+  --ag-foreground-color: var(--text-primary);
+  --ag-odd-row-background-color: var(--main-container-bg-color);
+  --ag-header-foreground-color: var(--text-primary);
+  --ag-header-background-color: var(--main-container-bg-color);
+  --ag-header-column-resize-handle-display: none;
+  --ag-font-size: 14px;
```
huong-li-nguyen, line 8: Some of these properties here get overwritten by more specific CSS selectors e.g. the font-size. So you might have to specify them again below in of the classes.

huong-li-nguyen (changes requested):
Hey you two @maxschulz-COL @AnnMarieW,

this already looks great! Looks a lot slicker than the DataTable already! 🚀  Applying the alpine theme and overwriting the properties there sound like the most feasible approach right now. I guess if users currently want to change that, they would have to use the `important!` tags to overwrite certain properties.

I've noticed a few things in terms of design deviations, which we can probably quickly update :) The other thing that would be good to understand is the re-sizing to the available width/height. I've added a screenshot - would be good to check if it's caused by us.

antonymilne (commented):
I'm not going to review this one properly since other people are better placed to do so... But just to say this looks amazing! Really nice work to all involved, great team work! ⭐ 💯

vizro-core/src/vizro/tables/dash_aggrid.py
nadijagraca: I just checked locally, and you can use --ag-row-height as AnnMarie suggested, but you have to delete #page-container from `#page-container .ag-theme-alpine.vizro` in order for it to work.  

If deleting #page-container breaks something in the styling, you can use `height: 40px !important;` in `#page-container .ag-theme-alpine .ag-header-row`.  (unfortunately, it will not work without `!important` tag). Everything seemed okay once I delete it, but do double check if you opt for this option.

vizro-core/src/vizro/static/css/aggrid.css
nadijagraca: I think once he deleted #page-container there will be no need to specify them below. Is there a reason why we added  #page-container to the name `#page-container .ag-theme-alpine.vizro `?

vizro-core/examples/default/app_dev.py
huong-li-nguyen: @maxschulz-COL - found it! 
If you change this to  `columnSize="responsiveSizeToFit"`, it should fit to the entire width.

https://dash.plotly.com/dash-ag-grid/column-sizing#size-to-fit-and-responsive-size-to-fit

vizro-core/src/vizro/tables/dash_aggrid.py
huong-li-nguyen: I think we should avoid using the `!important` tag as it makes it impossible for the user to change it. You can still achieve a row-height of 40px by using some of the CSS selectors that are not `.ag-theme-alpine.vizro`, but more specific than that 👍

vizro-core/src/vizro/static/css/aggrid.css
huong-li-nguyen: I think even if you remove the ID, you will still have to specify some of these attributes in the classes below e.g. remove the `#page-container` and double-check the font-size of the cell values. It's 15px instead of 14px because it gets overwritten by some of the more specific CSS selectors.

It's a bit fiddly but I think we have to try find the relevant CSS selector that's specific enough for us to change the property and still enables the user to overwrite it if they want to (e.g. by them using the important tag)

vizro-core/examples/default/app_dev.py
AnnMarieW: The `sizeToFit` prop is working as intended.  It's called once when the grid is created.  The `responsiiveSizeToFit` will adjust the columns as the screen size changes , but users often find that annoying because they will lose any of the column widths that they set themselves.  

The [Column flex](https://dash.plotly.com/dash-ag-grid/column-sizing#column-flex) might be a better default.

vizro-core/src/vizro/static/css/aggrid.css
maxschulz-COL: Hm good point - I have been thinking the same. I will speak to our designer about this, but for the time being let's leave them removed. The functionality is still there if one desires to resize columns - and it can be overwritten of course if it is very important 💪

vizro-core/src/vizro/tables/dash_aggrid.py
maxschulz-COL: Yes, this is actually already part of the docs PR

vizro-core/examples/default/app_dev.py
maxschulz-COL: Yes, this is alrady implemented as part of the docs PR

vizro-core/src/vizro/static/css/aggrid.css
AnnMarieW: You should be able to align the columns using the column `type` - see more information and examples in the [dash docs](https://dash.plotly.com/dash-ag-grid/column-definitions#right-aligned-and-numeric-columns)

Try changing the demo `app_dev.py` so the `defaultColDef` includes ` "type": "rightAligned"` like this:
```python
              figure=dash_ag_grid(
                    data_frame=df,
                    className="ag-theme-alpine vizro",
                    # className="ag-theme-custom-theme",
                    defaultColDef={"resizable": True, "sortable": True, "type": "rightAligned",},
                    columnSize="sizeToFit",
                ),
```

Only the headers are right aligned.  The Vizro CSS does't allow the columns to be right aligned
 
image

vizro-core/src/vizro/tables/dash_aggrid.py
maxschulz-COL: I just checked, and you are correct @nadijagraca . The row height can be adjusted when removing `#page-container`.

maxschulz-COL: So this now works, but the header height is strange.

I am setting it explicitly, it seems not overwritten, but also it seems to have no effect

vizro-core/examples/default/app_dev.py
maxschulz-COL: This is a tricky one. So before even going into this one, there are a few other questions of what we would like our default to be?

Do we want to our default behaviour even be size to fit? @huong-li-nguyen  The result look quite bad for big tables, but maybe there is a design guidance from Dan?:

image

If yes, I think I agree with @AnnMarieW , because the resizing can be quite annoying. I implemented the column flex as a default, I think it works well, and sizes to fit, but keeps proportions on resizing.

So the current proposed solution would be:
- size to fit space (let's discuss if this is desired)
- keep resized proportions upon window size adjustment with column flex
- user can of course overwrite this default to their liking


Wdyt?

vizro-core/src/vizro/static/css/aggrid.css
maxschulz-COL: Ah intersting! I guess we need to do that based on the content type though 🤔

vizro-core/examples/default/app_dev.py
AnnMarieW: Flex can work well.  I usually like to add a min width with size to fit.  But whatever default is chosen, it's good that it's easy to overwrite :slightly_smiling_face:

vizro-core/src/vizro/tables/dash_aggrid.py
AnnMarieW: The header height is a different selector I think.

vizro-core/src/vizro/static/css/aggrid.css
maxschulz-COL: Could you elaborate on what the Vizro CSS is going? Maybe we could change that. As it stands it looks a bit confusing :)

vizro-core/examples/default/app_dev.py
maxschulz-COL: That is a good idea, might be the best of both worlds to some extent

vizro-core/src/vizro/static/css/aggrid.css
```diff
@@ -0,0 +1,53 @@
...
+  --ag-icon-font-family: agGridBalham; /* confirmed, but hardly visible */
+  --ag-icon-size: 14px; /* confirmed */
+  --ag-row-height: 48px; /* This determines the row height, but not header */
+}
+/* General setting header row*/
+#page-container .ag-theme-alpine .ag-header-row {
+  color: var(--text-secondary);
+  font-weight: normal;
+  font-size: 14px;
+  height: 40px; /* this is not overwritten but has no effect */
```
maxschulz-COL, line 17: @AnnMarieW So I was under the impression that this is the handle, and it seems to not be overwritten, but at the same time changing it has no effect.

vizro-core/src/vizro/tables/dash_aggrid.py
maxschulz-COL: I thought I found it (see [here](https://github.com/mckinsey/vizro/pull/268/files#r1469886327))

vizro-core/src/vizro/static/css/aggrid.css
```diff
@@ -0,0 +1,53 @@
...
+  color: var(--text-primary);
+}
+
+#page-container .ag-theme-alpine .ag-row {
+  border-bottom: 1px solid var(--border-subtle-alpha-02);
+  /*height: 48px; !* This does not seem to set it to 48px due to element style *!*/
+}
+#page-container .ag-theme-alpine .ag-cell {
+  display: flex;
+  align-items: center;
```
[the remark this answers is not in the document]
AnnMarieW: @maxschulz-COL oops - please disregard.  I was messing with the CSS and I had something that was overriding the cell styling.  :see_no_evil:

vizro-core/src/vizro/static/css/aggrid.css
AnnMarieW: The handle I was referring to was the one in the header used to adjust the width by dragging.  It's controlled by 

```
 --ag-header-column-resize-handle-display: none; /* confirmed */
```
When I take this out the handle shows up (it's too bright, but that can be fixed)

image

vizro-core/src/vizro/tables/dash_aggrid.py
AnnMarieW: The header height can be set with 

```
  --ag-row-height: 48px; /* This determines the row height, but not header */
  --ag-header-height: 40px; /* This determines the header height */
```

More info here:  https://www.ag-grid.com/react-data-grid/global-style-customisation-headers/

vizro-core/examples/default/app_dev.py
maxschulz-COL: @AnnMarieW I have implemented the `flex` together with `minWidth` however it does not seem to extend the grid on large screens. Do you know how this could be done?

Li's suggestion (`responsiveSizeToFit`) fixes it in horizontal direction, but not in vertical - but we do not want to use it anyway due to the resizing

vizro-core/src/vizro/tables/dash_aggrid.py
maxschulz-COL: I attempted this, but it seems to have no effect. For instance when I set `--ag-header-height: 200px;` this is not visible.

AnnMarieW: I'm not sure what to say.  I added the following as line 11 in the  vizro-core/src/vizro/static/css/aggrid.css and it worked just fine.  It was a little hard to see because the height of the grid in the demo app is so small but I could definitely change the header height 
```
--ag-header-height: 200px
```

maxschulz-COL: I must have done something silly or not updated properly! I literally just did the exact same again it it works... Apologies for the confusion! 🙈

vizro-core/examples/default/app_dev.py
AnnMarieW: @maxschulz-COL Try putting the` "flex": 1, "minWidth": 70` in the `defaultColDef` instead of the `columnDefs`

maxschulz-COL: This does the trick, thanks for the suggestion ❤️ 

Do you know why this is? Just curious why setting it in the defaults changes how this appears on the page?

Is this due to when the grid actually calculates its column widths?

AnnMarieW: I'm not sure why it needs to be in the `defaultColDefs`, but you made a good guess.

huong-li-nguyen: I think the current behaviour works nicely! Just one question @maxschulz-COL - in the example above you added as a screenshot, I would have expected a horizontal scrollbar to appear automatically. Is this happening when there are too many columns? In general, I think the expected behaviour should be:

- Columns should optimally always fit their content
- If there are too many columns and their contents gets cut off -> horizontal scrollbar should appear
- When a user re-sizes the window, the Table should resize as well

vizro-core/src/vizro/tables/dash_aggrid.py
```diff
@@ -67,6 +68,8 @@ def dash_ag_grid(data_frame, **kwargs):
                 "buttons": ["apply", "reset"],
                 "closeOnApply": True,
             },
+            "flex": 1,
+            "minWidth": 70,
```
huong-li-nguyen, line 72: Can this also be done in the CSS files instead?

vizro-core/src/vizro/static/css/aggrid.css
```diff
@@ -0,0 +1,95 @@
+.ag-theme-quartz.vizro {
```
huong-li-nguyen, line 1: I've had to try out a few stuff myself and added my changes here: https://github.com/mckinsey/vizro/pull/329
You can take a look - I tried to add explanations on the relevant changes.

**Main changes include:**
- **Use AG-Grid variables where possible:** I've removed some of the properties by defining them globally in the ag-grid variables. They have plenty and I really liked their documentation and examples! This way we can avoid specifying certain properties several times.
- **Use our CSS variables for sizes**: Use our spacing, font-size, etc. variables where possible. It will be easier to change these globally then :) 
- **Add property specifications from Figma**: I've added some font specifications and hover font color ones where it was missing. Would be great if you could double-check! I am not sure if they are still visible to you? To access the dev version now, you need to have a license, so let me know if this doesn't work for you anymore.

In that draft PR I only looked at the header and rows - will comment on the other bits in another comment :)

huong-li-nguyen (commented):
Thank you for pushing through this! 🚀 This already looks great, I just have a few suggestions on refactoring some of the CSS and adding some properties which have been missing. 

Regarding some of your questions:

- **(hard) Create the column gaps as per designs** - No need to create the column gaps as per design. I've added a gap on both sides for this for now in the draft PR I attached. That suffices for now!
- **(minor) Fix the slight mismatch of header and design header on hover** - I think that's fixed in the draft PR right now by setting a line height?
- **(minor) Change the sorting icon to match exactly the designs** - No need to implement this. As you have it I think looks fine.

Given that you want to learn more CSS, I'll leave these things for you to fix still, but let me know if this gets too annoying at some point and Nadia or I should help😄 

- [ ] Apply styling to lens and calendar (in general dropdowns, input fields and buttons should look like ours + no round edges). As long as they look similar to our components all good

    Screenshot 2024-02-22 at 13 44 50
 
- [x] Whenever you click on a cell, there is this grey border appearing. See screenshot below. Would be great if you figure out how to get rid of it :)
    Screenshot 2024-02-22 at 17 07 09
MS: fixed with:
```css
#page-container
  .ag-theme-quartz
  .ag-cell-focus:not(.ag-cell-range-selected):focus-within {
  border-color: transparent;
}
```

- [ ] When you hover over the menu icon, it currently has a different grey background. Would be nice to just highlight it in white but no bg color (similar to how our icons work inside the nav bar)

    Screenshot 2024-02-22 at 17 07 17

- [x] Get rid of the second scroll bar. In the current dev example there are currently two scroll bars. I think one is coming from the table-container and then the ag-grid has its own as well. In that case best to get rid of one of the scroll bars, otherwise it's very confusing. And then style it properly. This file should help: https://github.com/mckinsey/vizro/blob/main/vizro-core/src/vizro/static/css/scroll_bar.css
MS: fixed with `"domLayout": "autoHeight",` and `"style": {"height": None},` - this also fills the entire vertical space

vizro-core/src/vizro/tables/dash_aggrid.py
```diff
@@ -67,6 +68,8 @@ def dash_ag_grid(data_frame, **kwargs):
                 "buttons": ["apply", "reset"],
                 "closeOnApply": True,
             },
+            "flex": 1,
```
maxschulz-COL, line 71: Regarding your comment on column size:
> I think the current behaviour works nicely! Just one question @maxschulz-COL - in the example above you added as a screenshot, I would have expected a horizontal scrollbar to appear automatically. Is this happening when there are too many columns? In general, I think the expected behaviour should be:

> - Columns should optimally always fit their content

This is as far as I understand linked to the property [`columnSize`](https://dash.plotly.com/dash-ag-grid/column-sizing#columnsize-property). The default is that the columns fit their content

> - If there are too many columns and their contents gets cut off -> horizontal scrollbar should appear

This is a consequence of the above default

> - When a user re-sizes the window, the Table should resize as well

We chose the option [`flex`](https://dash.plotly.com/dash-ag-grid/column-sizing#column-flex) for this - as it preserves column widths on resizing

As I understand, but please correct me if I am wrong @AnnMarieW, we can either:
- have columns size to their conent and a horizontal scroll bar appearing if there are too many/too long columns
OR
- have columns flexibly adjust to the table size, and preserve their size on table resizing

What I understand is not possible is that:
- columns fit to their content, but if its not enough to fill the entire grid, the columns get enlarged

vizro-core/src/vizro/tables/dash_aggrid.py
maxschulz-COL: As far as I understand no: https://dash.plotly.com/dash-ag-grid/column-sizing#column-flex

But I may not know what CSS can do :)

vizro-core/examples/default/app_dev.py
maxschulz-COL: Answered in this [comment](https://github.com/mckinsey/vizro/pull/268#discussion_r1504288071)

huong-li-nguyen (commented):
I didn't take a detailed look yet, will do that tomorrow :) But one thing you could already try implementing is the following:

The reason why some of the icons were incorrectly colored (lens+calendar) is because we're using a light theme while we turned on vizro-dark. This should be fixed by ensuring that that `ag-theme-quartz-dark` is provided as a className to the ag-grid when vizro_dark is turned on and `ag-theme-quartz` is provided as a className when vizro_light is turned on. @petar-qb wrote some similar client-side callbacks, so maybe you guys can connect :) 

But if you do that, then there is no need in replacing the icons for @nadijagraca anymore. They would be styled correctly automatically

When that is implemented, you could double-check the properties, because I think some overwrites might not be required anymore. But I can also look through it after when I review it.

huong-li-nguyen (approved):
Will leave the review for @nadijagraca and the team as I've pushed several changes.

It's still not perfect, but gets us 90% there. There are a couple of CSS properties which were difficult to customise because it seems like some ag-grid variables don't affect the elements as they are supposed to. Seems like some specificity conflicts between the themes..

I'll ask on the dash-ag-grid repository, but for now - this is good to go for me

nadijagraca (approved):
Looks really good. Awesome job! 🚀  🙌

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -50,7 +50,9 @@ class Table(VizroBaseModel):
     # Convenience wrapper/syntactic sugar.
     def __call__(self, **kwargs):
         kwargs.setdefault("data_frame", data_manager._get_component_data(self.id))
-        return self.figure(**kwargs)
+        figure = self.figure(**kwargs)
+        figure.id = self._callable_object_id
+        return figure
```
petar-qb: I love it!

PR #271, opened 2024-01-18
Add vizro ai task pipeline

Anna-Xiong:
## Description

**Add task pipeline for grouping vizro-ai components.**
- add pipeline class for adding different vizro-ai components for sequential run
- add pipeline manager for maintaining reusable pipelines

Pending tasks for Technical discussion about pipeline  have been added to tickets.
## Notice


vizro-ai/src/vizro_ai/task_pipeline/pipeline.py
```diff
@@ -0,0 +1,60 @@
...
+        """Initialize the Pipeline.
+
+        Args:
+            llm: The LLM instance to be used by components in the pipeline.
+        """
+        self.llm = llm
+        self.components = []
+        self.components_instances = {}
+
+    def add(self, component_class, input_keys: Union[List, None] = None, output_key: Union[str, None] = None):
```
maxschulz-COL, line 21:
```suggestion
    def add(self, component_class, input_keys: Optional[List[str]] = None, output_key: Optional[str] = None):
```
https://stackoverflow.com/questions/51710037/how-should-i-use-the-optional-type-hint

vizro-ai/src/vizro_ai/task_pipeline/pipeline.py
```diff
@@ -0,0 +1,60 @@
...
+        """Add a component class to the pipeline along with its input and output specifications.
+
+        Args:
+            component_class: The class of the component to be added to the pipeline.
+
+            input_keys: The keys or identifiers for the inputs that this component expects. These should match the
+            output keys of previous components in the pipeline, if applicable.
```
maxschulz-COL, line 28: Sorry pain in the butt, but can we indent the next line and not use whitespace between arguments

vizro-ai/src/vizro_ai/task_pipeline/pipeline.py
```diff
@@ -0,0 +1,60 @@
...
+        Returns:
+            Output of the last stage in pipeline.
+
+        """
+        context = initial_args
+        output = None
+        for component_class, input_keys, output_key in self.components:
+            component = self._lazy_get_component(component_class)
+            args = {key: context[key] for key in input_keys} if input_keys else {}
+            output = component.run(**args)
```
maxschulz-COL, line 50:
```suggestion
            kwargs = {key: context[key] for key in input_keys} if input_keys else {}
            output = component.run(**kwargs)
```

vizro-ai/src/vizro_ai/_vizro_ai.py
```diff
@@ -64,13 +63,17 @@ def _run_plot_tasks(
         self, df: pd.DataFrame, user_input: str, max_debug_retry: int = 3, explain: bool = False
     ) -> Dict[str, Any]:
         """Task execution."""
-        target_chart = self._lazy_get_component(GetChartSelection).run(df=df, chain_input=user_input)
-        df_code = self._lazy_get_component(GetDataFrameCraft).run(df=df, chain_input=user_input)
-        visual_code = self._lazy_get_component(GetVisualCode).run(
-            chain_input=user_input, chart_types=target_chart, df_code=df_code
+        self.pipeline_manager.llm = self.llm_to_use
```
maxschulz-COL, line 66: I am not 100%, but since we define `PipelineManager` as a class attribute, would this mean that setting the LLM would alter it for all instances of VizroAI as well? If yes, is that our desired behaviour?

vizro-ai/src/vizro_ai/task_pipeline/pipeline_manager.py
```diff
@@ -0,0 +1,37 @@
...
+from vizro_ai.components import (
+    GetChartSelection,
+    GetCustomChart,
+    GetDataFrameCraft,
+    GetVisualCode,
+)
+from vizro_ai.task_pipeline.pipeline import Pipeline
+
+
+class PipelineManager:
```
maxschulz-COL, line 12: Is this really a pipeline manager, or rather a pipeline collection? Is it best to store this in a class, or would it make more sense to just import pipelines directly as objects?

vizro-ai/src/vizro_ai/task_pipeline/__init_.py
maxschulz-COL, line 1: Do we not want to make this private? What user (atm) would import such a pipeline?

vizro-ai/src/vizro_ai/task_pipeline/pipeline_manager.py
Anna-Xiong: here we need to bridge the llm, llm would only be confirmed when the model is defined when VizroAI is initialized.

about the collection or manager, probably we can make this class to hold more logics in the future, like the for loop we want to plug in can probably also be made here, outside from the main class?

maxschulz-COL: Hm could be. I guess as long as it's private, we do not have to worry too much 💪

vizro-ai/src/vizro_ai/_vizro_ai.py
Anna-Xiong: yes i think we want to use the same model for the all of the LLM calls in each components / task pipelines once it is set in initialization of VizroAI? 

but probably it is better to fix this pipeline manager with the LLM outside of this plot tasks,  better in initialization of `VizroAI` if we have multiple tasks in future, adding the change now.

maxschulz-COL (approved):

vizro-ai/src/vizro_ai/_vizro_ai.py
maxschulz-COL: What I meant was, do we want the following case:

```py
vizro_ai_A = VizroAI(model="foo")

vizro_ai_B = VizroAI(model="bar")

# Then
vizro_AI_A.plot(df,"...")
# would use model="bar" no
```

As a class is a mutable object, and hence we would change it for all classes? Or am I missing something?

Anna-Xiong: i don't think so, objects including instance is mutable just through its own. If you have two instances of a class, mutating one does not inherently mutate the other. They have their own separate attributes and states.
```
class PipelineManager:
    def __int__(self,llm):
        self.llm = llm

class Test:
    _pipeline_manager = PipelineManager()
    def __init__(self, llm):
        self.llm = llm
        self._pipeline_manager.llm = llm

    def print_llm(self):
        print(self.llm)

test_a = Test("A")
test_b = Test("B")

test_a.print_llm()
test_b.print_llm()
# A
# B
```

huong-li-nguyen (approved):
I am currently not able to evaluate this properly 😄 So I'll just approve.
I'll join your calls in the next weeks to get a basic understanding 👍

PR #277, opened 2024-01-23
Fix scrolling and viewport bug on Safari

huong-li-nguyen:
## Description
- Fix bug on scrolling and viewport on Safari 
- Remove `MarkUpSafe` dependency (as it was there only for pen-testing, but we don't use it anywhere anymore)
- Squeezed quick tidy of docstrings in this PR as well

The issues appeared on Safari because of two properties:
- `overflow: hidden `- for context see https://teamtreehouse.com/community/overflow-hidden-not-working-in-ios-safari
- `calc(100%-Xpx)` - had to replace percentage units with viewport height. I'll double-check with @pruthvip15 whether this is the right approach, but for now it fixes the issue and doesn't seem to affect the other browsers

@l0uden  - this should fix the issue we've seen on Safari. Please double-check :) 
@nadijagraca - important to be aware of above such that we always double-check and fix before merge

## Screenshot
**Before (on Safari)**
- Header was not visible
- Screen was being cut-off and not scrollable

Screenshot 2024-01-23 at 09 33 29

**After**
Screenshot 2024-01-23 at 09 34 20

## Notice

### DIFF
--- a/vizro-core/examples/_dev/app.py
+++ b/vizro-core/examples/_dev/app.py
@@ -1,3 +1,5 @@
+"""Rough example used by developers."""
+
 import vizro.models as vm
 import vizro.plotly.express as px
 from vizro import Vizro
--- a/vizro-core/examples/_dev/app.py
+++ b/vizro-core/examples/_dev/app.py
@@ -1,7 +1,6 @@
 import vizro.models as vm
 import vizro.plotly.express as px
 from vizro import Vizro
-from vizro.actions import export_data
 from vizro.tables import dash_data_table
 
 df = px.data.gapminder()
@@ -46,7 +45,7 @@ dashboard = vm.Dashboard(
     title="Dashboard Title",
     pages=[
         table_and_container,
-    ]
+    ],
 )
 
 if __name__ == "__main__":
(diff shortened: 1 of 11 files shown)


vizro-core/examples/_dev/app.py
```diff
@@ -1,23 +1,53 @@
...
+table_and_container = vm.Page(
+    title="Table and Container",
     components=[
-        vm.Graph(figure=px.scatter(df, x="sepal_length", y="petal_width", color="species")),
-        vm.Graph(figure=px.histogram(df, x="sepal_width", color="species")),
-    ],
-    controls=[
-        vm.Filter(column="species"),
+        vm.Container(
+            title="Container w/ Table",
```
huong-li-nguyen, line 12: Will be reverted - just for demo purposes

huong-li-nguyen: That would also be fine for me 👍 I thought it would be just easier to delete the short default example for you guys rather than the longer dev examples I usually add here 😅 Fine either way to be honest

nadijagraca (approved):
Looks good. 🚀

l0uden (approved):
Tested and approved

Joseph-Perkins (approved):

PR #274, opened 2024-01-22
Temp fix dependency issues of kedro and pandas

Anna-Xiong:
## Description
- add pyarrow dependency to fix unit test error 
```
ERROR  - DeprecationWarning: 
Pyarrow will become a required dependency of pandas in the next major release of pandas (pandas 3.0),
(to allow more performant data types, such as the Arrow string type, and better interoperability with other libraries)
but was not found to be installed on your system.
If this would cause problems for you,
please provide us feedback at https://github.com/pandas-dev/pandas/issues/54466
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
```
- set temporary upper bound of pandas and kedro 

## Notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

  - I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
  - I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
  - I certify that the use of this contribution as authorized by the Apache 2.0 license does not violate the intellectual property rights of anyone else.
  - I have not referenced individuals, products or companies in any commits, directly or indirectly.
  - I have not added data or restricted code in any commits, directly or indirectly.

PR #279, opened 2024-01-23
Remove pin and pyarrow dependency

antonymilne:
## Description

A proper fix following https://github.com/mckinsey/vizro/pull/274:
* remove the `pyarrow` dependency - this is not something we want to specify as a dependency. Instead what we should do is ignore the deprecation warning from pandas (which was confusingly written so made it sound like we should add `pyarrow` as a dependency)
* remove upper pin for `kedro` and `pandas`. Our kedro test broke with the release of 0.19.2 due to https://github.com/kedro-org/kedro/pull/3451 and is fixed by adding `kedro-datasets` as a dependency

Also:
* moved some dependencies pinned by Snyk to optional ones since they only come about from `kedro`. These should never have been in the core dependencies, and actually it highlights a fundamental flaw in the use of Snyk that we shouldn't be responsible for transitive dependency versions at all really - to be discussed more another time

## Notice

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,5 @@
+# THIS IS NOT DESCRIBING A PACKAGE, but the DEV environment of this mono-repo
+# In order to install the packages of this mono-repo from source, refer to the pyproject.toml in the relevant folder
 [project]
 authors = [
   {name = "Vizro Team"}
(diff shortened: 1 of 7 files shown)


huong-li-nguyen (approved):
Thanks for fixing 👍

maxschulz-COL (approved):
Ah I like the move to optional dependencies!

PR #283, opened 2024-01-24
Apply `asserts_component_equal` to unit tests for components

nadijagraca:
## Description
Refactoring of unit tests for components where assert_component_equal is implemented. 

## Notice


huong-li-nguyen (commented):
I just quickly screened over - in general looks good 👍 

I think we shouldn't strip the className or parent_className though. If someone were to change it, and they are not the same, the components will look off. 

In the ticket for the assert_component_equal - could you add a checklist of what other test files need to be converted and check off the ones you cover here?

Then it's easier to see where others could pick up :)

vizro-core/tests/unit/vizro/models/_components/test_table.py
```diff
@@ -170,27 +137,44 @@ def test_pre_build_actions_underlying_table_id(self, dash_data_table_with_id, fi
...
         table = vm.Table(
-            id="text_table",
             figure=dash_data_table_with_id,
             actions=[filter_interaction_action],
         )
 
         table.pre_build()
+        built_table = table.build()
+
+        expected_table = dcc.Loading(
```
huong-li-nguyen, line 169:
```suggestion
        table = table.build()
        expected_table = dcc.Loading(
```

vizro-core/tests/unit/vizro/models/_components/form/test_checklist.py
```diff
@@ -148,9 +128,18 @@ def test_set_action_via_validator(self, identity_action_function):
...
     """Tests model build method."""
 
-    def test_checklist_build(self, expected_checklist):
-        checklist = Checklist(options=["A", "B", "C"], id="checklist_id", title="Title").build()
-
-        result = json.loads(json.dumps(checklist, cls=plotly.utils.PlotlyJSONEncoder))
-        expected = json.loads(json.dumps(expected_checklist, cls=plotly.utils.PlotlyJSONEncoder))
-        assert result == expected
+    def test_checklist_build(self):
+        checklist = Checklist(options=["A", "B", "C"], title="Title").build()
```
antonymilne, line 132:
```suggestion
        checklist = Checklist(id="checklist", options=["A", "B", "C"], title="Title").build()
```
For simple cases like this it's best to make the test as strict as possible so to try and not specify any `keys_to_strip`. This means you'll need to supply an `id` here.

vizro-core/tests/unit/vizro/models/_components/form/test_checklist.py
```diff
@@ -148,9 +128,18 @@ def test_set_action_via_validator(self, identity_action_function):
...
+                dcc.Checklist(
+                    options=["ALL", "A", "B", "C"],
+                    value=["ALL"],
+                    className="selector_body_checklist",
+                    persistence=True,
+                    persistence_type="session",
+                ),
+            ],
+        )
+        assert_component_equal(checklist, expected_checklist, keys_to_strip={"id", "className"})
```
antonymilne, line 145:
```suggestion
        assert_component_equal(checklist, expected_checklist)
```

antonymilne (commented):
This is a big improvement but still needs a bit of work I think. I left just one comment but it applies throughout. These tests should be as strict as possible, which basically means the process should be:
1. Don't specify `keys_to_strip` at all
2. Does the test pass?
    * yes: great, your work is done
    * no: figure out how to make the test pass - either add enough information (e.g. specify `id`) to make the output match exactly (preferred in easy cases) or add _only the things you need_ to `keys_to_strip` (preferred only in cases where it's awkward to get everything matching)

For all the cases here I think there shouldn't be a need for `keys_to_strip`. See https://github.com/mckinsey/vizro/commit/360e4bab3a30e2dffd6649448d4f7f09347b5511 for a nice example 🙂

vizro-core/tests/unit/vizro/models/_components/form/test_dropdown.py
```diff
@@ -188,16 +147,53 @@ def test_set_action_via_validator(self, identity_action_function):
...
-
-        result = json.loads(json.dumps(dropdown, cls=plotly.utils.PlotlyJSONEncoder))
-        expected = json.loads(json.dumps(expected_dropdown_without_all, cls=plotly.utils.PlotlyJSONEncoder))
-        assert result == expected
+    def test_dropdown_with_all_option(self):
+        dropdown = Dropdown(
+            options=["A", "B", "C"],
+            title="Title",
+            id="dropdown_id",
+        ).build()
```
huong-li-nguyen, line 155: Can this be put into one line? or is it exceeding the max length?

vizro-core/tests/unit/vizro/models/_components/test_graph.py
```diff
@@ -33,31 +32,6 @@ def standard_px_chart_with_str_dataframe():
...
-            config={
-                "autosizable": True,
-                "frameMargins": 0,
-                "responsive": True,
-            },
-            className="chart_container",
-        ),
-        color="grey",
-        parent_className="loading-container",
-    )
```
huong-li-nguyen, line 58: Have a general question for @antonymilne - is there any preference for whether we put these expected outputs as variables into the relevant unit tests or as fixtures?

Having it directly in the unit tests has the advantage that it's the closest to where it's being used (if only used there). Having it in the fixture makes the unit test easier to overview and people might tend to re-use these fixtures more.

If we do, then I would probably do this consistently throughout the tests here, because currently we mix. I personally prefer it to be consistent, but that's my cleaning up tick 🧹 However, if we don't care, then don't worry about it @nadijagraca  😄

huong-li-nguyen (approved):

antonymilne (approved):
Thanks for making the changes, this all looks good now!

vizro-core/tests/unit/vizro/models/_components/form/test_dropdown.py
huong-li-nguyen: Let's definitely do that 😄  @nadijagraca - can you just create a separate ticket for this? No need to implement this here already.

vizro-core/tests/unit/vizro/models/_components/test_graph.py
huong-li-nguyen: I like that! Let me put this into our coding guidelines (which we still have to align on 😄). @nadijagraca - could you then store all of the fixtures in variables (if they are only being used in that unit test)? I think you've done it for the majority already, so good to double-check everywhere (e.g. the slider currently still uses a fixture that is only being used in that unit test) 👍

vizro-core/tests/unit/vizro/models/_components/test_table.py
```diff
@@ -170,27 +137,46 @@ def test_pre_build_actions_underlying_table_id(self, dash_data_table_with_id, fi
...
-    def test_table_build_mandatory_only(self, standard_dash_table, expected_table):
-        table = vm.Table(
-            id="text_table",
-            figure=standard_dash_table,
-        )
-
+    def test_table_build_mandatory_only(self, standard_dash_table):
+        table = vm.Table(id="text_table", figure=standard_dash_table)
         table.pre_build()
+        table = table.build()
```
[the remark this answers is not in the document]
huong-li-nguyen: I've actually asked her to rename this to` table` so we are consistent with the rest of the tests (dropdown, checklist, slider etc.). I don't mind if we call it `built_table` or `table` though. So if we prefer that, it would be nice if we can do that for the rest of the tests also. Do you have any preference for this one?

Otherwise I would stick with `table `and change the remaining spot where you have it called `built_table`.

vizro-core/tests/unit/vizro/models/_components/test_graph.py
nadijagraca: Thank you both for you comments. I'll make sure to follow the guidelines in the future. The reason why I didn't use a variable for slider and range_slider is because component definition is quite long and I was afraid it will clutter the test. But I can move it to make it aligned with other tests.

vizro-core/tests/unit/vizro/models/_components/form/test_dropdown.py
antonymilne: Personally I think I prefer it but no strong feelings.

The alternative which would solve many existing cases and keep black with its default options is we just do a one off regex to remove all trailing commas now. Then any which are better suited to being on multiple lines can have the `,` manually added back in.

Anything which requires ongoing manual work or thought by developers or reviewers would be a bad system though. So after any initial work reformatting we shouldn't need to think about it again.

vizro-core/tests/unit/vizro/models/_components/test_table.py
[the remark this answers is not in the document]
antonymilne: Oh yes, let's call it  just `table` then and change the other case that doesn't match that. Agree that we should only use `built` if it's really necessary to disambiguate from other things.

vizro-core/tests/unit/vizro/models/_components/form/test_dropdown.py
huong-li-nguyen: Got it - that makes sense! Let me go through some of the cases now and if I miss some, you guys just add it back in manually whenever you see something. And then we'll implement the linter now 👍

PR #262, opened 2024-01-10
Add docs for `Container`

huong-li-nguyen:
## Description

I've considered the [old PR](https://github.com/mckinsey/vizro/pull/206) on this, but made some changes:
- Created an initial overview page for components (we had several places where it became clear that we needed a landing page for components as otherwise it was referencing the Card & Button docs)
- Changed the example for container 
- Updated docs on it

Will be merged after https://github.com/mckinsey/vizro/pull/254

## Screenshot
The overview page can definitely be improved further, but I wouldn't do that in this PR. So currently it's just a basic version.
Screenshot 2024-01-11 at 16 16 50

## Notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

  - I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
  - I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
  - I certify that the use of this contribution as authorized by the Apache 2.0 license does not violate the intellectual property rights of anyone else.
  - I have not referenced individuals, products or companies in any commits, directly or indirectly.
  - I have not added data or restricted code in any commits, directly or indirectly.

PR #284, opened 2024-01-24
Enhance docs for `Container`

huong-li-nguyen:
## Description
For reference, see previous discussion: https://github.com/mckinsey/vizro/pull/262#pullrequestreview-1840927682

## Notice


vizro-core/docs/pages/user_guides/container.md
```diff
@@ -105,3 +115,165 @@ To add a [`Container`][vizro.models.Container] to your page, do the following:
...
+        from vizro import Vizro
+
+        df = px.data.gapminder()
+
+        page = vm.Page(
+            title="Nested Containers",
+            components=[
+                vm.Container(
+                    title="Container Title",
+                    layout=vm.Layout(grid=[[0, 1], [0, 1]], , col_gap="80px"),
```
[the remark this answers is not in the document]
huong-li-nguyen: Good catch, and yes, that works too! Sometimes, I forget the simpler version of the layout 😆 I've also just remembered your comment about storing some of these configurations in variables, such that we don't have seven levels of parenthesis. So I've kept the nested containers in variables as well 👍

vizro-core/docs/pages/user_guides/container.md
```diff
@@ -105,3 +115,165 @@ To add a [`Container`][vizro.models.Container] to your page, do the following:
         [![Container]][Container]
 
     [Container]: ../../assets/user_guides/components/containers.png
+
```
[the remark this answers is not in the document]
huong-li-nguyen: I like it! 👍

vizro-core/docs/pages/user_guides/container.md
```diff
@@ -105,3 +115,165 @@ To add a [`Container`][vizro.models.Container] to your page, do the following:
...
+                  grid: [[0, 1], [0, 1]]
+                  col_gap: 80px
+                type: container
+                title: Container Title
+            title: Nested Containers
+        ```
+    === "Result"
+        [![NestedContainer]][NestedContainer]
+
+    [Container]: ../../assets/user_guides/components/nested_containers.png
```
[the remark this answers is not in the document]
huong-li-nguyen: Yeah, that makes sense! I've tried to make it clearer here now - feel like I'm repeating myself a bit though 😅 : https://github.com/mckinsey/vizro/pull/284/commits/383e08a441ee430067953f44078445f533d30348

Let me know if this works 🤔

vizro-core/docs/pages/user_guides/container.md
```diff
@@ -105,3 +126,173 @@ To add a [`Container`][vizro.models.Container] to your page, do the following:
...
+This nesting capability allows users to create more complex layouts and manage related components at any level of granularity.
+
+To create nested containers, simply add a `Container` to the `components` argument of another `Container`.
+
+!!! note
+
+    Note that an almost identical layout can also be achieved using [`Page.layout`](layouts.md), see the
+    [advanced grid example](layouts.md#grid-advanced-example) on how this can be done. Here we use `Container` instead
+    of `Page.layout` because we need finer control of selected grid areas (e.g. adding different titles to each subgrid
+    and have different spacing between groups of components).
```
maxschulz-COL, line 142:
```suggestion
  In the below example, we have two groups of charts (left and right) that we would like to arrange separately on the `Page` while also providing a separate title for the two groups. This is often the case when some charts are topically related. In this case it makes sense to use `Containers` instead of `Page.layout`.
  
  Note that an almost identical layout can also be achieved using solely the [`Page.layout`](layouts.md) -  see the [advanced grid example](layouts.md#grid-advanced-example) on how this can be done.
```

Here is an alternative which is more user friendly I think. Before I was not quite sure if I would have understood. Too much talk of nesting, hierarchical, grid etc :P

vizro-core/docs/pages/user_guides/container.md
```diff
@@ -105,3 +126,173 @@ To add a [`Container`][vizro.models.Container] to your page, do the following:
...
+                ),
+            ],
+        )
+
+        page = vm.Page(
+            title="Nested Containers",
+            components=[
+                vm.Container(
+                    title="Container Title",
+                    layout=vm.Layout(grid=[[0, 1]], col_gap="80px"),
```
maxschulz-COL, line 220: It's not even clear up to this point that these containers are nested. Maybe it makes sense to have a code pop-out comment here that shows that one could very easily (just 2 brackts or deleting the grid) out the two groups of charts below one another.

If we do, would maybe refer to that also in the above suggestion I made

vizro-core/docs/pages/user_guides/container.md
```diff
@@ -105,3 +126,173 @@ To add a [`Container`][vizro.models.Container] to your page, do the following:
...
+        df = px.data.gapminder()
+
+        nested_container_one = vm.Container(
+            title="Nested Container I",
+            layout=vm.Layout(grid=[[0, 1], [2, 2]]),
+            components=[
+                vm.Graph(
+                    figure=px.line(
+                        df,
+                        title="Graph 1 - Nested Container I",
```
maxschulz-COL, line 161: These titles do no appear fully in the screenshot (numbers are missing) which is actually not ideal, because it helps link the code to the outcome. Maybe we can shorten them so that they do?

vizro-core/docs/pages/user_guides/layouts.md
```diff
@@ -110,6 +114,137 @@ To customize the grid arrangement, provide the relevant grid specification:
 
     [Grid]: ../../assets/user_guides/layout/one_left_two_right.png
 
+### Grid - Advanced Example
+Generally, the `Layout` provides full control over the arrangement of top-level components within a Page,
+allowing arbitrarily granular control of the grid by creating larger grids.
+
+If you want to divide the grid into subgrids with finer control over these, you can use [Containers](container.md).
```
maxschulz-COL, line 121:
```suggestion
If you want to divide the grid into subgrids with finer control over these, you can use [`Containers`](container.md).
```

vizro-core/docs/pages/user_guides/container.md
```diff
@@ -1,13 +1,31 @@
...
+recommended approach if you just want to arrange components on a page with consistent row and/or column spacing.
+However, there are a few cases where you might want to use a `Container` instead:
+
+- If you want to split up your grid into subgrids to organize components together
+- If you want to add a title to your subgrids
+- If you want different row and column spacing between subgrids
+- If you want to apply controls to selected subgrids (will be supported soon)
+
+
+## Containers
```
maxschulz-COL, line 22:
```suggestion
## Basic usage
```
Otherwise it looks weird, especially on the navigation on the right

maxschulz-COL (approved):
Approval from my side, but I have a few suggestions that could improve this even more

vizro-core/docs/pages/user_guides/container.md
huong-li-nguyen: I've renamed to Basic Containers and Nested Containers to be consistent with the `parameters.md` 👍

vizro-core/docs/pages/user_guides/container.md
huong-li-nguyen: Yeah, it's because the titles do not automatically wrap. I've just shortened them now!

vizro-core/docs/pages/user_guides/container.md
huong-li-nguyen: Hmm.. could you clarify? I am not sure if I understand above. I've renamed the variables to `child_container_one` and `child_container_two`, which I think makes it quite clear that the components of the Container itself are Containers now. Does that help?

vizro-core/docs/pages/user_guides/container.md
huong-li-nguyen: Thanks! 🚀  I've changed the order and rephrased a bit. What I would like the users to read first is that a similar layout can be achieved in a simpler way. And then afterwards explaining why we've decided to go for the Containers in this case. WDYT?

  ```
  Note that an almost identical layout can also be achieved using solely the [`Page.layout`](layouts.md) - 
    see the [advanced grid example](layouts.md#grid-advanced-example) on how this can be done.
    
    Here we use the `Containers` instead of `Page.layout`, as we want to provide clearer visual separation between two chart groups.
    In the below example, we have two groups of charts (left and right) and want to assign separate titles to each group
    and increase the gap between them for enhanced visual clarity. Such an arrangement is particularly advantageous 
    when grouping charts that are thematically related.
```

File: vizro-core/docs/pages/user_guides/container.md (hunk 5 above)
Comment (huong-li-nguyen) [reply vizro-core/docs/pages/user_guides/container.md:233]: @antonymilne  - just had a chat with @maxschulz-COL and I get his confusion of why this example is used for nesting as theoretically we could also achieve the same layout by adding the two child container directly into `Page.components`.

Not sure what the right approach here is to make it clear that nesting is possible, but might not always be necessary? Should we actually just kill this section of nested containers and reuse the example here with two containers inside `Page.components`?

in general the nested containers make more sense when the Tabs come in. Before that one could probably achieve most of the layouts with top-level containers and page.layout only?

File: vizro-core/docs/pages/user_guides/container.md (hunk 4 above)
Comment (maxschulz-COL) [reply vizro-core/docs/pages/user_guides/container.md]: Hm I am not sure I like it. But I think we should also see it in the context of the below comment where after discussing with you I am even more of the opinion that the mention of nesting in combination with this example is actually confusing. So let's not iterate on this until we decide the below.

For this specific case:
- I think the admonition above the example looks visually not great, and it is also too big and long now
- The first sentence refers to something that follows a fair bit below, it does not make sense as it stands - I think actually the opposite: a user should understand first what they are supposed to be wanting to achieve, then see the example, then get notified (no need for big stand-out) that they can also achieve this with an alternative

My overall vote would go to:
- Rename this to "Advanced containers"
- kill the paragraph about nested containers (and move to next section - see last point)
- Use something like your second paragraph or my suggestion to frame the problem/idea
- Note subtly that you can achieve this elsehow
- Have another section about "Nested containers" with a 2-liner and maybe an inline example (ie not a big admonition, just some showcase of a `vm.Container` inside a `vm.Container`

File: vizro-core/docs/pages/user_guides/container.md (hunk 5 above)
Comment (maxschulz-COL) [reply vizro-core/docs/pages/user_guides/container.md:233]: See comment above - my vote for the resolution of this

File: vizro-core/docs/pages/user_guides/container.md [hunk 9]
```diff
@@ -1,13 +1,31 @@
...
+enhancing the ability to manage related components.
+
+## When to use containers
+In general, any arbitrarily granular layout can already be achieved using [`Page.layout`](layouts.md) alone and is our
+recommended approach if you just want to arrange components on a page with consistent row and/or column spacing.
+However, there are a few cases where you might want to use a `Container` instead:
+
+- If you want to split up your grid into subgrids to organize components together
+- If you want to add a title to your subgrids
+- If you want different row and column spacing between subgrids
```
antonymilne, line 18: This point about different spacing is a good one and not something I had thought of before 👍

vizro-core/docs/pages/user_guides/layouts.md
```diff
@@ -110,6 +114,137 @@ To customize the grid arrangement, provide the relevant grid specification:
 
     [Grid]: ../../assets/user_guides/layout/one_left_two_right.png
 
+### Grid - Advanced Example
+Generally, the `Layout` provides full control over the arrangement of top-level components within a Page,
```
antonymilne, line 118:
```suggestion
Generally, the `Layout` provides full control over the arrangement of top-level components within a page,
```

vizro-core/docs/pages/user_guides/container.md
antonymilne: I definitely also see the cause of confusion here. The problem is that the use of these nested containers is indeed quite limited until we add `controls` to the `Container` model. At the moment, in terms of the overall grid created, there's nothing you can do with nesting `Container` models that you can't already do with `Page.layout`.

I would still keep this example because the `col_gap` here does enable something you can't do just purely through `Page.layout`. And in future when we add `controls` the example will become much more interesting and clearer what the point of containers is.

antonymilne: Hmm yes I definitely also see the cause of confusion here and I also saw had the same thoughts as @maxschulz-COL when I read through it again. The problem is that the use of these nested containers is indeed quite limited until we add `controls` to the `Container` model. At the moment, in terms of the overall grid created, there's nothing you can do with nesting `Container` models that you can't already do with `Page.layout`.

Let me comment my suggestion on the other conversation to keep it all on one place.

vizro-core/docs/pages/user_guides/layouts.md
```diff
@@ -46,11 +46,11 @@ will automatically be stacked underneath each other. If that is your desired lay
 
 
 
-## Customizing the grid arrangement
+## Configuring the grid
```
antonymilne, line 49: I know it's annoying since I already approved the PR and would be good to get it merged... But I re-read this page and something I think is missing is a quick explanation of horizontal stacking. Currently we say
>  If no layout is specified, all charts/components will automatically be stacked underneath each other.

We should say also:
* this corresponds to `grid = [[0], [1], [2]]` because this specifies 3 equal rows, each of which contains a single component that takes up the whole width
* you can achieve stacking in columns with `grid = [[0, 1, 2]]` because this specifies 1 row that takes up the whole width, then divided into 3 equal columns

Maybe a short new section right above "Basic Example" called "stacking" or something would explain this?

vizro-core/docs/pages/user_guides/container.md
antonymilne: P.S. if you are attached to the current example after spending lots of time on it then fine to make a new section "Advanced containers" and keep it in there like @maxschulz-COL suggested 😀 Alternatively we can just shelve it for now and bring it back with controls when we have added them.

vizro-core/docs/pages/user_guides/container.md
```diff
@@ -105,3 +136,176 @@ To add a [`Container`][vizro.models.Container] to your page, do the following:
         [![Container]][Container]
 
     [Container]: ../../assets/user_guides/components/containers.png
+
```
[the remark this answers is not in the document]
huong-li-nguyen: I've added a comment to the code on line 42 and updated the Note to say it's equivalent to the grid you mention above 👍

vizro-core/docs/pages/user_guides/container.md
huong-li-nguyen: Thanks for your input - I would just shelve it for now! 👍

vizro-core/docs/pages/user_guides/container.md
```diff
@@ -1,13 +1,31 @@
...
+While there is currently no apparent difference in rendering, additional functionality will be added to the `Container` soon (e.g. controls specific to that container),
+enhancing the ability to manage related components.
+
+## When to use containers
+In general, any arbitrarily granular layout can already be achieved using [`Page.layout`](layouts.md) alone and is our
+recommended approach if you just want to arrange components on a page with consistent row and/or column spacing.
+However, there are a few cases where you might want to use a `Container` instead:
+
+- If you want to split up your grid into subgrids to organize components together
```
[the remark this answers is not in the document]
huong-li-nguyen: Updated the section as follows:

Screenshot 2024-01-30 at 18 57 22

vizro-core/docs/pages/user_guides/container.md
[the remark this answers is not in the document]
huong-li-nguyen: Screenshot 2024-01-30 at 18 59 51

vizro-core/docs/pages/user_guides/layouts.md
huong-li-nguyen: That makes sense! 💡 

Screenshot 2024-01-30 at 19 13 29

vizro-core/docs/pages/user_guides/container.md
```diff
@@ -11,15 +11,19 @@ enhancing the ability to manage related components.
 ## When to use containers
 In general, any arbitrarily granular layout can already be achieved using [`Page.layout`](layouts.md) alone and is our
 recommended approach if you just want to arrange components on a page with consistent row and/or column spacing.
-However, there are a few cases where you might want to use a `Container` instead:
+
+The `Page.layout` provides a grid, while the `Container` enables you to insert a `grid` into the component space on that page, allowing for more granular control by
```
antonymilne, line 15:
```suggestion
`Page.layout` has a `grid` argument that sets the overall layout of the page. `Container.layout` also has a `grid` argument. This enables you to insert a further `grid` into a component's space on the page, allowing for more granular control by
```

vizro-core/docs/pages/user_guides/container.md
```diff
@@ -39,18 +43,28 @@ To add a [`Container`][vizro.models.Container] to your page, do the following:
 
         page = vm.Page(
             title="Containers",
-            components=[
+            components=[  # (1)!
                 vm.Container(
                     title="Container I",
-                    layout=vm.Layout(grid=[[0, 1]]),  # (1)!
+                    layout=vm.Layout(grid=[[0, 1]]),
```
antonymilne, line 49:
```suggestion
                    layout=vm.Layout(grid=[[0, 1]]), # (2)!
```

vizro-core/docs/pages/user_guides/container.md
```diff
@@ -78,7 +92,7 @@ To add a [`Container`][vizro.models.Container] to your page, do the following:
         Vizro().build(dashboard).run()
         ```
 
-        1.  If you don't specify `layout`, it will default to `[[0], [1]]`, meaning the components will be arranged in rows.
+        1.  Note that the `Page.layout` argument is not specified here and will therefore default to `[[0], [1]]`, meaning the containers will be stacked in rows.
```
antonymilne, line 95:
```suggestion
        1.  Note that the `Page.layout` argument is not specified here and will therefore default to `[[0], [1]]`, meaning the containers will be stacked in rows.
         2. Stack the components inside this `Container` in columns.
```

vizro-core/docs/pages/user_guides/layouts.md
```diff
@@ -7,7 +7,7 @@ The [`Page`][vizro.models.Page] model accepts the `layout` argument, where you c
 
 ## Using the default layout
 The `layout` argument of the [`Page`][vizro.models.Page] model is optional. If no layout is specified, all charts/components
-will automatically be stacked underneath each other. If that is your desired layout, you can create your charts/components without providing a [`Layout`][vizro.models.Layout].
+will automatically be stacked horizontally. If that is your desired layout, you can create your charts/components without providing a [`Layout`][vizro.models.Layout].
```
antonymilne, line 10:
```suggestion
will automatically be stacked in rows down the page. If that is your desired layout, you can create your charts/components without providing a [`Layout`][vizro.models.Layout].
```

Due to the slight ambiguity mentioned below, best to start off with non-technical language when this is first mentioned I think.

vizro-core/docs/pages/user_guides/layouts.md
```diff
@@ -70,6 +70,13 @@ grid=[[0, 1, 3, 4],
       [2, 2, 3, 4]]
 ```
 
+### Stacking components
```
antonymilne, line 73: Are we using the terms "horizontal" and "vertical" the right way round here? To me it's kind of ambiguous which way round the definition is, because as a non-expert it sounds like horizontal stacking could mean either way round:
1. components are horizontal and stacked on top of each other (like you say here)
2. components are squashed together into one horizontal row (the opposite thing)

A quick google didn't make it clear to me which was the correct/most common definition here, so would be great if you could double check!

antonymilne (approved):

vizro-core/docs/pages/user_guides/layouts.md
huong-li-nguyen: Yeah, I see. What if we stick with the flexbox description? It's not completely ambiguous free, but then we stick with one known syntax: https://byteiota.com/flexbox-conatiner/

I could also add an image here to make it clearer and then we call it
- stacked in one row (flex-direction row means that they are stacked next to each other or vertically)
- stacked in one column (flex-direction column means when they are stacked on top of each other or horizontally)

vizro-core/docs/pages/user_guides/container.md
huong-li-nguyen: I just realised that this phrasing is basically the contrary of what flexbox does. If they are stacked below each other, they call it "stacked in one column", if they are stacked side by side, they call it "stacked in one row".

Maybe we use that instead?

vizro-core/docs/pages/user_guides/layouts.md
huong-li-nguyen: When I looked at infographics - the other way around seemed more common: https://www.c-sharpcorner.com/article/stack-and-grid-layouts-in-xamarin-forms/

**So 2) components are squashed together into one horizontal row (the opposite thing)**

huong-li-nguyen: I am for sure no designer, but I just quickly made this in Miro and we can add it here:

Screenshot 2024-01-31 at 10 25 38

antonymilne: That is also the way round that dbc does it (look at the section on "Stacking objects"): https://dash-bootstrap-components.opensource.faculty.ai/docs/components/layout/. So indeed it sounds like we should do it the opposite way round from how you currently have it.

I think a couple of pictures would definitely make this clearer 😅 We already have the relevant examples in https://vizro.readthedocs.io/en/stable/pages/user_guides/layouts/?h=layout#using-custom-layout-examples (which I guess is largely there for legacy reasons, like the `one_left_two_right` stuff), so we could just copy them from that table and label them using the stacking terminology.

vizro-core/docs/pages/user_guides/container.md
antonymilne: Yeah, that's much clearer I think 👍 To align with our new convention on the other page I'd suggest using the words horizontal and vertical as well as explaining non-ambiguously what it means like you do:
```suggestion
        1. Note that the `Page.layout` argument is not specified here and will therefore default to `[[0], [1]]`, meaning the containers will be **vertically stacked** down the page in one column.
        2. **Horizontally stack** the components side-by-side inside this `Container` in one row.
```

vizro-core/docs/pages/user_guides/layouts.md
huong-li-nguyen: Haha, beat you on this - see image above :) I'll add it in here then 👍

antonymilne: > I am for sure no designer, but I just quickly made this in Miro and we can add it here:
> 
> Screenshot 2024-01-31 at 10 25 38

This is quite beautiful I think and does the job perfectly 👌 I promote you to lead designer.

PR #289, opened 2024-01-25
[Feat] Implement AgGrid model

maxschulz-COL:
## Description
This PR is an iteration of discussion in https://github.com/mckinsey/vizro/pull/260. Here we decided to implement one model per `Callable` type.

- Example Dashboard contains main features
- Formatting of Data inspired by @AnnMarieW and the new features in AGGrid `31.0.0`
- no styling applied
- no tests
- no docs

Styling, docs and tests will be in separate PRs.

### Notable activated features
- Columns are resizeable
- Columns are sortable
- Columns are filterable (with an apply and reset button)
- Activated pagination
- Deactivated Row animation

### Cell formatting
- Defined 4 additional data types for easy column formatting - will be explained in the docs: `dollar`, `euro`, `percent`, `numeric` - see third page of dev example and Screenshot
- Filters adjust to cell content type: numeric, date, or string/other
- (to be discussed) Auto-format any detected `pd.datetime` column into a valid `dateString` ([required by AGGrid](https://dash.plotly.com/dash-ag-grid/date-filters#example:-date-filter) to function for dates out of the box)

## Screenshot
image

## Notice


vizro-core/examples/_dev/app.py
maxschulz-COL, line 1: Can be reverted after PR

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -108,40 +83,16 @@ def _get_parent_vizro_model(_underlying_callable_object_id: str) -> VizroBaseMod
...
     for ctd_filter_interaction in ctds_filter_interaction:
-        if "clickData" in ctd_filter_interaction:
-            data_frame = _apply_graph_filter_interaction(
-                data_frame=data_frame, target=target, ctd_filter_interaction=ctd_filter_interaction
-            )
-
-        if "active_cell" in ctd_filter_interaction and "derived_viewport_data" in ctd_filter_interaction:
-            data_frame = _apply_table_filter_interaction(
-                data_frame=data_frame, target=target, ctd_filter_interaction=ctd_filter_interaction
+        if "modelID" in ctd_filter_interaction:
```
petar-qb, line 90: Do we need this statement (`if "modelID" in ctd_filter_interaction:`) at all?

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -71,6 +75,39 @@ def __getitem__(self, arg_name: str):
...
+        customdata = ctd_click_data["value"]["points"][0]["customdata"]
+
+        for action in source_graph_actions:
+            if action.function._function.__name__ != "filter_interaction" or target not in action.function["targets"]:
+                continue
+            for custom_data_idx, column in enumerate(custom_data_columns):
+                data_frame = data_frame[data_frame[column].isin([customdata[custom_data_idx]])]
+
+        return data_frame
+
```
petar-qb, line 110: It looks we decided to go with the "Inserting 'actions info' and dash dependencies as a method of the model." approach. I would really like if you can update me, what happened with other approaches we were considering, and why we finally decided to go with this one.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -71,6 +75,39 @@ def __getitem__(self, arg_name: str):
...
+        customdata = ctd_click_data["value"]["points"][0]["customdata"]
+
+        for action in source_graph_actions:
+            if action.function._function.__name__ != "filter_interaction" or target not in action.function["targets"]:
+                continue
+            for custom_data_idx, column in enumerate(custom_data_columns):
+                data_frame = data_frame[data_frame[column].isin([customdata[custom_data_idx]])]
+
+        return data_frame
```
petar-qb, line 109: It looks like `_filter_interaction` method is pretty similar in Graph/Table/Grid model. What do you think about the idea of implementing `_extract_filtering_info` method per model instead (e.g. it can just return real selected `value` and maybe the `column` that has to be filtered), and conduct filtering inside the `_actions_utils._apply_filter_interaction`?

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -71,6 +75,39 @@ def __getitem__(self, arg_name: str):
             return self.type
         return self.figure[arg_name]
 
+    # Interaction methods
+    def _get_figure_interaction_input(self) -> Dict[str, State]:
+        """Required properties when using pre-defined `filter_interaction`."""
+        return {
+            "clickData": State(component_id=self.id, component_property="clickData"),
+            "modelID": State(component_id=self.id, component_property="id"),  # required, to determine triggered model
+        }
```
petar-qb, line 84: I know that solving "storing model-action info" is the biggest open-question we currently have in the Vizro backend. 

I'm just wondering should we implement `_get_action_input` and `_get_action_output` for every other model?

petar-qb (commented):
I left just a few general comments I'm interested in and I can review this PR in detail maybe tomorrow 😄.

vizro-core/src/vizro/actions/_actions_utils.py
maxschulz-COL: This was supposed to make the code more stable in case of custom components. Imagine someone making a custom model that doesn't require filtering functionality, so he chooses not to even create those callback required objects, in this case this function then does not fail. As far as I am aware the `_get_filtered_data` function is always called, so this would be necessary

vizro-core/src/vizro/models/_components/graph.py
maxschulz-COL: As mentioned briefly on a call:
- having different models is final for at least this case
- placing the actions info here is not, although it could be our final approach at least for the case of Callables inserted into models

vizro-core/src/vizro/actions/_actions_utils.py
petar-qb: `ctds_filter_interaction` contains info only for the components that have defined `filter_interaction`. Actually, only for components that: `hasattr(triggered_model, "_get_figure_interaction_input")`. See changes inside the function: `_get_inputs_of_figure_interactions`. 

However, we can leave this statement (`if "modelID" in ctd_filter_interaction:`), since it will probably be refactored in the actions V2 soon.

vizro-core/src/vizro/models/_components/graph.py
maxschulz-COL: That is an interesting idea! I think we should definitely discuss it. I guess the pros are:
- simplified and shortened code in the respective models
- avoiding repretition

The cons I can think of are:
- potentially wrong abstraction level given that e.g. the graph currently filters multiple columns, which the table/grid don't
- like this we are giving the user less control over the filtering

Open to both really :)

vizro-core/src/vizro/actions/_actions_utils.py
maxschulz-COL: Hm I see what you mean. 2 questions:
- do we prefer to fail if a user overwrites this method without adding the `modelID`
- or do we prefer the code not to fail, but the filtering also not to happen

As far as I understand, the answer to the above determines if we should leave the statement in or not. We should definitely not leave it in just because things are getting refactored soon

vizro-core/src/vizro/models/_components/graph.py
maxschulz-COL: Do you mean that for example the `Dropdown` will tell the code that it's input is the `State(value....)` etc?

petar-qb: Exactly, let's continue this "actions_info" conversation in the https://github.com/mckinsey/vizro/pull/289#discussion_r1495590983.

vizro-core/src/vizro/actions/_actions_utils.py
petar-qb: If we can't perform filter_interaction without `modelID` part of the input, it means we need to raise an exception or some kind of warning to the user. It really all depends again on the fact how do we deal with "actions_info", what is the part of the "actions_info" configuration, and what is the part of the "_actions_utils" functions. Does it makes sense to continue this conversation in the https://github.com/mckinsey/vizro/pull/289#discussion_r1495590983?

vizro-core/src/vizro/models/_components/graph.py
petar-qb: I agree that it is not easy to make this kind of abstraction in general, hence I call it - "actions_info" - the biggest open question we currently have.

If we make this kind of abstraction properly, we could say that the "filter_interaction" is just the same as the "_filter" action, which sounds perfectly correct to me. 

From top of my head: All "actions_info" we need to set for every Vizro model (e.g. Button, Card, Dropdown, Graph, Grid...) is:
- `_get_input_components` - e.g. {model_id: State(...), derived_viewport_data: State(...), active_cell: State(...)} OR {model_id: State(...), value: State(...)};
- `_extract_value_from_input` - some `function`,
- `_get_output_components` - e.g. {model_id: Output(...)}  

It means that every Vizro model will have implemented these methods in a form of (`class methods`, or some `actions_info: Dict[str, any]` argument). Also, these methods/arguments could be overwritten in case of custom component creation.

vizro-core/examples/_dev/app.py
antonymilne: No need to revert - this can be left however you like.

vizro-core/src/vizro/tables/dash_aggrid.py
```diff
@@ -0,0 +1,78 @@
+"""Module containing the standard implementation of `dash-ag-grid.AgGrid`."""
```
antonymilne, line 1:
```suggestion
"""Module containing the standard implementation of `dash_ag_grid.AgGrid`."""
```

vizro-core/src/vizro/tables/dash_aggrid.py
```diff
@@ -0,0 +1,78 @@
...
+        "baseDataType": "number",
+        "extendsDataType": "number",
+        "valueFormatter": {"function": "d3.format(',.1f')(params.value)"},
+    },
+}
+
+
+@capture("aggrid")
+def dash_ag_grid(data_frame=None, **kwargs):
+    """Implementation of `dash-ag-grid.AgGrid` with sensible defaults."""
```
antonymilne, line 51:
```suggestion
    """Implementation of `dash_ag_grid.AgGrid` with sensible defaults."""
```

vizro-core/src/vizro/tables/dash_aggrid.py
```diff
@@ -0,0 +1,78 @@
...
+            "sortable": True,
+            "filter": True,
+            "filterParams": {
+                "buttons": ["apply", "reset"],
+                "closeOnApply": True,
+            },
+        },
+        "dashGridOptions": {
+            "dataTypeDefinitions": DATA_TYPE_DEFINITIONS,
+            "animateRows": False,
```
antonymilne, line 73: Why don't we want this? Animated rows sounds cool 😀

vizro-core/src/vizro/tables/dash_aggrid.py
```diff
@@ -0,0 +1,78 @@
...
+    "numeric": {
+        "baseDataType": "number",
+        "extendsDataType": "number",
+        "valueFormatter": {"function": "d3.format(',.1f')(params.value)"},
+    },
+}
+
+
+@capture("aggrid")
+def dash_ag_grid(data_frame=None, **kwargs):
```
antonymilne, line 50:
```suggestion
def dash_ag_grid(data_frame, **kwargs):
```

Unless I missed something, this is not optional.

vizro-core/src/vizro/tables/dash_aggrid.py
```diff
@@ -0,0 +1,78 @@
...
+            },
+        },
+        "dashGridOptions": {
+            "dataTypeDefinitions": DATA_TYPE_DEFINITIONS,
+            "animateRows": False,
+            "pagination": True,
+        },
+    }
+    kwargs = _set_defaults_nested(kwargs, defaults)
+    return dag.AgGrid(**kwargs)
```
antonymilne, line 78: Just to check in case I missed something: AG Grid doesn't have a `data` argument or similar? It's all done through `rowData` instead?

vizro-core/src/vizro/tables/dash_aggrid.py
```diff
@@ -0,0 +1,78 @@
+"""Module containing the standard implementation of `dash-ag-grid.AgGrid`."""
+
+import dash_ag_grid as dag
+import pandas as pd
+
+from vizro.models.types import capture
+from vizro.tables._utils import _set_defaults_nested
+
+FORMAT_CURRENCY_EU = """d3.formatLocale({
```
antonymilne, line 9: This doesn't seem to be used?

vizro-core/src/vizro/models/types.py
```diff
@@ -239,8 +242,8 @@ class capture:
 
     """
 
-    def __init__(self, mode: Literal["graph", "action", "table"]):
-        """Instantiates the decorator to capture a function call. Valid modes are "graph", "table" and "action"."""
+    def __init__(self, mode: Literal["graph", "action", "table", "aggrid"]):
```
antonymilne, line 245:
```suggestion
    def __init__(self, mode: Literal["graph", "action", "table", "ag_grid"]):
```

I think we should have a `_` here (change applies in many places).

vizro-core/src/vizro/models/types.py
```diff
@@ -309,9 +312,26 @@ def wrapped(*args, **kwargs):
                     raise ValueError(f"{func.__name__} must supply a value to data_frame argument.") from exc
                 return captured_callable
 
+            return wrapped
+        elif self._mode == "aggrid":
```
antonymilne, line 316: Is this identical to the `table` case? Can we put it in the same `elif self._mode in ["table", "ag_grid"]`?

vizro-core/src/vizro/models/_components/aggrid.py
```diff
@@ -0,0 +1,121 @@
...
+from vizro.models import Action, VizroBaseModel
+from vizro.models._action._actions_chain import _action_validator_factory
+from vizro.models._components._components_utils import _process_callable_data_frame
+from vizro.models._models_utils import _log_call
+from vizro.models.types import CapturedCallable
+
+logger = logging.getLogger(__name__)
+
+
+class AGGrid(VizroBaseModel):
```
antonymilne, line 24: I think we should probably go for `AgGrid` here to match Dash.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -71,6 +75,39 @@ def __getitem__(self, arg_name: str):
             return self.type
         return self.figure[arg_name]
 
+    # Interaction methods
+    def _get_figure_interaction_input(self) -> Dict[str, State]:
```
antonymilne, line 79: Can we make this a `@property` called `_figure_interaction_input`? Similarly in other models.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -71,6 +75,39 @@ def __getitem__(self, arg_name: str):
...
+        """Required properties when using pre-defined `filter_interaction`."""
+        return {
+            "clickData": State(component_id=self.id, component_property="clickData"),
+            "modelID": State(component_id=self.id, component_property="id"),  # required, to determine triggered model
+        }
+
+    def _filter_interaction(
+        self, data_frame: pd.DataFrame, target: str, ctd_filter_interaction: Dict[str, CallbackTriggerDict]
```
antonymilne, line 87:
```suggestion
        self, target: str, ctd_filter_interaction: Dict[str, CallbackTriggerDict]
```
Can we get rid of this and just access this model's data_frame with `data_manager._get_component_data(str(self.id))`?

Similarly if this model instance is always the `source_graph_id` then let's do that with `self.id` rather than going through `ctd_click_data["id"]`.

One of the advantages of having this code in the model is that we have direct access to `self` now, so if it's possible then we should use it wherever possible rather than passing the arguments in manually or extracting from Dash callback context.

The same comment also applies for the other models.

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -59,13 +60,47 @@ def __getitem__(self, arg_name: str):
...
     def pre_build(self):
         if self.actions:
             kwargs = self.figure._arguments.copy()
 
             # This workaround is needed because the underlying table object requires a data_frame
-            kwargs["data_frame"] = DataFrame()
+            kwargs["data_frame"] = pd.DataFrame()
 
             # The underlying table object is pre-built, so we can fetch its ID.
             underlying_table_object = self.figure._function(**kwargs)
```
antonymilne, line 106: Not related to this PR, but @petar-qb do you remember we had a conversation a while ago about trying to simplify all this stuff? I think we should try and do that still, especially now it's duplicated for `AGGrid` and `Table`.

vizro-core/src/vizro/models/_components/aggrid.py
```diff
@@ -0,0 +1,121 @@
...
+    """Wrapper for `dash-ag-grid.AgGrid` to visualize grids in dashboard.
```
antonymilne, line 25:
```suggestion
    """Wrapper for `dash_ag_grid.AgGrid` to visualize grids in dashboard.
```

vizro-core/src/vizro/models/_components/aggrid.py
```diff
@@ -0,0 +1,121 @@
...
+
+            self._callable_object_id = underlying_aggrid_object.id
+
+    def build(self):
+        return dcc.Loading(
+            html.Div(
+                [
+                    html.H3(self.title, className="table-title") if self.title else None,
+                    html.Div(
+                        dash_table.DataTable(**({"id": self._callable_object_id} if self.actions else {})), id=self.id
```
antonymilne, line 113: I know this doesn't really do anything because it gets overwritten on page load, but it's confusing using `DataTable` here rather than `AGGrid`.

vizro-core/src/vizro/models/_components/graph.py
antonymilne: I would suggest we leave it as is for now and come back to it later. This is still a big open question indeed, and I think "Duplication is cheaper than the wrong abstraction" applies in this case.

Generally I think I agree with @petar-qb's idea, but there's still a lot of moving pieces here so hard to tell right now what the right solution is. So for now I think just focus on:
* getting these models working correctly
* initial actions v2 work to make those action `CapturedCallable` classes

And after that we can figure out what to do with all this actions info stuff.

Let's unresolve this conversation after merging though so we can easily refer back to it.

antonymilne (commented):
Generally looks good, although I didn't actually play around with using it yet 👍 Thanks for all the work on different versions of this, I know it's not been an easy task!

I left some initial comments so you don't wait too long for a review, but I don't think I will have much more to say after this round anyway, so long as we're confident it works ok.

What happens if a user does `AgGrid(figure=dash_data_table(...)` or `Table(figure=dash_ag_grid(...)`? I guess it renders and all works well apart from the filter interaction ?

vizro-core/src/vizro/models/_components/table.py
petar-qb: I really can't remember what we talked about exactly..
At least we can extract the implementation of this method and reuse it in the vm.AgGrid and vm.Table.

Also, I think we should get rid of "`if self.actions:`" statement in `pre_build` method and always conduct fetching of underlying id. In that way we can avoid raising a dash exception (something like: "Input of callback X doesn't exist."). This exception will be raised:
- if user provides an ID for the underlying table, 
- if the underlying table is an input (dash.State) of the custom action,
- if there is no "`actions`" defined for the vm.Table / vm.AgGrid

We should double check this assumption and if it's true we should also change one line in the `build` method. 
From:
```
                        dash_table.DataTable(**({"id": self._callable_object_id} if self.actions else {})), id=self.id
```
To:
```
                        dash_table.DataTable(id=self._callable_object_id), id=self.id
```

vizro-core/src/vizro/tables/dash_aggrid.py
AnnMarieW: Haha - that's my personal preference - I find the animation super annoying :face_with_spiral_eyes: 
You can give it a try -- the people at AG Grid think it makes for a "better user experience" :woman_shrugging:

vizro-core/src/vizro/tables/dash_aggrid.py
AnnMarieW: `rowData` is equivalent to  the `data` prop in the dash DataTable

vizro-core/src/vizro/models/_components/table.py
antonymilne: My idea was that we can remove basically all of this, like so:
```
Index: vizro-core/src/vizro/models/_components/table.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/vizro-core/src/vizro/models/_components/table.py b/vizro-core/src/vizro/models/_components/table.py
--- a/vizro-core/src/vizro/models/_components/table.py	(revision 4ea5b5f35cf9db4217537038d4f36ed4e8fc3ee4)
+++ b/vizro-core/src/vizro/models/_components/table.py	(date 1705597769502)
@@ -48,7 +48,9 @@
     def __call__(self, **kwargs):
         kwargs.setdefault("data_frame", data_manager._get_component_data(self.id))
-        return self.figure(**kwargs)
+        figure.id = self._callable_object_id
+        return figure
 
     def __getitem__(self, arg_name: str):
@@ -60,31 +62,14 @@
 
     @_log_call
     def pre_build(self):
-        if self.actions:
-            kwargs = self.figure._arguments.copy()
-
-            # This workaround is needed because the underlying table object requires a data_frame
-            kwargs["data_frame"] = DataFrame()
-
-            # The underlying table object is pre-built, so we can fetch its ID.
-            underlying_table_object = self.figure._function(**kwargs)
-
-            if not hasattr(underlying_table_object, "id"):
-                raise ValueError(
-                    "Underlying `Table` callable has no attribute 'id'. To enable actions triggered by the `Table`"
-                    " a valid 'id' has to be provided to the `Table` callable."
-                )
-
-            self._callable_object_id = underlying_table_object.id
+        self._callable_object_id = self.id + "***"
 
     def build(self):
         return dcc.Loading(
             html.Div(
                 [
-                    html.Div(
-                        dash_table.DataTable(**({"id": self._callable_object_id} if self.actions else {})), id=self.id
-                    ),
+                    html.Div(dash_table.DataTable(id=self._callable_object_id), id=self.id),
                 ],
                 className="table-container",
                 id=f"{self.id}_outer",
```

Definitely not for this PR, but just wanted to make sure we didn't forget it.

petar-qb: Ah yes, thanks for remembering 😄. This change really reduces the complexity of the table model ⚡.

huong-li-nguyen (commented):
I won't review this in detail as others already do :) 

As long as I remember - it would be amazing ⭐ if you could add the AG-grid example inside the **features dashboard** and add the yaml version of it (if applicable) as well. You can also replace the data table inside the **demo dashboard** with the ag-grid one already👍 

I'll leave that up to you though, whether you want to add it in this PR or your docs PR.

vizro-core/src/vizro/actions/_actions_utils.py
maxschulz-COL: ok dokes!

vizro-core/src/vizro/tables/dash_aggrid.py
maxschulz-COL: I was 50/50, it seems very snappy without, but somewhat fancy with it. Let's leave them out for now

vizro-core/src/vizro/tables/dash_aggrid.py
maxschulz-COL: Yes it is, it get's used in the `DATA_TYPE_DEFINITIONS` just below, but for clarity I separated the two

vizro-core/src/vizro/models/_components/aggrid.py
maxschulz-COL: I find it doesn't look so nice, and the original seems more AG Grid, but let's do it to align with dash

vizro-core/src/vizro/models/_components/graph.py
maxschulz-COL: I think this is not easily possible atm. 

We are already getting the data with `data_frame = data_manager._get_component_data(target)` in line 177 of `_action_utils.py`. This DF is then consecutively passed through filters.

If we were to load data afresh, we would skip the previous filters.

Besides, the triggered model provides the filtering information, the target component contains the to be filtered DF, so this would be the wrong DF anyway

vizro-core/src/vizro/models/_components/table.py
maxschulz-COL: Ok will keep this un-resolved but will not implement in this PR

vizro-core/src/vizro/models/_components/aggrid.py
maxschulz-COL: Going to see if we can remove this, but if I just exchange this for the `dag.AgGrid` the AG Grid doesn't render properly anymore. Going to investigate why

petar-qb: If you go with this solution: https://github.com/mckinsey/vizro/pull/289#discussion_r1497465481, does it work then? So just removing the `if self.actions:` line from `pre_build` method, and in the build method set:
`dash_table.DataTable(id=self._callable_object_id), id=self.id` line for Table, and
`dag.AgGrid(id=self._callable_object_id), id=self.id` line for AgGrid.

petar-qb: I just tested it with `dag.AgGrid(id=self._callable_object_id), id=self.id` approach, and it seems like backend properly returns data, but somehow it's not rendered on the UI.

petar-qb: @maxschulz-COL 
Everything seems to work as expected when "pagination=True" is defined 🤔.

After I figured out the following configuration works:
`dash_ag_grid(data_frame=pd.DataFrame(), id=self._callable_object_id)()`

I started by removing the default attributes and found that the minimum configuration required to make it work was:
`dag.AgGrid(id=self._callable_object_id, **{"dashGridOptions": {"pagination": True}}),`

petar-qb: Now I pulled the latest PR changes (in this commit `"pagination"=True` is removed from the `dash_ag_grid` default configuration) and it starts to work even with the `dag.AgGrid(id=self._callable_object_id)` configuration. 

It seems like it doesn't work when the object with the same ID initially doesn't have pagination included, and after the on_page_load mechanism it has. Somehow it makes sense to me now since `"pagination"=True` probably creates some mechanism per component (binds it with the component ID). If the pagination is additionally turned on for the already existing component then this backend mechanism isn't created, so it results with the empty data on the screen.

maxschulz-COL: Yes, it does make sense - what a funny coincidence. Well since we do not want pagination anymore, all good then :)

petar-qb: Exactly 😄

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
```diff
@@ -47,20 +47,8 @@ def _get_inputs_of_figure_interactions(
...
-                    "derived_viewport_data": State(
-                        component_id=triggered_model._callable_object_id, component_property="derived_viewport_data"
-                    ),
-                }
-            )
-        else:
-            inputs.append({"clickData": State(component_id=triggered_model.id, component_property="clickData")})
-
+        if hasattr(triggered_model, "_figure_interaction_input"):
+            inputs.append(triggered_model._figure_interaction_input)
```
petar-qb, line 51: Since this part of the code is executed in the build app phase and only for the models that have the filter_interaction actions attached, I would raise an exception here:
- if `_figure_interaction_input ` model method isn't defined. 

I would also raise an exception inside the `_apply_filter_interaction` function:
- if "modelID" doesn't exist inside the ctd_filter_interaction, or - (There is a change we will remove this limitation soon)
- if `_filter_interaction` model method isn't defined


This move covers the case when users create customTableComponent and attach filter_interaction action on it. I know it's almost impossible to expect someone will do that, but we should probably cover these cases as well. 
(I'm also fine with skipping solving this case if it's too much work right now)

petar-qb (approved):
It really looks good now. Thanks for all you hard work on this PR Max 🚀.
This PR shows some limitations we're going to overcome very soon 😄.

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
maxschulz-COL: Ok sounds good! I removed the if clause as it's not necessary anymore and added the relevant Errors

vizro-core/changelog.d/20240222_094020_maximilian_schulz_aggrid_separate_models.md
```diff
@@ -0,0 +1,48 @@
+<!--
```
antonymilne, line 1: Once the docs are written so we can link to them then this should definitely go in the Highlights sections.

vizro-core/src/vizro/models/_components/aggrid.py
antonymilne, line 1: This file should be called `ag_grid.py` for consistency.

vizro-core/src/vizro/models/_components/graph.py
antonymilne: Ah ok, I see. Please could you add a quick comment saying what `data_frame` refers to here?

vizro-core/src/vizro/tables/dash_aggrid.py
antonymilne, line 1: Let's call this `dash_ag_grid.py`.

In fact, unless I missed something, really both the `dash_*` files in this directory should be `_dash_*`, so that would be `_dash_ag_grid.py`. This is actually not just me being fussy but is quite important to change since the nesting here can confuse IDE imports - we explicitly _don't_ want people doing `from vizro.tables.dash_ag_grid import dash_ag_grid`. We just want `from vizro.tables import dash_ag_grid`.

antonymilne (approved):
💯 Amazing work! I know this has been a huge and lengthy task, thank you for all the work on it.

Still curious about the answer to this:
> What happens if a user does `AgGrid(figure=dash_data_table(...)` or `Table(figure=dash_ag_grid(...)`? I guess it renders and all works well apart from the filter interaction?

It would be great if you could demo this in a team learning session some time, so we can see all the crazy things that AG Grid is capable of. The docs examples look great -  I don't have time to try them out myself right now but would love to see them demoed.

vizro-core/tests/unit/vizro/models/_components/test_aggrid.py
```diff
@@ -0,0 +1,156 @@
...
+def dash_ag_grid_with_arguments():
+    return dash_ag_grid(data_frame=px.data.gapminder(), defaultColDef={"resizable": False, "sortable": False})
+
+
+@pytest.fixture
+def dash_ag_grid_with_str_dataframe():
+    return dash_ag_grid(data_frame="gapminder")
+
+
+class TestDunderMethodsAgGrid:
```
antonymilne, line 30: I think all these `test_create` things probably belong in a class that's not `TestDunderMethods` but instead `TestAgGridInstantation` like we do for other models.

vizro-core/tests/unit/vizro/models/_components/test_aggrid.py
```diff
@@ -0,0 +1,156 @@
...
+    def test_create_graph_mandatory_only(self, standard_ag_grid):
+        ag_grid = vm.AgGrid(figure=standard_ag_grid)
+
+        assert hasattr(ag_grid, "id")
+        assert ag_grid.type == "ag_grid"
+        assert ag_grid.figure == standard_ag_grid
+        assert ag_grid.actions == []
+
+    @pytest.mark.parametrize("id", ["id_1", "id_2"])
```
antonymilne, line 39: Why is this parameterised? The two cases are functionality identically aren't they?

vizro-core/tests/unit/vizro/models/_components/test_aggrid.py
```diff
@@ -0,0 +1,156 @@
...
+    def test_failed_ag_grid_with_no_captured_callable(self, standard_go_chart):
+        with pytest.raises(ValidationError, match="must provide a valid CapturedCallable object"):
+            vm.AgGrid(figure=standard_go_chart)
+
+    @pytest.mark.xfail(reason="This test is failing as we are not yet detecting different types of captured callables")
+    def test_failed_ag_grid_with_wrong_captured_callable(self, standard_px_chart):
+        with pytest.raises(ValidationError, match="must provide a valid ag_grid function vm.AgGrid"):
+            vm.AgGrid(figure=standard_px_chart)
+
+    def test_getitem_known_args(self, dash_ag_grid_with_arguments):
```
antonymilne, line 60: It's only this and the next test in this class that are testing the dunder methods as far as I can tell; the rest are instantiation tests.

vizro-core/tests/unit/vizro/models/_components/test_aggrid.py
```diff
@@ -0,0 +1,156 @@
...
+        with pytest.raises(KeyError, match="'data_frame'"):
+            ag_grid_with_str_df.figure["data_frame"]
+
+
+class TestPreBuildAgGrid:
+    def test_pre_build_no_actions_no_underlying_ag_grid_id(self, standard_ag_grid):
+        ag_grid = vm.AgGrid(id="text_ag_grid", figure=standard_ag_grid)
+        ag_grid.pre_build()
+
+        assert hasattr(ag_grid, "_callable_object_id") is False
```
antonymilne, line 104:
```suggestion
        assert not hasattr(ag_grid, "_callable_object_id")
```

vizro-core/tests/unit/vizro/tables/test_dash_ag_grid.py
```diff
@@ -0,0 +1,24 @@
...
+        "int": [4, 5, 6],
+        "float": [7.3, 8.2, 9.1],
+        "date": pd.to_datetime(["2021/01/01", "2021/01/02", "2021/01/03"]),
+    }
+)
+
+
+class TestDashAgGrid:
+    def test_dash_ag_grid(self):
+        grid = dash_ag_grid(data_frame=data)()
```
antonymilne, line 16: Can we do this using `assert_component_equals` instead?
```
assert_component_equals(grid, dag.AgGrid(columnDefs=..., ...), keys_to_ignore=...)
```

vizro-core/tests/unit/vizro/models/_components/test_aggrid.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+class TestAttributesAgGrid:
+    def test_ag_grid_filter_interaction_attributes(self, ag_grid_with_id):
+        ag_grid = vm.AgGrid(figure=ag_grid_with_id, title="Gapminder", actions=[])
+        assert hasattr(ag_grid, "_filter_interaction")
+        ag_grid.pre_build()
+        assert hasattr(ag_grid, "_filter_interaction_input")
+        assert "modelID" in ag_grid._filter_interaction_input
+
```
antonymilne, line 83:
```suggestion
```

This is too much of an implementation detail to test. I assume that if these didn't exist then some other actions tests would fail? If so then let's remove this one and the similar ones in the graph and table tests.

antonymilne (approved):
A few comments on the tests to look at, but it's minor stuff so feel free to merge when you've dealt with them 🙂

vizro-core/changelog.d/20240222_094020_maximilian_schulz_aggrid_separate_models.md
maxschulz-COL: I would add it in the docs PR - which would mark the official release I guess

vizro-core/tests/unit/vizro/models/_components/test_aggrid.py
maxschulz-COL: It's a bit tricky - I created them for the precise reason that other tests are not failing (at least in the current design)!

The reason for this is that we mock the callback context when "running" actions - so we fake the user clicking on something. If we did not do that the testing of actions would be a lot more complicated

You are correct in that the `assert hasattr(ag_grid, "_filter_interaction_input")` is unncessary in that sense, but I added it for completeness.

I would propose to remove that one then, and leave a comment why this is important

antonymilne: Ahhh ok, if the others tests aren't failing then we should definitely leave this here. I'm fine to leave all the `hasattr` there.

The fact that the actions tests _don't_ fail if you remove this property indicates a deficiency with our testing of actions, but given this implementation may well change anyway let's solve that another time as part of the actions v2 work @petar-qb.

vizro-core/tests/unit/vizro/models/_components/test_aggrid.py
maxschulz-COL: Did it for both AgGrid and Table

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -0,0 +1,121 @@
...
+        return self.figure[arg_name]
+
+    # Interaction methods
+    @property
+    def _filter_interaction_input(self):
+        """Required properties when using pre-defined `filter_interaction`."""
+        return {
+            "cellClicked": State(component_id=self._callable_object_id, component_property="cellClicked"),
+            "modelID": State(component_id=self.id, component_property="id"),  # required, to determine triggered model
+        }
```
petar-qb, line 70: Do we prefer comments above the line which it points to?
Same for the `vm.Table`.

vizro-core/tests/unit/vizro/models/_components/test_aggrid.py
maxschulz-COL: This was done the same on the table - is suppose it is to check that there is no hidden ID, and that it gets dynamically set, but it's a bit silly indeed

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -0,0 +1,121 @@
...
+
+            self._callable_object_id = underlying_aggrid_object.id
+
+    def build(self):
+        return dcc.Loading(
+            html.Div(
+                [
+                    html.H3(self.title, className="table-title") if self.title else None,
+                    html.Div(dag.AgGrid(**({"id": self._callable_object_id} if self.actions else {})), id=self.id),
+                ],
```
petar-qb, line 116: Max, can we fix one bug in this PR (and then I can solve this one [from the comment](https://github.com/mckinsey/vizro/pull/289/files#r1497888111))? The bug happens when no action is associated with the AgGrid/Table object, but this AgGrid/Table model is used as the input of some custom action. The bug says something like: "There is no component with the ID: {ID}, but it's used as dash.Input in the callback..."


All we need to avoid this bug to occur is:
1.  to change this line:
`html.Div(dag.AgGrid(**({"id": self._callable_object_id} if self.actions else {})), id=self.id),`
to
`html.Div(dag.AgGrid(id=self._callable_object_id), id=self.id),`

2. and to remove `if self.actions:` statement completely from the pre_build. 

We should apply this change in vm.Table as well.

vizro-core/tests/unit/vizro/actions/conftest.py
```diff
@@ -59,7 +59,9 @@ def managers_one_page_two_graphs_one_button(box_chart, scatter_chart):
 
 
 @pytest.fixture
-def managers_one_page_two_graphs_one_table_one_button(box_chart, scatter_chart, dash_data_table_with_id):
+def managers_one_page_two_graphs_one_table_one_aggrid_one_button(
+    box_chart, scatter_chart, dash_data_table_with_id, ag_grid_with_id
+):
```
petar-qb, line 64: Could you adjust the docstring below?

vizro-core/tests/unit/vizro/actions/test_filter_interaction_action.py
```diff
@@ -235,9 +261,38 @@ def test_table_filter_interaction_with_two_targets(
...
+            ((None, None, "Algeria"), (None, "Algeria"), (None, "Algeria")),
+            ((None, None, "Albania"), (None, "Albania"), (None, "Albania")),
+            ((None, None, "Argentina"), (None, "Argentina"), (None, "Argentina")),
+        ],
+        indirect=True,
+    )
+    def test_aggrid_filter_interaction_with_two_targets(
+        self, ctx_filter_interaction, target_scatter_filtered_continent, target_box_filtered_continent
+    ):
+        # to not overcrowd these tests with duplication, we use one general case here for the AG Grid
```
petar-qb, line 273:
```suggestion
        # To not overcrowd these tests with duplication, we use one general case here for the AG Grid
```

vizro-core/tests/unit/vizro/models/_components/test_aggrid.py
maxschulz-COL: Yes, so just FYI when we come back to this, so that we can remember: this is part of the build method of the `Action` model, so it will only get triggered then.

Currently we do not build the `Action` model in order to test the functionality - presumably because we fake the ctx

vizro-core/tests/unit/vizro/tables/test_dash_ag_grid.py
```diff
@@ -0,0 +1,33 @@
...
+        assert_component_equal(
+            grid,
+            dag.AgGrid(
+                columnDefs=[{"field": "cat"}, {"field": "int"}, {"field": "float"}, {"field": "date"}],
+                rowData=[
+                    {"cat": "a", "int": 4, "float": 7.3, "date": "2021-01-01"},
+                    {"cat": "b", "int": 5, "float": 8.2, "date": "2021-01-02"},
+                    {"cat": "c", "int": 6, "float": 9.1, "date": "2021-01-03"},
+                ],
+                # defaultColDef={"resizable": True, "sortable": True},
```
petar-qb, line 28: Should we remove this comments?

petar-qb (approved):
I've already approved this PR, but here's the final review. Everything is on its place except a few minor things I commented below.

vizro-core/src/vizro/models/_components/ag_grid.py
maxschulz-COL: Not sure - do you have a preference? If it fits inline I thought why not, but not sure if there is best practice on that

vizro-core/src/vizro/models/_components/ag_grid.py
maxschulz-COL: Ok done - will push in a sec. Can you then check if I understood this correctly?

petar-qb:
The initial bug is solved with the latest changes https://github.com/mckinsey/vizro/pull/289/commits/ec85e0c0dfcd5693d576e5fcc531f8bef88c3ca2 you pushed. Awesome work Max! 🤖

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -109,12 +108,17 @@ def pre_build(self):
             )
 
     def build(self):
-        dash_ag_grid_conf = {"id": self._callable_object_id} if hasattr(self, "_callable_object_id") else {}
+        # The pagination setting (and potentially others) only work when the initially built AgGrid has the same
+        # setting as the object that is built on-page-load and rendered finally.
+        dash_ag_grid_conf = self.figure._arguments.copy()
+        dash_ag_grid_conf["data_frame"] = pd.DataFrame()
+        if hasattr(self, "_callable_object_id"):
+            dash_ag_grid_conf["id"] = self._callable_object_id
```
petar-qb, line 116: This is very clean solution! I just think that the following two lines:
```
        if hasattr(self, "_callable_object_id"):
            dash_ag_grid_conf["id"] = self._callable_object_id
```
 don't make any difference as the `id` is already copied to the `dash_ag_grid_conf` in the previous lines. 

Also, it seems like it makes sense to push similar approach for the vm.Table - dash_data_table solution.

PR #265, opened 2024-01-11
Add unit tests for `Container`

huong-li-nguyen:
## Description
- Added unit tests for instantiation and build of `Container`
- Added unit tests for all helper functions in `_models_utils`
- Refactored shared tests between `Container` and `Page`

## Notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

  - I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
  - I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
  - I certify that the use of this contribution as authorized by the Apache 2.0 license does not violate the intellectual property rights of anyone else.
  - I have not referenced individuals, products or companies in any commits, directly or indirectly.
  - I have not added data or restricted code in any commits, directly or indirectly.

PR #291, opened 2024-01-26
Add unit tests for `Tabs`

huong-li-nguyen:
## Description
- Add unit tests for instantiation and build of Tabs
- Add unit tests for Tabs and actions

## Notice


vizro-core/tests/unit/vizro/actions/_callback_mapping/test_get_action_callback_mapping.py
```diff
@@ -29,51 +29,54 @@ def export_data():
 
 
 @pytest.fixture
-def managers_one_page_four_controls_three_figures_filter_interaction(request, dash_data_table_with_id):
+def config_for_testing_all_components_with_actions(request, dash_data_table_with_id):
     """Instantiates managers with one page that contains four controls, two graphs and filter interaction."""
```
huong-li-nguyen, line 33: I saw that we've done the same for the unit tests for the container, so I've just replicated here. I assume the rule here is that we test for the most complex configuration within a page and this fixture is then being used in all unit tests for the actions. I was wondering if this sufficiently covers it? Or if we need to actually create different configs covering:
- Just top-level components
- Top-level containers
- Nested containers
- Tabs with containers etc.

Or is this an overkill and we always assume that if the components are visible/accessible for the actions in the most complex config (so tabs with nested containers and components inside), that it will also work for simpler configs? @petar-qb @antonymilne

petar-qb: Probably we don't need to add any Containers or Tabs in this test at all (I know I've already done it for Containers), and we should test model_manager helper functions separately. 
 
However, I'm also fine by leaving this test as you have already implemented it and we can simplify this test later (by removing Container and Tabs models from it), when we implement tests for model_manager helper functions. (as a part of the https://github.com/mckinsey/vizro/issues/111)

vizro-core/tests/unit/vizro/models/_components/test_tabs.py
```diff
@@ -0,0 +1,74 @@
...
+        assert isinstance(tabs.tabs[0], vm.Container) and isinstance(tabs.tabs[1], vm.Container)
+        assert tabs.id == "tab-id"
+        assert tabs.type == "tabs"
+
+    def test_mandatory_tabs_missing(self):
+        with pytest.raises(ValidationError, match="field required"):
+            vm.Tabs(id="tab-id")
+
+
+class TestTabsBuildMethod:
```
maxschulz-COL, line 37: I have been a bit out of the loop with the `assert_component_equal` but this seems quite cumbersome. We repeat the same statement 6 times - this does not seem super intuitive and also not so easy to read what the actual expected outcome of the build method should be.

maxschulz-COL (approved):
Looks good from my side

vizro-core/tests/unit/vizro/models/_components/test_tabs.py
petar-qb: Probably we should prepare an entire expected fixture and write the test as:
```
def test_tabs_build(self, containers, expected):
        result = vm.Tabs(id="tab-id", tabs=containers).build()
        assert_component_equal(result, expected)
```

petar-qb (approved):
Looks good! 

After considering a [comment](https://github.com/mckinsey/vizro/pull/291/files#r1469334948), feel free to merge it.

vizro-core/tests/unit/vizro/models/_components/test_tabs.py
huong-li-nguyen: This is mostly due to the fact that putting everything into one `assert_component_equal` wouldn't allow us to test the precision we need e.g. the issue becomes clear when you take a look at the children of this highly nested component tree. I only want to ignore the `children` in the `assert_component_equal` at the lowest level (so ignore the content of the low-level `component.build()`, but I do not want to ignore the children of Tab and Tabslist etc. as these are created in the build)
Splitting it up allows us to test the precision we need (so selectively testing the components/classNames that are actually defined in the `Tabs.build()`.
See this [comment thread](https://github.com/mckinsey/vizro/pull/265#discussion_r1466268165) for reference.

vizro-core/tests/unit/vizro/models/_components/test_tabs.py
```diff
@@ -0,0 +1,74 @@
...
+        vm.Container(id="container-1", title="Title-1", components=[vm.Button()]),
+        vm.Container(id="container-2", title="Title-2", components=[vm.Button()]),
+    ]
+
+
+class TestTabsInstantiation:
+    """Tests model instantiation and the validators run at that time."""
+
+    def test_create_tabs_mandatory_only(self, containers):
+        tabs = vm.Tabs(id="tab-id", tabs=containers)
```
antonymilne, line 27:
```suggestion
        tabs = vm.Tabs(id="tabs-id", tabs=containers)
```

vizro-core/tests/unit/vizro/models/_components/test_tabs.py
```diff
@@ -0,0 +1,74 @@
...
+        assert isinstance(tabs.tabs[0], vm.Container) and isinstance(tabs.tabs[1], vm.Container)
```
antonymilne, line 28: FYI in an ideal world this would be a bit stricter:
```suggestion
        assert all(isinstance(tab, vm.Container) for tab in tabs.tabs) and len(tabs.tabs) == 2
```

Since the current test would pass with `tabs.tabs = [vm.Container(), vm.Container(), something, 123, vm.Container()]` when it shouldn't.

This is the case in lots of our tests and I don't care much, so it's just FYI for your personal enlightenment really - don't bother changing it unless you care a lot.

vizro-core/tests/unit/vizro/models/_components/test_tabs.py
```diff
@@ -0,0 +1,74 @@
...
+        assert tabs.type == "tabs"
+
+    def test_mandatory_tabs_missing(self):
+        with pytest.raises(ValidationError, match="field required"):
+            vm.Tabs(id="tab-id")
+
+
+class TestTabsBuildMethod:
+    def test_tabs_build(self, containers):
+        result = vm.Tabs(id="tab-id", tabs=containers).build()
```
antonymilne, line 39:
```suggestion
        result = vm.Tabs(id="tabs-id", tabs=containers).build()
```

vizro-core/tests/unit/vizro/models/_components/test_tabs.py
```diff
@@ -0,0 +1,74 @@
...
+        assert_component_equal(
+            result,
+            dmc.Tabs(id="tab-id", value="container-1", persistence=True, className="tabs"),
```
antonymilne, line 42: We should have `persistence_type="session"` here actually so should update the `Tabs` model to that.

vizro-core/tests/unit/vizro/models/_components/test_tabs.py
```diff
@@ -0,0 +1,74 @@
...
+        )
+        assert_component_equal(
+            result.children[1],
+            dmc.TabsPanel(className="tabs__panel", value="container-1"),
+            keys_to_strip={"children"},
+        )
+        assert_component_equal(
+            result.children[2],
+            dmc.TabsPanel(className="tabs__panel", value="container-2"),
+            keys_to_strip={"children"},
```
antonymilne, line 68: We can combine these two into a test for `assert_component_equal(result.children[1:], [..., ...])`.

vizro-core/tests/unit/vizro/models/_components/test_tabs.py
```diff
@@ -0,0 +1,74 @@
...
+        )
+        assert_component_equal(
+            result.children[1].children,
+            html.Div([html.H3(), html.Div()], className="tab__content"),
+            keys_to_strip={"children"},
+        )
```
antonymilne, line 74: This isn't testing what you think because the `html.H3(), html.Div()` is totally ignored since you have `keys_to_strip="children" - you could change them to anything and the test would still pass. In fact what you meant here is `html.Div(html.H3(), html.Div())` since there's another level of `html.Div` in `tab.build` itself.

I'd recommend doing this:
```suggestion
        assert_component_equal(
            [tab.children.children for tab in result.children[1:]],
            [html.Div(id="container-1"), html.Div(id="container-2")]
            keys_to_strip={"children"},
        )
```

vizro-core/tests/unit/vizro/models/_components/test_tabs.py
antonymilne: I suggested a couple of simplifications here but generally what you wrote was definitely the right strategy @huong-li-nguyen. I agree that a few comments about what the latter assert statements are for would be helpful.

Generally speaking having multiple assert statements in a test is completely fine, so long as it's clear what the point of them is. This is definitely a case where `assert_component_equal` doesn't work so nicely unfortunately so a couple of comments to clarify would be nice.

vizro-core/tests/unit/vizro/actions/_callback_mapping/test_get_action_callback_mapping.py
antonymilne: Agree with @petar-qb so this is fine how you have it here for now, and can simplify later as part of #111 👍

antonymilne (approved):
Really great job 🙌 These `assert_component_equal` ones were not so easy I know ⭐

vizro-core/tests/unit/vizro/models/_components/test_tabs.py
huong-li-nguyen: Ah, that's a really good point and one I hadn't considered yet! 👍

vizro-core/tests/unit/vizro/models/_components/test_tabs.py
huong-li-nguyen: Ah nice! I was even thinking about removing one as it's just repeated, but this solution is a lot nicer! ⭐

vizro-core/tests/unit/vizro/models/_components/test_tabs.py
huong-li-nguyen: Ahh, yes - that makes a lot of sense! I actually thought I did that 😆  Thank you! 🚀

PR #293, opened 2024-01-26
Changed html.P to html.Label

### LINKED ISSUE #290: Replace all `html.P()` with `html.Label()` in all form components
### Question

Context why this is beneficial: https://teamtreehouse.com/community/labels-vs-paragraphs#:~:text=The%20tag%20is%20just,in%20the%20CSS.

### Code/Examples

Replace `html.P(self.title)` with `html.Label(self.title)` for all form components (dropdown, checklist, etc.) and where applicable

### Other information

_No response_

### Which package?

vizro

### Package version

0.1.9

### Python version

3.9

### OS

_No response_

### Code of Conduct

- [X] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

AnnMarieW:
## Description
closes #290 

## Notice

### DIFF
--- a/vizro-core/src/vizro/static/css/typography.css
+++ b/vizro-core/src/vizro/static/css/typography.css
@@ -58,7 +58,7 @@ h4,
   text-decoration: underline;
 }
 
-p,
+p, label,
 .body-ui-02 {
   color: var(--text-secondary);
   font-size: var(--text-size-02);
--- a/vizro-core/src/vizro/models/_components/form/_user_input.py
+++ b/vizro-core/src/vizro/models/_components/form/_user_input.py
@@ -40,7 +40,7 @@ class UserInput(VizroBaseModel):
     def build(self):
         return html.Div(
             [
-                html.P(self.title) if self.title else None,
+                html.Label(self.title, htmlFor=self.id) if self.title else None,
                 dbc.Input(
                     id=self.id,
                     placeholder=self.placeholder,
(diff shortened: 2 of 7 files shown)


huong-li-nguyen (approved):
Looks great - thank you! 🚀 Also thanks for fixing the CSS at the same time 👍

vizro-core/src/vizro/static/css/typography.css
```diff
@@ -59,6 +59,7 @@ h4,
 }
 
 p,
+label,
```
antonymilne, line 62: @huong-li-nguyen does the `p` styling get used anywhere any more or is it redundant? What happens if we remove it? I only see it in some examples/tests now. I guess maybe it's what gets applied to `dcc.Markdown` also? Just curious...

antonymilne (approved):
Wow, that was fast! Thank you @AnnMarieW 🙏

vizro-core/src/vizro/static/css/typography.css
huong-li-nguyen: It is applied everywhere where we use `html.P` (e.g. the 404 error page) or even in our cards. In general, it's also better practice to have it in, especially for cases where people add their custom components and eventually have some paragraphs inside their components. Otherwise, they always have to use custom CSS to style any paragraphs. It is better if we ship this by default, but they can always overwrite it.

antonymilne: Cool, that was my thinking also 👍

AnnMarieW:
I tried to fix the conflicts, but it looks like I did something wrong and  I'll need some help getting this across the finish line :slightly_frowning_face:

PR #298, opened 2024-02-02
Add `TextArea` as a private component

huong-li-nguyen:
## Description
- Add `TextArea` as a private model (as it has been requested several times now as a custom component)
- Add unit tests for currently private models `UserInput` and `TextArea`
- Remove  "input_type" from `UserInput` for now
- Consolidate CSS between `TextArea` and `UserInput`
- Update `box-shadow `variables based on latest Figma design

Not related but squeezed this in:
- Tidy some docs (remove `Page.path` when not necessary) - In general we shouldn't add an `id` or `path` if it's not required in the config for the examples

## Notice


vizro-core/examples/_dev/app.py
```diff
@@ -3,77 +3,36 @@
...
-                y="lifeExp",
-                size="pop",
-                color="continent",
-            ),
-        ),
-    ],
-)
+# Only added to container.components directly for dev example
+vm.Container.add_type("components", UserInput)
+vm.Container.add_type("components", TextArea)
```
huong-li-nguyen, line 13: Not related to this PR, but we've already mentioned it, but worth discussing whether we even need a `Form` model or if we just want to fall back on the `Container` and add a default action if there is a Button inside. Will be relevant for the model interactions that are yet to come @maxschulz-COL

vizro-core/src/vizro/models/_components/form/_user_input.py
```diff
@@ -44,13 +44,15 @@ def build(self):
                 dbc.Input(
                     id=self.id,
                     placeholder=self.placeholder,
+                    # Note that "range" and "hidden" are not included as possible `input_type` as our recommended
```
antonymilne, line 47: Wouldn't the same be true for `"number"` also?

Actually I think we may well want to just fix `input_type="text"` here and not expose the field at all. Doesn't really matter for now since it's not public, but if you agree then let's just remove the field entirely.

vizro-core/src/vizro/models/_components/form/_text_area.py
```diff
@@ -0,0 +1,49 @@
...
+
+    Args:
+        type (Literal["text_area"]): Defaults to `"text_area"`.
+        title (str): Title to be displayed. Defaults to `""`.
+        placeholder (str): Default text to display in input field. Defaults to `""`.
+        actions (Optional[List[Action]]): Defaults to `[]`.
+    """
+
+    type: Literal["text_area"] = "text_area"
+    title: str = Field("", description="Title to be displayed")
```
antonymilne, line 27:
```suggestion
    # TODO: before making public consider naming this field (or giving an alias) label instead of title
    title: str = Field("", description="Title to be displayed")
```

vizro-core/src/vizro/models/_components/form/_user_input.py
```diff
@@ -14,22 +14,22 @@
...
         title (str): Title to be displayed. Defaults to `""`.
-        input_type (Literal["text", "number", "password", "email", "search", "tel", "url", "range", "hidden"]):
+        placeholder (str): Default text to display in input field. Defaults to `""`.
+        input_type (Literal["text", "number", "password", "email", "search", "tel", "url"]):
             Type of value to validate user input against. Defaults to `"text"`.
+        actions (Optional[List[Action]]): Defaults to `[]`.
     """
 
     type: Literal["user_input"] = "user_input"
     title: str = Field("", description="Title to be displayed")
```
antonymilne, line 29:
```suggestion
    # TODO: before making public consider naming this field (or giving an alias) label instead of title
    title: str = Field("", description="Title to be displayed")
```

vizro-core/src/vizro/models/_components/form/_text_area.py
```diff
@@ -0,0 +1,49 @@
...
+    actions: List[Action] = []
+
+    # Re-used validators
+    _set_actions = _action_validator_factory("value")
+
+    @_log_call
+    def build(self):
+        return html.Div(
+            [
+                html.Label(self.title, htmlFor=self.id) if self.title else None,
```
antonymilne, line 38: I guess we would make this `dbc.Label` in future if we're going all in on bootstrap. Judging by the docs it would make a very small difference in practice compared to `html.Label` though.

antonymilne (approved):
Looks good! 

FYI for when it does come to working out the form components properly, we should review https://wtforms.readthedocs.io/en/3.1.x/ and https://github.com/pydantic/FastUI/blob/main/demo/forms.py.

vizro-core/src/vizro/models/_components/form/_user_input.py
huong-li-nguyen: Yes, you are right! I think the only input types that are quite useful are `email` and `password`. This will most likely be used for authentication purposes though, where I think we currently recommend leveraging existing systems. 

So let me remove it for now and see how people might want to use it 👍

vizro-core/src/vizro/models/_components/form/_text_area.py
huong-li-nguyen: Yes, agree 👍  Would probably leave it for now, so it's consistent with the other components and then we change everything in one go.

vizro-core/examples/_dev/app.py
```diff
@@ -3,77 +3,29 @@
 import vizro.models as vm
 import vizro.plotly.express as px
 from vizro import Vizro
+from vizro.models._components.form._text_area import TextArea
+from vizro.models._components.form._user_input import UserInput
 
-gapminder = px.data.gapminder()
+iris = px.data.iris()
```
petar-qb, line 9: Do we need to undo the changes for this file before merging on the main? 

It seems that we've changed the file content in [this PR](https://github.com/mckinsey/vizro/pull/277). I'm wondering what is the final content for this example that we should keep.

vizro-core/src/vizro/models/_components/form/_text_area.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+
+    type: Literal["text_area"] = "text_area"
+    # TODO: before making public consider naming this field (or giving an alias) label instead of title
+    title: str = Field("", description="Title to be displayed")
+    placeholder: str = Field("", description="Default text to display in input field")
+    actions: List[Action] = []
+
+    # Re-used validators
+    _set_actions = _action_validator_factory("value")
```
petar-qb, line 33: This way of configuring TextArea and UserInput: 
- `_set_actions = _action_validator_factory("value")`
- `debouncing=True`

causes that if the "actions" property is defined, actions chain will be triggered when the enter key is pressed or `when the component loses focus`.

If we want not to trigger actions chain every time when the component loses focus, but only if the enter key is pressed, then we should set `_action_validator_factory` argument to `n_submit`.

petar-qb (approved):
Looks really good!

PR #305, opened 2024-02-08
Add ability push to main bump version after release

l0uden:
## Description

I've changed the script for creating bump PR after the release to push the code directly to `main` branch.
`Personal access token` for `vizro-svc` user was generated. It's stored in `Repository secrets` as `VIZRO_SVC_PAT`.

## Notice


huong-li-nguyen:
Hey @l0uden - thanks for this! Could you remind me what the next steps are and who needs to be involved?

I assume you need to reach out to Leon to allow the service user to bypass these protections. Anything else?

l0uden:
> Hey @l0uden - thanks for this! Could you remind me what the next steps are and who needs to be involved?
> 
> I assume you need to reach out to Leon to allow the service user to bypass these protections. Anything else?

Hey @huong-li-nguyen 

1. First I'll wait @maxschulz-COL, so he can share `personal access token` for the service user, because it connected to his phone. If this user will have ability to merge to `main` without extra permissions so that's it. 
2. If it doesn't I'll ask Leon about adding this user to the setting in the description (I can't do it manually)
3. If second step is not possible for any reason, I'll ask Leon whether we can use my or anyone else account for this purpose.

maxschulz-COL (approved):
In principle look good to me!

2 remaining thoughts:
- did you test this?
- should we do this after every PR, not just the release?

The second point is essentially providing us with a unique version for every PR merge. Not strictly required, but could be useful. What do you think @antonymilne @huong-li-nguyen @l0uden ?

l0uden:
> * did you test this?

Yes, I've tested it in the `vizro-qa` repo. With the same settings and user.

> * should we do this after every PR, not just the release?

I don't think that every merge to main should increase dev version. For example this PR do nothing to vizro library to increase even the dev version.
Looking for changes in the specific folders also will not work, because we need to create changelog for every PR (but maybe there's a way to exclude changelog folder from scan somehow).

.github/workflows/release-if-needed.yml
```diff
@@ -98,19 +101,14 @@ jobs:
...
           hatch run schema
           git config user.email "145135826+vizro-svc@users.noreply.github.com"
           git config user.name "Vizro Team"
           git add -A
-          git commit -m "version bump"
-          git push --set-upstream origin release/version_bump_next_minor/${formatted_date}
-          ./../tools/version-bump.sh mckinsey vizro ${{ secrets.GITHUB_TOKEN }} \
-          ${{needs.check-version.outputs.package_name}} \
-          ${{needs.check-version.outputs.package_version}} release/version_bump_next_minor/${formatted_date}
+          git commit -m "version bump after ${{needs.check-version.outputs.package_name}}-${{needs.check-version.outputs.package_version}} [skip ci]"
```
antonymilne, line 113:
```suggestion
          git commit -m "Bump ${{needs.check-version.outputs.package_name}} version to ${{hatch version}} [skip ci]"
```

Not sure what the right syntax here is but basically the output of `hatch version` shell command gives the new version, and it would be nice to have that in the commit message rather than the old version.

antonymilne: So please don't accept this suggestion without testing it out and correcting the `${{ hatch version }}` to whatever it should be 😀

antonymilne (approved):
This looks great, thank you for figuring it all out! Just a minor tweak to the commit message would be nice.

Re @maxschulz-COL's idea of bumping on every push to main: we should not do this, since the same functionality already exists: each "build" of `main` is labelled by a git commit hash. It's [already possible to `pip install` a specific commit directly](https://stackoverflow.com/questions/13685920/install-specific-git-commit-with-pip).  So effectively all we'd be doing is introducing some mapping from `.devN` to git commit, but this would also introduce lots of noise in the commit history, and we should aim to keep `main` commit history reasonably clean.

maxschulz-COL:
> This looks great, thank you for figuring it all out! Just a minor tweak to the commit message would be nice.
> 
> Re @maxschulz-COL's idea of bumping on every push to main: we should not do this, since the same functionality already exists: each "build" of `main` is labelled by a git commit hash. It's [already possible to `pip install` a specific commit directly](https://stackoverflow.com/questions/13685920/install-specific-git-commit-with-pip). So effectively all we'd be doing is introducing some mapping from `.devN` to git commit, but this would also introduce lots of noise in the commit history, and we should aim to keep `main` commit history reasonably clean.

I agree, but I was not referring to the ability to install a specific git commit, but rather to the ability to allow us/user to know precisely what version they actually have (if they install a different intermediate build).

I fully agree that this keep main unclean, and there is better solutions to it https://github.com/ofek/hatch-vcs. In fact I have tried the latter one back when I set this all up, but there was a known issue (that I think I documented in one of the PRs)

The above suggestion is essentially trying to emulate this.

antonymilne:
> I agree, but I was not referring to the ability to install a specific git commit, but rather to the ability to allow us/user to know precisely what version they actually have (if they install a different intermediate build).
> 
> I fully agree that this keep main unclean, and there is better solutions to it https://github.com/ofek/hatch-vcs. In fact I have tried the latter one back when I set this all up, but there was a known issue (that I think I documented in one of the PRs)
> 
> The above suggestion is essentially trying to emulate this.

Ah ok, that makes more sense. Overall I think this is too low value compared to the noise introduced in `main`, since nthe number of users who install intermediate builds is very low (we should continue with our frequent releases to ensure this remains the case, and also feel free to bump the dev version manually whenever we like if we want to supply users with a particular unreleased version).

.github/workflows/release-if-needed.yml
l0uden: Done

PR #309, opened 2024-02-14
[Feat] Enable temporal selector `DatePicker` 

nadijagraca:
## Description
Added temporal selector `DatePicker` for `Filter` and `Parameter`.  

DatePicker is a single model, but based on value of argument `range` will either return `dmc.DatePicker` or `dmc.DateRangePicker`. Default value of `range` is `True` and the model will return `dmc.DateRangePicker` if range is not provided, or if the selector is not defined. 

**Options where `DateRangePicker` is returned:**
1. User does not define the `Filter.selector` but filter `column` is `datetime` dtype.
2. User defines `vm.DatePicker` as `Filter.selector` without specifying `range` argument or by setting `range=True`
 
**Option where `DatePicker` is returned:**
1. User defines `vm.DatePicker` as `Filter.selector` and sets `vm.DatePicker` `range` argument as `False`. 

`vm.DatePicker.value` supported format are all inputs that could be translated to `datetime.date`. See more [here](https://docs.pydantic.dev/latest/api/standard_library_types/#datetime-types)

Done is separate pr: 
- [x] styling
- [x] docs
- [x] tests

## Screenshot
image

---

image

---

## Notice


vizro-core/examples/_dev/app.py
```diff
@@ -30,7 +34,62 @@
     ],
 )
 
-dashboard = vm.Dashboard(pages=[page])
+# CREATE FAKE DATA
```
nadijagraca, line 37: Changes will be reverted. Added the page for testing purpose.

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -0,0 +1,66 @@
...
+        actions (List[Action]): See [`Action`][vizro.models.Action]. Defaults to `[]`.
+
+    """
+
+    type: Literal["date_picker"] = "date_picker"
+    title: str = Field("", description="Title to be displayed.")
+    min_date: Optional[str] = Field(None, description="Start date value for date picker.")
+    max_date: Optional[str] = Field(None, description="End date value for date picker.")
+    description: str = Field("", description="Description to be displayed above the component.")
+    value: Optional[List[str]] = Field(None, description="Default date value for date picker", min_items=1, max_items=1)
```
huong-li-nguyen, line 40: I don't think they have to provide a list here? If it's a single-selector I would just allow a single value, then you also don't need the additional fields restrictions as it's coming natively from the typing
```suggestion
    value: Optional[date] = Field(None, description="Default date value for date picker")
```

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -0,0 +1,66 @@
...
+
+    actions = []
+
+    _input_property: str = "value"
```
huong-li-nguyen, line 44:
```suggestion
    _input_property:str = PrivateAttr("value")
```

vizro-core/src/vizro/models/_components/form/_form_utils.py
```diff
@@ -100,3 +101,37 @@ def set_default_marks(cls, marks, values):
...
+
+
+def validate_max_date(cls, max_date, values):
+    """Reusable validator for the "max_date" argument for date picker."""
+    if max_date is None:
+        return max_date
+
+    if values["min_date"] is not None and parse_date(max_date) < parse_date(values["min_date"]):
+        raise ValueError("Maximum value of date picker is required to be larger than min_date value.")
+    return max_date
```
huong-li-nguyen, line 137: This looks very similar to the `validate_max` function. Can we just consolidate the two? and they re-use `validate_max`?

vizro-core/src/vizro/models/_components/form/_form_utils.py
```diff
@@ -100,3 +101,37 @@ def set_default_marks(cls, marks, values):
...
+
+    if start_date >= end_date:
+        raise ValueError("Please provide a valid date range for value.")
+
+    if (values["min_date"] is not None and not start_date >= parse_date(values["min_date"])) or (
+        values["max_date"] is not None and not end_date <= parse_date(values["max_date"])
+    ):
+        raise ValueError("Please provide a valid value between the min and max value.")
+
+    return value
```
huong-li-nguyen, line 127: Similar comment as below - can we just refactor the `validate_slider_value` function and then re-use here?

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -0,0 +1,66 @@
...
+        init_value = self.value or self.min_date
+        return html.Div(
+            [
+                html.P(self.title) if self.title else None,
+                dmc.DatePicker(
+                    id=self.id,
+                    description=self.description,
+                    minDate=self.min_date,
+                    value=init_value,
+                    maxDate=self.max_date,
```
huong-li-nguyen, line 61: We should add:
```

persistence=True,
persistence_type="session",
```

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -0,0 +1,66 @@
...
+        description (str): Description to be displayed above the component.
+        actions (List[Action]): See [`Action`][vizro.models.Action]. Defaults to `[]`.
+
+    """
+
+    type: Literal["date_picker"] = "date_picker"
+    title: str = Field("", description="Title to be displayed.")
+    min_date: Optional[str] = Field(None, description="Start date value for date picker.")
+    max_date: Optional[str] = Field(None, description="End date value for date picker.")
+    description: str = Field("", description="Description to be displayed above the component.")
```
huong-li-nguyen, line 39: Hmm.. wondering if we even need this argument in the first go. i think I would remove it for now and see if people do want to use it. 

```suggestion
```

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -0,0 +1,66 @@
...
+    _validate_max_date = validator("max_date", allow_reuse=True)(validate_max_date)
+
+    def build(self):
+        init_value = self.value or self.min_date
+        return html.Div(
+            [
+                html.P(self.title) if self.title else None,
+                dmc.DatePicker(
+                    id=self.id,
+                    description=self.description,
```
huong-li-nguyen, line 58:
```suggestion
```

vizro-core/src/vizro/models/_components/form/date_range_picker.py
```diff
@@ -0,0 +1,71 @@
...
+except ImportError:  # pragma: no cov
+    from pydantic import Field, validator
+
+
+from vizro.models import VizroBaseModel
+from vizro.models._action._actions_chain import _action_validator_factory
+from vizro.models._components.form._form_utils import validate_date_picker_value, validate_max_date
+
+
+class DateRangePicker(VizroBaseModel):
```
huong-li-nguyen, line 17: Same comments I had on the DatePicker apply here (docstrings, typing, arguments etc.) :)

vizro-core/src/vizro/models/_components/form/date_range_picker.py
```diff
@@ -0,0 +1,71 @@
...
+        init_value = self.value or [self.min_date, self.max_date]  # type: ignore[list-item]
+        return html.Div(
+            [
+                html.P(self.title) if self.title else None,
+                dmc.DateRangePicker(
+                    id=self.id,
+                    description=self.description,
+                    minDate=self.min_date,
+                    value=init_value,
+                    allowSingleDateInRange=False,
```
huong-li-nguyen, line 65: Ah, this means we could actually just have one DateRangePicker that allows for single and multi-select if we set this to True?

@antonymilne - do we still want to have two separate models? Or are we actually fine having one model that does single and multi-select? If we go for the second option, we could just have this model here and set this to True. Visually they both look a bit different:

DatePicker:
Screenshot 2024-02-14 at 18 17 34

DateRangePicker if only one date is selected:
Screenshot 2024-02-14 at 18 16 35

```suggestion
                    allowSingleDateInRange=False,
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -92,8 +119,10 @@ def _set_targets(self):
 
     def _set_column_type(self):
         data_frame = data_manager._get_component_data(self.targets[0])
-        if isinstance(data_frame[self.column], pd.PeriodDtype) or is_numeric_dtype(data_frame[self.column]):
+        if is_numeric_dtype(data_frame[self.column]):
             self._column_type = "numerical"
+        elif pd.api.types.is_datetime64_any_dtype(pd.to_datetime(data_frame[self.column], errors="coerce")):
+            self._column_type = "temporal"
```
huong-li-nguyen, line 125: Could this be simplified to below?

```
if isinstance(data_frame[self.column], pd.PeriodDtype):
   self._column_type = "temporal"
elif is_numeric_dtype(data_frame[self.column]):
   self._column_type = "numerical"
else:
   self._column_type = "categorical"
```

vizro-core/src/vizro/models/types.py
```diff
@@ -334,7 +334,7 @@ class OptionsDictType(TypedDict):
 
 # All the below types rely on models and so must use ForwardRef (i.e. "Checklist" rather than actual Checklist class).
 SelectorType = Annotated[
-    Union["Checklist", "Dropdown", "RadioItems", "RangeSlider", "Slider"],
+    Union["Checklist", "Dropdown", "RadioItems", "RangeSlider", "Slider", "DateRangePicker", "DatePicker"],
```
huong-li-nguyen, line 337:
```suggestion
    Union["Checklist", "DatePicker", "DateRangePicker", "Dropdown", "RadioItems", "RangeSlider", "Slider"],
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -35,6 +47,21 @@ def _filter_isin(series: pd.Series, value: MultiValueType) -> pd.Series:
...
+
+    return series.between(start_date, end_date, inclusive="both")
+
+
+def _filter_date_isin(series: pd.Series, value: MultiValueType) -> pd.Series:
+    series = pd.to_datetime(series)
+    return series.isin(value)
+
+
+FILTER_FUNCTIONS = {RangeSlider: _filter_between, DateRangePicker: _filter_date_between, DatePicker: _filter_date_isin}
```
huong-li-nguyen, line 62: I haven't tried it out, so it might not work, but worth to give it a shot.

If instead of providing the datetime values as strings and then converting them here, what if you provide them directly as datetime objects? I think you wouldn't need these different filter functions then, but can then just re-use the existing ones `_filter_is_in` and `_filter_between`

The datetime object conversion should happen in the pydantic models before then

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -133,9 +162,14 @@ def _set_categorical_selectors_options(self):
...
 
+    def _set_filter_function(self):
+        default_filter_function = _filter_isin
+        filter_function = FILTER_FUNCTIONS.get(type(self.selector), default_filter_function)
+        return filter_function
+
     def _set_actions(self):
         if not self.selector.actions:
-            filter_function = _filter_between if isinstance(self.selector, RangeSlider) else _filter_isin
+            filter_function = self._set_filter_function()
```
huong-li-nguyen, line 172: Check comment on trying out to insert datetime objects directly. If that works, then eventually this doesn't have to be changed?

huong-li-nguyen (changes requested):
Hey @nadijagraca - thanks for your work on this! Have a few questions and suggestions :)

vizro-core/src/vizro/models/__init__.py
```diff
@@ -23,13 +23,15 @@
...
+    DateRangePicker=DateRangePicker,
     Tabs=Tabs,
+    DatePicker=DatePicker,
 )
 Navigation.update_forward_refs(Accordion=Accordion, NavBar=NavBar, NavLink=NavLink)
 Dashboard.update_forward_refs(Page=Page, Navigation=Navigation)
 NavBar.update_forward_refs(NavLink=NavLink)
 NavLink.update_forward_refs(Accordion=Accordion)
-
+Parameter.update_forward_refs(DateRangePicker=DateRangePicker, DatePicker=DatePicker)
```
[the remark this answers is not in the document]
nadijagraca: It is on my to-do list to investigate why it asks me to update forward ref when adding parameter, and later remove it.  I've added it so we can test filter functionality.

vizro-core/examples/_dev/app.py
antonymilne: It's fine to leave it in here, no need to worry about this becoming a mess - it's what it's there for 🙂

vizro-core/src/vizro/models/__init__.py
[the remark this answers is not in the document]
antonymilne: I mentioned this to @nadijagraca but if it's not easy to figure out how to remove this/why it's needed then don't worry, just leave it in there.

vizro-core/src/vizro/models/_components/form/date_range_picker.py
antonymilne: Hmmmm this is a very good question and I'm not sure. It would be neater if we could just have one model to keep things simple, but also `DateRangePicker` with a single date selected looks a bit weird.

It also raises another question which is now that I look at it, I don't know why we have a separate `RangeSlider` and `Slider` 🤔 As far as I can tell `RangeSlider` with a single element `value = [1]` would function identically to a `Slider`. So we maybe we should have simplified this to just a single `vm.Slider` model. Did we consider this before? I can't remember. @maxschulz-COL?

@AnnMarieW do you know if there's any difference in behaviour between `RangeSlider` with a single element `value` vs. `Slider`? Is there a reason there's two separate dcc components for these? Other than it's maybe easier for people to find the right component.

antonymilne:
This looks great! Just some answers to your open questions and then I will do a full review next week 🙂 

> * How to hand `value` argument in both date picker and date range picker? I followed the approach of slider and range_slider in this pr, meaning if `value` is not provided I take `min_date` allowed for date picker, and [`min_date`, `max_date`] for date range picker.

This sounds right to me 👍 

> * Do we keep `min_date`/`max_date` or do we replace it with `min` and `max` as in sliders?

Let's go for `min` and `max` to match the sliders.

> * Do we need to add more arguments to the components? I kept it minimal, but some of the additional args to add could be: `allowFreeInput`, `clearable` etc. For more information about possible args check: https://www.dash-mantine-components.com/components/datepicker

Definitely keep it minimal and just expose only the essential arguments like we do for slider/rangeslider. What we should do though is set values for these fields if we think the dmc defaults aren't suitable for us. But those wouldn't be configurable as model fields, just hardcoded into the component returned in the `build` method.

vizro-core/src/vizro/models/_components/form/date_range_picker.py
antonymilne: 💡 We could have a single `DatePicker` model but then it returns either `DateRangePicker` or `DatePicker` depending on whether the user sets some field to `True/False` or supplies two vs. one `value`?

AnnMarieW: Interesting question about using a single value in the Range Slider.  It seems to work, but it would be possible to change it to a range slider when updating it with two values in a callback.  So there would need to be some type of check to make sure that only a single value is returned.  I'm not sure if there would be any other unintended issues.

AnnMarieW:
Be sure to set the `dropdownPosition` prop.  It seems to be a workaround for this issue:  https://github.com/snehilvj/dash-mantine-components/issues/219

vizro-core/examples/_dev/app.py
```diff
@@ -30,7 +34,62 @@
...
+    title="Datepicker page",
+    components=[vm.Table(id="table", figure=dash_data_table(data_frame=data))],
+    controls=[
+        vm.Filter(
+            column="DATE_TIME",
+            selector=vm.DateRangePicker(
+                title="Pick a date",
+                min_date="2023-01-01",
+                value=["2024-01-01", "2024-03-01"],
+                max_date="2024-07-07",
```
petar-qb, line 85: Can you update this file so it works with the latest changes you made?

vizro-core/examples/_dev/app.py
```diff
@@ -30,7 +34,62 @@
...
+            ),
+            # selector=vm.DatePicker(title="Pick a date", min_date='2023-01-01', value=['2024-01-01']),
+        )
+    ],
```
petar-qb, line 89: Have you checked if DateRangePicker and DatePicker work with the vm.Parameter as well?

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -35,6 +47,21 @@ def _filter_isin(series: pd.Series, value: MultiValueType) -> pd.Series:
     return series.isin(value)
 
 
+def _filter_date_between(series: pd.Series, value: List[str]) -> pd.Series:
+    series = pd.to_datetime(series)
+    start_date, end_date = value
+
+    return series.between(start_date, end_date, inclusive="both")
```
petar-qb, line 54: I'm getting the following error here when I clear UI value from the vm.DateRangePicker:
`... not enough values to unpack (expected 2, got 1)`

vizro-core/src/vizro/models/_components/form/_form_utils.py
```diff
@@ -69,7 +69,7 @@ def validate_max(cls, max, values):
     return max
 
 
-def validate_slider_value(cls, value, values):
+def validate_range_value(cls, value, values):
```
petar-qb, line 72: We should probably update docstring below to something like:
"""Reusable validator for the "value" argument for range components."""

Let's also avoid using a word "slider/sliders" in the function implementation.

vizro-core/src/vizro/models/_components/form/date_range_picker.py
antonymilne: I've split this question off into a new discussion so as not derail things here: #318.

Currently my feeling is that this is a good idea, curious to hear what others think:
💡 We could have a single `DatePicker` model but then it returns either `DateRangePicker` or `DatePicker` depending on whether the user sets some field to `True/False` or supplies two vs. one `value`?

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -59,6 +59,8 @@ def build(self):
                     maxDate=self.max,
                     persistence=True,
                     persistence_type="session",
+                    dropdownPosition='bottom-start',
+                    clearable=False,
```
petar-qb: I'm still getting `... not enough values to unpack (expected 2, got 1)` exception when I click again on the "start date" from the UI after it's already selected. Here is the demo on how to recreate the bug: 

https://github.com/mckinsey/vizro/assets/108530920/0a4dfaac-904f-4184-81e0-1cc92932879e

QQ: Why did we decide to set `clearable=False` by default? 

Potential solution: 
We should define what is expected behaviour after the "start data" is clicked again (i.e. when the server receives `[None]`, instead of `[{start_date}, {end_date}]`).
 - Do we want to do nothing (probably just `raise dash.exceptions.PreventUpdate inside the `_filter_date_between` function would work here) or
 - We want to return an empty figure (return an empty `pd.Series` from the `_filter_date_between ` function). - This seems like a better solution to me. It's more intuitive in a case someone clears the DatePicker (by clicking on the "start date" again) and then selects another filter/parameter on the page.

vizro-core/src/vizro/models/_controls/filter.py
huong-li-nguyen: Did this not work @nadijagraca ?

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -133,9 +162,14 @@ def _set_categorical_selectors_options(self):
 
             self.selector.options = sorted(options)
 
+    def _set_filter_function(self):
+        default_filter_function = _filter_isin
+        filter_function = FILTER_FUNCTIONS.get(type(self.selector), default_filter_function)
+        return filter_function
```
huong-li-nguyen, line 168:
```suggestion
        filter_function = FILTER_FUNCTIONS.get(type(self.selector), _filter_isin)
        return filter_function
```

huong-li-nguyen (commented):
Have mostly tidy comments left 🧹 - so ensuring consistency between the other components, docstrings etc. The rest looks good to me 👍  

@nadijagraca - could you also update the PR description and title?

vizro-core/src/vizro/models/_components/form/date_picker.py
huong-li-nguyen: Generally, agree with @petar-qb 👍 🚀 

Could you guys just ensure the behaviour is consistent across the form components e.g. when I clear the Dropdown now - what's happening? https://vizro.mckinsey.com/relationship-analysis

It doesn't seem to do any of the two proposed solutions above 😄  Would be great to double-check 👍

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -35,6 +47,21 @@ def _filter_isin(series: pd.Series, value: MultiValueType) -> pd.Series:
...
+def _filter_date_between(series: pd.Series, value: List[str]) -> pd.Series:
+    series = pd.to_datetime(series)
+    start_date, end_date = value
+
+    return series.between(start_date, end_date, inclusive="both")
+
+
+def _filter_date_isin(series: pd.Series, value: MultiValueType) -> pd.Series:
+    series = pd.to_datetime(series)
+    return series.isin(value)
```
huong-li-nguyen, line 59: I still think that we should not need these extra functions, but should be able to re-use the existing ones. To get that to work, I think we have to convert the relevant column in the data set **before it gets into the filter function**. To illustrate what I mean with your current dev example:

This should work without any errors, but currently it breaks: 
I think we should encode that logic in that if the data column can be converted to a datetime object, then we do that automatically and apply the right default selector (DatePicker). Currently this is not the case, as we don't convert the data column automatically and this is also the reason why you currently need these two extra functions here. 

If we would convert the dtype for the column smartly as well, the above example should work out of the box (using the Datepicker automatically) and you might not need these two extra functions

```
page_1 = vm.Page(
    title="Datepicker page",
    components=[vm.Table(id="table", figure=dash_data_table(data_frame=data))],
    controls=[
        vm.Filter(column="DATE_TIME")
    ],
)
```

vizro-core/examples/_dev/app.py
huong-li-nguyen: Any updates on this? :)

vizro-core/src/vizro/models/_components/form/date_picker.py
nadijagraca: So, do we want to return the empty pd.Series in this case, and then check in a different pr the behaviour of other components? @huong-li-nguyen @petar-qb

vizro-core/src/vizro/models/__init__.py
```diff
@@ -23,13 +23,15 @@
     Graph=Graph,
     Parameter=Parameter,
     Table=Table,
+    DateRangePicker=DateRangePicker,
     Tabs=Tabs,
+    DatePicker=DatePicker,
```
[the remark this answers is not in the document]
nadijagraca: I've removed DatePicker and DateRangePicker as it is no longer necessary to update forward refs for Page.

vizro-core/src/vizro/models/_components/form/date_picker.py
huong-li-nguyen: I think we should the behaviour of other components now and replicate the same behaviour here. The thing i would leave for a different PR is if we want to deviate from the existing approach and optimise it.

petar-qb: Returning empty pd.Series in this case seems like the best solution since an empty vm.Dropdown and vm.Checklist as filter also causes that target figure doesn't contain any data.

vizro-core/examples/_dev/app.py
nadijagraca: It works with vm.Parameter as well. :)

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -0,0 +1,73 @@
...
+    Can be provided to [`Filter`][vizro.models.Filter] or [`Parameter`][vizro.models.Parameter].
+    Based on the underlying [`dmc.DatePicker`](https://www.dash-mantine-components.com/components/datepicker).
+
+    Args:
+        type (Literal["date_picker"]): Defaults to `"date_picker"`.
+        min (Optional[date]): Start date for date picker. Defaults to `None`.
+        max (Optional[date]): End date for date picker. Defaults to `None`.
+        value (Union[List[date], date]): Default date/dates for date picker. Defaults to `None`.
+        title (str): Title to be displayed. Defaults to `""`.
+        multi (bool): Boolean flag for displaying range picker. Default to `True`.
```
huong-li-nguyen, line 31: @nadijagraca @antonymilne - any preference in naming this argument? It's similar to the Dropdown but I don't find it very intuitive. 

Maybe` range=True/False?`

```suggestion
        multi (bool): Boolean flag for displaying range picker. Default to `True`.
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -124,6 +147,31 @@ def _set_slider_values(self):
             if self.selector.max is None:
                 self.selector.max = max(max_values)
 
+    def _set_date_picker_values(self):
+        if isinstance(self.selector, SELECTORS["temporal"]):
+            if self._column_type != "temporal":
```
huong-li-nguyen, line 152:
```suggestion
        if isinstance(self.selector, SELECTORS["temporal"]) and self._column_type != "temporal":
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -124,6 +147,31 @@ def _set_slider_values(self):
...
+            for target_id in self.targets:
+                data_frame = data_manager._get_component_data(target_id)
+                data_frame = self._convert_column_type(data_frame=data_frame)
+                min_values.append(data_frame[self.column].min())
+                max_values.append(data_frame[self.column].max())
+            if not is_datetime64_any_dtype(pd.Series(min_values)) or not is_datetime64_any_dtype(pd.Series(max_values)):
+                raise ValueError(
+                    f"Non-temporal values detected in the shared data column '{self.column}' for targeted charts. "
+                    f"Please ensure that the data column contains the same data type across all targeted charts."
+                )
```
huong-li-nguyen, line 169: Do we actually need this? I remember we have this for the range slider, but which case does this actually catch?
Because you already convert the relevant data column before right?

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -133,9 +181,19 @@ def _set_categorical_selectors_options(self):
 
             self.selector.options = sorted(options)
 
+    def _set_filter_function(self):
+        if isinstance(self.selector, RangeSlider) or (
+            isinstance(self.selector, DatePicker) and self.selector.multi is True
+        ):
+            filter_function = _filter_between
+        else:
+            filter_function = _filter_isin
```
huong-li-nguyen, line 190: Maybe convert this to a mapping dict as well? Similar to how we have it for the default selectors?

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -15,19 +15,32 @@
...
+SELECTOR_DEFAULTS = {"numerical": RangeSlider, "categorical": Dropdown, "temporal": DatePicker}
 
 # Ideally we might define these as NumericalSelectorType = Union[RangeSlider, Slider] etc., but that will not work
 # with isinstance checks.
-SELECTORS = {"numerical": (RangeSlider, Slider), "categorical": (Checklist, Dropdown, RadioItems)}
+SELECTORS = {
+    "numerical": (RangeSlider, Slider),
+    "categorical": (Checklist, Dropdown, RadioItems),
+    "temporal": DatePicker,
+}
```
huong-li-nguyen, line 37: Maybe we can actually combine these two mappings?

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -15,19 +15,32 @@
...
 
 
 def _filter_between(series: pd.Series, value: List[float]) -> pd.Series:
+    if len(value) != 2:
+        return pd.Series(False, index=series.index)
+
```
huong-li-nguyen, line 43: Can you remind me - what does this case catch? It's late and my brain is already a bit foggy 😅

vizro-core/src/vizro/models/_controls/filter.py
huong-li-nguyen: I haven't tested this, so it might be broken, but something like this:

```
 filter_mapping = {
        RangeSlider: _filter_between,
        DatePicker: _filter_between if self.selector.multi else _filter_isin,
        Dropdown: _filter_isin,
        RadioItems: _filter_isin,
        Checklist: _filter_isin
    }

# Then apply this directly:
...
if not self.selector.actions:
    filter_function = filter_mapping.get(self.selector, _filter_isin)
```

Only change it if you think it's an improvement though. I am a bit indifferent right now 😄

huong-li-nguyen (approved):
Thank you for your patience and implementing the comments! I think this looks a lot cleaner and clearer now! 🚀 Have some minor comments, but really up to you to implement it if you think it's an improvement. Otherwise just keep as is :) 

Also double-check the linting and CI tests - I think they are currently still failing :)

vizro-core/src/vizro/models/_components/form/date_picker.py
huong-li-nguyen: Just for my own tracking of what's happening now: 

- If you add the start-date but no end-date the figure does not get updated
- If you clear both inputs, then you get an empty figure

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -124,6 +147,31 @@ def _set_slider_values(self):
...
+                raise ValueError(
+                    f"Chosen selector {self.selector.type} is not compatible "
+                    f"with {self._column_type} column '{self.column}'."
+                )
+
+            min_values = []
+            max_values = []
+            for target_id in self.targets:
+                data_frame = data_manager._get_component_data(target_id)
+                data_frame = self._convert_column_type(data_frame=data_frame)
```
huong-li-nguyen, line 162: This can probably be done in a nicer way once the data refreshes are possible in the data manager. Would we store the converted data or do you think we shouldn't do that anyways @antonymilne ?

vizro-core/examples/_dev/app.py
```diff
@@ -45,7 +47,95 @@
...
+            selector=vm.DatePicker(
+                title="Pick a date",
+                min="2023-01-01",
+                value=["2024-01-01", "2024-03-01"],
+                max="2024-07-07",
+            ),
+            # selector=vm.DatePicker(title="Pick a date", min_date='2023-01-01', value=['2024-01-01'], multi=False),
+        )
+    ],
+)
```
petar-qb, line 105: I tried out numerous combinations and it works so smoothly 🚀. 

The only inconsistency I found is when I configure these two cases:
- multi=False, value=["2024-01-01", "2024-03-01"] and
- multi=True, value="2024-01-01"

So we should probably validate it.

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -0,0 +1,73 @@
...
+from vizro.models._action._actions_chain import _action_validator_factory
+from vizro.models._components.form._form_utils import validate_max, validate_range_value
+
+
+class DatePicker(VizroBaseModel):
+    """Temporal selector `DatePicker`.
+
+    Can be provided to [`Filter`][vizro.models.Filter] or [`Parameter`][vizro.models.Parameter].
+    Based on the underlying [`dmc.DatePicker`](https://www.dash-mantine-components.com/components/datepicker).
+
```
petar-qb, line 24: Should we add that it's based on the dmc.DateRangePicker as well?

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -0,0 +1,73 @@
...
+        date_picker = date_picker(
+            id=self.id,
+            minDate=self.min,
+            value=init_value,
+            maxDate=self.max,
+            persistence=True,
+            persistence_type="session",
+            dropdownPosition="bottom-start",  # dropdownPosition must be set to bottom-start as a workaround
+            # for issue: https://github.com/snehilvj/dash-mantine-components/issues/219
```
petar-qb, line 66: Maybe we should move these comments above the ` date_picker = date_picker(...)` line.

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: My understanding is that if the date picker is deleted, the input `value` argument becomes `[None]`.

We should probably improve this validation and check if the value is a list with two elements where both elements are numbers.

vizro-core/examples/_dev/app.py
petar-qb: I just found another weird behaviour. Both date pickers behave strange when the value is cleared, and when you navigate to another page and then get back to the page where the date picker is cleared.
- single date picker returns a default value (instead of being cleared as the vm.Dropdown does)
- rande date picker raises an exception

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: We should probably adjust value type hint accordingly then.

petar-qb (commented):
Awesome work 🚀. 
It really looks good, instead of a bug I found and commented below. When you fix it, let me know so I can hit the "Approve" 😄.

vizro-core/src/vizro/models/__init__.py
```diff
@@ -29,7 +29,7 @@
 Dashboard.update_forward_refs(Page=Page, Navigation=Navigation)
 NavBar.update_forward_refs(NavLink=NavLink)
 NavLink.update_forward_refs(Accordion=Accordion)
-
+Parameter.update_forward_refs(DatePicker=DatePicker)
```
antonymilne, line 32: Is this still needed? Fine if it is, but it just looks a bit weird since we don't have any other selectors here and don't have `Filter.update_forward_refs`.

vizro-core/src/vizro/models/_components/form/date_picker.py
antonymilne: Agree - let's rename `multi` to `range`.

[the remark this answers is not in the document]
antonymilne: I can definitely see this being useful, but let's keep it minimal for now and see how people use the component rather than adding it upfront. It's easy to modify with a custom component for anyone who does want to change it.

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -0,0 +1,73 @@
...
+    _validate_max = validator("max", allow_reuse=True)(validate_max)
+
+    def build(self):
+        init_value = self.value or ([self.min, self.max] if self.multi else self.min)  # type: ignore[list-item]
+        if self.multi:
+            date_picker = dmc.DateRangePicker
+        else:
+            date_picker = dmc.DatePicker
+
+        date_picker = date_picker(
```
antonymilne, line 58:
```suggestion
        date_picker_class = dmc.DateRangePicker if self.multi else dmc.DatePicker
        
        date_picker = date_picker_class(
```
Moving the conditional to one line is just a minor style thing. Renaming it important though or you have a variable name clash, which doesn't cause problems right now but is not a good idea as it can lead to super confusing error messages.

vizro-core/src/vizro/models/_controls/filter.py
antonymilne: You mean `SELECTORS` and `SELECTOR_DEFAULTS`? If so then I agree - we just say that the first entry in each tuple is the default selector for that key. The only place that `SELECTOR_DEFAULTS` is used can be removed and changed to `SELECTORS[self._column_type][0]`.

The catch here is that it's fragile because someone could change the order of e.g. `RangeSlider` and `Slider` without realising that it's changing the default selector. So if we make this change we need a comment right above `SELECTORS` explaining clearly that the first entry in each tuple is the default selector.

Note you'll need to swap the order of `categorical` to make `Dropdown` the first also. Hopefully if you don't do that then our tests will fail and tell us. If they don't then actually our tests aren't robust enough to catch this and we shouldn't make the change here at all.

vizro-core/src/vizro/models/_controls/filter.py
antonymilne: Hmm good question, I'm not sure. Definitely it feels a bit wrong having two calls to `_convert_column_type` at the moment, but also I'm not sure we should be persisting changes to data. So let's leave it this way for now.

vizro-core/src/vizro/models/_controls/filter.py
antonymilne: I would keep it simple, remove this function entirely and put the logic directly in `_set_actions`:
```
if isinstance(self.selector, RangeSlider) or isinstance(self.selector, DatePicker) and self.selector.range:
    filter_function = _filter_between
else:
    filter_function = _filter_isin
```

antonymilne (approved):
👏 This is wonderful, great work @nadijagraca and everyone else! I know it's not been an easy task but the resulting code really looks great 👍 

The only major change I have is to rename `multi` to `range`.

I didn't actually try it out but trust that between @nadijagraca @petar-qb and @huong-li-nguyen it's been tested enough.

I'll leave you guys to figure out any unresolved bugs and the `clearable`/`_filter_between` stuff, but just give me a shout if you want some help from me with it.

vizro-core/src/vizro/models/_controls/filter.py
huong-li-nguyen: Yes, that was what I was thinking about :) I think a comment should suffice because we rarely does this mapping

huong-li-nguyen:
@nadijagraca - let us know how it goes! I found several issues regarding the datepickerrange from the mantine library that are still open, all concerning persistence.. So if this is not easy to solve, then let's brainstorm next week what alternatives we have. As Petar said, I also hope that we find a way to fix this other than replacing it with the dcc version 😅  I think even if we would replace it with the dcc version, it should be straight-forward except for the styling 😓 But let's see!

If you haven't seen it yet:
- https://github.com/snehilvj/dash-mantine-components/issues/212
- https://github.com/snehilvj/dash-mantine-components/issues/185

vizro-core/examples/_dev/app.py
nadijagraca: Unfortunately, the bug you found is on the dmc component, as it produces the exact same exception when using pure dash. 
Possible option to mitigate this, is set `clearable=False` and disable clearing the datepicker. Pro's of this approach is we keep using dmc component and keep `persistence=True`.

huong-li-nguyen: That's fine for me 👍 But I remember that when I proposed that during stand-up, that there would be still a second issue remaining? Is this not the case anymore?

If we just need to set clearable=False, then I'm happy to do that 👍 Just add a comment of why we have it there and raise an issue in the dmc library as well. So in case they do fix it, we can also remove it :)

petar-qb: I agree with the workaround you suggested (`clearable=False` and `persistence=True`). 

However, if there is no rush to release the DatePicker feature, I would suggest that we try to solve this problem in a similar way to the one that appears in `vm.Slider` and `vm.RangeSlider`: 
Attempting to implement a client-side callback that is triggered on a `date_picker.value` change, with inputs=[`date_picker.value`, `temp_store.data`] with outputs=[`date_picker.value`] and a body that checks if the `date_picker_value` is null, and if it is, assign the `temp_store_data` to its value. 

I'm not a wizard in the dmc bug you found but I give 50% chance that this approach can solve the problem we have.

huong-li-nguyen: There is no rush, so happy for you guys to try that out if you think it could work 👍

vizro-core/src/vizro/models/__init__.py
nadijagraca: Removed as no loner needed.

vizro-core/src/vizro/models/_controls/filter.py
nadijagraca: Validations are no longer needed as date picker is no longer clearable.

vizro-core/src/vizro/models/_components/form/_form_utils.py
```diff
@@ -100,3 +101,21 @@ def set_default_marks(cls, marks, values):
...
+    if not range and isinstance(values["value"], date):
+        return range
+
+    if range and (len(values["value"]) == 1 or isinstance(values["value"], date)):
+        raise ValueError(
+            "The range value of the date picker is True, value argument needs to contain list of two date values."
+        )
+
+    if not range and isinstance(values["value"], list):
+        raise ValueError("The range value of the date picker is False, value argument needs to be of single date.")
```
petar-qb, line 121: For the following configuration the exception is raised in build phase - when I open the app in the browser.
```
 vm.Filter(
            column="DATE_TIME",
            selector=vm.DatePicker(
                title="Pick a date",
                min="2023-01-01",
                value="2024-01-01",
                max="2024-07-01",
            ),
        ),
```
And this configuration should raise a pydantic exception before as far as I know (because range isn't specified and its True by default but the value isn't a list of two elements).

petar-qb (approved):
This looks good to me! An awesome work full of investigation is behind you and thank you for that.

Just slightly update the PR description (replace `multi` to `range` and so on), pull the main branch and then it's ready for merging. 

QQ: Are we waiting for unit tests to be merged into this branch or we will merge this PR into the main as it is?

vizro-core/src/vizro/static/js/models/date_picker.js
```diff
@@ -0,0 +1,11 @@
...
+  if (value === null) {
+    return [input_store, input_store];
+  } else {
+    if (dash_clientside.callback_context.triggered.length === 0) {
+      return [input_store, input_store];
+    } else {
+      return [value, value];
+    }
+  }
+}
```
petar-qb, line 11:
```suggestion
export function _update_date_picker_values(value, input_store, self_data) {
  if (value === null || dash_clientside.callback_context.triggered.length === 0) {
    return [input_store, input_store];
  return [value, value];
}
```

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -0,0 +1,104 @@
...
+    range: bool = Field(True, description="Boolean flag for displaying range picker.")
+    actions: List[Action] = []
+
+    _input_property: str = PrivateAttr("value")
+    _set_actions = _action_validator_factory("value")
+
+    # Re-used validators
+    _validate_value = validator("value", allow_reuse=True)(validate_range_value)
+    _validate_max = validator("max", allow_reuse=True)(validate_max)
+    _validate_range = validator("range", allow_reuse=True, pre=True)(validate_date_picker_range)
```
petar-qb, line 52: The following case doesn't raise an exception in the initialisation phase but it should:
```
        vm.Filter(
            column="DATE_TIME",
            selector=vm.DatePicker(
                value="2024-01-01"
            )
        ),
```

There are also couple of examples I found that raise the IndexError exceptions instead of handled ValueError from pydantic which isn't the best practice. However we can fix these once we add tests.

Issue #285, opened 2024-01-24
No validation for Card.href can break page opening


l0uden:
### Description

If we create `path="/page page"` for any page it will be available in browser like `/page-page`.
At the same time if we create `Card.href="/page page"` we'll get 404 during page opening.

Looks like to synchronise this values we should apply same validation for `href` and `path`

### Which package?

vizro

### Package version

0.1.8

### Python version

3.9

### OS

Mac OS

### How to Reproduce

1. Create one page with `path="/page page"`
2. Create second page with `Card.href="/page page"`
3. Run the dashboard and click `Card` on the second page.

### Code of Conduct

- [X] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

antonymilne:
Currently this is the expected behaviour, but I think you're right we should change the behaviour.

There's two different things here:
* when you set `Page.path`, it's correct that we convert whitespace to `-` to make a clean valid URL for the page
* when you set `Card.href`, remember this can be a link to anything, not just a link inside your app. Hence it's valid to put in e.g. `Card.href = "http://www.example.com/something with space"` (space is not a valid character in a URL and it should be encoded as `%20`, but this encoding always happens automatically when you click the link so there's not really any point us doing it I think). Even if we did encode it, we should not convert spaces to `-` because it would contradict the correct HTML encoding of `%20`

The case you give here is a special (but common) case where you use `Card.href` to refer to a page in the app and hence use a relative link. We handle this as:
```
href=get_relative_path(self.href) if self.href.startswith("/") else self.href
```

I think we should change this:
1. put `self.href` through the same `clean_path` transformation that `Page.path` has in the case that it's a relative path. Hence writing `Card.href = "/page page"` would then work
2. even better, make it so that you can just write `Card.href = "page page"` without the leading `/` and it would still work. This would mean we need [a better check to see if the `href` is relative or absolute](https://stackoverflow.com/questions/8357098/how-can-i-check-if-a-url-is-absolute-using-python) and to add the `/` back in the case that it's not absolute and it's missing

This way a user will be able to put the page title directly in the `Card.href` without thinking about it even if it's got weird characters, i.e. this should work:
```
Page(title = "something weird %*")
Card(href = "something weird %*")
```

antonymilne:
After more thought we should not implement this feature for now. See #348 for explanation.

PR #311, opened 2024-02-14
Add default Bootstrap theme and overwrite CSS

huong-li-nguyen:
## Description

**This is also just a temporary POC fix to an issue we will solve differently long-term, changes included in this PR:**
- Add default Bootstrap theme
- Overwrite relevant CSS properties such that the Bootstrap theme has no effect on styling
- Remove CSS code that is not required anymore (as functionality is shipped automatically with Bootstrap theme now)
- Updated screenshots in the vizro-qa repo as there are new defaults for some elements

The current approach is not as different as our previous one, but has a few advantages:

**Remains the same:**
- `dbc.themes` still do not work properly in our code basis (in the sense of switching skins)
- When adding a `dbc.component` to our code basis we still need to use custom CSS to apply our design (eventually we need to overwrite more properties now)

**Improvement to previous version:**
- **Visual consistency:** Plugging a `dbc.theme `will have no visual effect now (with the currently existing components), as our CSS overwrites their CSS
- **Less code to maintain:** 
    - Including a `dbc.theme` by default has the advantage that all relevant CSS for the proper functioning of dbc components are correctly shipped, so when someone adds a dbc component as a custom component, it will work out of the box and they don't have to add CSS code to their assets folder anymore (same applies to us)
   - Many html elements that we did not have stylings for (e.g. ul, li, etc.) have a default Bootstrap styling applied now.
  
## Screenshot
**When plugging in a dbc.theme before:**
Screenshot 2024-02-14 at 20 46 34
Screenshot 2024-02-14 at 20 46 39

**Now:**
Screenshot 2024-02-14 at 20 47 10

Screenshot 2024-02-14 at 20 47 02

## Notice


vizro-core/src/vizro/models/_components/card.py
```diff
@@ -36,8 +35,8 @@ class Card(VizroBaseModel):
     def build(self):
         text = dcc.Markdown(self.text, className="card_text", dangerously_allow_html=False, id=self.id)
         button = html.Div(
-            dbc.Button(
-                href=get_relative_path(self.href) if self.href.startswith("/") else self.href, className="card_button"
+            dcc.Link(
+                href=get_relative_path(self.href) if self.href.startswith("/") else self.href, className="card-link"
```
huong-li-nguyen, line 39: It made more sense semantically here to convert it to a `dcc.Link`

vizro-core/src/vizro/static/css/accordion.css
```diff
@@ -1,39 +1,9 @@
-/* media queries necessary for opening and closing accordion*/
-@media (prefers-reduced-motion: reduce) {
```
huong-li-nguyen, line 2: All of this CSS code is not required anymore as it ships automatically by default with the bootstrap theme

huong-li-nguyen:
@antonymilne - this was initially supposed to be just a POC but shall we go ahead with this PR change now? Main advantages for me are:

- Lots of CSS functionality shipped automatically now for `dbc.components` (benefits us and users that add `dbc.components` as custom components)
- Default styling applied to semantic html elements that we haven't styled yet (e.g. a, li, ul, etc.)
- When someone does add a Bootstrap theme, in theory it shouldn't look completely off anymore as it just gets ignored as our CSS overwrites the Bootstrap ones for the components that are currently present

vizro-core/src/vizro/_vizro.py
```diff
@@ -24,8 +25,13 @@ def __init__(self, **kwargs):
 
         """
         self.dash = dash.Dash(**kwargs, use_pages=True, pages_folder="", title="Vizro")
-        self.dash.config.external_stylesheets.append(
-            "https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined"
+        self.dash.config.external_stylesheets.extend(
```
antonymilne, line 28: When we have other dbc themes working we should figure out how to handle this correctly, e.g. maybe we need to be inserting these to the beginning rather than end of `external_stylesheets` to make overrides work in the right order. But this is fine for now.

vizro-core/src/vizro/models/_components/card.py
antonymilne: This is massively better now but still feels hacky to me. Why can't we do it in what I would consider to be the "correct" way like this:
```
dcc.Link(text, href=...) if self.href is not None else text
```

Either way, we should change the variable names and class name so it's no longer `button`.

vizro-core/src/vizro/static/css/typography.css
```diff
@@ -50,17 +44,18 @@ h4,
...
 
 p,
 label,
-.body-ui-02 {
+h5,
+h6,
+ul,
+li,
+.body-ui-02,
+.body-ui-link-02 {
```
antonymilne, line 58: Just to understand as an example, what happens if we don't have this custom css here? Does it look really awful or just a little off?

vizro-core/src/vizro/models/_components/card.py
antonymilne: As for the navigation, our current solution feels quite abusive to me and so if we can fix it I think we should (doesn't have to be this PR though).

1. For now, how about setting `color="link"` on the `dbc.Button`? See https://dash-bootstrap-components.opensource.faculty.ai/docs/components/button/ for examples. This feels like it would be the right solution to me. Considering that the docs on `dbc.Accordion` include buttons inside the accordion, I don't think that's a naughty thing to do so long as it's clear that the button is operating as a link
2. If that doesn't work well, what about instead just using `dcc.Link`, dropping the `key` argument (unless I misunderstood something, it doesn't feel particularly important here) and doing the `active` styling with CSS classname instead 
3. Way outside scope for this PR, but possibly in future we should use [`NavLink`](https://dash-bootstrap-components.opensource.faculty.ai/docs/components/nav/) inside a proper `Nav`. This feels most correct and actually supports a more advanced `active` than `dbc.Button`

antonymilne (approved):
This seems like a big step in the right direction to me and I think we should go ahead with it 👍 

If we can now remove the "Adding a dash-bootstrap component" from the docs then let's do so. The "Dash Bootstrap Themes" warning is still relevant but could maybe be modified a bit now we know the direction of travel:
```
    Please note that Vizro is not yet compatible with [Dash Bootstrap Themes](https://dash-bootstrap-components.opensource.faculty.ai/docs/themes/).
    Adding a Bootstrap stylesheet will not have any effect on your Vizro app.
```

vizro-core/src/vizro/_vizro.py
huong-li-nguyen: Yes, I agree. I thought about that as well! For your information, usually, you load the Bootstrap theme and afterwards any custom CSS (ours, and theirs) because if the CSS selectors have the same specificity, the later one will be taken.

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -203,22 +206,22 @@ div.dashboard_container .custom-tooltip {
...
+  cursor: pointer;
+  position: absolute;
 }
 
 #collapse-icon.material-symbols-outlined:hover {
   color: var(--text-active);
 }
 
 .collapse-button-tooltip {
+  height: 24px;
```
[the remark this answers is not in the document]
huong-li-nguyen: I once did a quick search and I couldn't find an automatic linter that one could eventually add to the pre-commits. So we prioritised it back then. Nadija and I currently use [PyCharm's code formatting ](https://www.jetbrains.com/help/pycharm/settings-code-style-css.html)settings such that the CSS files get alphabetically ordered. The downside of that is that I have to click on it manually 😅  

For context, I did consider other sorting options, but we ended up with alphabetical. I really liked this article: https://jerrylowm.medium.com/alphabetize-your-css-properties-for-crying-out-loud-780eb1852153

vizro-core/src/vizro/static/css/typography.css
huong-li-nguyen: Depends what you consider as awful 😄 If you have some bullet points and some paragraph text in the card-text for example, you would see that the text is not consistently formatted within the card if this code is not in here. So I would consider it already as necessary for visual consistency. But it would not look completely off.

vizro-core/src/vizro/models/_components/card.py
huong-li-nguyen: I'll respond to your first request first, as the second one I need to try out first.

Yes, that's a great idea actually! That also works perfectly fine, I've implemented the change [here](https://github.com/mckinsey/vizro/pull/311/commits/ca213993b2da2a562c71832519353c1638eefb4c) and also got rid of one of the redundant htmls! Do you think it's worth implementing the CSS name changing from underscore to dashes already? Or better to do in one separate PR? Or not doing at all if we at some point switch to BEM / Bootstrap syntax?

huong-li-nguyen: On your second question:
1. Should be do-able and I could implement this in the current PR already.
2. I slightly remember @nadijagraca actually trying that out when she did implement the navigations. She also tried out the active CSS property as far as I remember, but it didn't seem to work? @nadijagraca - can you confirm? or did you not try that out? 
3. @nadijagraca - any reason why we didn't use NavLink in the first place? I somehow also remember there was something 😄

huong-li-nguyen:
> This seems like a big step in the right direction to me and I think we should go ahead with it 👍
> 
> If we can now remove the "Adding a dash-bootstrap component" from the docs then let's do so. The "Dash Bootstrap Themes" warning is still relevant but could maybe be modified a bit now we know the direction of travel:
> 
> ```
>     Please note that Vizro is not yet compatible with [Dash Bootstrap Themes](https://dash-bootstrap-components.opensource.faculty.ai/docs/themes/).
>     Adding a Bootstrap stylesheet will not have any effect on your Vizro app.
> ```

I've modified to below, but I think the warning is still relevant. However, I've removed the other warning as functionality should be automatically shipped now via the default dbc.theme included, and styling is required for any custom component that gets added by the user 👍 

```
Please note that Vizro is currently not compatible with [Dash Bootstrap Themes](https://dash-bootstrap-components.opensource.faculty.ai/docs/themes/).
    Adding a Bootstrap stylesheet will have no visual effect on the [components](https://vizro.readthedocs.io/en/stable/pages/user_guides/components/) included in Vizro.
```

vizro-core/src/vizro/models/_components/card.py
nadijagraca: 2. I did try out the `active` property, but it does not cover fully our use case. We need the styling to be permanent as long as the user is on that page. When using `active` property, the styling would go away as soon as you click anywhere on the page.  
3. I remember there were some problem with it in the old repo, but I cannot recall what was the issue. I didn't try implementing it in the new repo.

huong-li-nguyen: @antonymilne @nadijagraca - I actually just tried out the NavLink in the Accordion and that seems to work perfectly with the active CSS property only! This means we can actually replace the dbc.Buttons with dbc.NavLinks and could even remove the page-id argument from the builds because the highlighting of the active NavLink could now be achieved via CSS only. I would split this into a separate PR though that I can do next @antonymilne - so would leave the current version unchanged.

vizro-core/src/vizro/static/css/layout.css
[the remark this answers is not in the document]
antonymilne: Yeah, I'm all in for alphabetical ordering 👍 

I thought prettier would handle css too (it's mentioned on their site anyway), and https://github.com/pre-commit/mirrors-csslint looks promising - [they have an `order-alphabetical` option](https://github.com/CSSLint/csslint/wiki/Rules).

Not a big deal though, completely up to you 🙂

vizro-core/src/vizro/models/_components/card.py
antonymilne: > Yes, that's a great idea actually! That also works perfectly fine, I've implemented the change [here](https://github.com/mckinsey/vizro/pull/311/commits/ca213993b2da2a562c71832519353c1638eefb4c) and also got rid of one of the redundant htmls! Do you think it's worth implementing the CSS name changing from underscore to dashes already? Or better to do in one separate PR? Or not doing at all if we at some point switch to BEM / Bootstrap syntax?

Thank you, this looks much better 👍 If you think the underscore to dashes is a change worth doing regardless of whether/when we switch to BEM/Bootstrap then let's do it in a separate PR - I don't really mind if we make the change but have a weak preference in favour of doing so if it doesn't take us any further away from BEM/Bootstrap, which I don't think it does.

> @antonymilne @nadijagraca - I actually just tried out the NavLink in the Accordion and that seems to work perfectly with the active CSS property only! 🥳 This means we can actually replace the dbc.Buttons with dbc.NavLinks and could even remove the page-id argument from the builds because the highlighting of the active NavLink could now be achieved via CSS only. I would split this into a separate PR though that I can do next @antonymilne - so would leave the current version unchanged.

This is great news. Definitely let's do it in a separate PR since I'll have a couple more comments on it I think.

vizro-core/src/vizro/static/css/layout.css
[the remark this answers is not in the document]
huong-li-nguyen: Ahh, nice! I'll check that out, but would also leave it for another PR!

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -27,21 +26,22 @@ class Card(VizroBaseModel):
     text: str = Field(
         ..., description="Markdown string to create card title/text that should adhere to the CommonMark Spec."
     )
-    href: str = Field(
-        "",
+    href: Optional[str] = Field(
+        None,
```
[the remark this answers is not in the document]
antonymilne: I think the right solution here would be to keep as `""` but just change `if self.href is not None` to `if self.href`? This way the the `dcc.Link` should never get generated with an empty string.

AnnMarieW:
@huong-li-nguyen Nice work!  This is awesome!

In `dashboard.js` if you update this callback function, it will  will use the correct light/dark color mode for the Boootstrap theme:

```
export function _update_dashboard_theme(checked) {
  document.documentElement.setAttribute('data-bs-theme', checked ? 'light' : 'dark');
  return checked ? "vizro_light" : "vizro_dark";
}

```

vizro-core/src/vizro/models/_components/card.py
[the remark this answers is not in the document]
huong-li-nguyen: Ah yes, that works! Thank you! ❤️

huong-li-nguyen:
> @huong-li-nguyen Nice work! This is awesome!
> 
> In `dashboard.js` if you update this callback function, it will will use the correct light/dark color mode for the Boootstrap theme:
> 
> ```
> export function _update_dashboard_theme(checked) {
>   document.documentElement.setAttribute('data-bs-theme', checked ? 'light' : 'dark');
>   return checked ? "vizro_light" : "vizro_dark";
> }
> ```

Ahh, super smart! Thank you!!! 🚀 💯  Hope we can get a better long-term solution in place soon!

vizro-core/src/vizro/static/css/button.css
```diff
@@ -8,41 +10,28 @@
...
-  height: 32px;
-  letter-spacing: var(--letter-spacing-body-link-02);
-  line-height: var(--text-size-05);
-  padding: var(--spacing-01) var(--spacing-03);
+#page-container .button_primary:active {
+  background: var(--fill-active);
+  color: var(--text-contrast-primary);
 }
 
-.button_tertiary {
```
huong-li-nguyen, line 36: @nadijagraca - just FYI: I've removed the secondary and tertiary button because we didn't use it in our repository and I'd rather not have it in there then. Also I've removed the focus state. I had a quick catch-up with Pruthvi and he said that the design system will remove all focus states.

vizro-core/src/vizro/static/css/button.css
```diff
@@ -8,41 +10,28 @@
...
-  text-decoration-line: underline;
-}
-
-.button_primary:focus {
-  box-shadow: 0 0 0 2px var(--focus-focus) inset;
+#page-container .button_primary:hover {
+  background: linear-gradient(
+      var(--state-overlays-contrast-hover),
+      var(--state-overlays-contrast-hover)
+    ),
```
huong-li-nguyen, line 20: @nadijagraca - I can explain in our catch-up, but basically, we have been doing some of the hover states wrong 😄 It wasn't clear from the Figma specifications for me either though

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -35,13 +34,14 @@ class Card(VizroBaseModel):
...
-                href=get_relative_path(self.href) if self.href.startswith("/") else self.href, className="card_button"
-            ),
-            className="button_container",
+        card_content = (
+            dcc.Link(
+                text,
+                href=get_relative_path(self.href) if self.href.startswith("/") else self.href,
+                className="card-link",
+            )
+            if self.href
```
nadijagraca, line 43: It does not have to be part of this PR, but it would be good if we could provide card.href with similar validation we have for page.path. At this moment, if user has path with an empty space it will be replaced with "-". But if users tries to add the same path to the card.href it will not work and will lead to 404 page.

nadijagraca (approved):
Looks good, I just have one minor suggestion but does not have to be included in this PR. 🚀

vizro-core/src/vizro/models/_components/card.py
huong-li-nguyen: Yes, we are already aware! @l0uden already raised the issue [here](https://github.com/mckinsey/vizro/issues/285) - it's a ticket that's currently deprioritized but that anyone in the team can take on when they like and have capacity :)

PR #319, opened 2024-02-19
[QA] Add unit tests and docs for `DatePicker`

nadijagraca:
## Description
Here are the latest changes to the core code and tests that should allow DatePicker to run smoothly. 
NB: Filter interaction where the source is data in the datetime format will be implemented separately. 

## Relevant description
Added temporal selector `DatePicker` for `Filter` and `Parameter`.  

DatePicker is a single model, but based on value of argument `range` will either return `dmc.DatePicker` or `dmc.DateRangePicker`. Default value of `range` is `True` and the model will return `dmc.DateRangePicker` if range is not provided, or if the selector is not defined. 

**Options where `DateRangePicker` is returned:**
1. User does not define the `Filter.selector` but filter `column` is `datetime` dtype.
2. User defines `vm.DatePicker` as `Filter.selector` without specifying `range` argument or by setting `range=True`
 
**Option where `DatePicker` is returned:**
1. User defines `vm.DatePicker` as `Filter.selector` and sets `vm.DatePicker` `range` argument as `False`. 

`vm.DatePicker.value` supported format are all inputs that could be translated to `datetime.date`. See more [here](https://docs.pydantic.dev/latest/api/standard_library_types/#datetime-types)

## Screenshot
image

---

image

---

## Notice


vizro-core/docs/pages/user_guides/filters.md
```diff
@@ -65,7 +65,8 @@ automatically based on the target column, e.g. a dropdown for categorical data o
 ## Changing Selectors
```
stichbury, line 65:
```suggestion
## Changing selectors
```

stichbury: I'll be suggesting that we use sentence case (only first word capitalised) for subheadings and headings, but there's no style guide at present and there seems to be a mix on the site. Let's be consistent in this page and use it anyway, but no sweat about changing anything else until I run a consistency check across the docs 😅

vizro-core/docs/pages/user_guides/selectors.md
```diff
@@ -43,3 +43,12 @@ For more details, kindly refer to the documentation of the underlying dash compo
     [`dcc.RangeSlider`](https://dash.plotly.com/dash-core-components/rangeslider) component, which you can circumvent by adapting the `step` size accordingly.
 
 To enhance existing selectors, please see our How-to-guide on creating [custom components](custom_components.md)
+
+## Temporal Selectors
```
stichbury, line 47:
```suggestion
## Temporal selectors
```

vizro-core/docs/pages/user_guides/selectors.md
```diff
@@ -43,3 +43,12 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+
+For more details, kindly refer to the documentation of the underlying dash components:
```
stichbury, line 49:
```suggestion
For more details, refer to the Dash components documentation:
```

stichbury (approved):
LGTM 🌟 🏆

huong-li-nguyen (commented):
I didn't look through this in detail yet, but could you add an example to the **feature demo dashboard**? Would be great if we can have the pydantic + yaml version there 🙏

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -44,6 +56,16 @@ def _filter_isin(series: pd.Series, value: MultiValueType) -> pd.Series:
...
+def _filter_between_date(series: pd.Series, value: List[float]) -> pd.Series:
+    date_series, date_value = [_convert_to_date(value=elem) for elem in (series, value)]
+    return _filter_between(series=date_series, value=date_value)
+
+
+def _filter_isin_date(series: pd.Series, value: MultiValueType) -> pd.Series:
+    date_series, date_value = [_convert_to_date(value=elem) for elem in (series, value)]
+    return _filter_isin(series=date_series, value=date_value)
+
+
```
petar-qb, line 68: I would like to get an opinion on the two added filter functions. Does it seem like the right step or not? 

I decided to use two new filter functions instead of checking if `date_series` and `date_value` can be converted to datetime format (again) (e.g. inside the `_filter_between` and `_filter_isin` functions or inside the `_actions_utils.py`), because it doesn't seem like the best practice checking the same thing multiple times. Currently, this check is only conducted once, as a part of the `Filter.pre_build` stage (`_set_column_type` method).

Also, checking if the filter_function input argument `series` can be converted to the datetime could cause bugs since the input `series` is just the part of the already filtered data_frame (if more filters are on the screen).

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
```diff
@@ -111,6 +161,16 @@ def test_set_slider_values_incompatible_column_type(self, test_input, managers_o
         ):
             filter.pre_build()
 
+    def test_set_datepicker_values_incompatible_column_type(self, managers_one_page_two_graphs):
+        filter = vm.Filter(column="country", selector=vm.DatePicker())
+        model_manager["test_page"].controls = [filter]
+        with pytest.raises(
+            ValueError,
+            match=f"Chosen selector {vm.DatePicker().type} "
```
maxschulz-COL, line 169: I suppose what I meant earlier is that we could already raise this for any non datetime column to make our lives a bit easier

vizro-core/tests/unit/vizro/models/_controls/test_parameter.py
```diff
@@ -100,6 +100,16 @@ def test_set_categorical_selectors_with_missing_options(self, test_input):
         ):
             parameter.pre_build()
 
+    def test_set_datepicker_values_invalid(self):
+        parameter = Parameter(targets=["scatter_chart.x"], selector=vm.DatePicker())
```
maxschulz-COL, line 104: bit of a random one no? This is essentially testing the Selector again, and not the parameter no? We may actually have some tests like that here, so do not mind keeping it, just curious what you think

Should we not rather only test if the whoe parameter action works with a datepicker as selector?

maxschulz-COL (commented):
Good stuff! And I can see how allof this must have been quite tricky. Since I was not involved in previous reviews, forgive if I am saying something that has already been discussed/seems silly or obvious.

Given that we are not in a rush, I would say we try and implement some of the recommendations

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: Both approaches seem correct to me. The only question is, do we want to treat any type of date format as a date column or only treat the "YYYY-MM-DD" format as a date column, otherwise it's just a string and a categorical filter?

@antonymilne  @huong-li-nguyen @nadijagraca
What is you opinion? Currently, the date_picker will automatically be set as a Filter selector if the filtered column contains any date format (data such as: ["1-1-1952", "2-1-1952", "3-1-1952"])

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -36,6 +36,18 @@
 }
 
 
+def _convert_to_date(value):
```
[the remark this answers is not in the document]
maxschulz-COL: Ok fair enough, but I would say that it's not only for public functions that we have type hints! For mypy to be efficient, all function should be type hinted

vizro-core/tests/unit/vizro/models/_components/form/test_date_picker.py
```diff
@@ -0,0 +1,171 @@
...
+    def test_validate_datepicker_range_valid(self, range, value):
+        date_picker = vm.DatePicker(min="2024-01-01", max="2024-04-01", range=range, value=value)
+        assert date_picker.range == range
+
+    @pytest.mark.parametrize(
+        "range, value",
+        [
+            (True, "2024-01-01"),  # range True produces DateRangePicker, value needs to be list with 2 dates
+            (True, ["2024-01-01"]),  # range True produces DateRangePicker, value needs to be list with 2 dates
+            (False, ["2024-01-01"]),  # range False produces DatePicker, value needs to be single date
```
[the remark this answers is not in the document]
petar-qb: I didn't discussed this, and I guess that the `range` dictates the rules. @huong-li-nguyen @nadijagraca correct me if I'm wrong. 

Yes, it's possible to cast the value `list[str]` to `str` if the `range` is `False` or to set the range depending on what is the propagated value format. However that would increase the complexity for us and potentially for the user to understand I guess.

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
petar-qb: Sorry, but I don't understand. Isn't that we raise and exception for ant non datetime column.

```
    # FILE: filter.py
    def _set_date_picker_values(self):
        if isinstance(self.selector, SELECTORS["temporal"]):
            if self._column_type != "temporal":
                raise ValueError(
                    f"Chosen selector {self.selector.type} is not compatible "
                    f"with {self._column_type} column '{self.column}'."
                )
           ...
```

vizro-core/tests/unit/vizro/models/_components/form/test_date_picker.py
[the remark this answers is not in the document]
maxschulz-COL: I think it's less for smth that the user should understand, just one more option that would not fail. As in when they choose `range=False` but provide a list, if the list is `len(list) == 1`, we could make it work as a nice feature for the user

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
maxschulz-COL: I was referring to this comment: https://github.com/mckinsey/vizro/pull/319#discussion_r1517664691

vizro-core/src/vizro/models/_controls/filter.py
nadijagraca: If we already have a functionality to treat any type of date format as a date column I would leave it, but Max's approach works for me as well.

vizro-core/tests/unit/vizro/models/_controls/test_parameter.py
petar-qb: This test actually tests a newly added method `_set_date_picker_values` inside the Parameter model. 

I'm not entirely convinced about the necessity of testing the date picker as an input for the parameter action. This is because the parameter action is already tested and its functionality doesn't rely on the input type. The input is just propagated to the target without significant impact.

vizro-core/examples/features/app.py
```diff
@@ -404,6 +409,9 @@
             column="size",
             selector=vm.Slider(title="Slider (Tips - size)", step=1, value=2),
         ),
+        vm.Filter(
```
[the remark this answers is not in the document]
petar-qb: Thank you! I forgot to test the example after updating the file 😄. I believe this error is valid since `year' in this case is a datetime type column.

I don't think we want to allow our users to explicitly set any selector type if they want (regardless of column type). Hence we put validations like this. Do you think we should? 

I updated the column for this filter to `lifeExp`. 😄

vizro-core/tests/unit/vizro/models/_components/form/test_date_picker.py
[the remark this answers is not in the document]
nadijagraca: We could add an additional check in validate_range to accommodate Max's suggestion, but the main question is what combinations do we want to allow. 
If we allow `range=False` and `value=["2024-01-01"]`,  do we want to allow `range=True` and `value=["2024-01-01"]` and handle this case in the background as well?

I'm tilting a bit to keeping it simple and having users provide single date for when range=False and list of dates for range=True, but I'm okay with implementing the suggested changed if majority agrees.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -36,6 +36,18 @@
...
+    return pd.to_datetime(value, format="mixed")
+
+
+def _is_convertible_to_date(pd_series: pd.Series) -> bool:
```
antonymilne, line 43: Just FYI because I think we should get rid of this, but this is a very well written function 👏

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -44,6 +56,16 @@ def _filter_isin(series: pd.Series, value: MultiValueType) -> pd.Series:
     return series.isin(value)
 
 
+def _filter_between_date(series: pd.Series, value: List[str]) -> pd.Series:
+    date_series, date_value = [_convert_to_date(value=elem) for elem in (series, value)]
```
antonymilne, line 60: Also just an FYI because I think we should get rid of this, but this line is super confusing 😅 When it's just two unrelated elements like this I would never do it with a loop because trying to unpack this assignment in my head took way longer than if it was just written like this:
```suggestion
    date_series, date_value = _convert_to_date(series), _convert_to_date(value)
```

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: Here is a brief description of the conclusion:
 - If the column is `pd.api.types.is_datetime64_any_dtype` then it should go to `DatePicker` by default; otherwise it is treated as categorical so will go to `Dropdown` by default
 - No need to look for strings in `yyyy-mm-dd` or any other format. Just treat them as categorical.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -36,11 +36,15 @@
 }
 
 
-def _filter_between(series: pd.Series, value: List[float]) -> pd.Series:
+def _filter_between(series: pd.Series, value: Union[List[float], List[str]]) -> pd.Series:
+    if is_datetime64_any_dtype(series):
+        value = pd.to_datetime(value)
```
maxschulz-COL, line 41: Are we catching the case where this conversion does not work?

vizro-core/tests/unit/vizro/models/_components/form/test_date_picker.py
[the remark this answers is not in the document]
maxschulz-COL: It's fine for me, was just a suggestions. In your other case I would say the failure is clearer, because there is information missing with the second date

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -30,16 +30,30 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+
+- [dmc.DateRangePicker](https://www.dash-mantine-components.com/components/datepicker#daterangepicker)
+- [dmc.DatePicker](https://www.dash-mantine-components.com/components/datepicker)
+
+!!! note
+
+    [`DatePicker`][vizro.models.DatePicker] component is based on 2 underlying components dmc.DatePicker and dmc.DateRangePicker.
+    When [`DatePicker`][vizro.models.DatePicker] is configured with `range=False` argument, underlying component is dmc.DatePicker. Default value of `range` is `True`,
+    meaning [`DatePicker`][vizro.models.DatePicker] by default uses dmc.DateRangePicker.
+    When configuring the [`DatePicker`][vizro.models.DatePicker] make sure to provide your dates for `min`, `max` and `value` arguments in `"yyyy-mm-dd"` format or as datetime type.
```
antonymilne, line 57:
```suggestion
    [`DatePicker`][vizro.models.DatePicker] component is based on two underlying components `dmc.DatePicker` and `dmc.DateRangePicker`.
    When [`DatePicker`][vizro.models.DatePicker] is configured with `range=True` (the default), the underlying component is `dmc.DateRangePicker`; when `range=False` the component is `dmc.DatePicker`.
    When configuring the [`DatePicker`][vizro.models.DatePicker] make sure to provide your dates for `min`, `max` and `value` arguments in `"yyyy-mm-dd"` format or as datetime type.
```

It almost always makes sense to describe the default behaviour first.

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -30,16 +30,30 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
     To our knowledge, this is a current bug in the underlying [`dcc.Slider`](https://dash.plotly.com/dash-core-components/slider) and
     [`dcc.RangeSlider`](https://dash.plotly.com/dash-core-components/rangeslider) component, which you can circumvent by adapting the `step` size accordingly.
 
-To enhance existing selectors, please see our How-to-guide on creating [custom components](custom-components.md)
+## Temporal Selectors
+
+For more details, refer to the documentation of the underlying dash components:
+
+- [dmc.DateRangePicker](https://www.dash-mantine-components.com/components/datepicker#daterangepicker)
+- [dmc.DatePicker](https://www.dash-mantine-components.com/components/datepicker)
```
antonymilne, line 50:
```suggestion
- [`dmc.DateRangePicker`](https://www.dash-mantine-components.com/components/datepicker#daterangepicker)
- [`dmc.DatePicker`](https://www.dash-mantine-components.com/components/datepicker)
```

vizro-core/examples/features/app.py
```diff
@@ -365,19 +367,28 @@
...
             targets=["table-gapminder"],
-            column="year",
-            selector=vm.RangeSlider(title="Range Slider (Gapminder - year)", step=1, marks=None),
+            column="lifeExp",
+            selector=vm.RangeSlider(
+                title="Range Slider (Gapminder - lifeExp)",
+                step=1,
+                marks=None,
+                min=int(gapminder["lifeExp"].min()),
+                max=int(gapminder["lifeExp"].max()) + 1,
```
antonymilne, line 390: Do we need to set these `min`/`max` explicitly for some reason?

vizro-core/src/vizro/models/_controls/filter.py
maxschulz-COL: Or a different thought: when does this even receive a string? The [datepicker](https://github.com/mckinsey/vizro/blob/926037ce8048a79a50299e68c4b5f95e3a7d2b70/vizro-core/src/vizro/models/_components/form/date_picker.py#L41) only allows `date`, so why would this even be a string?

vizro-core/examples/features/yaml_version/app.py
```diff
@@ -10,9 +10,11 @@
 
 data_manager["tips"] = px.data.tips()
 gapminder = px.data.gapminder()
```
antonymilne, line 12: Do we still want this version with `datetime=False`?

vizro-core/pyproject.toml
```diff
@@ -56,7 +56,7 @@ exclude_lines = [
   "if __name__ == .__main__.:",
   "if TYPE_CHECKING:"
 ]
-fail_under = 90
+fail_under = 92
```
antonymilne, line 59: 🎉

vizro-core/src/vizro/models/_components/form/_form_utils.py
```diff
@@ -66,7 +66,7 @@ def validate_max(cls, max, values):
         return max
 
     if values["min"] is not None and max < values["min"]:
-        raise ValueError("Maximum value of slider is required to be larger than minimum value.")
+        raise ValueError("Maximum value of component is required to be larger than minimum value.")
```
antonymilne, line 69:
```suggestion
        raise ValueError("Maximum value of selector is required to be larger than minimum value.")
```

vizro-core/src/vizro/models/_controls/filter.py
antonymilne, line 41: Why do we need this? Won't `value` always be a `date` since it's cast to that by pydantic?

Similarly in `_filter_isin`.

vizro-core/tests/unit/vizro/conftest.py
```diff
@@ -13,6 +13,16 @@ def gapminder():
     return px.data.gapminder()
 
 
+@pytest.fixture
+def gapminder_with_datetime():
+    return px.data.gapminder(datetimes=True)
```
antonymilne, line 18: Similar question as in the features example - do we need both `gapminder` and `gapminder_with_datetime`?

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
```diff
@@ -111,6 +224,16 @@ def test_set_slider_values_incompatible_column_type(self, test_input, managers_o
...
+            f"is not compatible with categorical column '{filter.column}'.",
```
maxschulz-COL, line 233: Shall we just test for both, numerical and temporal, and check for a little broader error message?

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
```diff
@@ -132,6 +255,13 @@ def test_set_slider_values_defaults_min_max_none(self, test_input, gapminder, ma
         assert filter.selector.min == gapminder.lifeExp.min()
         assert filter.selector.max == gapminder.lifeExp.max()
 
+    def test_set_datepicker_values_defaults_min_max_none(self, gapminder_with_datetime, managers_one_page_two_graphs):
+        filter = vm.Filter(column="year", selector=vm.DatePicker())
+        model_manager["test_page"].controls = [filter]
+        filter.pre_build()
+        assert filter.selector.min == gapminder_with_datetime.year.min().to_pydatetime().date()
```
maxschulz-COL, line 262: Silly question, but where does it get it's min and max values. I could only find it for the slider....

vizro-core/tests/unit/vizro/models/_components/form/test_date_picker.py
```diff
@@ -0,0 +1,175 @@
...
+    @pytest.mark.parametrize(
+        "range, value",
+        [
+            (False, "2024-01-01"),
+            (False, date(2024, 1, 1)),
+            (True, ["2024-01-01", "2024-02-01"]),
+            (True, [date(2024, 1, 1), date(2024, 2, 1)]),
+        ],
+    )
+    def test_validate_datepicker_value_valid(self, range, value):
```
antonymilne, line 87: This test is pretty confusing to read because it's handling both the `range=True` case and the `range=False` case in one go and these things can return a list or a single value depending on the parameter.

Also I don't know what the point of testing both `"2024-01-01"` and `date(2024, 1, 1)` is? Isn't that just testing that pydantic is coercing `value` to `date`?

Personally I would simplify this into two tests:
```
def test_validate_datepicker_value_single_valid():
    date_picker = vm.DatePicker(min="2024-01-01", max="2024-02-01", range=False, value=...)
    assert date_picker.value == ...

def test_validate_datepicker_value_range_valid():
    date_picker = vm.DatePicker(min="2024-01-01", max="2024-02-01", range=False, value=...)
    assert date_picker.value == [..., ...]
```

In general it's fine to be less DRY and more explicit when writing tests than writing source code. e.g. `[v.date() for v in value_to_date]` is nice and general and would work for many cases but is less clear than just spelling out the two values you want here explicitly.

vizro-core/tests/unit/vizro/models/_controls/test_parameter.py
maxschulz-COL: Ok that makes sense!

vizro-core/tests/unit/vizro/models/_components/form/test_date_picker.py
```diff
@@ -0,0 +1,175 @@
...
+
+        assert date_picker.value == value_to_date
+
+    @pytest.mark.parametrize(
+        "range, value",
+        [
+            (False, "2024-01-01"),
+            (False, date(2024, 1, 1)),
+            (True, ["2024-01-01", "2024-02-01"]),
+            (True, [date(2024, 1, 1), date(2024, 2, 1)]),
```
antonymilne, line 104: Similar comment here about why we want to test both a string and a `date` object.

vizro-core/tests/unit/vizro/models/_controls/test_parameter.py
```diff
@@ -100,6 +100,16 @@ def test_set_categorical_selectors_with_missing_options(self, test_input):
         ):
             parameter.pre_build()
 
+    def test_set_datepicker_values_invalid(self):
```
maxschulz-COL, line 103:
```suggestion
    def test_datepicker_values_parameter_invalid(self):
```

vizro-core/tests/unit/vizro/models/_components/form/test_date_picker.py
```diff
@@ -0,0 +1,175 @@
...
+        ],
+    )
+    def test_validate_datepicker_range_valid(self, range, value):
```
antonymilne, line 120: Similar comment here about why we want to test both a string and a `date` object.

vizro-core/tests/unit/vizro/models/_components/form/test_date_picker.py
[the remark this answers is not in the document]
antonymilne: Agree with @nadijagraca - this is fine how it is, let's keep it simple in this case. This handling is consistent with how we handle e.g. `RadioItems` where `value` has to be strictly a single value rather than allowing `[value]`. And vice versa with `Checklist` if you provide `value = 1` we don't convert to `[1]` at the moment

vizro-core/examples/features/app.py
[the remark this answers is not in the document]
maxschulz-COL: No I think all good - just wanted to make you aware it didnt run

vizro-core/tests/unit/vizro/models/_components/form/test_date_picker.py
```diff
@@ -0,0 +1,175 @@
...
+        with pytest.raises(ValidationError):
+            vm.DatePicker(min="2024-01-01", max="2024-04-01", range=range, value=value)
+
+
+class TestBuildMethod:
+    @pytest.mark.parametrize(
+        "range, value",
+        [
+            (False, "2023-01-05"),
+            (False, date(2023, 1, 5)),
```
antonymilne, line 143: Similar comment here about why we want to test both a string and a date object.

vizro-core/tests/unit/vizro/models/_components/form/test_date_picker.py
```diff
@@ -0,0 +1,175 @@
...
+        ],
+    )
+    def test_validate_datepicker_value_valid(self, range, value):
+        date_picker = vm.DatePicker(min="2024-01-01", max="2024-02-01", range=range, value=value)
+
+        value_to_date = pd.to_datetime(value).to_pydatetime()
+        if isinstance(value_to_date, Collection):
+            value_to_date = [v.date() for v in value_to_date]
+        else:
+            value_to_date = value_to_date.date()
```
maxschulz-COL, line 94: We may have discussed this before, but this type of test still trips me up a little: we are doing a whole calculation to arrive at the expected resullt, which I think defeats the purpose of tests in some way.

Suggestion: split this into two tests (one for range and one for normal) and just test for the correct end result?

antonymilne (approved):
This looks great, thanks for all your work on this @nadijagraca and @petar-qb!

I just left some small comments and questions but nothing big.

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: If `is_datetime64_any_dtype(series)` is True, this means that the `value` comes from the DatePicker UI component, which is always in string format that can be converted to a datetime object. We know the value comes from the DatePicker UI component because `is_datetime64_any_dtype(series)` is True, and we only allow the DatePicker selector to be set.

petar-qb: > The [datepicker](https://github.com/mckinsey/vizro/blob/926037ce8048a79a50299e68c4b5f95e3a7d2b70/vizro-core/src/vizro/models/_components/form/date_picker.py#L41) only allows date, so why would this even be a string?

Yes, the DatePicker `value` allows values that are convertible to `date` see the [link](https://docs.pydantic.dev/latest/api/standard_library_types/#datetime-types), but inside the callback context, selected value always comes in the string format. More exactly in the 'yyyy-mm-dd' (or ['yyyy-mm-dd', 'yyyy-mm-dd']) format.

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
petar-qb: Good spot Max! I tried to refactor some tests for testing `column_type` <-> `selector` (in)compatibility.  

@huong-li-nguyen @antonymilne can you also help us to double check the desired requirements. 
Are the following statements correct? 
If yes, then we have a bug because currently we allow to set a `categorical selector` for `numerical/temporal column_type`.

1. column_type is `numerical`
  1.1. allowed selectors: (RangeSlider, Slider)
  1.2. disallowed selectors: (Dropdown, Checklist, RadioItems, DatePicker)
2. column_type is `temporal`
  2.1. allowed selectors: (DatePicker)
  2.2. disallowed selectors: (Dropdown, Checklist, RadioItems, RangeSlider, Slider)
3. column_type is `categorical` 
  3.1. allowed selectors: (Dropdown, Checklist, RadioItems)
  3.2. disallowed selectors: (RangeSlider, Slider, DatePicker)

huong-li-nguyen: 1 and 3 seem right.

For 2) What happens if you add temporal data into the RangeSlider or Slider? If the underlying dash component can visualise it, then I would allow it as a selector. If not, then not.

So generally, the disallowed selectors are the ones where the underlying dash component cannot handle that data type and therefore does not visualise it properly.

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
petar-qb: Inside the `Filter._set_date_picker_values` method, that is called inside the `Filter._pre_build` phase.

vizro-core/examples/features/app.py
petar-qb: I did it that way only so that min and max would be integers (it looks nicer on the screen).

vizro-core/pyproject.toml
petar-qb: It's even `93` now. 😄

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: Max has already asked the same in the https://github.com/mckinsey/vizro/pull/319#discussion_r1523452492. Let's continue the discussion there.

vizro-core/examples/features/app.py
huong-li-nguyen: I would probably remove that, otherwise the yaml and python version are out of sync, because you can't insert python functions in yaml. 

Could you double-check that the two versions (pydantic + yaml) output the same result?

petar-qb: You're right, the yaml example has the different output. So, let's revert these changes 😄.

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
antonymilne: I think this is probably too restrictive. Just because something is numerical doesn't mean it's continuous or that every value in the range is valid. Hence someone might wish to use a categorical selector for a numerical column.

e.g. I want a form component for someone to choose their favourite prime number out of 6, 28, 496, 8128. It would make more sense for this to be a categorical selector like `RadioItems` rather than a slider.

I think the requirements should be:
1. column_type is `numerical`
  1.1. allowed selectors: (RangeSlider, Slider, Dropdown, Checklist, RadioItems)
  1.2. disallowed selectors: (DatePicker)
2. column_type is `temporal`
  2.1. allowed selectors: (DatePicker, Dropdown, Checklist, RadioItems)
  2.2. disallowed selectors: (RangeSlider, Slider) - but allowed if these work ok for dates actually
3. column_type is `categorical` 
  3.1. allowed selectors: (Dropdown, Checklist, RadioItems)
  3.2. disallowed selectors: (RangeSlider, Slider, DatePicker)

Looked at another way:
* the categorical selectors (Checklist, Dropdown, RadioItems) should work for everything because you can treat _any_ data with finite number of values as categorical if you really want to.
* the numerical selector (RangeSlider, Slider) only makes sense for data that can be ordered (ordinal). _Generally speaking_ this is numbers and dates, but in theory it could be something else e.g. T-shirts sizes XS, S, M, L, XL
* the date selector DatePicker only makes sense for dates and nothing else

Currently the only restriction we have (pre-datepicker) is that numerical selector (RangeSlider, Slider) only works for numerical data. This is already slightly overly restrictive (due to T-shirt size example) but that's ok because it's an edge case that no one really cares about.

Basically we should let people use whatever selector they like _so long as the underlying functionality works ok_.  Our code for numerical selector relies on finding the `min` and `max` of a column, and for ordinal data like In the T-shirt size example this is tricky to do in pandas, so it's fine that we don't support it. Think of it as like duck typing: a selector has certain requirements that the data must obey, e.g. it defines a `min` or it has a finite number of values. So long as these requirements are met it's ok to use that sort of data in that selector.

vizro-core/examples/features/yaml_version/dashboard.yaml
```diff
@@ -429,6 +431,13 @@ pages:
         targets:
           - table-tips
         type: filter
+      - column: year
+        selector:
+          title: Date Picker (Gapminder - year)
```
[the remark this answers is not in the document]
petar-qb: The fix will come from https://github.com/mckinsey/vizro/pull/371.

vizro-core/pyproject.toml
maxschulz-COL: Then we should raise it to 93!

petar-qb: I already did it 😄

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -26,20 +26,106 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+For more details, refer to the documentation of the underlying Dash components:
+
+- [`dmc.DateRangePicker`](https://www.dash-mantine-components.com/components/datepicker#daterangepicker)
+- [`dmc.DatePicker`](https://www.dash-mantine-components.com/components/datepicker)
+
+!!! note
+
+    [`DatePicker`][vizro.models.DatePicker] component is based on two underlying components `dmc.DatePicker` and `dmc.DateRangePicker`.
+    When [`DatePicker`][vizro.models.DatePicker] is configured with `range=True` (the default), the underlying component is `dmc.DateRangePicker`; when `range=False` the component is `dmc.DatePicker`.
+    When configuring the [`DatePicker`][vizro.models.DatePicker] make sure to provide your dates for `min`, `max` and `value` arguments in `"yyyy-mm-dd"` format or as datetime type.
```
maxschulz-COL, line 59:
```suggestion
    The [`DatePicker`][vizro.models.DatePicker] component is based on two underlying components -  `dmc.DatePicker` and `dmc.DateRangePicker`.
    When the [`DatePicker`][vizro.models.DatePicker] is configured with `range=True` (the default), the underlying component is `dmc.DateRangePicker`. When `range=False` the underlying component is `dmc.DatePicker`.
    When configuring the [`DatePicker`][vizro.models.DatePicker] make sure to provide your dates for `min`, `max` and `value` arguments in `"yyyy-mm-dd"` format or as `datetime` type.
```
Can we add what `datatime` type is with a link. Like convertible by xy function or so

vizro-core/examples/features/app.py
```diff
@@ -365,19 +365,26 @@
...
     controls=[
         vm.Filter(
             targets=["table-gapminder"],
-            column="year",
-            selector=vm.RangeSlider(title="Range Slider (Gapminder - year)", step=1, marks=None),
+            column="lifeExp",
+            selector=vm.RangeSlider(
+                title="Range Slider (Gapminder - lifeExp)",
+                step=1,
+                marks=None,
```
huong-li-nguyen, line 386: Did you set the same for the yaml example? Otherwise remove or bring over to the yaml example

vizro-core/examples/features/app.py
```diff
@@ -404,6 +411,9 @@
             column="size",
             selector=vm.Slider(title="Slider (Tips - size)", step=1, value=2),
         ),
+        vm.Filter(
+            targets=["table-gapminder"], column="year", selector=vm.DatePicker(title="Date Picker (Gapminder - year)")
```
huong-li-nguyen, line 415: The year column is actually not a very nice example for this. I like your dev example more for example. What I would suggest is you use the `px.data.stocks()`, create a line chart and add it to this page. And this Filter will then take the date column and filter on that data set. Then just re-name to "Date Picker (Stocks - date)"

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -26,20 +26,106 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+
+To utilize `temporal` [`Filter`][vizro.models.Filter] selectors, the filtered column must be of `datetime` dtype,
```
maxschulz-COL, line 61:
```suggestion
To utilize temporal [`Filter`][vizro.models.Filter] selectors, the filtered column must be of `datetime` format,
```

vizro-core/examples/features/app.py
huong-li-nguyen: Don't forget to create the yaml example out of that as well then :)

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -26,20 +26,106 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+indicating that [pandas.api.types.is_datetime64_any_dtype()](https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_datetime64_any_dtype.html) must return `True` for the filtered column.
+
+!!! note
```
maxschulz-COL, line 64:
```suggestion
!!! tip
```

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -26,20 +26,106 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+
+    Dataframe column type can be changed to datetime using:
```
maxschulz-COL, line 66:
```suggestion
    `pd.DataFrame` column types can be changed to `datetime` using:
```
but please check this again, currently it renders ugly

maxschulz-COL: maybe even skip this, or make this a one liner with the first option only for simplification

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -26,20 +26,106 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+     - [pandas.to_datetime()](https://pandas.pydata.org/docs/reference/api/pandas.to_datetime.html) or
+     - [pandas.DataFrame.astype()](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.astype.html) with `datetime64` as the argument.
+
```
maxschulz-COL, line 69:
```suggestion

## Example of default Filter selectors
```

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -26,20 +26,106 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+Below is an example where all selectors are automatically set by default based on the type of the filtered column.
+
+!!! example "Default Filter selectors"
```
maxschulz-COL, line 72: Great idea of an example! I would use a well known DF though, so the yaml version is easier repeatable?

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -26,20 +26,106 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+To utilize `temporal` [`Filter`][vizro.models.Filter] selectors, the filtered column must be of `datetime` dtype,
+indicating that [pandas.api.types.is_datetime64_any_dtype()](https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_datetime64_any_dtype.html) must return `True` for the filtered column.
+
+!!! note
+
+    Dataframe column type can be changed to datetime using:
+     - [pandas.to_datetime()](https://pandas.pydata.org/docs/reference/api/pandas.to_datetime.html) or
+     - [pandas.DataFrame.astype()](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.astype.html) with `datetime64` as the argument.
+
+Below is an example where all selectors are automatically set by default based on the type of the filtered column.
```
huong-li-nguyen, line 70:
```suggestion
### Default selectors

If you don't specify a selector, a default selector is applied based on the data type of the provided data column. 

Default selectors for:
- categorical data: vm.Dropdown
- numerical data: vm.RangeSlider
- temporal data: vm.DatePicker(range=True)

...
```

huong-li-nguyen: Below that section you could add some information of how we identify categorical/numerical/temporal data, so basically what you have above with the pandas.api.types etc.

huong-li-nguyen (approved):
LGTM 👍 Have a few suggestion for the feature example dashboard and some re-shuffling of the docs

maxschulz-COL (commented):
Will now check the rest, but here my comments about the docs 💪

maxschulz-COL (approved):
Good from me too! If we address the docs and feature dashboard comments, then let's goooo 🚀

vizro-core/src/vizro/models/_controls/filter.py
antonymilne, line 41:
```suggestion
        # Each value will always have time 00:00:00. In order for the filter to include all times during
        # the end date value[1] we need to remove the time part of every value in series so that it's 00:00:00.
        value = pd.to_datetime(value)
        series = pd.to_datetime(series.dt.date)
```

This is my suggested solution here so that the same `return` line below works for both dates and numbers.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -36,11 +36,15 @@
...
     return series.between(value[0], value[1], inclusive="both")
 
 
 def _filter_isin(series: pd.Series, value: MultiValueType) -> pd.Series:
+    if is_datetime64_any_dtype(series):
+        value = pd.to_datetime(value)
```
antonymilne, line 47: I think we need the same conversion of `series` here too to ensure that all times in the given day are included.

vizro-core/docs/pages/user-guides/selectors.md
petar-qb: There is no well known DF [here](https://plotly.com/python-api-reference/generated/plotly.express.data.html) that contains fine temporal, categorical and numerical columns together.

The closest we have is `px.data.gapminder(datetimes=True)` which we avoid using (https://github.com/mckinsey/vizro/pull/319#discussion_r1525060748) in the examples since datetimes are around years with `Jan-01`. So to represent DatePicker selector we use `px.data.stocks` DF in the examples (e.g. features examples).

vizro-core/docs/pages/user-guides/selectors.md
petar-qb: Can you take a quick look at the docs and see if everything is fine now and if I still need to add a datetime type link?

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
petar-qb: Thanks for the input! I solved it, and it looks much cleaner to me now. What do you think? https://github.com/mckinsey/vizro/pull/319/commits/8f5a011b0a0e54fcac67552aab4941e4342d3a1b

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -26,20 +26,117 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+
+## Default selectors
+
+If you don't specify a selector, a default selector is applied based on the data type of the provided data column.
+
+Default selectors for:
+
+ - categorical data: vm.Dropdown
+ - numerical data: vm.RangeSlider
+ - temporal data: vm.DatePicker(range=True)
```
maxschulz-COL, line 66:
```suggestion
 - categorical data: [`Dropdown`][vizro.models.Dropdown]
 - numerical data: [`RangeSlider`][vizro.models.RangeSlider]
 - temporal data: [`DatePicker(range=True)`][vizro.models.DatePicker]```

File: vizro-core/docs/pages/user-guides/selectors.md [hunk 40]
```diff
@@ -26,20 +26,117 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+## Temporal Selectors
+
+For more details, refer to the documentation of the underlying Dash components:
+
+- [`dmc.DateRangePicker`](https://www.dash-mantine-components.com/components/datepicker#daterangepicker)
+- [`dmc.DatePicker`](https://www.dash-mantine-components.com/components/datepicker)
+
+!!! note
+
+    The [`DatePicker`][vizro.models.DatePicker] component is based on two underlying components -  `dmc.DatePicker` and `dmc.DateRangePicker`.
```
maxschulz-COL, line 54: This still renders weird for me:

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -26,20 +26,117 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+    When the [`DatePicker`][vizro.models.DatePicker] is configured with `range=True` (the default), the underlying component is `dmc.DateRangePicker`. When `range=False` the underlying component is `dmc.DatePicker`.
+    When configuring the [`DatePicker`][vizro.models.DatePicker] make sure to provide your dates for `min`, `max` and `value` arguments in `"yyyy-mm-dd"` format or as `datetime` type.
+
+## Default selectors
+
+If you don't specify a selector, a default selector is applied based on the data type of the provided data column.
+
```
maxschulz-COL, line 61:
```suggestion
If you don't specify a selector, a default selector is applied based on the data type of the provided column.

```

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -11,13 +11,13 @@ For instance, the [`Checklist`][vizro.models.Checklist] functions as a multi-opt
 the [`RadioItem`][vizro.models.RadioItems] serves as a single-option selector by default. However, the
 [`Dropdown`][vizro.models.Dropdown] can function as both a multi-option or single-option selector.
 
-For more details, kindly refer to the documentation of the underlying dash components:
+For more details, refer to the documentation of the underlying Dash components:
```
maxschulz-COL, line 14: Sorry to drag this out, but I have decided that I do not like the presentation here. What bugs me, and I have actively tripped over this, is that these are the docs about our selectors, but the most prominent links are to the underlying components.

What do you think of slightly restructuring this such that the bullets are smth a long those line:

```
For more information, refer to the API reference of the selector, or the documentation of its underlying Dash component:
- [`Checklist`][vizro.models.Checklist] based on [`dcc.Dropdown`](https://dash.plotly.com/dash-core-components/dropdown)
- ...
```

vizro-core/docs/pages/user-guides/selectors.md
maxschulz-COL: If we follow this guidance (https://github.com/mckinsey/vizro/pull/319/files#r1528241690), then we could also shorting this note significantly. I would then just add this explanation below the bullet point about this selector.

The note should then contain the only in my eyes important thing for the reader to note, which is the need to provide the correct date formats.

Given that this section is before the other, I wonder still how to make it clear that `datetime` actually is

vizro-core/docs/pages/user-guides/selectors.md
maxschulz-COL: And in stocks, we are missing the categorical column right?

vizro-core/src/vizro/static/js/models/date_picker.js
```diff
@@ -7,3 +7,13 @@ export function _update_date_picker_values(value, input_store) {
   }
   return [value, value];
 }
+
+export function _update_date_picker_position(clicks) {
```
maxschulz-COL, line 11: Where do we stand regarding the tests of our JS functions? We tested the framework, but haven't gotten to the implementation yet right?

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -35,12 +35,28 @@
     "temporal": (DatePicker,),
 }
 
+ALLOWED_SELECTORS = {
+    "numerical": SELECTORS["numerical"] + SELECTORS["categorical"],
+    "temporal": SELECTORS["temporal"] + SELECTORS["categorical"],
+    "categorical": SELECTORS["categorical"],
+}
+
```
antonymilne, line 43: Can we just modify `SELECTORS` to be this? Having a separate `SELECTORS` and `ALLOWED_SELECTORS` feels confusing - if we really need it then please add a comment explaining why 🙂 

Also would be good to link to my comment in this PR explaining here why we want categorical selectors in all the different types so we can refer back to it in future.

```suggestion
CATEGORICAL_SELECTORS = (Dropdown, ...)
SELECTORS = {
    "numerical": (RangeSlider, Slider, *CATEGORICAL_SELECTORS),
    "temporal": ...,
    "categorical": ...,
}
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -100,67 +116,47 @@ def _set_targets(self):
...
-                    f"Chosen selector {self.selector.type} is not compatible "
-                    f"with {self._column_type} column '{self.column}'."
-                )
+    def _validate_allowed_selector(self):
+        if not isinstance(self.selector, ALLOWED_SELECTORS[self._column_type]):
+            raise ValueError(
+                f"Chosen selector {self.selector.type} is not compatible "
+                f"with {self._column_type} column '{self.column}'. "
+                f"Allowed selectors for {self._column_type} column '{self.column}' "
+                f"are {ALLOWED_SELECTORS[self._column_type]}."
```
antonymilne, line 139: This is a bit too repetitive and verbose currently.
```suggestion
                f"Allowed selectors are {ALLOWED_SELECTORS[self._column_type]}."
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -100,67 +116,47 @@ def _set_targets(self):
...
-                self.selector.min = min(min_values)
-            if self.selector.max is None:
-                self.selector.max = max(max_values)
 
-    def _set_date_picker_values(self):
-        if isinstance(self.selector, SELECTORS["temporal"]):
-            if self._column_type != "temporal":
+            if not (is_numeric_dtype(pd.Series(min_values)) and is_numeric_dtype(pd.Series(max_values))) and not (
+                is_datetime64_any_dtype(pd.Series(min_values)) and is_datetime64_any_dtype(pd.Series(max_values))
+            ):
```
antonymilne, line 153: Maybe just personal preference, but I would De Morgan this since it took me a while to understand and doing it the other way round feels clearer:

```suggestion
            if not (is_numeric_dtype(pd.Series(min_values)) and is_numeric_dtype(pd.Series(max_values)) or                 is_datetime64_any_dtype(pd.Series(min_values)) and is_datetime64_any_dtype(pd.Series(max_values)))
            ):
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -100,67 +116,47 @@ def _set_targets(self):
...
                 raise ValueError(
-                    f"Chosen selector {self.selector.type} is not compatible "
-                    f"with {self._column_type} column '{self.column}'."
+                    f"Variable types detected in the shared data column '{self.column}' for targeted charts "
```
antonymilne, line 155:
```suggestion
                    f"Inconsistent types detected in the shared data column '{self.column}' for targeted charts "
```
"Variable" confused me here because I thought it meant "the type of the variable" rather than "not all the same" 😅

vizro-core/src/vizro/models/types.py
```diff
@@ -329,9 +331,9 @@ def wrapped(*args, **kwargs):
 
 
 # Types used for selector values and options. Note the docstrings here are rendered on the API reference.
-SingleValueType = Union[StrictBool, float, str]
+SingleValueType = Union[StrictBool, float, str, Timestamp]
```
antonymilne, line 334:
```suggestion
SingleValueType = Union[StrictBool, float, str, date]
```

Unless I missed something? Otherwise it's inconsistent with the model field type.

Similar for `MultiValueType`.

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
```diff
@@ -83,94 +219,166 @@ def test_set_targets_invalid(self, managers_one_page_two_graphs):
...
+        ],
+    )
+    def test_set_slider_values_shared_column_inconsistent_dtype(self, targets, managers_shared_column_different_dtype):
+        filter = vm.Filter(column="shared_column", targets=targets)
         model_manager["graphs_with_shared_column"].controls = [filter]
         with pytest.raises(
             ValueError,
-            match=f"Non-numeric values detected in the shared data column '{filter.column}' for targeted charts. "
-            f"Please ensure that the data column contains the same data type across all targeted charts.",
+            match=regex_escape(
```
antonymilne, line 307: Why is this needed? It just feels a bit weird seeing it imported as `regex_escape` rather than using it as `re.escape` or adding a couple of backslashes.

antonymilne (commented):
The tidy of `SELECTORS` generally looks like a big improvement 👍  I just had a few suggestions on it.

vizro-core/docs/pages/user-guides/filters.md
```diff
@@ -3,7 +3,7 @@
 This guide shows you how to add filters to your dashboard. One main way to interact with the charts/components on your page is by filtering the underlying data. A filter selects a subset of rows of a component's underlying DataFrame which alters the appearance of that component on the page.
 
 The [`Page`][vizro.models.Page] model accepts the `controls` argument, where you can enter a [`Filter`][vizro.models.Filter] model.
-This model allows the automatic creation of selectors (e.g. Dropdown, RadioItems, Slider, ...) that let a dashboard user filter with the charts/components on the screen.
+This model allows the automatic creation of [selectors](../user-guides/selectors.md) (e.g. Dropdown, RadioItems, Slider, ...) that let a dashboard user filter with the charts/components on the screen.
```
stichbury, line 6:
```suggestion
This model allows the automatic creation of [selectors](../user-guides/selectors.md) (e.g. Dropdown, RadioItems, Slider, ...) that operate on the charts/components on the screen.
```

vizro-core/docs/pages/user-guides/filters.md
```diff
@@ -14,7 +14,7 @@ To add a filter to your page, do the following:
 - configure the `column` argument, which denotes the target column to be filtered
 
 By default, all components on a page with such a `column` present will be filtered. The selector type will be chosen
-automatically based on the target column, e.g. a dropdown for categorical data or a range slider for numerical data.
+automatically based on the target column, e.g. a dropdown for categorical data, a range slider for numerical data or a date picker for temporal data.
```
stichbury, line 17:
```suggestion
automatically based on the target column, e.g. a dropdown for categorical data, a range slider for numerical data, or a date picker for temporal data.
```

vizro-core/docs/pages/user-guides/filters.md
```diff
@@ -65,7 +65,7 @@ automatically based on the target column, e.g. a dropdown for categorical data o
 ## Changing selectors
 
 If you want to have a different selector for your filter, you can provide the `selector` argument of the [`Filter`][vizro.models.Filter] with a different selector model.
-Currently available selectors are [`Checklist`][vizro.models.Checklist], [`Dropdown`][vizro.models.Dropdown], [`RadioItems`][vizro.models.RadioItems], [`RangeSlider`][vizro.models.RangeSlider] and [`Slider`][vizro.models.Slider].
+Currently available selectors are [`Checklist`][vizro.models.Checklist], [`Dropdown`][vizro.models.Dropdown], [`RadioItems`][vizro.models.RadioItems], [`RangeSlider`][vizro.models.RangeSlider], [`Slider`][vizro.models.Slider] and [`DatePicker`][vizro.models.DatePicker].
```
stichbury, line 68:
```suggestion
Currently available selectors are [`Checklist`][vizro.models.Checklist], [`Dropdown`][vizro.models.Dropdown], [`RadioItems`][vizro.models.RadioItems], [`RangeSlider`][vizro.models.RangeSlider], [`Slider`][vizro.models.Slider], and [`DatePicker`][vizro.models.DatePicker].
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -100,67 +116,47 @@ def _set_targets(self):
...
+        if not isinstance(self.selector, ALLOWED_SELECTORS[self._column_type]):
+            raise ValueError(
+                f"Chosen selector {self.selector.type} is not compatible "
+                f"with {self._column_type} column '{self.column}'. "
+                f"Allowed selectors for {self._column_type} column '{self.column}' "
+                f"are {ALLOWED_SELECTORS[self._column_type]}."
+            )
+
+    def _set_numerical_and_temporal_selectors_values(self):
+        if isinstance(self.selector, (SELECTORS["numerical"] + SELECTORS["temporal"])):
```
antonymilne, line 143:
```suggestion
        if isinstance(self.selector, SELECTORS["numerical"] + SELECTORS["temporal"]):
```

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -26,20 +26,117 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+
+Default selectors for:
+
+ - categorical data: vm.Dropdown
+ - numerical data: vm.RangeSlider
+ - temporal data: vm.DatePicker(range=True)
+
+Categorical selectors can be used independently of the data type of the column being filtered.
+
+To utilize numerical [`Filter`][vizro.models.Filter] selectors, the filtered column must be of `numeric` format,
```
stichbury, line 70:
```suggestion
To use numerical [`Filter`][vizro.models.Filter] selectors, the filtered column must be of `numeric` format,
```

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -26,20 +26,117 @@ For more details, kindly refer to the documentation of the underlying dash compo
...
+indicating that [pandas.api.types.is_numeric_dtype()](https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_numeric_dtype.html) must return `True` for the filtered column.
+
+To utilize temporal [`Filter`][vizro.models.Filter] selectors, the filtered column must be of `datetime` format,
```
stichbury, line 73:
```suggestion
To use temporal [`Filter`][vizro.models.Filter] selectors, the filtered column must be of `datetime` format,
```

vizro-core/docs/pages/user-guides/selectors.md
maxschulz-COL: Yes, that looks ok! Although my comment about shorting this still holds :)

vizro-core/src/vizro/static/js/models/date_picker.js
petar-qb: Only for a few client-side JS functions we implemented JS unit tests. We had a separate ticket for JS unit tests implementation that I can't find now, so I'll contact Joe.

vizro-core/docs/pages/user-guides/selectors.md
huong-li-nguyen: You can transform the stocks data set by changing it from wide to long, so the columns 'GOOG', 'AAPL', 'AMZN', 'FB', 'NFLX', 'MSFT' become categories in a column we call "Stocks" and then you have a numeric column with the values :)

Then you have a data set with date, categorical and numeric values.

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: I added the comment from the screenshot inside the `filter.py`:
image

The link from the comment (https://github.com/mckinsey/vizro/pull/319#discussion_r1524888171) actually points to the discussion we had in this PR above.

PR #326, opened 2024-02-21
Replace `dbc.Button` with `dbc.Link` in Navigation

huong-li-nguyen:
## Notice

### DIFF
--- a/vizro-core/src/vizro/models/_navigation/accordion.py
+++ b/vizro-core/src/vizro/models/_navigation/accordion.py
@@ -48,10 +48,10 @@ class Accordion(VizroBaseModel):
 
         accordion_items = []
         for page_group, page_members in self.pages.items():
-            accordion_buttons = self._create_accordion_buttons(pages=page_members)
+            nav_links = self._create_nav_links(pages=page_members)
             accordion_items.append(
                 dbc.AccordionItem(
-                    children=accordion_buttons,
+                    children=nav_links,
                     title=page_group.upper(),
                     class_name="accordion-item-header",
                     item_id=page_group,
@@ -71,15 +71,17 @@ class Accordion(VizroBaseModel):
                     persistence=True,
                     persistence_type="session",
                     always_open=True,
+                    # `active_item` is required to open the accordion automatically when navigating from a homepage
+                    # to any of the pages in the accordion.
                     active_item=active_item,
                 )
             ],
             id="nav-panel",
         )
 
-    def _create_accordion_buttons(self, pages):
-        """Creates a button for each provided page that is registered."""
-        accordion_buttons = []
+    def _create_nav_links(self, pages):
+        """Creates a `NavLink` for each provided page that is registered."""
+        nav_links = []
         for page_id in pages:
             try:
                 page = dash.page_registry[page_id]
@@ -87,13 +89,12 @@ class Accordion(VizroBaseModel):
                 raise KeyError(
                     f"Page with ID {page_id} cannot be found. Please add the page to `Dashboard.pages`"
                 ) from exc
-            accordion_buttons.append(
+            nav_links.append(
                 dbc.NavLink(
                     children=[page["name"]],
-                    key=page["relative_path"],
                     className="accordion-item-link",
                     active="partial",
                     href=page["relative_path"],
                 )
             )
-        return accordion_buttons
+        return nav_links
--- a/vizro-core/src/vizro/models/_navigation/accordion.py
+++ b/vizro-core/src/vizro/models/_navigation/accordion.py
@@ -48,7 +48,7 @@ class Accordion(VizroBaseModel):
 
         accordion_items = []
         for page_group, page_members in self.pages.items():
-            accordion_buttons = self._create_accordion_buttons(pages=page_members, active_page_id=active_page_id)
+            accordion_buttons = self._create_accordion_buttons(pages=page_members)
             accordion_items.append(
                 dbc.AccordionItem(
                     children=accordion_buttons,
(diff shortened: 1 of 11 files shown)


vizro-core/src/vizro/models/_navigation/accordion.py
```diff
@@ -71,29 +71,30 @@ def build(self, *, active_page_id=None):
...
-            accordion_buttons.append(
-                dbc.Button(
+            nav_links.append(
+                dbc.NavLink(
                     children=[page["name"]],
-                    key=page["relative_path"],
-                    className="accordion-item-button",
-                    active=page_id == active_page_id,
+                    className="accordion-item-link",
+                    active="partial",
```
huong-li-nguyen, line 96: This is actually a really nice feature of the `NavLink`, as we don't need the `active_page_id` for this anymore, but it will automatically detect it from the URL. Great push @antonymilne ! 🚀

huong-li-nguyen: There is one buggy behaviour though - while I think "partial" makes sense in the long-run especially if we might add URLs in the style of `../first-page/sub-page/container` in the future, I think switching to "exact" makes more sense for us now. With "partial" currently the homepage or the first page with href` /` is always active 😅  

Which doesn't work so nicely visually

vizro-core/src/vizro/models/_navigation/nav_link.py
```diff
@@ -73,13 +74,13 @@ def build(self, *, active_page_id=None):
...
+            active="partial",
         )
 
         # Only build the nav_selector (id="nav-panel") if the item is active.
         if item_active:
-            return html.Div([button, self._nav_selector.build(active_page_id=active_page_id)])
+            return html.Div([nav_link, self._nav_selector.build(active_page_id=active_page_id)])
 
-        return html.Div(button)
+        return nav_link
```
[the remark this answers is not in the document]
huong-li-nguyen: Never mind - solved! 😄 I'll add a comment why we need it there

petar-qb (approved):
After a detailed review, I can say that the PR is perfect. 🚀

vizro-core/src/vizro/models/_navigation/nav_link.py
```diff
@@ -73,13 +74,15 @@ def build(self, *, active_page_id=None):
...
             active=item_active,
         )
 
         # Only build the nav_selector (id="nav-panel") if the item is active.
         if item_active:
-            return html.Div([button, self._nav_selector.build(active_page_id=active_page_id)])
+            return html.Div([nav_link, self._nav_selector.build(active_page_id=active_page_id)])
 
-        return html.Div(button)
+        return html.Div(nav_link)  # html.Div required to access the nav_link via ID
```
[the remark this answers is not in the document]
huong-li-nguyen: Yes, of course :)

vizro-core/src/vizro/models/_navigation/nav_link.py
```diff
@@ -62,7 +63,7 @@ def build(self, *, active_page_id=None):
                 f"Page with ID {first_page_id} cannot be found. Please add the page to `Dashboard.pages`"
             ) from exc
 
-        button = dbc.Button(
+        nav_link = dbc.NavLink(
```
antonymilne, line 66: My decision to call this `vm.NavLink` feels vindicated now this actually returns a `dbc.NavLink` 😅

antonymilne (approved):
Huge improvement! 💯 I'm very pleased this worked out.

Are there any `dbc.Button` left in our codebase now apart from the `vm.Button` itself?

Some notes for the future, but not for now: 
1. Using `active="partial"` would indeed be nice if we enable hierarchical page URLs, since this would remove the need for `active_page_id` for the `NavLink`. It would still be needed for the accordion, but there's other ways we could get that working without passing the argument through.
2. I'd definitely like to try and migrate to a more "dbc native" style navigation where we use `dbc.Nav` etc. components more. This is a big task though, and might mean that we need to change some of our models, since I don't know whether the current `vm.NavLink`, `vm.Accordion` etc. will work so well then.

huong-li-nguyen:
> Huge improvement! 💯 I'm very pleased this worked out.
> 
> Are there any `dbc.Button` left in our codebase now apart from the `vm.Button` itself?
> 
> Some notes for the future, but not for now:
> 
> 1. Using `active="partial"` would indeed be nice if we enable hierarchical page URLs, since this would remove the need for `active_page_id` for the `NavLink`. It would still be needed for the accordion, but there's other ways we could get that working without passing the argument through.
> 2. I'd definitely like to try and migrate to a more "dbc native" style navigation where we use `dbc.Nav` etc. components more. This is a big task though, and might mean that we need to change some of our models, since I don't know whether the current `vm.NavLink`, `vm.Accordion` etc. will work so well then.



> Huge improvement! 💯 I'm very pleased this worked out.
> 
> Are there any `dbc.Button` left in our codebase now apart from the `vm.Button` itself?
> 
> Some notes for the future, but not for now:
> 
> 1. Using `active="partial"` would indeed be nice if we enable hierarchical page URLs, since this would remove the need for `active_page_id` for the `NavLink`. It would still be needed for the accordion, but there's other ways we could get that working without passing the argument through.
> 2. I'd definitely like to try and migrate to a more "dbc native" style navigation where we use `dbc.Nav` etc. components more. This is a big task though, and might mean that we need to change some of our models, since I don't know whether the current `vm.NavLink`, `vm.Accordion` etc. will work so well then.

No, there are no `dbc.Buttons` anymore 👍  We just use it for the actual Button component and for the 404 landing page where there is an actual Button as well.

PR #333, opened 2024-02-26
[Demo] Update features example dashboard with selectors

huong-li-nguyen:
## Description
- Adds a page with all selectors in the control panel on the left (as Filter only) to the features demo dashboard
- Consolidates all docstrings inside selectors
- Fix CSS for table pagination

## Screenshot
Screenshot 2024-02-26 at 18 10 32

## Notice

### DIFF
--- a/vizro-core/examples/features/app.py
+++ b/vizro-core/examples/features/app.py
@@ -3,6 +3,7 @@
 from time import sleep
 from typing import List, Literal, Optional
 
+import numpy as np
 import pandas as pd
 import plotly.graph_objects as go
 import vizro.models as vm
@@ -12,12 +13,11 @@ from vizro import Vizro
 from vizro.actions import export_data, filter_interaction
 from vizro.models.types import capture
 from vizro.tables import dash_data_table
-import numpy as np
 
 iris = px.data.iris()
 gapminder = px.data.gapminder()
 tips = px.data.tips()
-tips["smoker"] = np.where(tips["smoker"] == 'Yes', 1, 0)
+tips["smoker"] = np.where(tips["smoker"] == "Yes", 1, 0)
 gapminder_2007 = px.data.gapminder().query("year == 2007")
 waterfall_df = pd.DataFrame(
     {
@@ -346,7 +346,8 @@ selectors = vm.Page(
     title="Selectors",
     layout=vm.Layout(grid=[[0], [1], [1], [1], [2], [2], [2]], row_min_height="170px", row_gap="24px"),
     components=[
-        vm.Card(text="""
+        vm.Card(
+            text="""
         A selector can be used within the **Parameter** or **Filter** component to allow the user to select a value.
 
         The following selectors are available:
@@ -356,8 +357,11 @@ selectors = vm.Page(
         * RangeSlider (**numerical** multi option selector only)
         * Slider (**numerical** single option selector only)
 
-        """),
-        vm.Table(id="table-gapminder", figure=dash_data_table(data_frame=gapminder, page_size=10), title="Gapminder Data"),
+        """
+        ),
+        vm.Table(
+            id="table-gapminder", figure=dash_data_table(data_frame=gapminder, page_size=10), title="Gapminder Data"
+        ),
         vm.Table(id="table-tips", figure=dash_data_table(data_frame=tips, page_size=10), title="Tips Data"),
     ],
     controls=[
@@ -394,7 +398,7 @@ selectors = vm.Page(
     ],
 )
 
-    # ACTIONS ---------------------------------------------------------------------
+# ACTIONS ---------------------------------------------------------------------
 export_data_action = vm.Page(
     title="Export data",
     components=[
--- a/vizro-core/examples/features/app.py
+++ b/vizro-core/examples/features/app.py
@@ -3,7 +3,6 @@
 from time import sleep
 from typing import List, Literal, Optional
 
-import numpy as np
 import pandas as pd
 import plotly.graph_objects as go
 import vizro.models as vm
@@ -17,7 +16,6 @@ from vizro.tables import dash_data_table
 iris = px.data.iris()
 gapminder = px.data.gapminder()
 tips = px.data.tips()
-tips["smoker"] = np.where(tips["smoker"] == "Yes", 1, 0)
 gapminder_2007 = px.data.gapminder().query("year == 2007")
 waterfall_df = pd.DataFrame(
     {
--- a/vizro-core/examples/features/app.py
+++ b/vizro-core/examples/features/app.py
@@ -50,8 +50,15 @@ home = vm.Page(
 
                 ### Controls
 
-                Vizro has two different control types **filters** and **parameters**.
+                Vizro has two different control types **Filter** and **Parameter**.
 
+                You can use any pre-existing selector inside the **Filter** or **Parameter**:
+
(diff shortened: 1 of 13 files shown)


vizro-core/examples/features/app.py
```diff
@@ -14,6 +15,9 @@
 from vizro.tables import dash_data_table
 
 iris = px.data.iris()
+gapminder = px.data.gapminder()
+tips = px.data.tips()
+tips["smoker"] = np.where(tips["smoker"] == "Yes", 1, 0)
```
petar-qb, line 20: Can you change the line below from:
`gapminder_2007 = px.data.gapminder().query("year == 2007")`
to 
`gapminder_2007 = gapminder.query("year == 2007")`
(and define `gapminder_2007` exactly below the line where `gapminder` is defined)

vizro-core/examples/features/yaml_version/app.py
```diff
@@ -2,14 +2,20 @@
...
 from vizro.managers import data_manager
 from vizro.models import Dashboard
 
+tips = px.data.tips()
+tips["smoker"] = np.where(tips["smoker"] == "Yes", 1, 0)
 data_manager["iris"] = px.data.iris()
 data_manager["gapminder_2007"] = px.data.gapminder().query("year == 2007")
+data_manager["gapminder"] = px.data.gapminder()
+data_manager["tips"] = tips
+
```
petar-qb, line 18: Here is a suggestion for resorting data preparation code lines, similar like in the comment above.
```suggestion
tips = px.data.tips()
tips["smoker"] = np.where(tips["smoker"] == "Yes", 1, 0)
data_manager["tips"] = tips
gapminder = px.data.gapminder()
data_manager["gapminder"] = gapminder
data_manager["gapminder_2007"] = gapminder.query("year == 2007")
data_manager["iris"] = px.data.iris()
```

petar-qb (approved):
Awesome as usual! 🧹 🚀

vizro-core/examples/features/yaml_version/app.py
huong-li-nguyen: Yes! - Thank you! 🚀 🧹

nadijagraca (approved):
Looks good. 🚀

vizro-core/examples/features/app.py
antonymilne, line 20:
```suggestion
tips["smoker"] = tips["smoker"] == "Yes"
```
Would work ok depending on how we and plotly handle `True/False` in pandas series. So don't just accept this suggestion without checking it because it might break everything 😅

vizro-core/examples/features/app.py
```diff
@@ -46,8 +50,15 @@
 
                 ### Controls
 
-                Vizro has two different control types **filters** and **parameters**.
+                Vizro has two different control types **Filter** and **Parameter**.
 
+                You can use any pre-existing selector inside the **Filter** or **Parameter**:
+
+                * Dropdown
```
antonymilne, line 57: Let's remember to add the date picker here when it's ready.

vizro-core/examples/features/app.py
```diff
@@ -331,6 +342,62 @@
...
+        ),
+        vm.Filter(
+            targets=["table-tips"],
+            column="sex",
+            selector=vm.RadioItems(title="Radio Items (Tips - sex)"),
+        ),
+        vm.Filter(
+            targets=["table-tips"],
+            column="smoker",
+            selector=vm.Slider(title="Slider (Tips - smoker)", step=1, value=1),
```
antonymilne, line 396: This is not the best example of a `Slider` given that it's just a binary choice. Do we have any real, continuous numerical column we could use instead here?

vizro-core/examples/features/app.py
antonymilne: Also see later comment on whether we want to be filtering on this column at all.

antonymilne (approved):
Looks good! I think we should improve the `Slider` example if easy to do so, but no worries if not.

vizro-core/examples/features/app.py
maxschulz-COL: Or maybe use it as a parameter? I think filtering on a single number will always be tricky, but having opacity on a graph is always a good one e.g.

In general I think having a graph on this page would be good

vizro-core/src/vizro/models/_components/form/_text_area.py
```diff
@@ -14,7 +14,9 @@
 
 
 class TextArea(VizroBaseModel):
-    """Component provided to `Form` to allow user multi-line text input.
+    """Multi-line text input component `TextArea`.
+
+    Based on the underlying [`dcc.TextArea`](https://dash.plotly.com/dash-core-components/textarea).
```
maxschulz-COL, line 19: Love it!

vizro-core/src/vizro/static/css/table.css
```diff
@@ -52,6 +52,37 @@
   padding: 4px 0;
 }
 
+#right-side .dash-table-container .previous-next-container button,
```
maxschulz-COL, line 55: We may want to also apply some of this to the pagination of the AgGrid

maxschulz-COL (approved):
Lgtm! If I could choose, I would probably work more with a graph, and ideally actually only one component, but that may be tough with fitting all the options in.

huong-li-nguyen:
> Lgtm! If I could choose, I would probably work more with a graph, and ideally actually only one component, but that may be tough with fitting all the options in.

Yeah, that's what I had before, but it did not work well if we want to have all selectors on one page 😄  They are contradicting each other / intersecting - which made the interactions horrible with one graph.

vizro-core/src/vizro/static/css/table.css
huong-li-nguyen: Would be good if you could double-check that in your CSS PR. Just add pagination and see if the CSS is messed up there. I just squeezed this in here, as it was obviously wrong in the sample screen

vizro-core/examples/features/app.py
huong-li-nguyen: @nadijagraca

vizro-core/examples/features/app.py
huong-li-nguyen: @antonymilne - I find it a bit tricky because the slider will usually reduce the number of observations drastically 😄 Which is why I didn't want to use the "tips" column as if you would use that for the slider, there would be normally only one observation left. If you combine it with the categorical single-select dropdown, they usually lead up to an empty table. Hence, why I just transformed the binary column to a numeric one, as this at least left some observations for the other selectors. I can also just add another chart below with another data set and apply the Slider only there?

antonymilne: Yeah, that makes sense. What about the `size` column for the tips dataset though? Does that lead to an empty table too often?

If that doesn't work well then fine to leave how you have it with a binary slider I think.

vizro-core/examples/features/app.py
huong-li-nguyen: I removed it entirely 👍

vizro-core/examples/features/app.py
huong-li-nguyen: I went for the size column - I just played around with some default value settings for the other selectors such that we have some nice data and now it works fine 👍

PR #346, opened 2024-03-04
[Feat] Add CSS styling to single and multi `DatePicker`

huong-li-nguyen:
## Description
- Add CSS styling to single and multi `DatePicker`
- Add hover, selected, and focus state for user-input
- Add dark and light theme

## Screenshot
**Original Figma designs:**
Screenshot 2024-03-04 at 22 17 46

**Current implementation (some deviations exist):**
Screenshot 2024-03-04 at 22 15 41 Screenshot 2024-03-04 at 22 15 11
Screenshot 2024-03-04 at 22 15 34 Screenshot 2024-03-04 at 22 15 04
 Screenshot 2024-03-04 at 22 15 23 Screenshot 2024-03-04 at 22 14 56

## Notice


petar-qb (commented):
Awesome work! 
There are a few UI "enhancements" I'd like to see implemented if they make sense for you.

- Text colour differentiation for years, months or dates that can't be selected (not in the min-max range). In the screenshot example below, only 2023 and 2024 are years in the allowed range.
image


- Left roundness on start_date, right roundness on the end_date selected as the original dmc components have. (I know it's not suggested by the original Figma design) 
- Also some text colour differentiation between current month name, days in a week and dates that can be selected would look nice by my opinion (I know it's not suggested by the original Figma design) 
image
image

huong-li-nguyen:
Hey @petar-qb,

thanks for your input!

1) Should be do-able and definitely what we should do. I've implemented it for the days already, but apparently not for the months/years. Have to see what we can do there.

2) This would be against our design guidelines, as we don't use round edges anywhere. I know it can look nice in some cases, but for the sake of consistency, I wouldn't recommend doing that.

3) Do you mind explaining that more? Currently there is a text color / bg color differentiation between selected dates, dates in range and disabled dates. Where would you like to add more coloring?

petar-qb:
---
> 1. Should be do-able and definitely what we should do! Great catch 🚀 I've implemented it for the days already, but apparently not for the months/years. Implemented it in this commit now: https://github.com/mckinsey/vizro/commit/ddd9658ec76bee7cc2307dc2dc1ff6ee3cbd0cb4

I like how the month and year texts are disabled for those outside the min-max range. Although, I can't see the difference in text color (disabled text) for days outside the min-max range. In the following screenshot, only the date 2024-07-01 is within the min-max range but it has the same text colour as all the other dates in July 2024.

image

---
> 2. This would be agains our designed guidelines...

Makes sense. 

---
> 3. Do you mind explaining that more? Currently there is a text color / bg color differentiation between selected dates, dates in range and disabled dates. Where would you like to add more coloring? If you refer to the font color for the month, the weekdays and the date these are intentionally the same in Figma as the main focus for the user should be on the selected day range. Hence, only the selected dates have different font colors :)

Yes, I was referring to the font colour for the month, the weekdays and the date. Then it makes sense if they are intentionally made to have the same text colour.

huong-li-nguyen:
Ah yes, I see! Thank you!! Fixed here: https://github.com/mckinsey/vizro/pull/346/commits/b5d1fcec66aec0438d074ff8d0e071b9aed05de0


Screenshot 2024-03-05 at 13 32 12

petar-qb (approved):
The last thing I found is that the DatePicker looks weird once it's treated as a `Page.component`.

image
  

Other than that, everything looks good 🚀

huong-li-nguyen:
> The last thing I found is that the DatePicker looks weird once it's treated as a `Page.component`.
> 
> image
> Other than that, everything looks good 🚀

That's the expected behaviour :) The calendar will always extend to it's max-content width only while the dropdown selector takes all the entire available width (so it's consistent with the behaviour of all other selectors when added to a page). Currently users would have to update their container width for that. This might change when we work on the form components, but for now that's the expected behaviour 👍

antonymilne (approved):
I didn't do a careful check of everything since @petar-qb did an amazing job of that already (thank you, it's really valuable to have such thorough reviews!), but skimming through the screenshots and discussion this looks great. If you think someone else needs to check again more carefully then please do say so, but otherwise I trust you and Petar did a wonderful job here 💯

PR #347, opened 2024-03-05
[Snyk] Security upgrade kedro from 0.18.14 to 0.19.3

maxschulz-COL:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to fix one or more vulnerable packages in the `pip` dependencies of this project.</h3>

#### Changes included in this PR

- Changes to the following files to upgrade the vulnerable dependencies to a fixed version:
    - vizro-core/snyk/requirements.txt

<details>
<summary>⚠️ <b>Warning</b></summary>

```
pandas 1.3.5 requires numpy, which is not installed.
kedro 0.18.14 has requirement importlib-metadata<5.0,>=3.6; python_version < "3.8", but you have importlib-metadata 6.7.0.

```
</details>

##### By pinning:
Severity                   | Priority Score (*)                   | Issue                   | Upgrade                   | Breaking Change                   | Exploit Maturity
:-------------------------:|-------------------------|:-------------------------|:-------------------------|:-------------------------|:-------------------------
medium severity  |  **551/1000**  <br/> **Why?** Recently disclosed, Has a fix available, CVSS 5.3  | Arbitrary File Write via Archive Extraction (Zip Slip) <br/>[SNYK-PYTHON-KEDRO-6370687](https://snyk.io/vuln/SNYK-PYTHON-KEDRO-6370687) |  `kedro:` <br> `0.18.14 -> 0.19.3` <br>  |  No  | No Known Exploit 

(*) Note that the real score may have changed since the PR was raised.

Some vulnerabilities couldn't be fully fixed and so Snyk will still find them when the project is tested again. This may be because the vulnerability existed within more than one direct dependency, but not all of the affected dependencies could be upgraded.

Check the changes in this PR to ensure they won't cause issues with your project.

------------

**Note:** *You are seeing this because you or someone else with access to this repository has authorized Snyk to open fix PRs.*

For more information:  
🧐 [View latest project report](https://app.snyk.io/org/vizro/project/bcb0f21a-98d6-4665-8101-e4ea8ac33049?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr)

🛠 [Adjust project settings](https://app.snyk.io/org/vizro/project/bcb0f21a-98d6-4665-8101-e4ea8ac33049?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr/settings)

[...]

Issue #226, opened 2023-12-18
Automatically generated navigation homepage


antonymilne:
The Homepage on https://vizro.mckinsey.com/ is currently written entirely manually but follows a very clear formula. Following #176 we now have a field `Page.description` and so all of the ingredients are in place with the exception of the page image. Somehow we should make it possible to automatically generate such a homepage. In theory this sounds pretty easy to do.

**Major considerations**
* What would be the right user API for it? A new `HomePage` model that subclasses `Page`?  A `Homepage` model that includes a `Page` model? Just some Python helper functions? We'd like to follow our paradigm of extensibility and customisability and enable something that's possible from YAML too 🤔 
* How does this all relate to `Navigation` and different ways of grouping pages in hierarchy?

**Minor considerations**
* How does page image fit into this? Should we change these to Google material icons? Do we want a `Page.image` field that's also used in meta tags (originally existed in #176 but removed)? Or to automatically pick up an image with filename `page.id`? Could we generate preview images of the pages automatically as discussed in final point of [this comment](https://github.com/mckinsey/vizro/issues/176#issuecomment-1824109735) and subsequent responses?

PR #348, opened 2024-03-05
[Bug] Consolidate validation for `Card.href` and `Page.path`

huong-li-nguyen:
## Description
- Fix https://github.com/mckinsey/vizro/issues/285 by consolidating and  validation for `Card.href` and `Page.path`
- Replace `dcc.Link` with `dbc.NavLink `(this seemed to be broken for external links before - we might have to add more tests here @l0uden)
- The following are valid inputs for the `Card.href` now:
    - https://google.com
    - http://google.com
    - /page-title
    - page-title  (didn't work previously) 
    - Page Title (didn't work previously) 
- Change doc and demo examples to use the more intuitive way of linking to pages via the `Card.href`

## Notice


vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -20,3 +21,28 @@ def _validate_min_length(cls, field):
...
+
+    Absolute URLs: https://www.google.com, http://www.google.com
+    Relative URLs: /first-page, first-page, www.google.com
+
+    """
+    parsed_url = urlparse(url)
+    return not bool(parsed_url.netloc)
+
+
+def _clean_url(url, allowed_characters):
```
maxschulz-COL, line 37:
```suggestion
def _clean_url(url: str, allowed_characters: str) -> str:
```

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -20,3 +21,28 @@ def _validate_min_length(cls, field):
     if not field:
         raise ValueError("Ensure this value has at least 1 item.")
     return field
+
+
+def _is_relative_url(url):
```
maxschulz-COL, line 26:
```suggestion
def _is_relative_url(url: str) -> bool:
```

vizro-core/examples/_dev/app.py
```diff
@@ -1,112 +1,89 @@
...
+            text="""
+            ### http
+            Lorem ipsum dolor sit amet consectetur adipisicing elit. Maxime mollitia,
+            molestiae quas vel sint commodi repudiandae consequuntur voluptatum laborum
+            numquam blanditiis harum quisquam eius sed odit fugiat iusto fuga praesentium
+            option, eaque rerum!
+            """,
+            href="http://www.google.com",
+        ),
+        vm.Card(
```
maxschulz-COL, line 72: Is this one supposed to work, I get 404

vizro-core/changelog.d/20240305_142454_huong_li_nguyen_fix_card_validation.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Consolidate validation for `Card.href` and `Page.path`. ([#347](https://github.com/mckinsey/vizro/pull/347))
```
maxschulz-COL, line 40:
```suggestion
- Allow spaces in `Card.href` and `Page.path` strings - e.g. `Page Title` now works as reference to a page. ([#347](https://github.com/mckinsey/vizro/pull/347))
```

maxschulz-COL (approved):
Lgtm, just a few minor edits!

vizro-core/examples/_dev/app.py
huong-li-nguyen: No, that's the expected result as you have it :) Just wanted to demo, that www.google.com is not considered an absolute path 👍

vizro-core/changelog.d/20240305_142454_huong_li_nguyen_fix_card_validation.md
huong-li-nguyen: Spaces did work already in `Page.path`, so I think I wouldn't take over the first part. The second part I've split up in a separate changelog entry now, as this is changed functionality and doesn't belong to fixed. I agree that it's worth mentioning 👍

vizro-core/docs/pages/user-guides/card-button.md
```diff
@@ -409,7 +409,7 @@ To create a navigation card, do the following:
 
 - Insert the [`Card`][vizro.models.Card] into the `components` argument of the [`Page`][vizro.models.Page]
 - Provide the `text` parameter with a title and some description
-- Provide the `href` parameter (relative or absolute URL)
+- Provide the `href` parameter (relative or absolute URL). In case it refers to a `Page` inside the `Dashboard`, you can insert the `Page.title` directly.
```
stichbury, line 412:
```suggestion
- Provide the `href` parameter (relative or absolute URL). If it refers to a `Page` inside the `Dashboard`, you can insert the `Page.title` directly.
```

stichbury (commented):
Docs LGTM! 🙏

vizro-core/changelog.d/20240305_142454_huong_li_nguyen_fix_card_validation.md
maxschulz-COL: Ok cool! I would actually remove the fixed bit then as I think that it is not really interesting to a user who needs to decide if it's worth upgrading Vizro to a newer version. Less is more here I feel :)

huong-li-nguyen: For me, these are actually two different points. The fixed one does fix a bug that has been reported by Alexey and was the main point for this PR: https://github.com/mckinsey/vizro/issues/285

The fact that you can provide a `Page.title` directly to `Card.href` is an add-on as I got rid of the requirement having to provide a `/` + the point above.

vizro-core/changelog.d/20240305_142454_huong_li_nguyen_fix_card_validation.md
```diff
@@ -0,0 +1,46 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Align validators for `Card.href` and `Page.path` to ensure consistent navigation linking. ([#347](https://github.com/mckinsey/vizro/pull/347))
```
maxschulz-COL, line 39:
```suggestion
- Fix [bug](https://github.com/mckinsey/vizro/issues/285) which breaks page opening when providing the same `str to `Card.href` and `Page.path`, when that `str` is not formatted as a valid `path`. ([#347](https://github.com/mckinsey/vizro/pull/347))
```

I think it would be advantageous to communicate the bug then, and give an idea if the user would be affected. Previously I have to guess what this really means.

Overall I would suggest linking the bug in the PR description, and labelling the PR as such as I wasn't aware this was a bug even.

vizro-core/changelog.d/20240305_142454_huong_li_nguyen_fix_card_validation.md
```diff
@@ -0,0 +1,46 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Allow `Page.title` to be inserted directly into `Card.href` as a relative link. ([#347](https://github.com/mckinsey/vizro/pull/347))
```
maxschulz-COL, line 28:
```suggestion
- Allow `Page.title` to be inserted directly into `Card.href` as a relative link without the need to provide a `/`. ([#347](https://github.com/mckinsey/vizro/pull/347))
```

Then I think it's super clear :) I first read the sentence as: I insert the title as a relative link so I need to add a `/`

vizro-core/changelog.d/20240305_142454_huong_li_nguyen_fix_card_validation.md
huong-li-nguyen: I can add it to the PR description as well 👍 

However, the PR was already labeled as a bug and the issue ticket was also connected before such that it automatically closes the issue if this PR gets merged - you can see it in the development section of this PR! :) 

Screenshot 2024-03-06 at 09 19 06

vizro-core/changelog.d/20240305_142454_huong_li_nguyen_fix_card_validation.md
huong-li-nguyen: Or we just shorten to: Allow `Page.title` to be inserted directly into `Card.href` ?

Because it's not only the slash that prevented that from doing that before

vizro-core/changelog.d/20240305_142454_huong_li_nguyen_fix_card_validation.md
maxschulz-COL: Ah mea culpa 🙏 I find it always weirdly hidden on the right there haha. Awesome!

huong-li-nguyen: I've rephrased a bit to make it shorter, but kept your message:

Fix [bug](https://github.com/mckinsey/vizro/issues/285) where navigating from a `Card` to `Page` failed when an unformatted `Page.path` was used in `Card.href`. ([#347](https://github.com/mckinsey/vizro/pull/347))

vizro-core/changelog.d/20240305_142454_huong_li_nguyen_fix_card_validation.md
```diff
@@ -0,0 +1,46 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Allow `Page.title` to be inserted directly into `Card.href`. ([#348](https://github.com/mckinsey/vizro/pull/348))
```
antonymilne, line 28: Technically this is `Page.id` but I think the changelog is clearer how you have it so don't change it.

vizro-core/docs/pages/user-guides/card-button.md
```diff
@@ -409,7 +409,7 @@ To create a navigation card, do the following:
 
 - Insert the [`Card`][vizro.models.Card] into the `components` argument of the [`Page`][vizro.models.Page]
 - Provide the `text` parameter with a title and some description
-- Provide the `href` parameter (relative or absolute URL)
+- Provide the `href` parameter (relative or absolute URL). If it refers to a `Page` inside the `Dashboard`, you can insert the `Page.title` directly.
```
antonymilne, line 412: Here I think it does matter that we are correct. I'd leave as `Page.title` as the first thing you mention, since 99% of the time that's correct and easier to understand for a user than `Page.id`. But then add something afterwards saying that technically it's `Page.id` and link to https://vizro.readthedocs.io/en/latest/pages/user-guides/pages/#customizing-the-page-url.

vizro-core/examples/features/app.py
```diff
@@ -68,7 +68,7 @@
 
                 Standard predefined actions are made available including **export data** and **filter interactions**.
                 """,
-            href="/export-data",
+            href="Export data",
```
antonymilne, line 71: Is this the right capitalisation? I know it will convert to lowercase either way but would be nice to match exactly with the `Page.title` ("Export Data" or "Export data" - I don't know which it is).

antonymilne (commented):
I got halfway through reviewing and was going to say that this is perfect, great tests, etc... and then I realised that maybe my proposed solution was not quite right after all 😬 

Let me ponder briefly and then come back to this tomorrow. The comments about docs will still apply even if we change solution though.

It's a great bit of coding even if I suggested the wrong solution the first place though 👏 😅

vizro-core/examples/features/app.py
huong-li-nguyen: Yes, it's the right capitalisation - I was also trying to sync it 😄 So the page.title is actually "Export data"

huong-li-nguyen:
Closing as we decided to not enable this feature.

antonymilne:
In more detail...

The problem with the proposed solution for `Card.href` is that really we should be referring to the page id and not the page title. But the page id is not necessarily the same as the page path since (a) there’s discrepancy of whether / is scrubbed from the path depending on whether you set the `page.path` manually or whether it's taken from the `id` automatically (b) the annoying homepage case where `id`/`title` and path don’t match up (maybe changed in #366) and (c) we let users set `page.path` manually (which in hindsight was a mistake I think since I’m pretty sure no one does it now and it makes this sort of thing a bit awkward :grimacing:). So that means that `page.path` could be completely different from `page.id`, in which case the solution proposed in this PR won't work at all.

So really the way we should be doing this is:
* user supplies page id
* look it up in mode manager to find relevant page
* get relative path out of that page and use that for the link

https://github.com/mckinsey/vizro/pull/348/commits/09305d69638dd80a1369691bbd9c7a9edaa196ea shows how we might do this. While it's not that hard to do, it does add some complexity and at some point in the future I'd hope that the current `href` behaviour of `Card` might be removed at all in favour of #226.

If we do add this in future to lookup by `page.id` we should also consider whether to remove the functionality that providing a relative path would work so that the only options are to provide `page.id` or a complete `href`. This might be a bit tidier but would be a breaking change.

PR #349, opened 2024-03-06
[Docs] Add change to enable 404 page automation

stichbury:
## Description
Small change to follow [instructions in RTD guide](https://docs.readthedocs.io/en/stable/reference/404-not-found.html#tool-integration) and enable a 404 page. Setting the `site_url` to the base canonical URL for the docs is sufficient for the tool to build us a `404.html`, which we need now I've moved some pages.

## Notice


vizro-core/mkdocs.yml
```diff
@@ -1,5 +1,5 @@
 site_name: Vizro
-site_url: https://mckinsey.github.io/vizro/
+site_url: https://vizro.readthedocs.io/en/stable/
```
huong-li-nguyen, line 2: Do we need something similar for the docs on vizro-ai?

stichbury: Good point! I forgot those docs (as usual). Let me add it.

huong-li-nguyen: No worries, I always forget about them too 😆

huong-li-nguyen (approved):
LGTM! 👍 

Just a clarifying question - after this PR, a link like this: https://vizro.readthedocs.io/en/stable/pages/user_guides/custom_actions/#interacting-with-dashboard-inputs-and-outputs

should now lead to a 404 page right? Because it seems like this section doesn't exist anymore?

stichbury:
> Just a clarifying question - after this PR, a link like this: https://vizro.readthedocs.io/en/stable/pages/user_guides/custom_actions/#interacting-with-dashboard-inputs-and-outputs
> 
> should now lead to a 404 page right? Because it seems like this section doesn't exist anymore? Ah no wait, this seems to exist, but the forward reference doesn't seem to work with this one?

The links to sections within a page should not break a link if the section is renamed or removed. From my testing, if the page exists, it'll be displayed but when the subsection isn't found, it just falls back to show the top of the page.

It sounds like that's another redirect issue, which seems to be ongoing today, but that page may not be in my top 20 list. Maybe I'll just add them all for safety!

petar-qb (approved):
Looks good!

vizro-ai/mkdocs.yml
```diff
@@ -1,5 +1,5 @@
 site_name: Vizro-AI
-site_url: https://mckinsey.github.io/vizro-ai/
+site_url: https://vizro.readthedocs.io/en/stable/
```
huong-li-nguyen, line 2: This might be very confusing, but vizro-ai doesn't have a stable version but only a latest. Best to double-check with @maxschulz-COL if we actually want to change that:

```suggestion
site_url: https://vizro-ai.readthedocs.io/en/latest/
```

maxschulz-COL: Yes, this is important. Thanks for pointing this out @huong-li-nguyen!

Vizro-ai does not have stable docs, but only latest, as RTD does not (easily? didn't check in a wile) allow multiple sets of docs in a monorepo that also link to that repos release cycle.

Let's discuss more in person, as this also somewhat links to the discussion if we should have two sets of docs.

stichbury: Ah yes, good point about the latest vs stable. I feel like all the signs are pointing to building docs together in one or setting up a proper subproject in RTD so that the `vizro-ai` docs can have its own release cadence and then latest + stable docs. I am not sure how easily this can be done with MkDocs compared to Sphinx but will investigate. I would, however, very much like to look at whether it's worth having separate docs at all, so will take a more holistic view and compare both approaches before trying to solve for just one.

@maxschulz-COL The `site_url` setting is to set the canonical URL, which in SEO terms means that, where identical content appears in more than one place (e.g. repo in markdown and published as HTML on RTD) it tells search engines which to surface in preference. So we do actually want the `site_url` to be the docs rather than the repo. This change is the correct thing to do for that, and has the benefit of fixing the 404 page alongside!

PR #354, opened 2024-03-06
[Bug] Fix AgGrid inspect console error

maxschulz-COL:
## Description
Fixes small bug due to non-existent column type in cell data definition

Additionally using this PR to re-add the release highlights as I assume we won't do another release before the tests are ready

## Screenshot
Screenshot 2024-03-06 at 19 38 11

## Notice


huong-li-nguyen (commented):
I was just about to ask how this got merged with the console errors, because usually @l0uden tests catch that 😄 But then remembered that we didn't add them yet.

Are there any other console errors popping up or is it the only one? @l0uden @maxschulz-COL

maxschulz-COL:
> Are there any other console errors popping up or is it the only one? @l0uden @maxschulz-COL

None that I know of!

petar-qb (approved):
Well spotted! 👍 

I've also spotted this one occurring, but have no idea if it can even be solved on our end.
image

huong-li-nguyen:
> > Are there any other console errors popping up or is it the only one? @l0uden @maxschulz-COL
> 
> None that I know of!

I'll let @l0uden review this one, because often there are console errors found in his tests, while they don't appear locally on mine 😄 If he approves, I think you're good to go 👍

l0uden:
@huong-li-nguyen, my tests won't find this because it is `Warning` and not `Severe` error. But anyway my tests for AGgrid are not main, so this errors could be missed. 
I will take a look I my tests if I can spot this kind of errors.

l0uden:
I've added checks for warnings in my branch. Thanks!

It is ok for me.

Will we investigate @petar-qb's console error in this PR?

l0uden (approved):

antonymilne:
Regardless of @l0uden's tests, shouldn't this have failed the integration tests on this repo?
https://github.com/mckinsey/vizro/blob/main/vizro-core/tests/integration/test_examples.py#L49

I'm a bit concerned if this isn't working because it definitely used to be 😅 And I would expect it to catch warnings as well as errors.

@petar-qb don't know whether that error is on our side or not, but at a glance my guess would be it's something to do with handing missing/None values in a table?

l0uden:
@antonymilne, dash's `get_logs` method is looking only for `Severe` errors. We can't change it. That is why this tests will never catch this `Warning`.

l0uden:
I've managed to reproduce @petar-qb's error. 
To do it you should open page with AGgrid and then immediately click different page. In this case error occur.
If you leave AGgrid page open for 1-2 seconds, there will be no such errors. 

It looks like it connected to the case when smth not loaded on the page and we switched to another. 
Hope it helps to investigate this problem and find if it is a major one.

PR #358, opened 2024-03-11
[Bug] Fix path to built-in Vizro assets

antonymilne:
## Description

There was a subtle bug where we were using `routes_pathname_prefix` somewhere we should have used `requests_pathname_prefix`. What this meant was that in some deployments our built-in Vizro js and css would not be found.

* `requests_pathname_prefix` is "is a local URL prefix for file requests. Must end with `routes_pathname_prefix`" (from Dash docs). "the prefix for the AJAX calls that originate from the client (the web browser)" (from comment in Dash source code). This is what's used when injecting flask server into Dash as in [Host on a route of existing Flask app](https://community.plotly.com/t/how-to-add-your-dash-app-to-flask/51870/2#host-on-a-route-of-existing-flask-app)
* `routes_pathname_prefix` is "a local URL prefix for JSON requests. Must start and end  with `'/'`" (from Dash docs). "the prefix for the API routes on the backend (this flask server)" (from comment in Dash source code). This is what's used when mounting Dash onto a server using middleware (could be WSGI like flask or ASGI like FastAPI that uses `WSGIMiddleware`) as in [`DispatcherMiddleware`](https://community.plotly.com/t/how-to-add-your-dash-app-to-flask/51870/2#dispatchermiddleware)
*  setting `url_base_pathname = x` is equivalent to setting both `requests_pathname_prefix = x` and `routes_pathname_prefix = x`

Previously things worked ok when `requests_pathname_prefix == routes_pathname_prefix` but not otherwise. Now it works ok - tested on:
* injecting flask server into Vizro
* mounting Vizro into Flask app
* mounting Vizro into FastAPI app

## Notice


vizro-core/src/vizro/_vizro.py
```diff
@@ -37,25 +37,26 @@ def __init__(self, **kwargs):
...
-        blueprint_prefix = self.dash.config.routes_pathname_prefix.replace("/", "_").replace(".", "_")
+        routes_pathname_prefix = self.dash.config.routes_pathname_prefix
+        blueprint_prefix = routes_pathname_prefix.replace("/", "_").replace(".", "_")
         self.dash.server.register_blueprint(
             flask.Blueprint(
                 f"{blueprint_prefix}vizro_assets",
                 self.dash.config.name,
                 static_folder=vizro_assets_folder,
-                static_url_path=static_url_path,
+                static_url_path=routes_pathname_prefix + STATIC_URL_PREFIX,
```
antonymilne, line 59: This is the only line that's actually changed functionally because it now uses `routes_pathname_prefix` and not `requests_pathname_prefix`.

The other changes are just refactoring now that the `static_url_prefix` is no longer required.

maxschulz-COL (approved):

huong-li-nguyen (approved):
LGTM 👍

PR #366, opened 2024-03-12
[Tidy] Replace `dash.page_registry` as page lookup with `model_manager`

huong-li-nguyen:
## Description
- Replace page lookup in `dash.page_registry` with `model_manager` to have one source of truth for page lookups
- Update unit and integration tests

## Notice


vizro-core/src/vizro/models/_navigation/accordion.py
```diff
@@ -79,22 +79,17 @@ def build(self, *, active_page_id=None):
...
-                    f"Page with ID {page_id} cannot be found. Please add the page to `Dashboard.pages`"
-                ) from exc
             nav_links.append(
                 dbc.NavLink(
-                    children=[page["name"]],
+                    children=page_id,
                     className="accordion-item-link",
                     active="exact",
-                    href=page["relative_path"],
+                    href=model_manager[ModelID(str(page_id))].path,
```
huong-li-nguyen, line 92: @antonymilne - had to change this one to take the `page.path` from the model_manager in case someone sets a different `page.path` than the` page.id`

vizro-core/src/vizro/models/_page.py
```diff
@@ -67,10 +67,10 @@ def set_id(cls, values):
     def set_path(cls, path, values) -> str:
         # Based on how Github generates anchor links - see:
         # https://stackoverflow.com/questions/72536973/how-are-github-markdown-anchor-links-constructed.
-        def clean_path(path, allowed_characters):
+        def clean_path(path: str, allowed_characters: str) -> str:
             path = path.strip().lower().replace(" ", "-")
             path = "".join(character for character in path if character.isalnum() or character in allowed_characters)
-            return "/" + path
+            return path if path.startswith("/") else "/" + path
```
huong-li-nguyen, line 73: Had to add this in, otherwise the first href of the landing page was "//"

vizro-core/tests/integration/test_navigation.py
```diff
@@ -8,6 +8,7 @@
 @pytest.fixture
 def dashboard_result(request):
     # Inject the navigation into the dashboard. Note we need to call request.param since they are all lambda functions.
+    Vizro._reset()
```
huong-li-nguyen, line 10: I don't 100% understand why this needs to be called at the beginning instead of where the current `model_manager._clear(`) call is, so not sure if this is the correct solution @antonymilne, but this seems to fix the failing integration tests.

 The solution in the comments also does not work, so yielding the dashboard and then calling `Vizro._reset()` after. Somehow this only works if `Vizro._reset()` is called before the dashboard definition. But i am wondering why this didn't have interdependence issues before..

antonymilne: Yeah, this is correct how you have it. It's definitely not quite right overall but I don't think we'll easily be able to get it any better than this because I spent ages before trying to improve it and couldn't. I'll just updated and moved the comment in [f6b3279](https://github.com/mckinsey/vizro/pull/366/commits/f6b3279537f1be7442374173fefe3d7a0c63b472) but good work figuring this out 👍

vizro-core/src/vizro/models/_navigation/accordion.py
```diff
@@ -79,22 +79,18 @@ def build(self, *, active_page_id=None):
...
-                ) from exc
+            page = model_manager[ModelID(str(page_id))]
             nav_links.append(
                 dbc.NavLink(
-                    children=[page["name"]],
+                    children=page.title,
                     className="accordion-item-link",
                     active="exact",
-                    href=page["relative_path"],
+                    href=page.path,
```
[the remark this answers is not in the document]
huong-li-nguyen: Is this required such that it can deployed elsewhere as well?

So to have the below working?
`"requests_pathname_prefix might be unset and so a relative URL like /page-2 can just be /page-2. However, when the app is deployed to a URL like /my-dash-app, then relative_path will be /my-dash-app/page-2."`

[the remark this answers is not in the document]
antonymilne: Yes exactly! Otherwise if you deploy at `www.example.com/my-dash-app` and then click a link to `/page-2` then you'll end up at `www.example.com/page-2` rather than `www.example.com/my-dash-app/page-2`.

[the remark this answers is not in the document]
huong-li-nguyen: @petar-qb @maxschulz-COL @nadijagraca @lingyielia - good to know ⏫

huong-li-nguyen:
> @petar-qb:
> 
> 1. any problem with making the change [b4e0d3b](https://github.com/mckinsey/vizro/commit/b4e0d3b770b6462bcbcea6ca0476f787a57ad365)? Note the `dash.page_registry` will always be a subset of `model_manager` (in future they will be equal, but right now they're not necessarily)

I can't answer the second one, but the first one for sure leads to issues that I am not sure if we can get rid of quickly @antonymilne @petar-qb. Alexeys screenshot tests are currently all breaking and you can replicate it by creating a page with a graph as a component, but **not including it** in the dashboard.pages. You will get a bunch of error messages now that an ID cannot be found. 

I assume this happens because it's trying to run some action chains (connected to graphs) from pages that are inside the model manager but not inside the dashboard. Therefore it cannot find the relevant pages as they only exist in the model manager.

Screenshot 2024-03-13 at 12 45 32

petar-qb:
> > @petar-qb:
> > 
> > 1. any problem with making the change [b4e0d3b](https://github.com/mckinsey/vizro/commit/b4e0d3b770b6462bcbcea6ca0476f787a57ad365)? Note the `dash.page_registry` will always be a subset of `model_manager` (in future they will be equal, but right now they're not necessarily)
> 
> I can't answer the second one, but the first one for sure leads to issues that I am not sure if we can get rid of quickly @antonymilne @petar-qb. Alexeys screenshot tests are currently all breaking and you can replicate it by creating a **page with a graph as a component**, but **not including it** in the dashboard.pages. You will get a bunch of error messages now that an ID cannot be found.
> 
> I assume this happens because it's trying to run some action chains (connected to graphs) from pages that are inside the model manager but not inside the dashboard. Therefore it cannot find the relevant pages as they only exist in the model manager.

Exactly this bug you recreated Li is the reason why I look at dash.page_registry as a source of truth while constructing the action engine. I'm afraid we have to revert [b4e0d3b](https://github.com/mckinsey/vizro/commit/b4e0d3b770b6462bcbcea6ca0476f787a57ad365) changes since it doesn't work. There is no easy and reliable way to fix this behaviour, but one of the ways on how we can fix it could be to  remove non-registred page models from the `model_manager` in the  `Dashboard._pre_build()` method or similar 😄.

antonymilne:
Cool, thank you @huong-li-nguyen and @petar-qb! That all makes sense. I've reverted the change with [f065f46](https://github.com/mckinsey/vizro/pull/366/commits/f065f46b239cc23d803489f15e5312fd374de17a). We can fix this properly when I finally get around to https://github.com/mckinsey/vizro/issues/111.

huong-li-nguyen:
Big compliment for @l0uden tests by the way 🚀 All of the demo dashboards and unit + integrations were working fine and didn't raise anything. I only understood the issue because in the screenshot tests there were a bunch of console errors and I tried to replicate the setup. Took me only a few minutes to understand that it was caused by a page with a graph not being added to the `dashboard.pages` 👍

maxschulz-COL (approved):
LGTM! Definitely an improvement to have one source of truth!

petar-qb:
> any reason this code is in _action_loop_utils.py rather than moving to model_manager where you moved similar things before?

@antonymilne you are right. It seems like we can get rid od `_get_actions_chains_on_all_pages` util function since the similar is implemented in the `model_manager`.

maxschulz-COL:
> Big compliment for @l0uden tests by the way 🚀 All of the demo dashboards and unit + integrations were working fine and didn't raise anything. I only understood the issue because in the screenshot tests there were a bunch of console errors and I tried to replicate the setup. Took me only a few minutes to understand that it was caused by a page with a graph not being added to the `dashboard.pages` 👍

Amazing! Could we even move such a test to our integration tests? I.e if we were to have an additional page created in one of the tested examples, that should create a console error that is caught 🤔 ?

PR #384, opened 2024-03-25
[Tidy] Add `vizro-bootstrap` stylesheet and remove custom CSS

huong-li-nguyen:
## Description
- Add WIP version of static vizro-bootstrap CSS file (it's not complete yet, but only contains selected components)
- Remove custom CSS for `Accordion`, `Button`, `Card`, `Typography`, `NavBar`
- Refactor code for `Button`, `Card`, `NavBar`,  `NavLink`, 

---

### For Users
This change should only affect your app if you have been using custom CSS. If you've been using custom CSS and referenced any CSS classNames that have now been removed from our stylesheet, your components might look off. In this case, you need to target the correct [Bootstrap CSS classNames](https://www.w3schools.com/bootstrap/bootstrap_ref_all_classes.asp) instead. 

---

### For Reviewers
Don't be scared of the big number of added code lines. This is mainly due to adding the `vizro-bootstrap.min.css` as a physical file into our repository, which you don't have to review. In the long term, this file might be hosted elsewhere, and then we need to link the URL to our external stylesheets (1 line). 

**Context**
The `vizro-bootstrap.min.css` is the output of the following private repository: https://github.com/McK-Private/vizro-bootstrap. It's currently private, but we might also open-source it (not in any near PR). Currently, it's a WIP version where the CSS is created using SASS (for context: https://blog.hubspot.com/website/how-to-override-bootstrap-css#how-to-customize-bootstrap-sass). The vizro-bootstrap CSS file **does not contain** the correct design for all components. Instead, we will gradually extend it.

**What does it mean for our future development?**
**1) Preferred choice of components:** Always prefer the `dbc` component if the functionality is the same. In some cases, it makes sense to deviate from that choice, e.g., if the component is not available in dbc or functionality is significantly better in the other library. In this case, custom CSS needs to be added (short-term), but even better, a translation layer for CSS needs to be created to map bootstrap properties to CSS classNames of another component library.

**2) Adding custom CSS:** No custom CSS should be added anymore when adding new components. Everything should go via the bootstrap theme, so we have one single source of truth. In some cases, it might require custom CSS, but this needs to be decided on a case-by-case basis. The general guideline is that all CSS needs to be created via the Bootstrap theme.

@antonymilne  - I think we have two options in terms of the current deployment
1) Put `vizro-bootstrap.min.css` inside the static folder of vizro-core and load the static file into the external stylesheets (as currently in this PR)
2) Put `vizro-bootstrap.min.css` outside of vizro-core, so it's not part of the package, e.g., inside a folder called `dist`. The advantage of this approach is that it doesn't increase the size of our package, and we can still reference the CSS file via a CDN link: https://medium.com/javarevisited/how-to-host-your-repository-js-css-on-open-source-cdn-jsdelivr-4de252d6fbad 

I have a slight preference for 1) Now after reading through the arguments on why Dash decided to server their assets locally as a default: https://github.com/plotly/dash/issues/469

## Notice

  - I certify that (a) this contribution is my original creation and (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
  - I have not referenced individuals, products, or companies in any commits, directly or indirectly.


huong-li-nguyen:
> @antonymilne - I think we have two options in terms of current deployment
> 
> 1. Put `vizro-bootstrap.min.css` inside the static folder of vizro-core and load in the static file into the external stylesheets (as currently in this PR - it's not my preference though)
> 2. Put `vizro-bootstrap.min.css` outside of vizro-core, so it's not part of the package e.g. inside a folder called `dist`. The advantage of this approach is that it doesn't increase the size of our package and we can still reference the CSS file via a CDN link: https://medium.com/javarevisited/how-to-host-your-repository-js-css-on-open-source-cdn-jsdelivr-4de252d6fbad (feels cleaner to me, but then I have to split up this PR as linking the CDN would only work if the file is already present on main)
> 
> With option 2) I could also frequently update the CSS file without having to wait for releasing vizro-core for the changes to be present. The only disadvantage I see is that we have already "a link" that people might pull. Given that we haven't decided yet on how to host this (and if the free CDN link from github will be our long-term solution), I am not sure if this will break for people in the future.


@AnnMarieW  - I think you also went for option 2 right? Based on your files, it also looks like the file is just present in your repository and then you use the free CDN link from github to link it to your stylesheets?

Do I understand correctly?
- This the free Github CDN link you link to your external stylesheets: https://cdn.jsdelivr.net/gh/AnnMarieW/dash-bootstrap-templates/dbc.min.css
- This is the actual file that just lives in your repository: https://github.com/AnnMarieW/dash-bootstrap-templates/blob/main/dbc.css

Screenshot 2024-04-09 at 15 33 54

huong-li-nguyen:
@pruthvip15 - I'll let you know when it's possible, but at some point, I would like you to pull main into this PR or do any commit so the two of us will pop up as code contributors for this PR ❤️ It will be squashed and merged in the end, but your contribution should be visible then :)

.prettierignore
```diff
@@ -1,2 +1,4 @@
 **/docs/
 **/.devcontainer/devcontainer.json
+**/static/css/vizro-bootstrap.min.css
+**/static/css/vizro-bootstrap.min.map
```
pruthvip15, line 4: @huong-li-nguyen double checking, should this be 

```/vizro-bootstrap.min.map``` or ```/vizro-bootstrap.min.css.map```

File: .pre-commit-config.yaml [hunk 2]
```diff
@@ -33,6 +33,7 @@ repos:
     rev: v2.2.6
     hooks:
       - id: codespell
+        args: ["--skip=*.min.css.map, *.min.css"]
```
pruthvip15, line 36: @huong-li-nguyen  here we might have to change accordingly

huong-li-nguyen: Ah yes - if you've renamed the file, then we just have to reflect the name that we use for the map file.

.prettierignore
huong-li-nguyen: I think `.min.css.map` is the common syntax: https://stackoverflow.com/questions/21504611/what-are-the-map-files-used-for-in-bootstrap-3-x

vizro-core/src/vizro/_vizro.py
```diff
@@ -39,6 +35,10 @@ def __init__(self, **kwargs):
         vizro_assets_folder = Path(__file__).with_name("static")
         requests_pathname_prefix = self.dash.config.requests_pathname_prefix
         vizro_css = [requests_pathname_prefix + path for path in self._get_external_assets(vizro_assets_folder, "css")]
+
+        # Ensure vizro-bootstrap.min.css is loaded in first to allow overwrites
+        vizro_css.sort(key=lambda x: not x.endswith("vizro-bootstrap.min.css"))
+
```
huong-li-nguyen, line 41: Just FYI
Screenshot 2024-04-10 at 18 57 32

vizro-core/src/vizro/static/css/vizro-bootstrap.min.map
```diff
@@ -0,0 +1 @@
...AA;EACA,gBAAA;EACA,kBAAA;EACA,oBAAA;EACA,wBAAA;EACA,iBAAA;EI1CE,0BAAA;AlFmkYF;;AkFhkYA;EJ2CI,eAAA;EACJ,qBAAA;EACA,kBAAA;EACA,gBAAA;EACA,kBAAA;EACA,oBAAA;EACA,iBAAA;EACA,iBAAA;EI/CE,0BAAA;AlFykYF;;AkFtkYA;EJ2DI,eAAA;EACJ,qBAAA;EACA,kBAAA;EACA,gBAAA;EACA,kBAAA;EACA,oBAAA;EACA,wBAAA;EACA,iBAAA;EI/DE,0BAAA;AlF+kYF;;AkF5kYA;EJqDI,eAAA;EACJ,qBAAA;EACA,kBAAA;EACA,gBAAA;EACA,kBAAA;EACA,oBAAA;EACA,wBAAA;EACA,iBAAA;EIzDE,eAAA;EACA,0BAAA;AlFqlYF;;AkFllYA;EJ8CI,eAAA;EACJ,qBAAA;EACA,kBAAA;EACA,gBAAA;EACA,kBAAA;EACA,oBAAA;EACA,wBAAA;EACA,iBAAA;EIlDE,eAAA;EACA,0BAAA;AlF2lYF;;AkFvlYA;EJkFI,eAAA;EACJ,qBAAA;EACA,kBAAA;EACA,gBAAA;EACA,kBAAA;EACA,oBAAA;EACA,wBAAA;EACA,iBAAA;EItFE,4BAAA;AlFgmYF;;AkF7lYA;EJiEI,eAAA;EACJ,0BAAA;EACA,kBAAA;EACA,gBAAA;EACA,kBAAA;EACA,oBAAA;EACA,wBAAA;EACA,iBAAA;A9EgiYA;;AkFpmYA,0GAAA;AAEA,wGAAA;AACA;EACE,yBAAA;AlFsmYF;;AkFnmYA;EACE,qCAAA;AlFsmYF;;AkFnmYA;EACE,uCAAA;AlFsmYF;;AmFlqYA;EAEE,2CAAA;AnFoqYF;AmFlqYE;EACE,UAAA;EACA,mBAAA;EACA,aAAA;EACA,YAAA;EACA,uBAAA;EACA,4BAAA;AnFoqYJ;A+E7qYI;EIYE,0BAAA;AnFoqYN;;AGrqYA;EACI,sCAAA;EACA,aAAA;EACA,WAAA;AHwqYJ;;AGrqYA;EACI,sCAAA;EACA,aAAA;EACA,WAAA;AHwqYJ;;AGrqYA;EACI,sCAAA;EACA,aAAA;EACA,WAAA;AHwqYJ","file":"vizro-bootstrap.min.css"}
```
huong-li-nguyen, line 1: For reviewers - if you are wondering what this is for: https://stackoverflow.com/questions/21504611/what-are-the-map-files-used-for-in-bootstrap-3-x

vizro-core/src/vizro/_vizro.py
```diff
@@ -39,6 +35,10 @@ def __init__(self, **kwargs):
         vizro_assets_folder = Path(__file__).with_name("static")
         requests_pathname_prefix = self.dash.config.requests_pathname_prefix
         vizro_css = [requests_pathname_prefix + path for path in self._get_external_assets(vizro_assets_folder, "css")]
+
+        # Ensure vizro-bootstrap.min.css is loaded in first to allow overwrites
+        vizro_css.sort(key=lambda x: not x.endswith("vizro-bootstrap.min.css"))
```
pruthvip15, line 40: 🚀 🚀 🚀 🚀 🚀 🚀

vizro-core/src/vizro/_vizro.py
```diff
@@ -28,9 +27,6 @@ def __init__(self, **kwargs):
         self.dash.config.external_stylesheets.extend(
```
huong-li-nguyen, line 27: @antonymilne - this is important to understand: Our current set-up of specifying external_stylesheets

If a user would like to use another bootstrap theme, this would not work:
`Vizro(external_stylesheets=[dbc.themes.BOOTSTRAP]).build(dashboard).run()`

The current code in `_vizro.py`  always extends to that list, so our `vizro-bootstrap.css` would be read in after the `dbc.themes.Bootstrap`, so the user-defined bootstrap theme has no effect. At this point in time, it's definitely our desired result, as otherwise everything looks completely off.

However, just know, that in case someone still wants to overwrite it, it has to be done like this (right now):

```
  app.dash.config.external_stylesheets.extend([dbc.themes.BOOTSTRAP])
  app.run()
```

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -32,8 +31,4 @@ class Button(VizroBaseModel):
 
     @_log_call
     def build(self):
-        return html.Div(
-            dbc.Button(id=self.id, children=self.text, className="button_primary"),
-            className="button_container",
-            id=f"{self.id}_outer",
-        )
+        return dbc.Button(id=self.id, children=self.text)
```
antonymilne, line 34: This sort of tidying is great and I think a sign we're doing the right thing 👍 The more we can do this, the better. When I compare the DOM to what it used to look like on VizX or even when we first released Vizro, it's so much cleaner now and that is great. We should definitely continue pushing in this direction for nice simple, semantic HTML without unnecessary nesting. 💯

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -34,15 +34,16 @@ class Card(VizroBaseModel):
...
     @_log_call
     def build(self):
-        text = dcc.Markdown(self.text, className="card_text", dangerously_allow_html=False, id=self.id)
-        card_content = (
-            dbc.NavLink(
-                text,
-                href=get_relative_path(self.href) if self.href.startswith("/") else self.href,
-                className="card-link",
+        text = dcc.Markdown(self.text, dangerously_allow_html=False, id=self.id)
+        card_content = dbc.CardBody(
```
antonymilne, line 38: Suggest refactoring this to remove the change here and instead put the `dbc.CardBody` in the return statement just to make it a bit more readable (if I understood the change correctly):
```
return dbc.Card(dbc.CardBody(card_content), className=card_class, id=f"{self.id}_outer")
```

vizro-core/src/vizro/_vizro.py
antonymilne: Yeah, this is indeed right for now I think. Probably we'll need to change the logic here a bit if and when we introduce CDN as an option anyway.

antonymilne (approved):
This is AMAZING work, thank you @huong-li-nguyen and @pruthvip15 also who I know has been very instrumental behind the scenes here 🚀 This is a big step forward I think.

For now I'm happy to stick with option 1 like you have here where we just serve the stylesheet locally because that's easy and always works. Then we can figure out later how to handle CDN, `serve_locally` option, whether to provide a `vizro.themes.stylesheet` shortcut like bootstrap does, whether to try and put it in bootswatch/dbc, etc. etc.

I guess there might be a few follow on PRs like this while I'm on holiday - feel free to merge them all if they're just doing similar work like you're doing here, no need to wait for my review 🚀

vizro-core/src/vizro/models/_components/card.py
huong-li-nguyen: Let me actually remove the `dbc.CardBody` altogether. I think the `card_content` can be inserted directly into the `dbc.Card` 👍

nadijagraca (approved):
This looks really good. Awesome job! 🚀  🙌

Issue #383, opened 2024-03-25
CSS of AG Grid `floatingFilter` option buggy


maxschulz-COL:
### Description
Floating filter not looking ideal. This sort of buggy behaviour may arise with a few other AG Grid options that we have not explicitly tested. In an ideal world we would build our own custom theme that takes care of all potential options automatically.

![image](https://github.com/mckinsey/vizro/assets/83698606/feb15fd5-3be6-43c1-bb5d-cc3765612011)

### Expected behavior

Input field not cutting over the continuous line

### Which package?

vizro

### Package version

0.1.13

### Python version

3

### OS

Mac

### How to Reproduce

```python
import vizro.models as vm
import vizro.plotly.express as px
from vizro import Vizro
from vizro.tables import dash_ag_grid

df = px.data.gapminder(datetimes=True)

page = vm.Page(
    title="Enhanced AG Grid",
    components=[
        vm.AgGrid(
            title="Dash AG Grid",
            figure=dash_ag_grid(
                data_frame=df,
                columnDefs=[
                    {"field" : "country", 'floatingFilter': True},
                    {"field" : "continent"},
                    {"field" : "year"},
                    {"field" : "lifeExp", "cellDataType": "numeric"},
                    {"field" : "pop", "cellDataType": "numeric"},
                    {"field" : "gdpPercap", "cellDataType": "euro"},]

            )
        ),
    ],
    controls=[vm.Filter(column="continent")],
)
dashboard = vm.Dashboard(pages=[page])

Vizro().build(dashboard).run(port = 8051)
```

### Code of Conduct

- [X] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

huong-li-nguyen:
Thanks for raising! I took a look into this one, and this should be quick to fix in terms of CSS. 

In terms of UI, if we want to turn on the `floatingFilter` in any of the examples in the docs, I think it's better if we simultaneously turn off the filter menu option in the header. Otherwise, it's confusing as the filter is present in both the header and the floatingFilter. However, I checked the configurations and somehow that's not possible - I've raised an issue on Dash here: https://github.com/plotly/dash-ag-grid/issues/280

So for now, we can definitely fix the CSS as it's quick to do and otherwise it's unusable. But I wouldn't add an example with the `floatingFilter` in any of the official docs just yet

PR #388, opened 2024-03-26
[Bug] Fix CSS for `floatingFilter` in `AgGrid`

huong-li-nguyen:
## Description
- Fixes https://github.com/mckinsey/vizro/issues/383

As already discussed, I think we need to evaluate on a case-by-case basis whether we fix the relevant CSS or whether the user needs to add the CSS themselves. In general, the default ag-grid theme should look fine without having to apply any custom CSS (e.g. the datepicker looks different than ours, but it looks fine still). The disadvantage of trying to cover every case is that our custom CSS grows and at some point gets difficult to maintain, because our overwrites might conflict. Long-term wise we should probably create a proper Figma design for the ag-grid, but that's currently not in the scope.

 In this case, I agree we should fix it, as otherwise it's completely misaligned and it doesn't add much CSS.

## Screenshot
Screenshot 2024-03-26 at 12 56 09

## Notice


vizro-core/examples/_dev/app.py
```diff
@@ -1,181 +1,33 @@
...
-            column="time",
-            selector=NewDropdown(
-                options=[datetime.datetime(2024, 1, 1) + datetime.timedelta(days=i) for i in range(31)]
+        vm.AgGrid(
+            title="Dash AG Grid",
+            figure=dash_ag_grid(
+                data_frame=df,
+                columnDefs=[
+                    {"field": "country", "floatingFilter": True, "suppressHeaderMenuButton": True},
+                    {"field": "continent", "floatingFilter": True, "suppressHeaderMenuButton": True},
```
huong-li-nguyen, line 19: Note: See comment [here](https://github.com/mckinsey/vizro/issues/383#issuecomment-2020211906). Optimally we would like to suppress the header menu button for the fields where the floatingFilter is turned on. However, this doesn't seem to work (also not in a pure dash app). So for now we can fix the CSS but I wouldn't add the floatingFilter as an example to any of the docs / demo apps just yet.

maxschulz-COL (approved):
Amazing! I would even add this as a fix to the fixed section. It definitely makes a difference to users. Thanks for taking care of this 🙏

nadijagraca (approved):
Looks great. 🚀

PR #393, opened 2024-03-27
[Tidy] Improve semantics of `NavBar` and `Accordion`

huong-li-nguyen:
## Description
- Replace `html.Div` with `dbc.Nav` and `dbc.Navbar` (for context on [why we should do the switch](https://stackoverflow.com/questions/36097957/what-is-the-difference-between-nav-and-div))

## Notice


vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -32,8 +32,13 @@
...
 def _all_hidden(components: List[Component]):
-    """Returns True if all `components` are either None and/or have hidden=True."""
-    return all(component is None or getattr(component, "hidden", False) for component in components)
+    """Returns True if all `components` are either None and/or have hidden=True and/or className contains `d-none`."""
+    return all(
+        component is None
+        or getattr(component, "hidden", False)
+        or "d-none" in getattr(component, "className", "d-inline")
+        for component in components
+    )
```
huong-li-nguyen, line 41: @antonymilne - I know we've changed this back and forth 😅 I tried to do everything in this commit so it's easier to see which changes are required: https://github.com/mckinsey/vizro/pull/393/commits/74031d621cf2201fda4709f51cfa945e3ba54eab
Essentially the `dbc.Nav` / `dbc.Navbar` do not have a `hidden` argument, so we have to go back to one of the previous approaches of providing a `className` for these. Note I am using the classNames from Bootstrap utilities: https://getbootstrap.com/docs/5.0/utilities/display/. Returning None wouldn't work in this case, as we need to access the the html elements `nav-bar` and ´nav-panel´ via their ID later on.

vizro-core/src/vizro/models/_navigation/accordion.py
```diff
@@ -44,7 +42,7 @@ def build(self, *, active_page_id=None):
         # Note build does not return _NavBuildType but just a single html.Div with id="nav-panel".
         # Hide navigation panel if there is only one page
         if len(list(itertools.chain(*self.pages.values()))) == 1:
-            return html.Div(hidden=True, id="nav-panel")
+            return dbc.Nav(id="nav-panel", className="d-none invisible")
```
huong-li-nguyen, line 45: `d-none` is equivalent to `display: none`
`invisible` is equivalent to `visibility: hidden`

https://getbootstrap.com/docs/5.0/utilities/display/

vizro-core/src/vizro/models/_dashboard.py
antonymilne: This looks good, thanks for making the change so clear just in one commit! 🚀 

Just one thing I'm curious about: why do we put `className="d-none invisible"` and not just `d-none`?

antonymilne (approved):
This is great, thank you!

maxschulz-COL (approved):
👍

PR #402, opened 2024-04-03
[Docs] Update build docs settings to use `--strict` 

stichbury:
## Description
I've updated the docs build settings to use `--strict` and catch any warnings as they arise.

Unfortunately, they also catch the griffe warnings in mkdocstrings. Could use some help from the engineering team with these please, all in the same file:

```text
WARNING -  griffe: src/vizro/_vizro.py:23: No type or annotation for parameter 'kwargs'
WARNING -  griffe: src/vizro/_vizro.py:70: No type or annotation for returned value 'Vizro'
WARNING -  griffe: src/vizro/_vizro.py:88: No type or annotation for parameter 'args'
WARNING -  griffe: src/vizro/_vizro.py:89: No type or annotation for parameter 'kwargs'
```

```python
class Vizro:
    """The main class of the `vizro` package."""

    def __init__(self, **kwargs):
        """Initializes Dash app, stored in `self.dash`.

        Args:
            kwargs: Passed through to `Dash.__init__`, e.g. `assets_folder`, `url_base_pathname`. See
                [Dash documentation](https://dash.plotly.com/reference#dash.dash) for possible arguments.

        """

...
def build(self, dashboard: Dashboard):
        """Builds the `dashboard`.

        Args:
            dashboard (Dashboard): [`Dashboard`][vizro.models.Dashboard] object.

        Returns:
            Vizro: App object

        """
...

def run(self, *args, **kwargs):  # if type annotated, mkdocstring stops seeing the class
        """Runs the dashboard.

        Args:
            args: Passed through to `dash.run`.
            kwargs: Passed through to `dash.run`.

        """


```

How would I go about fixing this? I tried adding the * / ** in the docstring (as suggested on stack overflow) but still got the warning.

cc @huong-li-nguyen @antonymilne 

## Notice

### DIFF
--- a/vizro-ai/.readthedocs.yaml
+++ b/vizro-ai/.readthedocs.yaml
@@ -11,6 +11,6 @@ build:
     python: "3.11"
   commands:
     - pip install hatch
-    - cd vizro-ai/ && hatch run docs:mkdocs build
+    - cd vizro-ai/ && hatch run docs:mkdocs build --strict
     - mkdir --parents $READTHEDOCS_OUTPUT
     - mv vizro-ai/site/ $READTHEDOCS_OUTPUT/html
--- a/vizro-ai/changelog.d/20240404_112532_jo_stichbury_update_build_docs_settings.md
+++ b/vizro-ai/changelog.d/20240404_112532_jo_stichbury_update_build_docs_settings.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Deprecated
+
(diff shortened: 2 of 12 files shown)


antonymilne:
```
WARNING -  griffe: src/vizro/_vizro.py:25: No type or annotation for parameter 'kwargs'
WARNING -  griffe: src/vizro/_vizro.py:74: No type or annotation for returned value 'Vizro'
WARNING -  griffe: src/vizro/_vizro.py:92: No type or annotation for parameter 'args'
WARNING -  griffe: src/vizro/_vizro.py:93: No type or annotation for parameter 'kwargs'
```

The 2nd of these is worth trying to fix because the docstring for `build` looks wrong. It should be this:
```
from typing_extensions import Self

def build(self, dashboard: Dashboard) -> Self:
        """Builds the `dashboard`.

        Args:
            dashboard (Dashboard): [`Dashboard`][vizro.models.Dashboard] object.

        Returns:
            Vizro app
        """
```

The other 3 could probably be fixed with some fiddling but I don't think is worth doing because type-hinting these variables doesn't benefit anyone in reality, and maybe it's not even so easy to get working (see the comment `# if type annotated, mkdocstring stops seeing the class`).

We are going to need some way to ignore griffe's warnings anyway because it currently raises lots of warnings that we don't care about (basically everything of the form `WARNING -  griffe: src/vizro/models/_components/form/slider.py:36: Parameter 'actions' does not appear in the function signature`). So I think we should figure out how to best ignore griffe here also.

stichbury:
> We are going to need some way to ignore griffe's warnings anyway because it currently raises lots of warnings that we don't care about (basically everything of the form `WARNING - griffe: src/vizro/models/_components/form/slider.py:36: Parameter 'actions' does not appear in the function signature`). So I think we should figure out how to best ignore griffe here also.

I've fixed that one! Well, suppressed it. It's the one I can fix, so that's killed over 100 warnings.

However, it's just these 4 left (or 3 if I fix the one you describe). I couldn't find a generic "ignore" but will keep looking.

huong-li-nguyen (approved):
This looks good - amazing job for figuring it all out! 🚀 🔥 

I suspect the linting step will fail which is not connected to this PR but the mypy version being updated recently. For now, just change it back to the previous version in the pre-commit-config.yaml, because I've quickly checked the errors and they are not quick to fix 😅 

Screenshot 2024-04-04 at 12 04 22

huong-li-nguyen:
Just saw that the linting step still fails, it doesn't show me these warnings locally anymore so I think the Github actions eventually use some cached hatch environment? We can quickly check with @l0uden and the team during stand-up

nadijagraca (approved):
LGTM. ⭐

PR #406, opened 2024-04-05
[CI] Fix mypy linting

antonymilne:
## Description

#397 bumped our pre-commit mypy from 1.8.0 to 1.9.0, which seemed to flag new errors that were still there even after reversion to 0.18.0 in #403.

The reason for these new errors was **not** actually the change to mypy version but that there was a new pydantic release at about the same time. Pinning the pydantic version used by mypy fixes this.

There is no caching in our CI. There is caching on pre-commit.ci but it works well and will always be running the correct version specified in our .pre-commit-config.yaml. So both our CI and pre-commit.ci are still perfect sources of truth and working correctly. Behaviour locally should be the identical to these, but just in case it's not then `pre-commit clean` should fix it by clearing your local cache.

## Notice

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,7 +72,7 @@ repos:
       - id: mypy
         files: ^vizro-core/src/
         additional_dependencies:
-          - pydantic>=1.10.13, <2 # deliberately pinned to <2 until we bump our pydantic requirement to strictly >=2
+          - pydantic==1.10.14 #, <2 # deliberately pinned to <2 until we bump our pydantic requirement to strictly >=2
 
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -121,3 +125,4 @@ ci:
     - bandit
     - mypy
     - gitleaks
+    - vale
```
antonymilne, line 128: FYI @maxschulz-COL @stichbury this should have been on the skip list for https://pre-commit.ci/ already, because vale just flags errors rather than fixing them. The only things we run on pre-commit.ci are things that can fix stuff.

huong-li-nguyen (approved):
Ahh, amazing!! 🚀

PR #407, opened 2024-04-05
[Bug] Turn on `autocolorscale` for assignment of correct continuous color palettes

huong-li-nguyen:
## Description
- Previously continous color palettes were not correctly assigned to Plotly charts, which is now fixed by turning autocolorscale to True
- See the example dev to understand:
    - Our diverging color palette is assigned if the data has positive and negative values
    - Our sequential color palette is assigned if the data has positive values only
    - Our sequentialminus color palette is assigned if the data has negative values only
- Squeezed in another fix for a CSS bug that has been triggering me everytime I saw a screenshot 😄  It's removing these blue lines when you focus on it

Screenshot 2024-04-05 at 13 27 47

## Screenshot

**Run dev example:**

Screenshot 2024-04-05 at 13 56 30

## Notice

### DIFF
--- a/vizro-core/changelog.d/20240405_135655_huong_li_nguyen_change_theme_default.md
+++ b/vizro-core/changelog.d/20240405_135655_huong_li_nguyen_change_theme_default.md
@@ -1,0 +1,47 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Turn on `autocolorscale` for color palettes to be assigned correctly in Plotly charts. ([#407](https://github.com/mckinsey/vizro/pull/407))
+
+<!--
+### Security
+
+- A bullet item for the Security category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
--- a/vizro-core/examples/_dev/app.py
+++ b/vizro-core/examples/_dev/app.py
@@ -1,31 +1,57 @@
 """Example to show dashboard configuration."""
 
+from typing import Optional
+
+import pandas as pd
 import vizro.models as vm
 import vizro.plotly.express as px
 from vizro import Vizro
-from vizro.tables import dash_ag_grid
+from vizro.models.types import capture
 
-df = px.data.gapminder()
+gapminder_pos = px.data.gapminder()
+gapminder_neg = px.data.gapminder()
+gapminder_mixed = px.data.gapminder()
+gapminder_neg["lifeExp"] = gapminder_neg["lifeExp"] * (-1)
+gapminder_mixed["lifeExp"][:40] = gapminder_mixed["lifeExp"][:40] * (-1)
+
+
+@capture("graph")
+def variable_map(data_frame: pd.DataFrame = None, color: Optional[str] = None):
+    """Custom choropleth figure that needs post update calls."""
+    fig = px.choropleth(
+        data_frame,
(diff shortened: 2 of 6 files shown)


petar-qb:
To be honest, I don't fully understand why we need this change 😄. Below is a screenshot for the same application running from the `main` branch and it somehow seems more natural to me to understand because the same colour palette is always used and users don't have to look carefully at the colour palette legend to understand which colour means higher and which colour means lower values.

This means that if we add a constant number (eg +15) to each cell element, the same chart colours will be displayed per country on the map (and only the legend mark label values will change). However, that's just my opinion and I don't know if there's any UX rule that says otherwise. 😄

image

Also, isn't that red colours should be used for negative and blue for positive values in your "Mixed Life Expectancy" example?

petar-qb:
@huong-li-nguyen , thanks a lot for the explanation and the link (the article is excellent!). All these changes make sense to me now 😄.

P.S. I've updated the screenshot from the PR description.

petar-qb (approved):

antonymilne (commented):
Thanks for making the change, this is a huge improvement! I knew something like `coloraxis_autocolorscale=True` should exist so great to see it does exactly what we need 😀 

Just one thing though: do we definitely want the colors this way round? So that blue = high and red = low. To me this feels backwards since it's the opposite of a temperature scale that goes from blue = cold to red = hot.

Definitely I agree that we should be consistent between sequential and diverging palettes so that the same color always means the same thing, but to me these should be completely reversed everywhere.

maxschulz-COL (approved):
⭐ Love it! Reminds me of our discussion back about the colors of the scatter plot! This is really how it should be :)

Have you checked if this affects any custom choices? I would assume `color_continuous_scale` and `color_continuous_midpoint` still work (e.g. when I would like the middle of my diverging scale to be a certain value)

maxschulz-COL:
> Just one thing though: do we definitely want the colors this way round? So that blue = high and red = low. To me this feels backwards since it's the opposite of a temperature scale that goes from blue = cold to red = hot so it would feel more natural to me to have blue = low and red = high.
> 
> Definitely I agree that we should be consistent between sequential and diverging palettes so that the same color always means the same thing, but to me these should be completely reversed everywhere.

Tricky - I feel like that since we have this automation specifically for negativ to positive numbers, I am more comfortable with red = negativ, blue = positive (green would probably make this even more clear...)

huong-li-nguyen:
> > So when using diverging color scales, take an extra minute to think about your color choices. Which ones would be the most intuitive colors for your readers (e.g. [party colors](https://blog.datawrapper.de/partycolors/), or red = bad)? And how can you make clear which color means which extreme? Label your color legend well and/or [turn your title into a color key](https://academy.datawrapper.de/article/190-how-to-turn-your-title-into-a-color-key), and use annotations in the chart to remind readers what they see.

@antonymilne - the answer is it depends 😄 I did ask Joe and Dan about the current choice and they both agreed that blue represents positive values, and red represents negative values and intuitively I also thought so. However, I definitely understand your confusion and I don't think this color sequence is crystal clear. But the direction of color sequence not only depends on the +/n sign but actually the context of your data.

In the example below, the negative values are represented by green, and the positive values by red. Now you might think: Shouldn't it be vice versa? In this case, it depends. If the desired parameter/outcome is actually a negative value (let's say -40% is our target), then green color is correctly assigned to the negative values, because green is usually thought of as the positive outcome color. Positive in the sense of desired result.

Screenshot 2024-04-08 at 12 28 02

Regarding @maxschulz-COL comment on green-red, we always try to avoid that due to accessibility concerns (red-green color blindness which is actually quite common), so that's we use blue-red.

I don't think the diverging red-cyan will always represent the color sequence correctly because it depends on the data context and the example above also shows that negative values do not always represent negative outcomes. So I don't think we can get it right for every use case. I would therefore stay with the current choice as it's Dan's preference.

huong-li-nguyen:
> Have you checked if this affects any custom choices? I would assume `color_continuous_scale` and `color_continuous_midpoint` still work (e.g. when I would like the middle of my diverging scale to be a certain value)

Very good point @maxschulz-COL !! I just tried that out and it is **not compatible** 😭  So maybe that was the reason we didn't turn it on in the past 🤔 Here for context: https://stackoverflow.com/questions/61083382/plotly-make-subplots-doesnt-keep-selected-colors 

This means howver, we have two choices to make:

1. We have `autocolorscale=True` and automatically provide a sequential and diverging color sequence based on the data, but if a user wants to provide their own color sequence via `color_continuous_scale` or other params, they explicitly have to turn off `autocolorscale=False` in their update layout calls.

2. We don't turn on `autocolorscale=False`, which enables the user to use `color_continuous_scale` in their charts without having to explicitly do a post update call, but this means we can only provide one default continuous color scale which would be SEQUENTIAL_CYAN. 

Both approaches have pros and cons..any preferences? @antonymilne @maxschulz-COL @petar-qb @Joseph-Perkins

huong-li-nguyen:
> Both approaches have pros and cons..any preferences? @antonymilne @maxschulz-COL @petar-qb @Joseph-Perkins

We decided to go for option 2) for easy use for users to still customise their color sequences.

vizro-core/changelog.d/20240405_135655_huong_li_nguyen_change_theme_default.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Change default continuous color scale to `SEQUENTIAL_CYAN´. ([#407](https://github.com/mckinsey/vizro/pull/407))
```
antonymilne, line 28:
```suggestion
- Change default continuous color scale to `SEQUENTIAL_CYAN`. ([#407](https://github.com/mckinsey/vizro/pull/407))
```

antonymilne (approved):
Re-approving with the new solution 😀 Thanks for the excelelnt code comments that make it all clear!

PR #425, opened 2024-04-16
[QA] Ignore `pytest.PytestUnhandledThreadExceptionWarning`

l0uden:
## Description
Ignoring `pytest.PytestUnhandledThreadExceptionWarning` in the integration tests as it is flaky right now and this warning doesn't affect anything

## Notice

### DIFF
--- a/vizro-core/changelog.d/20240416_134105_alexey_snigir_ignore_pytest_thread_warning.md
+++ b/vizro-core/changelog.d/20240416_134105_alexey_snigir_ignore_pytest_thread_warning.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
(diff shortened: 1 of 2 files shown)


petar-qb (approved):
This looks good! 

Thanks for temporarily masking the annoying CI bug. We should probably get rid of this warning, but until then, it's fine to ignore it. 😄

huong-li-nguyen (approved):
Do you think we need to investigate this further and actually fix it? Or do you think this is safe to ignore and is our long-term solution?

Depending on that could you add a ticket for any required follow-ups? 🙏

l0uden:
> Do you think we need to investigate this further and actually fix it? Or do you think this is safe to ignore and is our long-term solution?
> 
> Depending on that could you add a ticket for any required follow-ups? 🙏

I think it is safe to ignore, because it is happening in the teardown of the test. So it's a long-term.

l0uden:
> This looks good!
> 
> Thanks for temporarily masking the annoying CI bug. We should probably get rid of this warning, but until then, it's fine to ignore it. 😄

I  think it is not temp solution because this thing happens after test passed, on teardown. In our case we're not checking anything on this stage so can skip the warning

PR #433, opened 2024-04-22
[Tidy] Update dashboard layout

huong-li-nguyen:
## Notice


vizro-core/src/vizro/static/css/layout.css
```diff
@@ -170,17 +171,10 @@
   color: var(--text-active);
 }
 
-.collapse-button-tooltip {
-  height: 24px;
-  left: -12px;
-  position: relative;
-  top: 36px;
-}
-
```
huong-li-nguyen, line 179: @nadijagraca - I've removed this CSS className because it didn't seem to affect anything? Could you double-check for me if that's fine to do or if there was a reason why we had this class in.

nadijagraca: We had this class to position the tooltip, but it does seem it has been fixed with the latest changes so it is safe to remove.

vizro-core/src/vizro/static/js/models/dashboard.js
```diff
@@ -20,7 +20,6 @@ export function _collapse_nav_panel(n_clicks, is_open) {
         false,
         {
           transform: "rotate(180deg)",
-          left: "4px",
```
nadijagraca, line 23: Why was this removed? 
The `left: "4px"` setting is intended to position the collapse icon so that it remains fully visible when the left side is collapsed. Without this you can only see half of the icon, or this is the intended behaviour?
We can change it to `left: "2px"` so it does not overlap with the page title.

vizro-core/src/vizro/static/js/models/dashboard.js
```diff
@@ -33,7 +32,6 @@ export function _collapse_nav_panel(n_clicks, is_open) {
       false,
       {
         transform: "rotate(180deg)",
-        left: "4px",
```
nadijagraca, line 36: Same as above.

vizro-core/src/vizro/static/js/models/dashboard.js
nadijagraca: Without `left: "4px"` the collapse icon looks like this when collapsed. 

Screenshot 2024-04-23 at 10 23 16

huong-li-nguyen: Yes, this is the intention that only half of the icon is visible. This is still to be discussed though. I see the advantages of both. The half-visible one doesn't change the position of the icon, so naturally "the effect" is only to collapse the side-panel without a jump in the icons position. 

I do get the reservations of the icon being only half-visible though. So this default behaviour is still to be discussed w/ J. We haven't decided on a default yet.

huong-li-nguyen: We decided to keep it like that 👍

nadijagraca (approved):
Looks really sleek. 🚀

Joseph-Perkins (approved):

Issue #435, opened 2024-04-24
Custom AG Grid function overwrites pandas.DataFrame provided as input with an empty pandas.DataFrame


pablo-fence:
### Description

`data_frame` is an empty pandas.DataFrame within the scope of custom AG grid function `matrix_aggrid()`, see line debug print on line 28 `print("Original DataFrame size:", data_frame.shape)` .

`data_frame` is not empty just prior to the function call in debug print line 71.

### Which package?

vizro

### Package version

0.1.15

### Python version

3.12.1

### OS

macOS sonoma 14.4

### How to Reproduce

1. Install package dependencies in provided requirements.txt
[requirements.txt](https://github.com/mckinsey/vizro/files/15084571/requirements.txt)

2. Run the following code
```
import pandas as pd
import vizro.models as vm
from vizro.models.types import capture
from dash_ag_grid import AgGrid
from typing import List

# Sample data generator
def generate_sample_data():
    # Creating a DataFrame with the necessary structure
    data = {
        'client_id': range(1, 11),
        'cohort': ['2020-Q1'] * 5 + ['2020-Q2'] * 5,
        'relative_month': [0, 1, 2, 3, 4] * 2,
        'amount_paid': [100, 150, 200, 250, 300] * 2,
        'initial_balance': [500, 500, 500, 500, 500] * 2
    }
    return pd.DataFrame(data)

@capture('ag_grid')
def matrix_aggrid(data_frame, id_column="client_id", row_dimension_column="cohort",
                  column_dimension_column="relative_month", numerator_column="amount_paid",
                  denominator_column="initial_balance", cell_values="numerator_column", periods_to_show=[0, 12]):
    print("Original DataFrame size:", data_frame.shape)
    min_cohort, max_cohort = periods_to_show
    data_frame[column_dimension_column] = data_frame[column_dimension_column].astype(float)
    mask = (data_frame[column_dimension_column] >= min_cohort) & (data_frame[column_dimension_column] <= max_cohort)
    data_frame = data_frame[mask].copy()

    initials = data_frame[[id_column, row_dimension_column, denominator_column]].drop_duplicates().groupby(row_dimension_column)[denominator_column].sum().reset_index()
    gpd = data_frame.groupby([id_column, row_dimension_column, column_dimension_column])[numerator_column].last().reset_index()
    gpd2 = gpd.groupby([row_dimension_column, column_dimension_column])[numerator_column].sum().reset_index()
[...]

PR #439, opened 2024-04-26
[Bug] Fix bug that prevented custom tables/grids with column referral

maxschulz-COL:
## Description
Closes https://github.com/mckinsey/vizro/issues/435 and https://github.com/McK-Internal/vizro-internal/issues/747

We now send the full DF during the build phase (instead of an empty `pd.DataFrame`). This could have performance consequences for very large DFs, but on the grand scheme of things I think it probably the better solution.

And alternative would be to send an empty DF with the columns present, but that may lead to other bugs down the road.

## Notice


huong-li-nguyen:
Thanks for fixing so quickly! 🚀  What "potential bugs down the road" do you imagine? I actually like the alternative solution approach - it's also the one we've implemented on VizX actually. I think it did not lead to any major bugs, and the performance issue was more severe.

maxschulz-COL:
> Thanks for fixing so quickly! 🚀 What "potential bugs down the road" do you imagine? I actually like the alternative solution approach - it's also the one we've implemented on VizX actually. I think it did not lead to any major bugs (as far as I remember), and the performance issue was more severe.

So of the top of my head I could imagine that someone not only refers to a column, but also a specific row. This then would fail, as there are no rows present.

This seems to be a case of a wider range of things where people refer or rely on the size of the data, or generally anything regarding the original DF they provide.

One could argue that it is bad design to write a custom grid/table like that, but the confusion of the user showed that people simply do not expect the data to be different at any point from what they originally provide

> Do we know how other tools deal with this? e.g. do they load in the data in batches? or do they always load in the entire dataset?

So the AgGrid for example has the opportunity to [enable batch loading in infinite scroll](https://www.ag-grid.com/javascript-data-grid/infinite-scrolling/#turning-on-infinite-scrolling), but I think that defeats a little bit the point. In principle we are not converned with loading the entire data, that is what we do anyway once the page loads with all filters/parameters etc. It is simply this initial building that immediately gets overwritten.

Ideally we would want to just create an entirely different loading component (like for `graph`), but it has been shown (see comment above the build lines in both table and grid) that having different settings for the objects causes problems (the case we observed was the pagination setting). Remember the long discussion we had where you said, just write "Do not change this line" :). So this is related to that.

huong-li-nguyen:
Thanks for clarifying!

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -111,7 +111,7 @@ def build(self):
                 # here) must have the same setting as the object that is built by the on-page-load mechanism using
                 # with the user settings and rendered finally. Otherwise the grid is not rendered correctly.
                 # Hence be careful when editing the line below.
-                html.Div(self.__call__(data_frame=pd.DataFrame()), id=self.id, className="table-container"),
+                html.Div(self.__call__(), id=self.id, className="table-container"),
```
petar-qb, line 114: Maybe something like this could work:

```
 def build(self):
        clientside_callback(...)
        grid = self.__call__()
        grid["rowData"] = {}  # or `grid["rowData"] = pd.DataFrame()`
        return dcc.Loading([..., html.Div(grid, ...), ...], ...)
```

This should send the original data_frame into the `figure` callable, but also return an empty DataFrame with all calculated configuration on the `build load`. Then the UI chart goes into the Loading state , and finally the `update_figures` returns filtered data + parametrised configuration.

maxschulz-COL: It is an interesting idea that I will try out

antonymilne:
I would like to review this so please don't merge yet 🙂

vizro-core/src/vizro/models/_components/table.py
```diff
@@ -108,7 +108,7 @@ def build(self):
             html.Div(
                 [
                     html.H3(self.title, className="table-title") if self.title else None,
-                    html.Div(self.__call__(data_frame=pd.DataFrame()), id=self.id),
+                    html.Div(self.__call__(), id=self.id),
```
antonymilne, line 111: Please could you add an explanatory comment here that refers to the one in `AgGrid.build` or just copies and pastes it in? (I don't know if the pagination stuff is also relevant here or if it's just done the same way to match AgGrid, but I was confused just reading through the code why we did this here rather than just returning an empty `DataTable`)

antonymilne (approved):
tl;dr: thanks @petar-qb and @maxschulz-COL for your great work on this. Hopefully longer term we will somehow have an entirely better system here but for now this looks good.

---

I prefer the solution @petar-qb suggested in https://github.com/mckinsey/vizro/pull/439/files#r1580918187 so would be great if you could try it out. But the current solution is fine too.

I am not hugely concerned right now about the performance hit this incurs, but I am not keen on it either. In general I don't much like the "double loading" we currently have where we do everything once with an empty dataframe (or the real thing, like now) and then immediately override it with the real data. This PR unfortunately, but necessarily, adds another layer of unsatisfactoriness to that scheme 😬 

That isn't meant as a criticism of this PR or the existing system, since I know the current on-page-load system has its merits and lots of good reasoning behind it. I tried to come up with some improvements before and couldn't. I just hope that as part of the actions v2 work we can somehow improve this scheme though 🤞

vizro-core/src/vizro/models/_components/ag_grid.py
maxschulz-COL: I tried this in https://github.com/mckinsey/vizro/pull/439/commits/4c3924653c2db277d9200dbd010005d043a1a6f6, and it _almost works_.

There seem to be no issues with pagination for example, but unfortunately it seems to break the persistence settings.

Presumably that is because it "forgets" the filters when an intermediate AGGrid has no data. You can check out the complete changes at the commit above. Try setting a filter for the first AGGrid and change page. Upon returning the filter is gone, although persistence is set to true.

antonymilne: Ah ok, thanks for trying it out! I'm fine with your current solution then. Hopefully we can improve this some time in the future.

antonymilne (approved):

PR #446, opened 2024-04-30
[Bug] Fix reloading issues of the AG Grid

maxschulz-COL:
## Description
This PR closes the following issues:
- https://github.com/McK-Internal/vizro-internal/issues/670
- https://github.com/McK-Internal/vizro-internal/issues/660
- https://github.com/McK-Internal/vizro-internal/issues/747 (this would be already closed by https://github.com/mckinsey/vizro/pull/439)

After some investigation, it looks like there is some confirmed buggy behaviour:
- column size arguments not consistently rendering on page refresh or change (e.g. a user providing `columnSize="autoSize"` will not see his results consistently)
- AG Grid filter persistence not working properly on page refresh

The identified two  underlying reasons are:
- not providing the full DF in the initial build method of the AG Grid (again, similar topic that https://github.com/McK-Internal/vizro-internal/issues/747 adresses)
- providing defaults to the `_dash_ag_grid` functions that concern columns settings that could potentially conflict with user chosen values such as `columnSize="autoSize"`

It has to be noted that this behaviour is not always easy to investigate, as there are many moving pieces etc. This PR presents to options for investigation:
- `app.py` - Vizro app setup to investigate different options.
- `app2.py` (will be deleted later) Pure Dash App to test what "bugs" carry over

### Proposed solution
- https://github.com/McK-Internal/vizro-internal/issues/747
- Removing any column width defaults from `_dash_ag_grid`

Those two changes fix all observed problems for me. This is the current status of the code, and it would be good if reviewers could test out if the see the same with the various settings possible.

### How to review
Run `hatch run example` and test the various options proposed in `app.py`. Particular focus should be on:
- if I choose a column setting as a user, is that always reflected on screen, even after page change, and page refresh
- same for persistence settings

`app2.py` and the additional files under `pages` is just there in case we want to try out things in a pure Dash app. I will delete once we agree on a solution.

## Notice


vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -111,7 +111,7 @@ def build(self):
                 # here) must have the same setting as the object that is built by the on-page-load mechanism using
                 # with the user settings and rendered finally. Otherwise the grid is not rendered correctly.
                 # Hence be careful when editing the line below.
-                html.Div(self.__call__(data_frame=pd.DataFrame()), id=self.id, className="table-container"),
+                html.Div(self.__call__(), id=self.id, className="table-container"),
```
petar-qb, line 114: Should we apply the same changes in the `Table.__call__` as well?

petar-qb (approved):
@maxschulz-COL thank you for all your work on this PR. I know how hard it was to test all the cases and options and to find what was causing the persistence issues. 

By returning the original `data_frame` from`figure.build` and with other changes (like the default dash_ag_grid configuration) you really solved all the mentioned issues. 

I tried to find any inconsistency by altering arguments such as: persistence, persistence_type, persisted_props, columnSize and defaultColDef and found that the problem with the `filterModel` disappeared completely.

However `columnSize` doesn't work if configured in combination withsomething like `defaultColDef={"flex": 1, "minWidth": 200},`. This is really an edge case where the user can expect that the column size is auto-sized but that the width is never lower than 200. Also, if `columnSize="autoSize"` is set in the combination with `persisted_props=["columnSize"]`, "autoSize" doesn't work at all. This seems strange to me since, `columnSize` and `filerModel` should behave same since they are both dag.AgGrid properties.

I would go with this solution (hence the approval from my side), but we should also create a new ticket to optimise the page load process since it doesn't seem like a perfect solution to return the original data_frame from `figure.build` and to overwrite it immediately afterwards with `on-page-load`.

antonymilne:
FYI @petar-qb since you were away at the time, just for reference see my comment here: https://github.com/mckinsey/vizro/pull/439#pullrequestreview-2030995568.

maxschulz-COL:
> However `columnSize` doesn't work if configured in combination with something like `defaultColDef={"flex": 1, "minWidth": 200},`. This is really an edge case where the user can expect that the column size is auto-sized but that the width is never lower than 200. Also, if `columnSize="autoSize"` is set in the combination with `persisted_props=["columnSize"]`, "autoSize" doesn't work at all. This seems strange to me since, `columnSize` and `filerModel` should behave same since they are both dag.AgGrid properties.

@petar-qb If I understand it correctly, you are essentially referring to user created `AgGrid` inconsistencies right? My point above was that if a user configures "contradicting" options, then it is ok that it fails, **as long as hidden defaults are not the cause of it**.

If this is what you also meant, then let's go with the proposed solution indeed. Any user question we can refer to the Dash AG Grid, because we are not altering that implementation in any way and thus the answer can only lie there.

petar-qb:
> If I understand it correctly, you are essentially referring to user created `AgGrid` inconsistencies right? My point above was that if a user configures "contradicting" options, then it is ok that it fails, **as long as hidden defaults are not the cause of it**.
> 
> If this is what you also meant, then let's go with the proposed solution indeed.

I totally agree, and as I said: **I would go with this solution (hence the approval from my side)** 😄

vizro-core/src/vizro/models/_components/ag_grid.py
maxschulz-COL: This is already done in a previous PR 💪

antonymilne (approved):

PR #455, opened 2024-05-06
[Tidy] Store `page.id` in outer page container to enable page specific styling

huong-li-nguyen:
## Description

- Store `page.id` in outer page container to enable page specific stylings (e.g. this now enables hiding the page-title selectively on a given page)
- Change `page-container` to be a className instead of an ID
- Add quick CSS fix that aligns table and graph titles when next to each other (see dev example)
Screenshot 2024-05-06 at 14 07 02

## Notice


vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -20,3 +20,25 @@ def _validate_min_length(cls, field):
     if not field:
         raise ValueError("Ensure this value has at least 1 item.")
     return field
+
+
+# Based on how Github generates anchor links - see:
+# https://stackoverflow.com/questions/72536973/how-are-github-markdown-anchor-links-constructed.
+def _clean_path(path: str, allowed_characters: str, enable_url: bool = True) -> str:
```
huong-li-nguyen, line 27: This function had to be refactored for the sake of creating valid CSS IDs e.g. a valid CSS ID cannot start with` / `

vizro-core/tests/unit/vizro/models/test_models_utils.py
```diff
@@ -21,3 +22,37 @@ def test_check_for_valid_component_types(self, model_with_layout):
...
+        ("Page Title!", "page-title"),
+        ("Another123Title", "another123title"),
+        ("Some_Example%Title", "some_exampletitle"),
+        ("!@#$%^&*()_+=", "_"),
+        ("", ""),
+    ],
+)
+def test_clean_path_with_url_disabled(test_string, expected):
+    cleaned_string = _clean_path(path=test_string, allowed_characters="-_/", enable_url=False)
+    assert cleaned_string == expected
```
yury-fedotov, line 58: Was opening an issue just now and randomly came across this :) You can do multiple `parametrize` on the same test, which is like a nested for loop, so those two tests can be joined into one if you like.

```suggestion
@pytest.mark.parametrize("enable_url", [True, False])
@pytest.mark.parametrize(
    "test_string, expected",
    [
        ("Title", "title"),
        ("Page Title", "page-title"),
        ("Page Title!", "page-title"),
        ("Another123Title", "another123title"),
        ("Some_Example%Title", "some_exampletitle"),
        ("!@#$%^&*()_+=", "_"),
        ("", ""),
    ],
)
def test_clean_path_with_url_enabled(test_string, expected, enable_url):
    cleaned_string = _clean_path(path=test_string, allowed_characters="-_/", enable_url=enable_url)
    expected_output = f"/{expected}" if enable_url else expected
    assert cleaned_string == expected_output
```

huong-li-nguyen: Haha, amazing! 🚀 I actually parameterised all four arguments before the way you proposed it above, but I thought it wasn't easy to read. I like your suggestion above though, as it is still kept very simple! Thanks!!

yury-fedotov: Welcome. Also might be helpful to have type hints in test functions signatures too (that would show that last param is `bool` while two others are `str`, which is probably what you meant by "difficult to read").

But I didn't add it in my example to keep it isolated to the double parametrization only.

huong-li-nguyen: That's a great suggestion 👍 We do the type hints on the actual functions we test so we have it there, but I don't think we do it on the test function signatures yet. 

I had this before and then a long list of use cases 😄 In the end I thought, it became more difficult to read and understand because in the end `enable_url=True` just adds a `/` at the beginning, but that didn't become immediately clear from screening through all the use cases. Didn't think about just adding a simple if/else - haha! Thanks for the suggestion and your first contribution! 🚀 

```
@pytest.mark.parametrize(
    "test_string, allowed_characters, enable_url, expected",
    [
        ("Title", "-_/", True, "/title"),
        ("Title", "-_/", False, "title"),
        ...
    ],
)
def test_clean_path_with_url_enabled(test_string, allowed_characters, enable_url, expected):
    cleaned_string = _clean_path(path=test_string, allowed_characters=allowed_characters, enable_url=enable_url)
    assert cleaned_string == expected_output
```

nadijagraca (approved):
LGTM. ⭐

vizro-core/docs/pages/user-guides/assets.md
```diff
@@ -87,6 +87,71 @@ For reference, see the [Vizro CSS files](https://github.com/mckinsey/vizro/tree/
...
+              type: card
+          title: Page with shown title
+        ```
+    === "Result"
+         [![PageTitle]][PageTitle]
+
+    [PageTitle]: ../../assets/user_guides/assets/css_page_title.png
+
 
 ## Overwrite CSS properties in selective components
```
antonymilne, line 156: Is this section still correct? e.g. it talks a lot about `_outer`, and I'm not sure whether that's still used much after your recent bootstrap refactorings?

vizro-core/docs/pages/user-guides/assets.md
```diff
@@ -87,6 +87,71 @@ For reference, see the [Vizro CSS files](https://github.com/mckinsey/vizro/tree/
     [AssetsCSS]: ../../assets/user_guides/assets/css_change.png
 
 
+## Overwrite CSS properties for specific pages
+Each page is assigned a unique identifier (`page.id`), which is sanitized and then stored as the ID of an outer `<div>` element.
+This identifier is formatted according to CSS naming conventions by removing non-alphanumeric characters and replacing spaces with dashes.
+For instance, if a `page.id` is "Variable Analysis," it will be stored as a CSS ID of "variable-analysis."
```
antonymilne, line 93:
```suggestion
For instance, if a `page.id` is `Variable Analysis` then it has CSS ID `variable-analysis`.
```

As it stands the "," and "." are very confusing here because it looks like you're replacing "," with ".".

Not sure whether we should convert to lowercase, but let's discuss 🤔

huong-li-nguyen: Oh, I don't even know how this got in - probably my automatic grammar checker😆 There is not supposed to be a comma 😄

vizro-core/docs/pages/user-guides/assets.md
huong-li-nguyen: This still holds, but it gets a bit confusing now because it does not hold for all components. Let's discuss in our call, but just to show you what I mean:

Here we don't have the `_outer` suffix for an ID, because we can provide the self.id directly to the component that gets outputted:

`dbc.Button(id=self.id, children=self.text)`

However, here we still need the `_outer` suffix because of two reasons:

1. We once aligned that the `self.id` is being provided to the inner component that might be reactive to callbacks. In the example of the card it is the `dcc.Markdown` component. Because of that we cannot provide the `self.id` to the outermost component. 
2. Let's say someone wants to change the card background-color but only for a selective card and not for all. Here we need to provide an `id` to the outermost wrapper of the card so we can selectively target all inner components. And because we can't use `self.id`, we started using `self.id_outer`.

Happy to change this current system though, just not sure what's the best approach here 🤔 

`dbc.Card(card_content, className=card_class, id=f"{self.id}_outer")`

antonymilne: Yeah, that is what I expected but I didn't know how prevalent the `_outer` suffix was any more.

I think the current system basically makes sense (using `self.id` for the "core" reactive component rather than the outer container) but we should _consider_ (but maybe not do...) removing the `_outer` ones if they're just there for CSS selectors [in favour of using `has()`](https://stackoverflow.com/questions/1014861/is-there-a-css-parent-selector).

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -225,7 +230,7 @@ def _arrange_page_divs(self, page_divs: _PageDivsType):
 
         page_header = html.Div(page_header_divs, id="page-header", hidden=_all_hidden(page_header_divs))
         page_main = html.Div([collapsable_left_side, collapsable_icon, right_side], id="page-main")
-        return html.Div([page_header, page_main], id="page-container")
+        return html.Div([page_header, page_main], id=page_divs["page-id"].children, className="page-container")
```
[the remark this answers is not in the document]
huong-li-nguyen: Ah yes! Implemented here: https://github.com/mckinsey/vizro/pull/455/commits/ba9966e050ad1070f9e99dd3f0c0a50da6ded65f

vizro-core/docs/pages/user-guides/assets.md
huong-li-nguyen: Yes, I like that! Let's discuss in our call later. But I would also be in favour of killing the entire _outer system and just updating the docs 👍

vizro-core/src/vizro/models/_models_utils.py
huong-li-nguyen: Just for my own tracking: We've decided to **not clean** the path, but instead add to the docs that the user has to provide a valid `id`. I'll add to the docs what we consider as `valid`:
- Use lowercase letters and hyphens
- Avoid using special characters or spaces
- Provide a unique `id`

vizro-core/docs/pages/user-guides/assets.md
```diff
@@ -87,6 +87,70 @@ For reference, see the [Vizro CSS files](https://github.com/mckinsey/vizro/tree/
...
 
 
+## Overwrite CSS properties for specific pages
+If you want to style specific pages only, you need to provide a valid `id` to the page you want to modify.
+This unique `id` enables you to use CSS selectors to target specific pages and their components for styling.
+
+An HTML `id` is considered valid if the following conditions are met:
+- Use lowercase letters and hyphens
+- Avoid using special characters or spaces
+- Provide a unique `id`
```
antonymilne, line 97:
```suggestion
To style components for only a specific page, use the page's `id` in CSS selectors. By default, this is the [same as the page `title`](pages.md), but such a value might not be a valid CSS identifier. A suitable `id` must be unique across all models in the dashboard and should contain only alphanumeric characters, hyphens (`-`) and underscores (`_`). In particular, note that spaces are _not_ allowed.
```

This guidance is general about `id` really applies everywhere of course, not just to page's. Feel free to move it now or in the future when we restructure the page as discussed earlier.

The only special part that applies specifically to page is just that the `id` is automatically set and hence more likely to be inappropriate for CSS identifiers due to having a space in it.

vizro-core/docs/pages/user-guides/assets.md
```diff
@@ -87,6 +87,70 @@ For reference, see the [Vizro CSS files](https://github.com/mckinsey/vizro/tree/
...
+
+Suppose you want to hide the page title on one page only. Here's how you can achieve this:
+1. Provide a valid `id` to the `Page`, for example `Page(id="page-with-hidden-title", title="Page with hidden title", ...)`
+2. Identify the CSS class or ID you need to target. To hide the page title, you need to hide the container with the ID `right-header`.
+3. Use the `id` in combination with CSS selectors to change the relevant CSS properties
+4. Add your custom css file to the `assets` folder as explained above
```
antonymilne, line 103:
```suggestion
1. Provide a valid `id` to the `Page`, for example `Page(id="page-with-hidden-title", title="Page with hidden title", ...)`.
2. Identify the CSS class or ID you need to target. To hide the page title, you need to hide the container with the ID `right-header`.
3. Use the `id` in combination with CSS selectors to change the relevant CSS properties.
4. Add your custom css file to the `assets` folder as explained above.
```

antonymilne (approved):
Thank you for the changes, looks good!

huong-li-nguyen:
@stichbury - do you want to take a final look at the docs changes? :) I'll merge after your approval 👍

vizro-core/docs/pages/user-guides/assets.md
```diff
@@ -87,6 +87,64 @@ For reference, see the [Vizro CSS files](https://github.com/mckinsey/vizro/tree/
     [AssetsCSS]: ../../assets/user_guides/assets/css_change.png
 
 
+## Overwrite CSS properties for specific pages
+To style components for only a specific page, use the page's `id` in CSS selectors. By default, this is the [same as the page `title`](pages.md), but such a value might not be a valid CSS identifier. A suitable `id` must be unique across all models in the dashboard and should contain only alphanumeric characters, hyphens (`-`) and underscores (`_`). In particular, note that spaces are _not_ allowed.
```
stichbury, line 91:
```suggestion
To style components for a specific page, use the page's `id` in CSS selectors. By default, this is the [same as the page `title`](pages.md), but such a value might not be a valid CSS identifier. A suitable `id` must be unique across all models in the dashboard and should contain only alphanumeric characters, hyphens (`-`) and underscores (`_`). In particular, note that spaces are _not_ allowed.
```

stichbury (approved):
LGTM, thank you 🙏

PR #461, opened 2024-05-07
[Docs] Make dynamic data example more visual

huong-li-nguyen:
## Description
I was initially thinking about replacing the entire example with an API data connection that updates every second. For an update every second, the stock market data seems to be the most suitable one: https://www.alphavantage.co/documentation/

However, this would require the user to create a free API key first to be able to run our examples as we cannot publish our free API key. The stock market data can update every second, but the visualization changes might be minimal, so I am not sure if this will come across in a GIF.

We might want to come back to this when we create a more sophisticated demo dashboard with API connections. I think there it will make more sense than for this small demo in the docs. So for now I would just make the current dynamic data example **more visual**, as the changes in the dots are a bit difficult to see and it's not immediately clear that the update is caused by a page refresh.

**Changes include:**

- Removed the `id` from the Graph where it was not required
- Removed the `controls` from the page as they were not used in the GIF and I think they didn't add anything to the page? 
- Replaced the scatter with a more visual violin chart
- Included the browser inside the GIF to make it more clear that the graph updates on page refresh

## Notice


nadijagraca (approved):
LGTM. ⭐  I agree with removing `controls`, as you said it didn't add anything to the example.

antonymilne (approved):
Definitely an improvement on the current docs, although I agree that in future hopefully we can come up with an even better example. Once we have parameterised data loading with https://github.com/McK-Internal/vizro-internal/issues/753 then we'll eventually need a good example for that also, and the iris example might not be enough then. Did you also considering using the weather forecast that you mentioned before?

> Removed the controls from the page as they were not used in the GIF and I think they didn't add anything to the page?

Agree, although worth noting just for your own interest that the filters work nicely on page refresh (i.e. the filter option is maintained across data reloading).

> Replaced the scatter with a more visual violin chart

Much as I <3 violin chart, I suspect that a box plot would be more suitable here since since once you've categorised by species some of the distributions don't have many points so don't make much sense as continuous distributions. Like this:
image

huong-li-nguyen:
> Much as I ❤️ violin chart, I suspect that a box plot would be more suitable here since since once you've categorised by species some of the distributions don't have many points so don't make much sense as continuous distributions. Like this: !

Ah yes, I see what you mean! I'll replace it with a boxplot.

To your other question - I did consider the weather forecast first, but there users would also have to set up a free API key first and it does not update every second, but maybe every hour. The weather forecast doesn't change significantly, so for something to become visually clear in a GIF you need something that is volatile and updates per second or minute at least, so we can record the GIF and then just speed it up 😄 I was also thinking about combining it with some filters e.g. filter per hour/year/month and then show the visual changes, but I think it then mixes up the dynamic data loading + filtering. 

So in the end, I still liked your example (picking random data points) the best in terms of showing what dynamic data does as an introduction. But i agree that we need to create more sophisticated demo dashboards, but there the visual change per second will also not be the main goal as it is here.

PR #477, opened 2024-05-13
[Docs] Two small docs changes for improved usability

stichbury:
## Description
Following discussion with @antonymilne 

* Add a link to the user examples from the main examples page, in case people don't spot the user examples page (which requires them to understand how the left hand nav works). By adding a section with a title, it brings it into the RHS table of contents, so more obvious.
* Add a mention of Plotly community forums. Have I got the right category @antonymilne ?
    * TBH that whole contribution guide needs a solid revision and I have this on my list, just not yet, so if it makes sense to do so, I can park this change for now and add it in later. Alternatively, we could put something about these forums behind a "Got a question?" link on the front page of the docs, and point straight through to the Plotly forums there, so it's really obvious. LMK if you'd like to try that and see how many queries get raised. 

## Notice

### DIFF
--- a/vizro-core/docs/pages/examples/examples.md
+++ b/vizro-core/docs/pages/examples/examples.md
@@ -23,3 +23,7 @@ The [`examples/features` folder](https://github.com/mckinsey/vizro/tree/main/viz
 !!! info "Ways to configure a dashboard"
 
     In most cases, the pydantic model should be your preferred method of dashboard configuration, but you can also [define a dashboard with YAML, JSON, or as a Python dictionary](../user-guides/dashboard.md).
+
+## Examples from Vizro users
+
+We maintain a separate, curated page of [videos, blog posts, and examples of Vizro usage from our community](./your-examples.md).
--- a/vizro-core/docs/pages/examples/examples.md
+++ b/vizro-core/docs/pages/examples/examples.md
@@ -29,4 +29,3 @@ The [`examples/features` folder](https://github.com/mckinsey/vizro/tree/main/viz
 ## Examples from Vizro users
 
 We maintain a separate, curated page of [videos, blog posts, and examples of Vizro usage from our community](your-examples.md).
-
--- a/vizro-core/docs/pages/examples/examples.md
+++ b/vizro-core/docs/pages/examples/examples.md
@@ -1,4 +1,6 @@
-# Vizro examples
+# Example dashboard code
+
+This page explains the dashboard code examples found in the `examples` folder of the [Vizro GitHub repository](https://github.com/mckinsey/vizro/tree/main/vizro-core/examples).
 
 ## Vizro live demo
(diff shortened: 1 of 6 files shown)


petar-qb (approved):

vizro-core/docs/pages/examples/examples.md
```diff
@@ -23,3 +23,7 @@ The [`examples/features` folder](https://github.com/mckinsey/vizro/tree/main/viz
 !!! info "Ways to configure a dashboard"
 
     In most cases, the pydantic model should be your preferred method of dashboard configuration, but you can also [define a dashboard with YAML, JSON, or as a Python dictionary](../user-guides/dashboard.md).
+
+## Examples from Vizro users
+
+We maintain a separate, curated page of [videos, blog posts, and examples of Vizro usage from our community](./your-examples.md).
```
antonymilne, line 29:
```suggestion
We maintain a separate, curated page of [videos, blog posts, and examples of Vizro usage from our community](your-examples.md).
```

Very minor, has no real impact on anything, but just makes things a bit simpler. It seems like we always do link this way apart from two remaining `./` style links in `first-dashboard.md`, so if you can be bothered then please can you make the change there too?

antonymilne:
Thanks for following up on these things! Definitely a good improvement.

I still find the repetition of "Examples" in the left-hand menu a bit jarring, especially since it doesn't match the title of the page (unlike the "Examples from vizro users" page, which does). So if you can figure out a nicer way to do that it would be great 😀 
image
Maybe we should even just put this all in one page for now instead of splitting it up? Up to you.

> Add a mention of Plotly community forums. Have I got the right category @antonymilne ?
Yes, looks good!

> TBH that whole contribution guide needs a solid revision and I have this on my list, just not yet, so if it makes sense to do so, I can park this change for now and add it in later. Alternatively, we could put something about these forums behind a "Got a question?" link on the front page of the docs, and point straight through to the Plotly forums there, so it's really obvious. LMK if you'd like to try that and see how many queries get raised.

I like this idea! The link is definitely quite buried at the moment so would be interesting to see if there's more posts there if we put it on the front page. Or we could have a shortcut from our docs front page to the section of the contribution guidelines which suggest asking a general question on Github issues or the plotly forums. Like you say, the contribution guidelines in general are in need of a revision, and probably people should be able to find where they can ask questions without considering it a "contribution". Very happy to leave perfecting this for future changes, and for now it's useful to just have a quick and easy change to make more obvious what our two support routes are (Github issues, plotly forums) however you think is best to do so.

antonymilne (approved):
Feel free to merge whenever you're happy with it! No need to check again 🙂 My suggestions are just minor stuff and I leave it up to you to decide how/when/whether to make more changes here.

stichbury:
> I still find the repetition of "Examples" in the left-hand menu a bit jarring, especially since it doesn't match the title of the page (unlike the "Examples from vizro users" page, which does). So if you can figure out a nicer way to do that it would be great 😀

That's a really good point. I've made a slight change to the title and the table of contents listing, so now it looks as follows:

image

It's probably still a bit repetitive but I've retained the intent to make it obvious that there are two separate pages, and I think we should keep to that because one will become an "awesome" list at some point (and ideally put into a separate repo or hosted in GitHub rather than docs) while the other page is more a description of what we provide in the repo or as official examples.

PR #479, opened 2024-05-14
[Feat] Make dynamic data loading functions accept arguments

antonymilne:
## Description

This is step 1 out of 2 in completing https://github.com/McK-Internal/vizro-internal/issues/753/ to allow for parametrisation of dynamic data from the frontend.

Here I just implement the changes needed to the data manager. I anticipated this would be as simple as adding `*args, **kwargs` in a couple of places, but it turned out to be much more complicated than that. In short:
* for memoization to work fully (so with `def f(a=1): ...`, we need `f(a=1)` and `f(1)` and `f()` to all have same cache key) flask-caching relies on inspecting the signature of the function being memoized
* previously we memoized the `_DynamicData.load` bound method, but it is basically impossible for this to be done in a way that mirrors the signature of the underlying `load_data` function (details in https://github.com/GrahamDumpleton/wrapt/issues/263)
* hence I've moved the memoization inside the `load` function itself
* this means that the memoized function could be either a function or a bound method (e.g. in the important case of kedro datasets)
* handling bound methods with flask-caching is tricky because it's difficult to set independent timeouts without lots of hacking
* eventually I found the cleanest way to achieve this was to handle everything as a `partial` function named with the data source name. This ensures that both bound methods and functions are handled correctly, with caches keyed by data source name
* small consequence of this: using `partial` functions now works so I've removed the error raised in `DataManager.__setitem__`

Also I've written many more tests to check this all works. These are a bit slow because they run for many different functions and include `time.sleep`. If they become annoying to run in CI we can tag them as slow and split them out or skip them or something.

## To do in this PR
- [x] @petar-qb to do manual test with redis 🙏  **No need to test anything that's new functionality involving passing arguments, just to check I didn't break anything that worked before.** I suggest you pick a few representative cases from https://github.com/mckinsey/vizro/pull/398#issuecomment-2049540014 and see what the cache keys look like since they'll be a bit different from before. Then please say if anything looks amiss

## To do in following PR #482 
- [ ] @petar-qb to add step 2 of 2 and tests in completing https://github.com/McK-Internal/vizro-internal/issues/753/ 
- [ ] @antonymilne and @petar-qb to write docs
- [ ] @petar-qb to do manual test with redis to check new functionality **including passing arguments**. Since you'll be familiar with what they look like now after testing this current PR, hopefully it'll be easy to check it all makes sense when we test in the following PR
- [ ] add to changelog

## To do separately

- [x] Make new ticket to tidy `log_call`

## Notice


vizro-core/tests/unit/vizro/integrations/kedro/fixtures/test_catalog.yaml
```diff
@@ -3,5 +3,5 @@ companies:
   filepath: companies.json
 
 reviews:
-  type: pickle.PickleDataset
+  type: pandas.JSONDataset
```
antonymilne, line 6:
```suggestion
  type: pickle.PickleDataset
```

vizro-core/src/vizro/managers/_data_manager.py
```diff
@@ -3,74 +3,50 @@
...
+from typing import Callable, Dict, Optional, Union
 
 import pandas as pd
 import wrapt
 from flask_caching import Cache
 
 from vizro.managers._managers_utils import _state_modifier
 
-logger = logging.getLogger(__name__)
+
```
antonymilne, line 17: @petar-qb as with https://github.com/mckinsey/vizro/pull/479 I suggest you run an example app _without_ the following environment variables that are set in our hatch default environment:
```
DASH_DEBUG = "true"
VIZRO_LOG_LEVEL = "DEBUG"
```

That will make the logs clearer and easier to understand.

vizro-core/src/vizro/managers/_data_manager.py
```diff
@@ -3,74 +3,50 @@
...
-    __func__: Callable[..., Any]
-
-    def __call__(self, *args, **kwargs): ...
-
-
-# wrapt.decorator is the cleanest way to decorate a bound method when instance properties (here instance.timeout)
-# are required as arguments. The enabled argument bypasses the entire decorator when data_manager._cache_has_app
-# is False. This prevents Flask-Caching from trying to access a Flask app when one doesn't yet exist.
-@wrapt.decorator(enabled=lambda: data_manager._cache_has_app)
-def memoize(wrapped: BoundMethod, instance: _DynamicData, args, kwargs):
```
antonymilne, line 40: This is now done inside `_DynamicData.load_data`.

vizro-core/src/vizro/managers/_data_manager.py
```diff
@@ -87,47 +63,39 @@ class _DynamicData:
...
-        different so the cache will not match up.
-
-        flask_caching make it possible to set a __cached_id__ attribute to handle this so that repr can be set
-        independently of cache key, but this doesn't seem to be well documented or work well, so it's better to rely
-        on __repr__.
-        """
-        # Note that using repr(self.__load_data) is not good since it depends on the id of self.__load_data and so
-        # would not be consistent across processes.
-        return f"{self.__class__.__name__}({self._name!r}, {self.__load_data.__qualname__})"
+    def load(self, *args, **kwargs) -> pd.DataFrame:
```
antonymilne, line 77: This line is basically the key change; the other stuff is just needed to get flask-caching working properly.

petar-qb (approved):
After thorough code (and tests) inspection and manual testing with a few the most representative combinations from https://github.com/mckinsey/vizro/pull/398#issuecomment-2049540014, I didn't spot anything that amiss. As expected 😄 

Also, many thanks for all your code/tests and PR comments. They made my life easier.

vizro-core/src/vizro/managers/_data_manager.py
```diff
@@ -197,18 +162,35 @@ def __init__(self):
...
+            # running with --preload.
+            # __module__ is also required in flask_caching.utils.function_namespace and not defined for partial
+            # functions in some versions of Python.
+            # update_wrapper ensures that __module__, __name__, __qualname__, __annotations__ and __doc__ are
+            # assigned to the new partial(data) the same as they were in data. This isn't strictly necessary but makes
+            # inspecting these functions easier.
+            data = functools.update_wrapper(partial(data), data)
+            data.__module__ = getattr(data, "__module__", "<nomodule>")
+            data.__name__ = ".".join([getattr(data, "__name__", "<unnamed>"), name])
+            data.__qualname__ = ".".join([getattr(data, "__qualname__", "<unnamed>"), name])
```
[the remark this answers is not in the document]
antonymilne: Amazing, I'm pleased to hear it 👍 The old cache keys were definitely a bit weird.

vizro-core/src/vizro/managers/_data_manager.py
```diff
@@ -197,18 +162,35 @@ def __init__(self):
...
+            self.__data[name] = _DynamicData(data)
```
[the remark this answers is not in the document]
antonymilne: I tried this out in https://github.com/mckinsey/vizro/pull/479/commits/191868fcda33a36a388dd2bee2e1ea75bef3afa6 and it almost works but not quite.

Rewriting the function signature is not too bad since it's very similar to [this wrapt example](https://wrapt.readthedocs.io/en/latest/decorators.html#signature-changing-decorators). The biggest catch is actually associated with a note I added to one of the tests:
> When the cache for one set of memoized arguments expires, the cache for the whole data source expires, even for other values of memoized arguments.

This means that `test_independent_dynamic_data_callable_with_timeout` starts failing, because if you have the following then the `timeout` setting applies to both `data_x` and `data_y`.
```
data_manager["data_x"] = RandomData().load
data_manager["data_y"] = RandomData().load
data_manager["data_y"].timeout = 1
```

This is a big problem when it comes to kedro datasets which all work according to this pattern.

This means that for the case of bound methods we still need to pass in functions rather than bound methods, and those functions still need to be labelled according to the underlying data source name. It's also a problem for `lambda` and `partial` functions. We also still need to assign a `__qualname__` anyway in order to be able to handle `partial` function at alls, which is [something we probably do want to be able to do](https://github.com/mckinsey/vizro/pull/451#issuecomment-2111138606). So most of the ugly `__qualname__` stuff is still needed anyway.

Hence I've reverted trying this out. But thank you for the excellent comment, it was very useful to investigate!

Joseph-Perkins (approved):

PR #484, opened 2024-05-15
[CI] Github tests trigger for CircleCI

l0uden:
## Description

Here we'll get rid of CircleCI pipeline for vizro repo.
New GitHub Action will trigger tests pipeline instead.

`test-integration-vizro-ai` and `CircleCI tests trigger` GitHub Actions will always pass for the forked repos.

`test-integration-vizro-ai` will run for every push to `vizro-ai` folder except `docs`. And will run for every merge to `main` for changes in any folder.

Connected with https://github.com/mckinsey/vizro-qa/pull/89

## Notice


.github/workflows/circleci-trigger.yml
```diff
@@ -0,0 +1,89 @@
...
+    name: CircleCI tests trigger
+    environment: circleci_secrets
+    runs-on: ubuntu-latest
+    steps:
+      - name: Ping CircleCI pipeline
+        run: |
+          if [[ "${{ github.ref_type }}" == "tag" ]]; then
+            github_tag=${{ github.ref_name }}
+          else github_tag=""
+          fi
```
antonymilne, line 25: Do we need this? Do we ever trigger this job from a tag rather than a branch?

.github/workflows/circleci-trigger.yml
```diff
@@ -0,0 +1,89 @@
...
+            --data '{"branch":"${{ github.head_ref }}", "parameters":{"branch":"${{ github.head_ref }}", "vizro_branch": "${{ github.head_ref }}", "tag": "$github_tag"}}' \
+          | jq -r '.id'
+          )
+          if [[ "$CREATED_PIPELINE" == "null" ]]; then
+            CREATED_PIPELINE=$(curl --request POST \
+            --url "${{ secrets.QA_PIPELINE_URL }}" \
+            --header "Circle-Token: ${{ secrets.CIRCLECI_API_KEY }}" \
+            --header "content-type: application/json" \
+            --data '{"branch":"main", "parameters":{"branch":"main", "vizro_branch": "${{ github.head_ref }}", "tag": "$github_tag"}}' \
+            | jq -r '.id'
```
antonymilne, line 40: What's the different between these two? One is triggered on merge to `main` and one is in the PR itself?

.github/workflows/circleci-trigger.yml
```diff
@@ -0,0 +1,89 @@
...
+            do
+              sleep 15
+              created_workflow_status_updated=$(curl --request GET \
+                  --url "https://circleci.com/api/v2/pipeline/$CREATED_PIPELINE/workflow" \
+                  --header "Circle-Token: ${{ secrets.CIRCLECI_API_KEY }}" \
+                  --header "content-type: application/json" \
+                | jq -r '.items[0].status'
+              )
+              echo $created_workflow_status_updated
+            done
```
antonymilne, line 74: Is it possible to refactor this somehow to be less repetitive? Currently it looks a lot like:

```
created_workflow_status = do_something()

if [[ "$created_workflow_status" == "running" ]]; then
  created_workflow_status_updated = do_something()
  while [[ "$created_workflow_status_updated" == "running" ]]
  do
    sleep 15
    created_workflow_status_update = do_something()
  done
```

Can't we just do as something like this? As rewritten quickly by ChatGPT plus me.

```
get_workflow_status() {
  curl --silent --request GET \
       --url "https://circleci.com/api/v2/pipeline/$CREATED_PIPELINE/workflow" \
       --header "Circle-Token: ${{ secrets.CIRCLECI_API_KEY }}" \
       --header "content-type: application/json" \
    | jq -r '.items[0].status'
}

while created_workflow_status=$(get_workflow_status); [[ "$created_workflow_status" == "running" ]]; do
  echo $created_workflow_status
  sleep 15
done
```

antonymilne (commented):
A few initial questions for now, but we might need to schedule some time together to discuss because I'm a bit confused a bit the moment 😅

.github/workflows/circleci-trigger.yml
l0uden: Yes, we have release tests for the new tags.

antonymilne: Ah ok. That leads to more questions though, sorry... 😅 
1. How would these get triggered? I see there's a step `Tag main and create release on github` that creates the tag in our release workflow, but then there's nothing in the `on:` section of this file that will respond to it?
2. What do we actually want to run before a release? It seems a bit random to run just the QA tests but nothing else? I think we should either:
a. decide we want to test everything before release so add triggers into other GHA jobs
b. decide we don't want to test anything before release

Option b is fine for me since all the checks will run in the PR that merges to main anyway. But If we want to do option a then probably the best way is to modify our release job so that it runs all required jobs as `needs` before doing the release steps: https://stackoverflow.com/a/71489231.

.github/workflows/circleci-trigger.yml
antonymilne: ok, makes sense. Please could you add some comments to explain this? And maybe refactor also to make it clearer what's happening. The JSON payload is also pretty hard to read here - since we use `jq` to parse JSON already, maybe we can also use it to generate the JSON? Something like this, again mainly written by ChatGPT...

```
create_circleci_pipeline() {
  local branch=$1
  local vizro_branch=$2
  local tag=$3

  local json_data=$(jq -n --arg branch "$branch" --arg vizro_branch "$vizro_branch" --arg tag "$tag" '{branch: $branch, parameters: {branch: $branch, vizro_branch: $vizro_branch, tag: $tag}}')

  curl --request POST \
    --url "${{ secrets.QA_PIPELINE_URL }}" \
    --header "Circle-Token: ${{ secrets.CIRCLECI_API_KEY }}" \
    --header "content-type: application/json" \
    --data "$json_data" \
  | jq -r '.id'
}

CREATED_PIPELINE=$(create_circleci_pipeline "${{ github.head_ref }}" "${{ github.head_ref }}" "$github_tag")

if [[ "$CREATED_PIPELINE" == "null" ]]; then
  CREATED_PIPELINE=$(create_circleci_pipeline "main" "${{ github.head_ref }}" "$github_tag")
fi
```

l0uden: Nice suggestion! Thanks! Did it

.github/workflows/circleci-trigger.yml
l0uden: Done with it. Nice one!

.github/workflows/circleci-trigger.yml
antonymilne: For the record, we are removing this check. It doesn't seem to test anything useful, especially given how the `hatch build` process works so that all files in `src/vizro` that are on git are always included.

antonymilne (approved):
Looks good, thanks for all your work on this @l0uden.

I made a few small commits just to make it more consistent with how we do other GHA on this repo:

1. [Small renaming](https://github.com/mckinsey/vizro/pull/484/commits/7b6434bd4abcabb7090c59f47a26b1f2b4e18147)
2. [Remove remaining references to tag and unnecessary arguments](https://github.com/mckinsey/vizro/pull/484/commits/4c2c9fe29bf753d8c661c882312e3598e2946bff) - this one is the only bit that changes functionality. I think when you removed the `tag` stuff you forgot to remove some parts, so please check this
3. [Tiny refactor](https://github.com/mckinsey/vizro/pull/484/commits/3fba3e7a94eac1a85de01246be393188c761efdf)

maxschulz-COL (approved):

PR #487, opened 2024-05-16
[Feat] Add new dcc.Loading features in dash 2.17

### LINKED ISSUE #486: New features available for dcc.Loading
### Which package?

vizro

### What's the problem this feature will solve?

Hi Vizro team :wave: 

I just wanted to let you know about some cool new features available for `dcc.Loading` in dash 2.17.0 

It's  now possible to:
 - Delay showing the spinner for a specified number of milliseconds - which eliminates the  flickering when the loading time is very short.  
 - Make the background visible and  customizable with CSS -- like adding `blur` or `opacity`.  This makes the transition between loading and ready a lot smoother.
 - Use your own custom spinner rather than the built-in spinners provided.

You can find the new features in the [dcc.Loading](https://dash.plotly.com/dash-core-components/loading) section of the dash docs.

I'll do a draft PR to apply a few new settings, and you can check it out by running the demo  app.   If you are interested in adding this,  I'd be happy to help fine-tune it.

### Describe the solution you'd like

As described above

### Alternative Solutions

None

### Additional context

None

### Code of Conduct

- [X] I agree to follow the  [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

AnnMarieW:
## Description

Closes #486 

This PR is a quick demo of a couple new features in dcc.Loading available in dash 2.17. This is just to make it easier to see how it it might work in Vizro apps.

It adds 
- `delay_show=500` to reduce the flickering on short loading times
- `overlay_style={"visibility": "visible", "filter": "blur(2px)"}` as an example of styling the  background.

I'm interested to hear what you think!

## Notice

### DIFF
--- a/vizro-core/changelog.d/20240523_085359_amward_update_dcc_loading.md
+++ b/vizro-core/changelog.d/20240523_085359_amward_update_dcc_loading.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Made components visible with added opacity while loading.  Updated the minimum Dash version requirement to 2.17.0 to incorporate these new dcc.Loading features.  ([#487](https://github.com/mckinsey/vizro/pull/487))
+
+-->
+
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Security
+
+- A bullet item for the Security category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
--- a/vizro-core/changelog.d/20240523_085359_amward_update_dcc_loading.md
+++ b/vizro-core/changelog.d/20240523_085359_amward_update_dcc_loading.md
@@ -22,19 +22,18 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 - A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
 
 -->
-
+<!--
 ### Changed
 
-- Made components visible with added opacity while loading. Updated the minimum Dash version requirement to 2.17.0 to incorporate these new dcc.Loading features. ([#487](https://github.com/mckinsey/vizro/pull/487))
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
 
 -->
-
+<!--
 ### Deprecated
 
 - A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
 
 -->
-
 <!--
 ### Fixed
 
--- a/vizro-core/changelog.d/20240523_085359_amward_update_dcc_loading.md
+++ b/vizro-core/changelog.d/20240523_085359_amward_update_dcc_loading.md
@@ -25,7 +25,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 ### Changed
 
-- Made components visible with added opacity while loading.  Updated the minimum Dash version requirement to 2.17.0 to incorporate these new dcc.Loading features.  ([#487](https://github.com/mckinsey/vizro/pull/487))
+- Made components visible with added opacity while loading. Updated the minimum Dash version requirement to 2.17.0 to incorporate these new dcc.Loading features. ([#487](https://github.com/mckinsey/vizro/pull/487))
 
 -->
 
@@ -34,6 +34,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 - A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
 
 -->
+
 <!--
 ### Fixed
 
--- a/vizro-core/src/vizro/models/_components/ag_grid.py
+++ b/vizro-core/src/vizro/models/_components/ag_grid.py
@@ -118,5 +118,5 @@ class AgGrid(VizroBaseModel):
             color="grey",
             parent_className="loading-container",
             delay_show=500,
-            overlay_style={"visibility": "visible", "filter": "blur(2px)"}
+            overlay_style={"visibility": "visible", "filter": "blur(2px)"},
         )
--- a/vizro-core/src/vizro/models/_components/ag_grid.py
+++ b/vizro-core/src/vizro/models/_components/ag_grid.py
@@ -117,7 +117,5 @@ class AgGrid(VizroBaseModel):
             ],
             color="grey",
             parent_className="loading-container",
-            delay_show=500,
-            delay_hide=500,
             overlay_style={"visibility": "visible", "opacity": 0.3},
         )
--- a/vizro-core/src/vizro/models/_components/ag_grid.py
+++ b/vizro-core/src/vizro/models/_components/ag_grid.py
@@ -117,4 +117,6 @@ class AgGrid(VizroBaseModel):
             ],
             color="grey",
             parent_className="loading-container",
+            delay_show=500,
+            overlay_style={"visibility": "visible", "filter": "blur(2px)"}
         )
--- a/vizro-core/src/vizro/models/_components/ag_grid.py
+++ b/vizro-core/src/vizro/models/_components/ag_grid.py
@@ -118,5 +118,6 @@ class AgGrid(VizroBaseModel):
             color="grey",
             parent_className="loading-container",
             delay_show=500,
+            delay_hide=500,
             overlay_style={"visibility": "visible", "filter": "blur(2px)"},
         )
--- a/vizro-core/src/vizro/models/_components/ag_grid.py
+++ b/vizro-core/src/vizro/models/_components/ag_grid.py
@@ -119,5 +119,5 @@ class AgGrid(VizroBaseModel):
             parent_className="loading-container",
             delay_show=500,
             delay_hide=500,
-            overlay_style={"visibility": "visible", "filter": "blur(2px)"},
+            overlay_style={"visibility": "visible", "opacity": 0.3},
         )
--- a/vizro-core/src/vizro/models/_components/graph.py
+++ b/vizro-core/src/vizro/models/_components/graph.py
@@ -136,8 +136,6 @@ class Graph(VizroBaseModel):
             ),
             color="grey",
             parent_className="loading-container",
-            delay_show=500,
-            delay_hide=500,
             overlay_style={"visibility": "visible", "opacity": 0.3},
         )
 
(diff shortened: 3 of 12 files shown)


vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -136,6 +136,8 @@ def build(self):
             ),
             color="grey",
             parent_className="loading-container",
+            delay_show=500,
+            overlay_style={"visibility": "visible", "filter": "blur(2px)"},
```
huong-li-nguyen, line 140:
```suggestion
```

The `overlay_style` is such a nice feature! Amazing job @AnnMarieW 🚀 

I would have loved to implement a similar example as in your PR for the loading spinner, providing some opacity on the figure and having the loading spinner overlayed on top. This looks so nice visually! 🎨 

However, I just realized that this is not possible with Vizro, as we always always output an empty figure before the actual figure is loaded. So, the loading spinner is always displayed on top of an empty background. So we currently can't leverage `overlay_style` as I would like to. This might change in the future, in which case I would definitely go for some opacity on the figure 👍 

Screenshot 2024-05-21 at 10 25 07

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -117,4 +117,6 @@ def build(self):
             ],
             color="grey",
             parent_className="loading-container",
+            delay_show=500,
```
huong-li-nguyen, line 120: @AnnMarieW—I have a quick question about the `delay_show`. I've read through your documentation (great examples, by the way! 🚀 ) and your original PR on Dash. From what I understand, it reduces the flickering because the loading spinner is only shown after 500ms. If the loading time is short, the loading spinner might not even appear, did I understand that correctly? So, I guess we need to find the sweet spot in terms of time here.

When running the demo dashboards (`hatch run example demo`) with `delay_show=500`, it seems to flicker a bit more now as the loading spinner is displayed for a shorter time. Are there any other advantages for the `delay` arguments that might not be visually clear from the Vizro examples?

huong-li-nguyen (commented):
First of all, congratulations @AnnMarieW on getting these features into Dash! 🚀 ⭐  From a visual perspective, I think they can improve the loading spinners significantly. 

I left a couple of comments as one big visual advantage of the new loading spinner feature doesn't really come across in Vizro. That's an issue on our side though. I'll double-check with @antonymilne if this might change in the future. 

The other visual advantage of reducing the flickering - I would love to understand better as from the current Vizro demo examples it seems to flicker a bit more, but I think it's because we need to fine-tune the ms for delay_show?

antonymilne:
Hello @AnnMarieW! Thanks very much for making the Github issue and raising this PR. And many congratulations on https://github.com/plotly/dash/pull/2760! It's a really amazing PR ⭐ 

@huong-li-nguyen Following on from the slack conversation and just to make sure we're all on the same page here, because I had to remind myself of some previous work I'd done to stop graphs flickering (https://github.com/mckinsey/vizro/pull/166/) and what our current scheme is... There's currently two times when `dcc.Loading` is relevant:
1. when `on-page-load` action replaces the empty graph (or full datatable/AgGrid as @petar-qb explained) with the "real" one containing filtered data
2. when you change some parameter/filter data on the screen

So even though `overlay_style` might not make any difference to graphs in case 1, it may well do so in case 2 if the filter/parameter takes a long time to apply (we can add some `time.sleep` in to try it out) so I think is probably valuable already.

`delay_show` I think should also improve things already (again, for non-graph in case 1 and for graph/table/grid  components in case 2). So if it's somehow making flickering worse that would be good to understand what's happening.

Just for the record, some other things that might be useful from https://github.com/plotly/dash/pull/2760 which @AnnMarieW didn't change here would be:
1. the `target_components` property, if you can remember why that might have been useful to us (I can't) - see our conversation with Ann Marie on [dcc.Loading PR](https://community.plotly.com/t/dcc-loading-pr/82455) on the plotly forums!
2. `display="show"/"hide"` might help us get rid of `on-page-load` as per https://github.com/mckinsey/vizro/pull/439, but this is a longer term thing

vizro-core/src/vizro/models/_components/graph.py
antonymilne: I think this might already be useful for us actually once the figure has already loaded and you apply a slow filter/parameter 🙂

vizro-core/src/vizro/models/_components/ag_grid.py
huong-li-nguyen: Yes, I agree! I've tried both with our example demo 😄 but faced the issue you described. The `delay_hide` artificially delays, which we don't want from a UX perspective. The faster, the better. 

The `delay_show` sounds lovely, but I couldn't find the right time, so it looks like it's flickering more because the loader appears for a shorter amount of time. I think J. prefers to have the loaders visible, even briefly, to let people know something is happening.

vizro-core/src/vizro/models/_components/graph.py
huong-li-nguyen: Ah, yes, you are right! I tried out the timeout as you suggested; otherwise, it's not visible. In this case, @AnnMarieW, I would change it to below—that suffices for us! 💯 

```suggestion
       overlay_style={"visibility": "visible", "opacity": 0.3},
```

huong-li-nguyen:
@antonymilne - good that you mention it. The second case was not on my mind anymore 😄 In that case the `overlay_style` does work for us!

On `delay_show` / `delay_hide`, I would love to get @AnnMarieW opinion. I find it tricky to find the sweet spot of time that works well for us. See PR comments. But maybe there are some other advantages which are not visually visible here?

I've looked into our previous discussion before and also thought that `target_components` looks promising! However, I think we always return the entire fig whenever we trigger any dashboard interaction + we don't know beforehand what users might want to encode as a trigger for a custom action. So I am not sure if we could leverage this partial triggering of the dcc.Loading component with our current system of actions. @petar-qb - wdyt?

vizro-core/src/vizro/models/_components/ag_grid.py
AnnMarieW: Yes, finding the sweet spot can be a little tricky.

Note that you can include both the `delay_show` and the `delay_hide`.  

I'll make the change to also include `delay_hide=500`  This makes it so that once the spinner is triggered it will show for a min of 500ms.  So that should help with the flash when the callback takes around 600ms

AnnMarieW:
And regarding the `target_components` prop - I agree, I don't think it's applicable with Vizro.

huong-li-nguyen (approved):
Defining `delay_show` and `delay_hide` now works nicely for our demos - thanks @AnnMarieW!

Approved from my side :) Could you add the opacity in as suggested? It's not visible in the demos, but it will be whenever a filter/parameter takes a long time! 

The other potential remaining things:
- [ ] Fix unit tests - I think you just need to bring over your changes to the unit tests as well, as the components are tested for equality.
- [ ] Ignore all of the vizro-ai related tests (they currently don't run through, I'll take them out for your PR to merge, whenever it's ready 👍 )
- [ ] Add a changelog file and mention the increase in the lower bound due to leveraging new features from the newest Dash version

petar-qb:
First of all, @AnnMarieW congrats for the great Dash contribution, and also a big thank you for the Vizro contribution ❤️.

Regarding this PR, I like the `overlay_style={"visibility": "visible", "opacity": 0.3},` as the default Vizro figure configuration and really think it has a place in our codebase. 

Also, I've tried a lot of `delay_show/hide` timings combinations and it's quite difficult to find the right timings given we don't know how long the callback will take to run. 

So as I understood, these are the goals we want to achieve by utilising the `delay_show/hide`:
 1. To avoid a `figure` flickering when the answer comes "immediately" from the server (by utilising the `delay_show` prop),
 2. To avoid a `loading-icon` flickering when the answer comes "immediately" after the `delay_show` time passes (by utilising the `delay_hide` prop).

So, if `delay_show=500, delay_hide=500` props are set and server response time is:

- up to 500 ms:
  - The initial loading page "triggers" `delay_hide` and users wait additionally 500 ms to get fully visible response. It means that the final (fully filtered and parametrised) response is visible to the user for additionally 500 ms but it is in blurred mode. So it's visible, but it seems like some calculations are still happening (but they don't). 
  - There is no loading icon when filter/parameter are triggered. This is maybe something we want to achieve.

- from 500 ms and up:
  - The on_page_load behaves same as described above (in the "up to 500 ms" section)
  - When filtering process is triggered, the `delay_show=500` shows up. It means that after user's filter value selection, only after 500ms the loading icon will appear. By my opinion this looks buggy, and looks like frontend part of the app isn't optimised properly.
  - Also, the final (fully filtered) response is potentially (if server response comes exactly after `delay_show` passes) visible for additional 500 ms (because of `delay_hide`) but in blurred mode.

So, the same effect is keep popping up even if I change  `delay_show/hide` number of milliseconds. 
Because of everything listed above, my opinion is that we shouldn't hardcode the `delay_show/hide` (because we can't foresee server response time for all the apps created using Vizro) and should let our users to do it on themself (by creating the custom component) with the desired number of delay milliseconds.

What do you think? @AnnMarieW @antonymilne @huong-li-nguyen

AnnMarieW:
Hi @petar-qb 

Yes, you are correct with the objectives of the `delay_show` and the `delay_hide` props. The main purpose is to prevent the loading spinner from showing for a short time which can appear as a flash on the screen.

Taking the Viro demo site as an example,  currently, the loading spinner is displayed on every user interaction.  Since the dataset is small, the spinner is displayed for less than one second in most cases.

Adding  `delay_show` of 500ms makes the spinner show up less than half the time (when  I run the app locally).  However, if the callback takes slightly longer, let's say 700ms, then the spinner is back to having the flashing effect because it will be visible for 200ms.  

Adding the `delay_hide` can prevent the flash, because you can specify the minimum amount of time to display the spinner once it becomes visible.  However, the `delay_hide` prop looks the best when the underlying components are not visible while loading (the default) to handle the case described above where the component might be updated before the `delay_hide` time has elapsed. 

It is tricky to get right, and it also depends a lot on the app and the preferences of the users.   I agree that it might be best to let people add it as a custom component.  Would you like me to close this PR now?

huong-li-nguyen:
Hey @AnnMarieW ,

I've talked with both @antonymilne  and @petar-qb. You can still go ahead with the PR and add the  `overlay_style={"visibility": "visible", "opacity": 0.3}` :) 

As Petar said, let's not define delay_show/delay_hide for now 👍

vizro-core/src/vizro/models/_components/ag_grid.py
huong-li-nguyen: As aligned with the rest, let's remove `delay_show` and `delay_hide` and define the overlay_style only 👍

vizro-core/changelog.d/20240523_085359_amward_update_dcc_loading.md
```diff
@@ -0,0 +1,49 @@
...
+-->
+
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+<!--
+### Fixed
```
huong-li-nguyen, line 39: No need to change the changelog entry 👍 

```suggestion
-->
<!--
### Added

- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))

-->
<!--
### Changed

- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))

-->
<!--
### Deprecated

- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))

-->
<!--
### Fixed
```

huong-li-nguyen (commented):
There is one place where you would have to additionally change the dependency for the integration tests to pass:

https://github.com/mckinsey/vizro/blob/main/vizro-core/hatch.toml#L91

Just change it to `dash==2.17.0` on that line :)

antonymilne (approved):
Thank you very much for this @AnnMarieW and for all the careful reviewing @huong-li-nguyen and @petar-qb 🙏

AnnMarieW:
Hey everyone,

I think adding the opacity is a nice addition and will make the transition between "loading" and "ready" smoother.   However, there is still a flicker when the loading time is very short.

Could you please try one more thing before closing this?  I just tried setting the `delay_show=200` and I think it looks great!  Check out the  demo on the Benchmark Analysis page where you click on the grid and it updates the graph.  This setting completely eliminates the flicker.  And 200ms short enough that I don't think it will have the same problem of shifting the flicker as noted earlier.

And of course if you prefer not to add this, I'm still delighted about the addition of the new loading overlay style :slightly_smiling_face:

petar-qb:
> Hey everyone,
> 
> I think adding the opacity is a nice addition and will make the transition between "loading" and "ready" smoother. However, there is still a flicker when the loading time is very short.
> 
> Could you please try one more thing before closing this? I just tried setting the `delay_show=200` and I think it looks great! Check out the demo on the Benchmark Analysis page where you click on the grid and it updates the graph. This setting completely eliminates the flicker. And 200ms short enough that I don't think it will have the same problem of shifting the flicker as noted earlier.
> 
> And of course if you prefer not to add this, I'm still delighted about the addition of the new loading overlay style 🙂

Hey @AnnMarieW.

I tried it out and got that the "initial page load" flickers equally as it flickers without setting the `delay_show=200`. Also, there is no loading icon (and no flickering) while applying filters/parameters. However, it works better only when you run the Vizro demo app locally, and we can't guarantee how all other users examples will work. So, I put `time.sleep(0.15)` inside the filtering method and found that the app flickered even more.

Since we don't know the performance of users apps built using Vizro (since it depends on many factors), I wouldn't hardcode any delay, and just push the overlay argument.

petar-qb (approved):
@AnnMarieW thanks again for your contribution! 🥇 

I pushed some unit tests to make the CI 🟢 and the PR is now ready to go.

PR #488, opened 2024-05-16
[Feat] get vizro ai customized text output

Anna-Xiong:
## Description

we will have a `dataclass` for `Plot` components.
```
@dataclass
class Plot:
    """Data class about a vizro ai plot."""

    code_string: str
    fig: go.Figure = field(default=None)
    business_insights: str = field(default=None)
    code_explanation: str = field(default=None)
```

here is our updated `plot()` call.

```
    def plot(
        self,
        df: pd.DataFrame,
        user_input: str,
        explain: bool = False,
        max_debug_retry: int = 3,
        return_plot_components: bool = False,
    ) -> Union[go.Figure, Plot]:
```

- `explain` controls whether to run extra LLM to get text insights of business and code explanation
- `return_plot_components` controls whether to Plot dataclass or return only plotly fig object

Pending:
- Integration test needs to be updated, another PR will address integration tests and        
```
 # TODO Tentative for integration test, will be updated/removed for new tests
        if self._return_all_text:
            return output_dict
```
will be deleted afterwards.

  -


maxschulz-COL:
I like it! I think this should have the exact same API as the `plot` function, and it does!

A couple of ideas for suggestions:
- maybe tune the name so we know the return type: `get_plot_dict`?
- Warning is nice, but I think not even necessary, could even skip it

Other than that, think this is great 💪 let's do it

Will do a final review once you finalised it, thanks!

Anna-Xiong:
> * maybe tune the name so we know the return type: `get_plot_dict`?
I am leaning towards including output in the name, probably get_plot_output_dict if it is not too long?
> * Warning is nice, but I think not even necessary, could even skip it
 Here we are using INFO instead of WARNING, it would be nice to notify user they haven't turned on explain and won't be able to get business insights and code explanation.

antonymilne (commented):
tl;dr: this doesn't feel right to me, but maybe it's the best solution for now. I don't have much context on it so happy to go with what you guys think is best. We can always change in the future since I imagine the API here may well change anyway in a breaking way when we add the dashboard functionality.

---

Sorry it took so long to review this but I wanted to give it a proper look!

To be honest I am not a big fan of the change, but also not a big fan of the current alternatives. I don't have much context on what we want to achieve here or why we need to return these additional objects, but on the face of it this feels weird to me. Maybe it's ok as a temporary solution but I don't think it's right in the long term. Removing it in future would be a breaking change, but we can be much more relaxed about on VizroAI than on Vizro. So basically I don't have a strong objection if you guys think it's the right thing to do, but it doesn't feel right to me as a long term solution.

### Things that feel wrong

1. API: returning a dictionary for something that is structured data with fixed fields (see e.g. https://www.attrs.org/en/stable/why.html#dicts, https://stackoverflow.com/questions/354883/alternatives-for-returning-multiple-values-from-a-python-function). Here I think the only sensible options are namedtuple or dataclass. And out of those you can't really go wrong with dataclass I think.

2. API: Two functions with the same signature that do something basically identical but one returns a subset of the other. Sometimes this makes sense, but here I think it does not (sorry @maxschulz-COL. And also @Anna-Xiong because I said I basically agreed with Max on this before, but that was before I had read through more of the details).

3. Implementation (so less important than the above two points): this is a case where the code could be much more DRY. There's two functions that do something basically identical but are copy and pasted. If `plot` is effectively a "shortcut wrapper" for `get_plot_outputs` then probably it should be implemented as `def plot(...): return get_plot_outputs(...)["fig_object"]`.

### What would feel better?

_In an ideal world_ where `plot` didn't already exist and return `go.Figure` then probably something like this:

```
@dataclass
class Plot:
    # I'm just taking the same property names as you use now, even if they could maybe be improved a bit
    business_insights: str
    code_explanation: str
    code_string: str
    fig: go.Figure

class VizroAI:
    def plot(...) -> Plot:
        ...

# use as:
plot = VizroAI().plot()
print(plot.business_insights)
vm.Graph(figure=plot.figure)
```

Then the user has access to everything they need, can choose what to take from the output object, etc. without needing nearly-duplicated methods.

It might be worth thinking about how things will evolve in future, e.g. if there's a new `dashboard` method. What would the different patterns look like then?

```
# Following pattern proposed here - potentially many duplicated methods
class VizroAI:
    def plot(...):
        ...

    def get_plot_outputs(...):
        ...

    def dashboard(...):
        ...

    def get_dashboard_outputs(...):
        ...

# Following my above patten 
class VizroAI:
    def plot(...) -> Plot:
        ...

    def dashboard(...) -> Dashboard: # maybe a bad name for class since will be confused with vizro.models.Dashboard, but you get the idea
    # Here it means a dataclass containing a vm.Dashboard object and maybe other things like the code string
        ...
```

## For now

_In the non-ideal world we live in_ where `plot` does already exist and return `go.Figure` and that was a change made recently (? I think) then we should probably not do the above since it's a breaking change. What would then be  the options?

```
# Same as here but don't use dict as return type and refactor to be more DRY
class VizroAI:
    def plot() -> go.Figure: 
     # ultimately if this is a higher-level entry point than get_plot_outputs, it probably shouldn't have as many arguments as get_plot_outputs. But you can't remove arguments for now without it being breaking
        return self.get_plot_outputs.figure

    def get_plot_outputs() -> Plot:
        ...

# The alternative Anna mentioned before where a flag determines the return type
class VizroAI:
    def plot(verbose: bool) -> Union[go.Figure, Plot]: # rename verbose to something that makes sense
        ...
```

tbh I don't much like either of these but would decide based on what arguments to `plot` vs. `get_plot_outputs` would be:
* if they should remain the same signature then probably do the `Union` return type and one `plot` method
* if `get_plot_outputs` might in future have more arguments then probably do as two methods and leave `plot` as the "shortcut wrapper"

maxschulz-COL:
Great review from Antony, and on a more thorough review I agree with some but not all points. Let me try to summarize:

### Regarding what would feel better

I disagree that we should return a `dataclass` for `plot` - a core idea for simplicity is that `plot` returns a figure, and in jupyter that figure is automatically displayed.

### Regarding the things that feel wrong

#### Regarding 1

Yes, I 100% agree and I am annoyed I didn't think of it before given that this is exactly what is done in DS tools. So for the component collection, yes 100% dataclass.

#### Regarding 2

Does not feel great, but I think here it makes sense, because `.plot` is a very special function... It returns a fig object for simplicity (and that should remain as is, as in jupyter it shows the fig automatically and it is the core functionality), but in a jupyter environment, it also shows the insights and code snippet above the fig. That is potentially what is most confusing, but at the same time probably the best way to handle a complicated situation.

Thus, the second function with the same function signature was intended to return the same (not a subset) things, but in a way that is simple to further use (e.g. if you would like to just have the code string, or if you would like to return the insights separately in a chatbot).

Here is the 4 combinations, and what I think it should do:

_In Jupter_:
- `plot(...,explain=False)` -> Return fig object, that is automatically displayed
- `plot(...,explain=True)` ->  Return fig object, that is automatically displayed, with explanation displayed above, not altering the return type

_not in jupyter_:
- `plot(...,explain=False)` -> Return fig object, that can be displayed if the user chooses `fig.show()` themselves
- `plot(...,explain=True)` -> Return fig object, that can be displayed if the user chooses `fig.show()` themselves. _Print the explanation to the terminal_ (this is new, and I think this would make `.plot` more consistent

If the above is ensured, we just have to find a way to deliver all output to the user in an alternative way (which as said in 1 should definitely be a `dataclass`. I think there is broadly three options:

1. Variable return type based on a flag, e.g. `.plot(...,return_components = True)`
2. Alternative function (current proposed solution)
3. Alternative function but private method, so no pubic API (somewhat current solution)
4. (Scrap this idea, and just return components - A's preferred solution, which is disagree with)


#### Regarding 3
Yes, I think that would be wise, but I felt it was not the major focus for now

### What is my actual preferred option
I have tried to show above that `plot` is not a subset of `get_plot_outputs`, but a sister function. That would suggest (I think) a variable return type. But it makes sense as you say to look into the future:
- will there be additional things to return that do not make sense to be part of `.plot`
- will there be situations, where `plot` interacts with other parts of `VizroAI`, will it be a burden to sync with a potential different function (thinking of iteration, usage in `.dashboard`, etc)
- do we need to think about iteration, will that be part of the same function?

Long story short - we simply cannot tell. Thus I prefer to have a single function with a variable return type that is clear to maintain: it returns either the fig with all other components shown alongside OR it returns a dataclass with all the components as attributes.

That is also easy to explain in docs, and easy to use. If `plot` becomes involved in other parts of the program, at least we only have to think about `plot` and not something else.

antonymilne:
Great points @maxschulz-COL! You are absolutely right that it's good to automatically show the figure in a Jupyter notebook, which I hadn't thought of before. That would actually still be possible with my "ideal" solution above since you could delegate the relevant methods to the underlying figure:
```
@dataclass
class Plot:
    # I'm just taking the same property names as you use now, even if they could maybe be improved a bit
    business_insights: str
    code_explanation: str
    code_string: str
    fig: go.Figure

    # Whatever methods are needed
    def _repr_html_(...):
        return fig._repr_html_(...)
```

This is very similar to one of the prototypes I had for our `_DashboardReadyFigure` which had similar requirements of needing to behave like a `go.Figure` in Jupyter while requiring extra behaviour for use in a dashboard. There I went for inheritance over composition in the end. One of the reasons for this is it turned out that while it's possible to do the above it's actually a bit painful to implement. And given we have the constraint of needing to return a `go.Figure` already given the change in vizro-ai 0.2.0, let's not do this.

We could actually also follow the subclassing approach like we do with `_DashboardReadyFigure` and do `class Plot(go.Figure)` and put a `business_insights` property into that class to get the best of both worlds where you get something that behaves like `go.Figure` but also exposes additional properties. This is pretty ugly and also a little awkward to implement so I don't _think_ I'm a fan of it but might work ok here 🤔 

Assuming we don't like that then I agree with your solution 1 here that we have a single `plot` function but with an argument that alters the return type. But I find the existence of both `explain` and `return_components` a bit confusing and wonder if we could just use the `explain` argument for both?

1. `plot(..., explain=False)`: return fig object and don't show any explanation, just like it does now.
2. `plot(..., explain=True)`: return `dataclass` (different from now). If in Jupyter then display explanation on screen (as done now) and also fig on screen (needed since we no longer directly return the figure object). If outside Jupyter then personally I don't care whether we show the explanation or not, since it's available in the return dataclass anyway.

My assumption here is that in 99% of cases `explain` is basically equivalent to `return_components`. The only cases that this solution renders impossible would be:
* `explain=True` and `return_components=False`: you want to see the explanation but don't want to use it further. With my above solution you set `explain=True` and get extra components in your return object and you can just ignore those if you like so this seems fine since the graph is still displayed
* `explain=False` and `return_components=True`: again just set `explain=True`. The only difference here is the explanation gets shown (at least in Jupyter notebooks) when maybe you don't want that, but that seems ok to me also.

maxschulz-COL:
> Assuming we don't like that then I agree with your solution 1 here that we have a single `plot` function but with an argument that alters the return type. But I find the existence of both `explain` and `return_components` a bit confusing and wonder if we could just use the `explain` argument for both?
> 
> 1. `plot(..., explain=False)`: return fig object and don't show any explanation, just like it does now.
> 2. `plot(..., explain=True)`: return `dataclass` (different from now). If in Jupyter then display explanation on screen (as done now) and also fig on screen (needed since we no longer directly return the figure object). If outside Jupyter then personally I don't care whether we show the explanation or not, since it's available in the return dataclass anyway.

Happy with this! Would this not be another breaking change, because now the return type of the plot function with `explain=True` is actually `go.Figure` and it would change to `dataclass`?

antonymilne:
Haha, I knew you would ask that 😀 And it's definitely the right question to ask.

Technically yes, it's a breaking change. But breaking changes are always a bit of a grey area, relative to how things are actually used and not clear cut. Within the context of how people use VizroAI, would anyone be doing `explain=True` and using the `go.Figure` directly? My feeling is probably no: if you want to use the figure then you would do `explain=False` (the default, so undoubtedly the most common anyway). The result for Jupyter users would look the same because we'd still show the plot in the notebook. The problem would only come if you're actually consuming that plot in some way with `explain=True`.

Given that we only just did the breaking release 0.2.0 and changed the return type there I would be inclined to do this more as a "new feature" rather than breaking change, but it's definitely debatable. I don't mind really - up to you guys to think about how many people this actually would break the workflow for.

maxschulz-COL:
> Given that we only just did the breaking release 0.2.0 and changed the return type there I would be inclined to do this more as a "new feature" rather than breaking change, but it's definitely debatable. I don't mind really - up to you guys to think about how many people this actually would break the workflow for.

I am ok with it! Let's see what @Anna-Xiong thinks

Anna-Xiong:
Thanks for the thorough reviews and suggestions on enhancing the plot function of VizroAI. After considering all the inputs, I have updated the code accordingly and have a few point would like to contribute here:

**Agreement on Single plot Call:**
I agree with using a single `plot` method that adjusts return type based on the provided arguments. The reason is the same as our first POC: given that even if we have two calls, there would be a duplicate of almost all arguments.

**Support for Dataclass:**
Using a dataclass to return structured data is a great idea and easier to use than dictionary for sure. updated that already.

**Concerns with only using explain:**
However, I have reservations about using only `explain` parameter to control all aspects of the output. 
1. Our users might have requirements where they need the `figure object` and `code string` to integrate into chat application without the explanation or vice versa.  In fact, that is one of the most common requirement we have seen and this is also what our chat application uses as well. 
2. This also involves calling extra layer of LLM or not(cost and speed relevant). 
In short, I would summarize the responsibility of `explain` and `return_plot_components` as below:
- `explain` only controls whether to run extra LLM to get text insights of business and code explanation
- `return_plot_components` controls whether to Plot dataclass or return only plotly fig object

I would be strongly leaning towards additional flags here that control the type of the output without merging them into a single parameter `explain`. 

If we are not ok for this adding extra argument change, I would think it is better to have a separate `get_plot_components()` method that
-  `VizroAI.get_plot_components()` could return the dataclass so that we don't mix what `explain` controls, 
- and make `VizroAI.plot()` a short cut , it can return the `fig` from `dataclass` returned from `VizroAI.get_plot_components()`

**Decision on Pylint Rule Adjustment:**
Lastly, as we expand our this `VizroAI.plot()`function, the additional argument brings our total to six, surpassing the current pylint rule of five arguments per function. We need to decide if we should adjust our linting rules to accommodate this change. 


**Code Updates:**
I've made some updates to the existing prototype to reflect our discussions so far. 

We care more about auto rendering in jupyter environment, so if users are using jupyter, the utils has been refactoring to take care of auto rendering of fig. **No matter what users specify in `plot()` call, the fig is always rendered in jupyter, and not rendered automatically in other environment, which they can add one line of code to show the fig(exact same as our current `plot()` call, no breaking change).



- `plot(...,explain=False, return_plot_components= True)` -> Return dataclass Plot that consists of `code_string` and `fig`, the fig is automatically displayed if in jupyter
- `plot(...,explain=False, return_plot_components= False)` -> Return `fig` directly, the fig is automatically displayed if in jupyter.
- `plot(...,explain=True, return_plot_components= True)` -> Return dataclass `Plot` that consists of `code_string` and `fig`, `business_insights` , and `code_explanation`.  The fig is automatically displayed if in jupyter and the Markdown is rendered for text information.
-  `plot(...,explain=True, return_plot_components= False)` -> Return `fig` directly, The explnation marthe fig is automatically displayed if in jupyter and the Markdown is rendered for text information.

Also `return_plot_components` will be default to False, so we don't cause any breaking changes from the users' side.

Let's take a look at the code and share your thoughts on these points, especially regarding the flexibility of the output controls and the pylint rule adjustment.

maxschulz-COL:
> **Concerns with only using explain:** However, I have reservations about using only `explain` parameter to control all aspects of the output.
> 
> 1. Our users might have requirements where they need the `figure object` and `code string` to integrate into chat application without the explanation or vice versa.  In fact, that is one of the most common requirement we have seen and this is also what our chat application uses as well.
> 2. This also involves calling extra layer of LLM or not(cost and speed relevant).
>    In short, I would summarize the responsibility of `explain` and `return_plot_components` as below:
> 
> * `explain` only controls whether to run extra LLM to get text insights of business and code explanation
> * `return_plot_components` controls whether to `Plot dataclass` or return only `plotly fig object`
> 
> I would be strongly leaning towards additional flags here that control the type of the output without merging them into a single parameter `explain`.

This is a very valid point, and I had not thought of that. I think this makes a strong case for actually adding an extra parameter. It also avoids the "breaking change".

> If we are not ok for this adding extra argument change, I would think it is better to have a separate `get_plot_components()` method that


> **Decision on Pylint Rule Adjustment:** Lastly, as we expand our this `VizroAI.plot()`function, the additional argument brings our total to six, surpassing the current pylint rule of five arguments per function. We need to decide if we should adjust our linting rules to accommodate this change.

I think it is ok to add an ignore here - I think that pylint rule is generally good to have, but can feel too strict at times.

> **Code Updates:** I've made some updates to the existing prototype to reflect our discussions so far.
> 
> We care more about auto rendering in jupyter environment, so if users are using jupyter, the utils has been refactoring to take care of auto rendering of fig. **No matter what users specify in `plot()` call, the fig is always rendered in jupyter, and not rendered automatically in other environment, which they can add one line of code to show the fig**(exact same as our current `plot()` call, no breaking change).
> 
> * `plot(...,explain=False, return_plot_components= True)` -> Return dataclass Plot that consists of `code_string` and `fig`, the fig is automatically displayed if in jupyter
> * `plot(...,explain=False, return_plot_components= False)` -> Return `fig` directly, the fig is automatically displayed if in jupyter.
> * `plot(...,explain=True, return_plot_components= True)` -> Return dataclass `Plot` that consists of `code_string` and `fig`, `business_insights` , and `code_explanation`.  The fig is automatically displayed and the Markdown is rendered for text information if in jupyter
> * `plot(...,explain=True, return_plot_components= False)` -> Return `fig` directly, The `fig` is automatically displayed and the Markdown is rendered for text information if in jupyter

Didn't have time to check the details - is the mechanism that render the `fig` in the cases where `return_plot_components= False` the original Jupyter one? Just curious on the details.

> Also `return_plot_components` will be default to False, so we don't cause any breaking changes from the users' side.

Great!

> Let's take a look at the code and share your thoughts on these points, especially regarding the flexibility of the output controls and the pylint rule adjustment.

Overall I think we should do it as you say, and you are right, it is one of the most requested features, and it separates responsibility of arguments.

vizro-ai/src/vizro_ai/_vizro_ai.py
```diff
@@ -76,18 +78,29 @@ def _run_plot_tasks(
...
+                "Chart creation failed. Retry debugging has reached maximum limit. Try to rephrase the prompt, "
+                "or try to select a different model. Fallout response is provided: \n\n" + code_string
             )
-
-        return {
-            "business_insights": business_insights,
-            "code_explanation": code_explanation,
-            "code_string": code_string,
-        }
+        else:
```
antonymilne, line 87: No need for this `else` here (I'm surprised if the linter doesn't point this out actually) since the `if` clause raises an exception.

vizro-ai/src/vizro_ai/_vizro_ai.py
```diff
@@ -76,18 +78,29 @@ def _run_plot_tasks(
...
+                    chain_input=user_input, code_snippet=code_string
+                )
+
+                return Plot(
+                    code_string=code_string,
+                    fig=fig_object,
+                    business_insights=business_insights,
+                    code_explanation=code_explanation,
+                )
+            else:
```
antonymilne, line 102: Also no need for the `else` here since the `if` clause has `return`.

vizro-ai/src/vizro_ai/_vizro_ai.py
```diff
@@ -99,41 +112,53 @@ def _get_chart_code(self, df: pd.DataFrame, user_input: str) -> str:
...
-        code_explanation = output_dict.get("code_explanation")
+        vizro_plot = self._run_plot_tasks(
+            df=df, user_input=user_input, explain=explain, max_debug_retry=max_debug_retry
+        )
 
-        if code_string.startswith("Failed to debug code"):
-            raise DebugFailure(
-                "Chart creation failed. Retry debugging has reached maximum limit. Try to rephrase the prompt, "
-                "or try to select a different model. Fallout response is provided: \n\n" + code_string
+        if explain is False:
```
antonymilne, line 143:
```suggestion
        if not explain:
```

antonymilne (approved):
Thanks for all the explanation and changes to the code @Anna-Xiong. It all makes sense to me and looks much better I think. I agree with @maxschulz-COL - no need to worry about the pylint warning, just ignore it like you do.

I have a few minor suggestions, mainly about the names of things. Would be good if @maxschulz-COL could also review and think about the names when he gets a chance 🙂

vizro-ai/src/vizro_ai/_vizro_ai.py
Anna-Xiong: usually when there are too many if else i prefer to make them more clearly by not omitting else, but ok to delete it as well if that is the style we follow.

vizro-ai/src/vizro_ai/_vizro_ai.py
```diff
@@ -99,41 +112,53 @@ def _get_chart_code(self, df: pd.DataFrame, user_input: str) -> str:
...
-    def plot(
-        self, df: pd.DataFrame, user_input: str, explain: bool = False, max_debug_retry: int = 3
-    ) -> Union[go.Figure, Dict[str, Any]]:
+    def plot(  # pylint: disable=too-many-arguments  # noqa: PLR0913
+        self,
+        df: pd.DataFrame,
+        user_input: str,
+        explain: bool = False,
+        max_debug_retry: int = 3,
+        return_plot_components: bool = False,
```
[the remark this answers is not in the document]
Anna-Xiong: how about `return_elements`

Anna-Xiong:
@antonymilne @maxschulz-COL 

Thank you for reviewing it again! 

> Didn't have time to check the details - is the mechanism that render the fig in the cases where return_plot_components= False the original Jupyter one? Just curious on the details.

- Yes, it will render the fig where return_plot_components= False, and return a direct fig as well.

Then most questions remaining here would be relevant to naming of arguments and dataclass. 
- I followed @antonymilne suggestions on dataclass attributes and this new one looks simpler and easier to use, but probably use `return_elements` instead of `return_details`  would be better for the argument in `plot()` call? @maxschulz-COL do you have any other suggestion on names?

vizro-ai/src/vizro_ai/_vizro_ai.py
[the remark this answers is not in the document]
maxschulz-COL: ir even `return_dataclass`, but happy with the above as well!

vizro-ai/changelog.d/20240603_173503_anna_xiong_get_vizro_ai_customized_output.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Enable feature to get all possible outputs from VizroAI.plot() using by specifying 'return_elements = True' ([#488](https://github.com/mckinsey/vizro/pull/488)). It returns a datalcass that contains code string, figure object, business insights, and code explanation.
```
maxschulz-COL, line 22:
```suggestion
- Add argument 'return_elements` to `VizroAI.plot()` that, when set to `True`, changes the return type to a `dataclass` containing the code string, figure object, business insights, and code explanation. ([#488](https://github.com/mckinsey/vizro/pull/488))
```

vizro-ai/src/vizro_ai/utils/helper.py
```diff
@@ -9,6 +10,16 @@
 from .safeguard import _safeguard_check
 
 
+@dataclass
+class Plot:
```
maxschulz-COL, line 14:
```suggestion
class PlotOutputs:
```

vizro-ai/src/vizro_ai/utils/helper.py
```diff
@@ -9,6 +10,16 @@
...
+    """Data class about a vizro ai plot."""
```
maxschulz-COL, line 15:
```suggestion
    """Dataclass containing all possible `VizroAI.plot()` output."""
```

vizro-ai/src/vizro_ai/utils/helper.py
```diff
@@ -49,22 +60,29 @@ def _debug_helper(
...
 
     Args:
         code: code string to be executed
         local_args: additional local arguments
+        show_fig: boolean flag indicating if fig will be rendered automatically
         is_notebook_env: boolean flag indicating if code is run in Jupyter notebook
 
     Returns:
-        go.Figure
+        Plotly go figure
```
maxschulz-COL, line 74:
```suggestion
        go.Figure
```
I think that is clearer actually seeing it's an element of `graph_objects`

vizro-ai/src/vizro_ai/utils/helper.py
```diff
@@ -75,21 +93,14 @@ def _exec_code_and_retrieve_fig(
     return dashboard_ready_fig
 
 
-def _exec_fig_code_display_markdown(
-    df: pd.DataFrame, code_snippet: str, biz_insights: str, code_explain: str
-) -> go.Figure:
-    # TODO change default test str to other
+def _display_markdown(code_snippet: str, biz_insights: str, code_explain: str) -> None:
```
maxschulz-COL, line 96: I think this is a very good refactoring!

vizro-ai/src/vizro_ai/_vizro_ai.py
```diff
@@ -99,41 +112,51 @@ def _get_chart_code(self, df: pd.DataFrame, user_input: str) -> str:
...
+        return_elements: bool = False,
+    ) -> Union[go.Figure, Plot]:
         """Plot visuals using vizro via english descriptions, english to chart translation.
 
         Args:
             df: The dataframe to be analyzed.
             user_input: User questions or descriptions of the desired visual.
             explain: Flag to include explanation in response.
             max_debug_retry: Maximum number of retries to debug errors. Defaults to `3`.
+            return_elements: Flag to return plot dataclass that includes all components.
```
maxschulz-COL, line 133:
```suggestion
            return_elements: Flag to return dataclass that includes all components produced.
```

maxschulz-COL (approved):
⭐ This is really great stuff, and I love how clean it all has become. Definitely feels like a better solution!! Well done, and thanks for guarding against my premature opinions ;)

antonymilne:
Just a final suggestion on the names: `return_elements` is ok but I think I'd prefer `return_full` overall if that makes sense for you two also? @maxschulz-COL @Anna-Xiong

Anna-Xiong:
> Just a final suggestion on the names: `return_elements` is ok but I think I'd prefer `return_full` overall if that makes sense for you two also? @maxschulz-COL @Anna-Xiong

I am ok either way, but slightly towards `return_elements` or `return_full_elements`? Let's see what @maxschulz-COL thinks?

vizro-ai/src/vizro_ai/_vizro_ai.py
Anna-Xiong: sometimes i always feel lint detection is a bit mysterious, yes you are right, it should detect it by this rule:https://pylint.readthedocs.io/en/stable/user_guide/messages/refactor/no-else-raise.html? 

Recently i did not get this complaint anymore somehow

maxschulz-COL:
> > Just a final suggestion on the names: `return_elements` is ok but I think I'd prefer `return_full` overall if that makes sense for you two also? @maxschulz-COL @Anna-Xiong
> 
> I am ok either way, but slightly towards `return_elements` or `return_full_elements`? Let's see what @maxschulz-COL thinks?

I prefer `return_elements` - I find that is really what it is.

antonymilne:
Fine by me, let's go for `return_elements` then 🙂

PR #497, opened 2024-05-27
[QA] Fix flaky `data_manager` unit tests

petar-qb:
## Description
See [issue](https://github.com/McK-Internal/vizro-internal/issues/957) for more.

Changes introduced in this PR:
- deterministically solve the problem of unstable data_manager tests
- speed-up data_manager unit tests a lot

## Notice


vizro-core/tests/unit/vizro/managers/test_data_manager.py
```diff
@@ -9,11 +10,26 @@
...
 from asserts import assert_frame_not_equal
 from flask_caching import Cache
+from freezegun import freeze_time
 from pandas.testing import assert_frame_equal
 from vizro import Vizro
 from vizro.managers import data_manager
 
 
+# TODO: Consider putting this inside the global 'tests/conftest.py' file with a scope of "session"?
+@pytest.fixture(autouse=True)
```
huong-li-nguyen, line 20: Do you mind explaining how this speeds up the data_manager unit tests? I am a noob in this so it's not immediately clear from the code for me 😄

petar-qb: So, the following lines of code ensure that every unit test in the `test_data_manager.py` have pinned system time to "2024-01-01 12:00:00". So the time is frozen inside the test (from the beginning to the end of the test), and time doesn't tick. 
```
@pytest.fixture(autouse=True)
def freeze_and_patch_time(monkeypatch):
    with freeze_time("2024-01-01 12:00:00", tick=False) as frozen_time:
        yield
```
It means that if you run `datetime.datetime.now()` or `time.time()` (or similar) inside the test, the returned value will be "2024-01-01 12:00:00". It's a crucial move as flask-caching library relies a lot on the system time (see the[issue](https://github.com/McK-Internal/vizro-internal/issues/957) description). So, if `data_frame` timeout is set like this `data_manager["data_1"].timeout=15`, the expiration time (calculated from the flask-caching) for the `data_1` will be 2024:01:01-12:00:15 now.

In the current tests implementation we use `time.sleep(N)` to ensure that data_frame cache expires. But, it means that the most of our unit tests will fail because time doesn't tick and cached data_frame will never expire. So, to ensure that time.sleep(N) moves the time for `N` seconds, I patched time.sleep function so `it moves the frozen time for N seconds`. To implement this I used the following code:
```
def advance_time(seconds):
            # Advance the frozen time by the given number of seconds
            frozen_time.move_to(frozen_time.time_to_freeze + datetime.timedelta(seconds=seconds))

        # Patch time.sleep to use advance_time instead of actual sleeping
        monkeypatch.setattr(time, "sleep", advance_time)
```

So, the speedup comes as a side product of the time.sleep patching as time.sleep doesn't really "sleep" for N seconds but it triggers the `advance_time` function that moves the frozen time for N seconds instantly.

antonymilne (approved):
@petar-qb this is  reallyawesome work 💯 Not only have you fixed the flakiness, the data manager tests have gone from taking ~35s (which was always a source of mild irritation for me) to around 3.5s! I've tested the latest version I pushed by running `pytest test_data_manager.py` 100 times. Before the change, that would fail ~10% of the time, and now there's not a single failure 🥳 

I made a couple of commits, see what you think. My reasoning is:
* we should try to keep tests as close to real/unmocked behaviour as possible, and only tests in data manager use `time.sleep`
* hence let's not make it `autouse=True` or increase the fixture scope outside where it's strictly needed and instead keep it as explicitly opt-in
* set `tick=True` for similar reasons to ensure time is only frozen for the purpose of fast-forwarding `time.sleep`  and does not artificially make subsequent calls to `.load` seem simultaneous
* monkeypatching the `time.sleep` was super clever but I would again prefer to be explicit, especially given that `freezegun` comes with the `.tick` method for exactly this, and monkeypatching is always quite fragile (e.g. `from time import sleep; sleep()` isn't the same as `import time; time.sleep()`)
* the downside is that any future tests that use `time.sleep` will not automatically benefit, but I think this is ok because I don't anticipate there being many such things (just tests of caching really), and I'd rather be explicit here rather than it being too magical

petar-qb:
@antonymilne thanks for taking the PR to a new level. I really like how the tests cases are now much more realistic (with timeouts of 300s and 100s). I agree with all your bullets from the comment above, but unfortunately I can't approve the PR I created 😄.

vizro-core/hatch.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
   "coverage[toml]>=6.5",
   "pytest",
   "pytest-mock",
+  "freezegun>=1.5.0",
```
petar-qb, line 24: The `TickingDateTimeFactory.tick()` (we use it as `freezer.tick()`) [has been supported](https://github.com/spulec/freezegun/commit/e126e111cc64d6da957167bf4ec316266613c687) since version `1.5.0`, hence this limitation.

vizro-core/tests/unit/vizro/managers/test_data_manager.py
huong-li-nguyen: Thanks for the explanation - this helps a lot! 🚀

huong-li-nguyen (approved):
LGTM! Thank you guys 💪

Issue #109, opened 2023-10-12
Rerun Vizro Dashboard in Jupyter doesn't work


noklam:
### Description

> ValidationError: 1 validation error for Graph
> figure
>   Component with id=scatter_chart already exists and is mapped to dataset 140638475834896. Components must uniquely map to a dataset across the whole dashboard. If you are working from a Jupyter Notebook, please either restart the kernel, or use 'from vizro.managers import data_manager; data_manager._reset()`. (type=value_error)

 'from vizro.managers import data_manager; data_manager._reset()`. After I follow the instruction and rerun the cell, I run into a new error.

> DuplicateIDError: Model with id=scatter_chart already exists. Models must have a unique id across the whole dashboard. If you are working from a Jupyter Notebook, please either restart the kernel, or use 'from vizro.managers import model_manager; model_manager._reset()`.

After I run 'from vizro.managers import model_manager; model_manager._reset()`, no error is raised but it doesn't show the dashboard UI anymore.

![image](https://github.com/mckinsey/vizro/assets/18221871/7d1e2f36-61de-45bb-b53b-03ceeb1e6bf4)

### Expected behavior

I expect the dashboard render properly. Workaround is restart the Kernel which work fine but is considerably slower.

### vizro version

0.1.4

### Python version

3.8.5

### OS

MacOS

### How to Reproduce

Follow the tutorial and run it in a notebook twice.

https://vizro.readthedocs.io/en/stable/pages/tutorials/first_dashboard/#4-explore-further

### Code of Conduct

- [X] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

maxschulz-COL:
Hey @noklam,

thanks so much for getting in touch :) I was able to reproduce this issue, and it is indeed a bug. We are working on a smoother Jupyter experience already (see https://github.com/mckinsey/vizro/issues/65), but it looks like that the short term solution described in this issue does not capture this case.

So in the short term we would recommend going via the approach of restarting the kernel, this is as you say slower, but the most reliable method. We will get back once there is a more comprehensive solution.

maxschulz-COL:
Reproducing this error works, although atm the error message does not show for me. Code can be found at: https://github.com/mckinsey/vizro/tree/bug/109

Observations:
- page registries are different but not empty
- comparing the global variable `dash` seems to suggest that only `supplied_layout` and `layout` are differing, and in that, only the IDs
[...]

PR #519, opened 2024-06-07
[POC] KPI Infinite scroll AgGrid

petar-qb:
## Description

- Implementation of [AgGrid infinite-row-model](https://dash.plotly.com/dash-ag-grid/infinite-row-model) for the KPI example due to speedup.

FYI:
 - Dash [AgGrid pagination](https://dash.plotly.com/dash-ag-grid/pagination) stores all data on the client side. Therefore, the recommended implementation for big data is to use `infinite row model`. The downsides of this approach are that the             sorting and filtering interface options must be set to False (disabled) (because only a small part of the source data is visible on the client side).
 
Additionally:
- For the further optimisation of the initial page loading, we should:
  -  or reimplement the `Page.pre_build()` so it doesn't trigger the on_page_load mechanism if there are only tables and grids and without any controls on the page,
  -  or we should reimplement the AgGrid/Table to initially return an empty pd.DataFrame.
- The new custom ag-grid (`infinite_scroll_ag_grid`) implementation is introduced since the predefined one (`dash_ag_grid`) doesn't allow to skip the following code execution which significantly slows down the page loading time:
```
...
"rowData": data_frame.apply(
            lambda x: (
                x.dt.strftime("%Y-%m-%d")  # set date columns to `dateString` for AG Grid filtering to function
                if pd.api.types.is_datetime64_any_dtype(x)
                else x
            )
        ).to_dict("records"),
...
```

Open Question:
- Should we make the `infinite_scroll_dash_ag_grid` public?

## Notice

### DIFF
--- a/vizro-core/examples/kpi/app.py
+++ b/vizro-core/examples/kpi/app.py
@@ -247,16 +247,98 @@ dashboard = vm.Dashboard(
 
 
 # CALLBACKS -----------------------------------------------------------------------------------
+operators = {
+    "greaterThanOrEqual": "ge",
+    "lessThanOrEqual": "le",
+    "lessThan": "lt",
+    "greaterThan": "gt",
+    "notEqual": "ne",
+    "equals": "eq",
+}
+
+
+def filter_df(df, data, col):
+    if data["filterType"] == "date":
+        crit1 = data["dateFrom"]
+        crit1 = pd.Series(crit1).astype(df[col].dtype)[0]
+        if "dateTo" in data:
+            crit2 = data["dateTo"]
+            crit2 = pd.Series(crit2).astype(df[col].dtype)[0]
+    else:
+        crit1 = data["filter"]
+        crit1 = pd.Series(crit1).astype(df[col].dtype)[0]
+        if "filterTo" in data:
+            crit2 = data["filterTo"]
+            crit2 = pd.Series(crit2).astype(df[col].dtype)[0]
+    if data["type"] == "contains":
+        df = df.loc[df[col].str.contains(crit1)]
+    elif data["type"] == "notContains":
+        df = df.loc[~df[col].str.contains(crit1)]
+    elif data["type"] == "startsWith":
+        df = df.loc[df[col].str.startswith(crit1)]
+    elif data["type"] == "notStartsWith":
+        df = df.loc[~df[col].str.startswith(crit1)]
+    elif data["type"] == "endsWith":
+        df = df.loc[df[col].str.endswith(crit1)]
+    elif data["type"] == "notEndsWith":
+        df = df.loc[~df[col].str.endswith(crit1)]
+    elif data["type"] == "inRange":
+        if data["filterType"] == "date":
+            df = df.loc[df[col].astype("datetime64[ns]").between_time(crit1, crit2)]
+        else:
+            df = df.loc[df[col].between(crit1, crit2)]
+    elif data["type"] == "blank":
+        df = df.loc[df[col].isnull()]
+    elif data["type"] == "notBlank":
+        df = df.loc[~df[col].isnull()]
+    else:
+        df = df.loc[getattr(df[col], operators[data["type"]])(crit1)]
+    return df
+
+
 @callback(
     Output("kpi_grid", "getRowsResponse"),
     Input("kpi_grid", "getRowsRequest"),
 )
 def infinite_scroll(request):
-    """Infinite scroll callback mechanism for the AG Grid."""
-    if request is None:
-        return no_update
-    partial = df_complaints.iloc[request["startRow"] : request["endRow"]]
-    return {"rowData": partial.to_dict("records"), "rowCount": len(df_complaints.index)}
+    dff = df_complaints.copy()
+
+    if request:
+        if request["filterModel"]:
+            fils = request["filterModel"]
+            for k in fils:
+                try:
+                    if "operator" in fils[k]:
+                        if fils[k]["operator"] == "AND":
+                            dff = filter_df(dff, fils[k]["condition1"], k)
+                            dff = filter_df(dff, fils[k]["condition2"], k)
+                        else:
+                            dff1 = filter_df(dff, fils[k]["condition1"], k)
+                            dff2 = filter_df(dff, fils[k]["condition2"], k)
+                            dff = pd.concat([dff1, dff2])
+                    else:
+                        dff = filter_df(dff, fils[k], k)
+                except:
+                    pass
+            dff = dff
+
+        if request["sortModel"]:
+            sorting = []
+            asc = []
+            for sort in request["sortModel"]:
+                sorting.append(sort["colId"])
+                if sort["sort"] == "asc":
+                    asc.append(True)
+                else:
+                    asc.append(False)
+            dff = dff.sort_values(by=sorting, ascending=asc)
+
+        lines = len(dff.index)
+        if lines == 0:
+            lines = 1
+
+        partial = dff.iloc[request["startRow"] : request["endRow"]]
+        return {"rowData": partial.to_dict("records"), "rowCount": lines}
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


huong-li-nguyen (commented):
Hey @petar-qb - thanks for the quick investigation! 🚀 

I would like to keep this PR open as a draft for discussion when @antonymilne is back, but I wouldn't like to merge it into the kpi dashboard if this can only be implemented when we turn off filtering and sorting.

The page with the ag-grid doesn't have any controls on purpose, as it looks nicer if the table takes up all space and people don't have to switch between the side panel and main panel. But that's why we do need the filtering and sorting functionality on that table. I would prefer the slower speed over not having filtering/sorting enabled in this case.

maxschulz-COL:
> Hey @petar-qb - thanks for the quick investigation! 🚀
> 
> I would like to keep this PR open as a draft for discussion when @antonymilne is back, but I wouldn't like to merge it into the kpi dashboard if this can only be implemented when we turn off filtering and sorting.
> 
> The page with the ag-grid doesn't have any controls on purpose, as it looks nicer if the table takes up all space and people don't have to switch between the side panel and main panel. But that's why we do need the filtering and sorting functionality on that table. I would prefer the slower speed over not having filtering/sorting enabled in this case.

I am sure you have looked at this @petar-qb, but just in case you had not seen: https://dash.plotly.com/dash-ag-grid/infinite-scroll#infinite-scroll-with-sort-and-filter

Would this work?

petar-qb:
> I am sure you have looked at this @petar-qb, but just in case you had not seen: https://dash.plotly.com/dash-ag-grid/infinite-scroll#infinite-scroll-with-sort-and-filter
> 
> Would this work?

@maxschulz-COL Thanks for pointing this out (I forgot even though I saw it). I don't know how good idea it is to push this code for the demo example (since its complexity could confuse our users and repeal them to use Vizro).

However, let me investigate how generic this code is on Monday. If it is, maybe we can implicitly enable this code if the infinite scroll is turned on.

petar-qb:
@maxschulz-COL @huong-li-nguyen @antonymilne 

The code from the [link](https://dash.plotly.com/dash-ag-grid/infinite-scroll#infinite-scroll-with-sort-and-filter) works 🎉. 

What do you think, do we want to predefined the `infinite_scroll_dash_ag_grid`?

maxschulz-COL:
> @maxschulz-COL @huong-li-nguyen @antonymilne
> 
> The code from the [link](https://dash.plotly.com/dash-ag-grid/infinite-scroll#infinite-scroll-with-sort-and-filter) works 🎉.
> 
> What do you think, do we want to predefined the `infinite_scroll_dash_ag_grid`?

I am not opposed to having another grid callable, it would be in line with what we are doing for KPI cards. 

But I think we need to abstract the entire callback logic away. Ideally we also have only a minimum set of defaults.

Another route would be to enhance the existing callable for the case that someone chooses infinite scroll: e.g. we could make this an argument of the `dash_ag_grid` and when chosen `True`, we would take care of the rest. Not sure I like it, but it would reduce the number of `CapturedCallable` we produce, but could lead to other maintenance issue.

Overall I think I am in favour if we can prove that this setting improves the performance noticeably. If we can enable it in an easy way, then I think this is exactly what Vizro stands for

antonymilne:
Sorry for the very delayed response to this but hopefully better late than never... 🤞 Just to check we're all on the same page here, I think this is where things stand:
* default AgGrid uses client-side row model that downloads all data to browser at start. No callbacks involved
* infinite row model _without_ sorting or filtering can be achieved with [one simple callback](https://dash.plotly.com/dash-ag-grid/infinite-row-model#simple-example:-no-sorting-or-filtering)
* infinite row model _with_ sorting or filtering can be achieved with a [much more complicated callback](https://dash.plotly.com/dash-ag-grid/infinite-scroll#infinite-scroll-with-sort-and-filter)

I think @maxschulz-COL was describing above two possible APIs:
1. keep our `dash_ag_grid` with client-side row model, no callbacks + new `dash_ag_grid_infinite` with infinite row model and callback
2. keep our `dash_ag_grid` with client-side row model, no callbacks. Intercept the argument `rowModelType="infinite"` and use that to make it infinite row model and callback

In theory I agree with either API, but I'm not sure how easily we can actually implement either one? The main problem I see here is **where do you define the callback?** 
1. if it's outside `dash_ag_grid` then you don't have correct access to `data_frame` and so can't get the correct data (after filtes/parameters on control panel). If you don't have any controls then it's possible to do a hack here to just do `data_manager["iris"].load()` directly in the function but not a good general solution
2. if it's inside the `dash_ag_grid` then will this actually work? We would be redefining a callback every time the function is run, which might work ok but doesn't feel right[^1]. Possibly there's a way to avoid this and look to see if the callback is already defined though 🤔 

Am I missing something here? If not then next steps here would be to investigate 2, which would be useful anyway due to https://github.com/mckinsey/vizro/issues/109.

[^1]: we already do this every Dash page navigation callback since this rebuilds the page and hence e.g. the rangeslider callbacks. This already feels quite wrong but seems to work ok. See also https://github.com/mckinsey/vizro/issues/109#issuecomment-1768715486)

antonymilne:
The other thing I'm curious about and maybe @AnnMarieW can answer... Where does all the code in [this example](https://dash.plotly.com/dash-ag-grid/infinite-scroll#infinite-scroll-with-sort-and-filter) come from? How well tested and generic is this? i.e. could we safely apply it to any dataframe, or is it really just suitable for a few representative examples?

AnnMarieW:
> The other thing I'm curious about and maybe @AnnMarieW can answer... Where does all the code in [this example](https://dash.plotly.com/dash-ag-grid/infinite-scroll#infinite-scroll-with-sort-and-filter) come from? How well tested and generic is this? i.e. could we safely apply it to any dataframe, or is it really just suitable for a few representative examples?

The example is adapted from the example in the [AG Grid docs](https://www.ag-grid.com/react-data-grid/infinite-scrolling/#example-sorting-filtering-and-selection).  It's designed to be more generic than that example and should work in most cases.

This example was provided by @BSd3v who is the lead developer of dash-ag-grid.  Hey Bryan, do you have any other comments?

BSd3v:
Hello, All,

While this code is robust, and I believe it will work for most cases, it hasnt been tried on every single dataset, or even a broad spectrum of datasets for that matter. I fear that this could run into some of the enterprise features, such as filter sets and advanced filters.

I think if you want to go this route, you have it be a function that can be passed to the component if needed to get more advanced than this, where this first function is set as a default.

---

On a side note, the first post mentions pagination and being clientside. While this is the case in a default setting, it is possible to utilize pagination with an infinite row model as well. Just throwing that out there. 😄

antonymilne:
Thanks so much for chipping in @AnnMarieW and @BSd3v, very helpful! We are indeed considering doing pagination with the infinite row model and just wondering whether it would be a good idea to build the code in https://dash.plotly.com/dash-ag-grid/infinite-scroll#infinite-scroll-with-sort-and-filter into Vizro.

There are other technical problems on the Vizro side here so not sure whether we'll actually end up doing it but would be good to know in principle whether you would mind us doing so @BSd3v? 🙂 

> I fear that this could run into some of the enterprise features, such as filter sets and advanced filters.

This is a very good point and something that had occurred to me too. We'll need to make sure we don't clash with them.

> I think if you want to go this route, you have it be a function that can be passed to the component if needed to get more advanced than this, where this first function is set as a default.

This should indeed be possible through a user defining their own custom AgGrid function, although I imagine in practice not many people would.

huong-li-nguyen:
For my own tracking - blocked until this is solved: https://github.com/McK-Internal/vizro-internal/issues/1100

huong-li-nguyen:
Is this still relevant based on the new speed optimisations for the Tables and Graph? :)

petar-qb:
> Is this still relevant based on the new speed optimisations for the Tables and Graph? :)

Only a half of the "ideally fast ag_grid" problem is solved with the https://github.com/mckinsey/vizro/pull/644.

The other half involves enabling server-side pagination e.g. by enabling `infinite_scroll_ag_grid` figure function, and this is blocked until we investigate already mentioned -> https://github.com/McK-Internal/vizro-internal/issues/1100

PR #523, opened 2024-06-13
[Tidy] Update Graph theme using clientside_callback

petar-qb:
## Description
The PR addresses two issues:
- [#995](https://github.com/McK-Internal/vizro-internal/issues/995) -> "Investigate whether we need to configure overlay_style differently or have to remove it",
- [#74](https://github.com/McK-Internal/vizro-internal/issues/74) -> "Rewrite the 'update_graph_theme' callback into client-side callback"

After a little investigation into the overlay style issue, I found that removing it would be a quick and easy fix for the unlovely grey background in the `graphs` that appears when the theme is changed. However, there is a better solution which idea comes from one old ticket [#74](https://github.com/McK-Internal/vizro-internal/issues/74) (that also is planned to be addressed in the actions v2 (Unstaged phase) process). 

Here are some of the benefits that this PR brings:
- There is no more unlovely grey background in the graphs when the theme is changed.
- The way how the "theme changing system" works for graphs is aligned with the way how it already works for AgGrid and Dashboard (clientside_callback). 
- Changing theme process is faster when the app is deployed.
- There are three http requests to the server when a new page is opened (main branch). With these changes there will be only two (which is the ideal case I was always trying to achieve):
  1. Dash pages request and
  2. on_page_load request.

## Screenshot
Download videos and see visual difference: 

**from the main branch:**
https://github.com/mckinsey/vizro/assets/108530920/fd0905a7-80e5-4a76-bf78-d1fb11632a9c

**from the tidy/update_graph_theme_clientside branch:**
https://github.com/mckinsey/vizro/assets/108530920/6baec66f-b1d7-4d72-af20-bc341c7ab502

## Notice


vizro-core/examples/_dev/app.py
```diff
@@ -5,26 +5,41 @@
 from vizro import Vizro
 from vizro.tables import dash_ag_grid, dash_data_table
 
-df = px.data.gapminder()
+NUMBER_OF_COMPONENTS = 64
```
huong-li-nguyen, line 8: Not sure if this is related to the large number of components, but when I run the dev example and change the themes it gives me a bunch of `id` errors 🤔  Could you double-check?

Screenshot 2024-06-13 at 13 57 28

huong-li-nguyen (commented):
Tremendous job @petar-qb - thanks for doing this! 🚀  Really satisfying to see things getting cleaned up and I think this is a huge improvement! 😄 

I got a couple of id errors when I run the dev example - not sure if it's just my environment, but could you double-check? I definitely checked and my hatch environment did sync.

maxschulz-COL:
I started reviewing, but for me the theme switch does not work at all now for figures, see below screenshot:
image

vizro-core/src/vizro/static/js/models/dashboard.js
```diff
@@ -12,10 +12,25 @@ export function _update_ag_grid_theme(checked) {
...
+  Plotly.relayout(plotly_graph, { template: theme_to_apply });
+
+  return dash_clientside.no_update;
+}
+
 export function _collapse_nav_panel(n_clicks, is_open) {
   if (!n_clicks) {
     /* Automatically collapses left-side if xs and s-devices are detected*/
-    if (window.innerWidth < 576 || window.innerHeight < 576) {
+    if (window.innerWidth < 176 || window.innerHeight < 176) {
```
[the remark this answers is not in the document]
petar-qb: I usually decrease this size number while developing so the navbar doesn't collapse when the chrome dev tools is opened, and I forgot to revert it. 😄

vizro-core/examples/_dev/app.py
petar-qb: Ahh good catch. Does it work now?

maxschulz-COL (approved):
⭐ Amazing @petar-qb  - love that you took a simple fix and did some major refactoring!

Have only one small question for my own understanding!

vizro-core/examples/_dev/app.py
maxschulz-COL: For me yes!

maxschulz-COL:
One other question out of interest: I think previously we could change the graph theme mid animation (e.g. on the animated gapminder tutorial). That should still be possible right, because we are just changing the layout, not the entire object?

petar-qb:
> One other question out of interest: I think previously we could change the graph theme mid animation (e.g. on the animated gapminder tutorial). That should still be possible right, because we are just changing the layout, not the entire object?

You're right. The following JS code (that's added in this PR) enables exactly this `Plotly.relayout(plotly_graph, { template: theme_to_apply });`.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -142,7 +153,8 @@ def build(self):
         return html.Div(
             id="dashboard-container",
             children=[
-                html.Div(vizro.__version__, id="vizro_version", hidden=True),
+                html.Div(id="vizro_version", children=vizro.__version__, hidden=True),
+                dcc.Store(id="vizro_themes", data={"dark": themes.dark, "light": themes.light}),
```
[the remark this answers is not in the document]
petar-qb: Yes, we are now storing two entire theme templates on the client side. However (as we can see from the screenshot below), this only uses about 7KB of memory. This really shouldn't slow down the first load.

I just tested it and didn't notice (looking at the dev tools network timeouts) any difference in the initial loading time between running the same app from the main and the feature branch.

image

vizro-core/examples/_dev/app.py
petar-qb: FYI: The following change (in the `Graph.build clientside_callback`) makes it to work properly now 😄 
```python
            # Output here to ensure that the callback is only triggered if the graph exists on the currently open page.
            output=[Output(self.id, "figure")],
```

huong-li-nguyen (approved):
LGTM 👍

antonymilne (commented):
This is an amazing change, thank you for doing it! So nice to get all this stuff done 👏 

I just have a few questions but nothing big.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -116,6 +116,14 @@ def _filter_interaction(
 
     @_log_call
     def build(self):
+        clientside_callback(
```
antonymilne, line 119: So now for the example app with 64 graphs when you change theme there's 64 clientside callbacks that execute rather than 1 serverside callback. Performance looks great judging by the videos, but I'm curious  if you know how they're actually executed on the clientside - is it sequentially or in parallel?

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -116,6 +116,14 @@ def _filter_interaction(
...
+            ClientsideFunction(namespace="clientside", function_name="update_graph_theme"),
+            # Output here to ensure that the callback is only triggered if the graph exists on the currently open page.
+            output=[Output(self.id, "figure")],
+            inputs=[Input("theme_selector", "checked"), State("vizro_themes", "data"), State(self.id, "id")],
```
antonymilne, line 123: Can we instead do this by passing `State(self.id, "figure")` and then no need for the `document    .getElementById(graph_id)` in the callback function?

It would also be really nice if it were possible to actually use the `Output(self.id, "figure")` rather than needing to do a `no_update` - would that be possible if we passed in the graph itself as `State` rather than  just the id?

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -1,7 +1,7 @@
 import logging
```
antonymilne, line 1: Putting comment here because I can't put it elsewhere... Can we get rid of the `def _update_theme` now?
1. What happens if you remove the function altogether? Do we get flickering or is it ok?
2. Failing that, let's put it in `__call__` directly - the only reason it was done as a separate function before was for reuse, but now it's only used once

vizro-core/src/vizro/models/_components/graph.py
petar-qb: Testing performances was exactly the reason why I've added 64 graphs.

Javascript is executed in a single thread so it means that everything is executed sequentially (you can literally see how graphs are executed one by one with a debugger). I was aware of that, so I tested the example with 64 graphs. However, even with 64 graphs, everything happens invisibly fast. 😄 

(There is maybe an option to return the Promise from the clientside_callback, but there is nothing "awaitable" (as far as I know) in the function that could be optimised using the concurrent programming and the famous async/await syntax).

vizro-core/src/vizro/models/_components/graph.py
petar-qb: Ideally, we should only affect the `layout` of the figure (not the `data` of the figure). One reason is performance (if the chart displays large data) and the other is that we don't want to redraw the entire chart because we want to allow the chart theme to change properly even when the chart is in the middle of an animation (e.g. on an animated gapminder tutorial). And, because `Patch` object is not available in the clientside_callbacks, I handled it using the `Plotly.relayout`.

vizro-core/src/vizro/models/_components/graph.py
petar-qb: Removing theme setting in `Graph.__call__` (and adding `"graph.figure"` as a clientside_callback `_update_graph_theme` dcc.Input) causes graphs to flicker. However, let me do a little PR by moving the `_update_theme` directly to `Graph.__call__`.

antonymilne: Cool, thank you for checking and raising the PR! 👍

vizro-core/src/vizro/models/_components/graph.py
antonymilne: Cool, that makes sense and is what I suspected.

Is it possible to rewrite the `State(self.id, "id")` to use `State(self.id, "figure")` directly instead of doing the `document.getElementById(graph_id)`? No problem if not, but if it works then you could add it to #536

petar-qb: I don't think this could work. See the `Plotly.relayout` documentation -> https://plotly.com/javascript/plotlyjs-function-reference/#plotlyrelayout. Its first argument has to be or DOM node or its ID (and the figure is just a JSON obj sent from the server). So, as we don't have the ID for this DOM node, we have to fetch the entire DOM node with the:
```js
  const plotly_graph = document
    .getElementById(graph_id)
    .querySelector(".js-plotly-plot");
```

PR #530, opened 2024-06-17
[QA] Update `tenacity` dependency to fix failing unit tests

Anna-Xiong:
## Description

Fix failing vizro ai unit tests:
Set tenacity upper bound to avoid issue described in https://github.com/langchain-ai/langchain/issues/22972.

## Notice

### DIFF
--- a/vizro-ai/pyproject.toml
+++ b/vizro-ai/pyproject.toml
@@ -17,6 +17,7 @@ dependencies = [
   "pandas",
   "tabulate",
   "openai>=1.0.0",
+  "tenacity<=8.3.0",  # TODO update when tenacity 8.4.0 is compatible with langchain
   "langchain>=0.1.0, <0.3.0",  # TODO update all LLMChain class and remove upper bound
   "langchain-openai",
   "python-dotenv>=1.0.0",  # TODO decide env var management to see if we need this
--- a/vizro-ai/pyproject.toml
+++ b/vizro-ai/pyproject.toml
@@ -17,7 +17,7 @@ dependencies = [
   "pandas",
   "tabulate",
   "openai>=1.0.0",
-  "tenacity==8.4.1",  # TODO update when tenacity 8.4.0 is compatible with langchain
+  "tenacity=!=8.4.1",  # https://github.com/langchain-ai/langchain/issues/22972
   "langchain>=0.1.0, <0.3.0",  # TODO update all LLMChain class and remove upper bound
   "langchain-openai",
   "python-dotenv>=1.0.0",  # TODO decide env var management to see if we need this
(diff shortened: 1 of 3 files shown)


maxschulz-COL (approved):
Thanks!

antonymilne:
What is the version that actually causes a problem here? If it's 8.4.0 then a much better version specifier would be `!=8.4.0`, so that when a fixed version, say 8.4.1, comes out then you don't need to change the requirement again.

Anna-Xiong:
> What is the version that actually causes a problem here? If it's 8.4.0 then a much better version specifier would be `!=8.4.0`, so that when a fixed version, say 8.4.1, comes out then you don't need to change the requirement again.

It is the newly released version 8.4.0(  only released a few hours ago), they now have even newer version 8.4.1 i can try to see if it resolves the issue, otherwise probably let's pin under 8.4.0?

vizro-ai/snyk/requirements.txt
```diff
@@ -1,6 +1,7 @@
 pandas
 tabulate
 openai>=1.0.0
+tenacity==8.4.1
```
huong-li-nguyen, line 4: Seems like the latest version seems to be working 🥳  Do you want to apply the suggested change from @antonymilne? Should be good to go then 👍 

```suggestion
tenacity!=8.4.1
```

vizro-ai/pyproject.toml
```diff
@@ -17,6 +17,7 @@ dependencies = [
   "pandas",
   "tabulate",
   "openai>=1.0.0",
+  "tenacity==8.4.1",  # TODO update when tenacity 8.4.0 is compatible with langchain
```
huong-li-nguyen, line 20:
```suggestion
  "tenacity=!=8.4.1",  # https://github.com/langchain-ai/langchain/issues/22972
```

huong-li-nguyen (approved):

antonymilne (changes requested):
If the tenacity 8.4.1 has been released and fixes the problem then there should be no need for any changes at all here and we shouldn't add a transitive dependency to our requirements. Anyone doing `pip install vizro-ai` will get the latest version of tenacity unless there's some weird conflict that we know of that forces them into the problematic 8.4.0. So best to just not do anything here I think and close the PR.

huong-li-nguyen:
> If the tenacity 8.4.1 has been released and fixes the problem then there should be no need for any changes at all here and we shouldn't add a transitive dependency to our requirements. Anyone doing `pip install vizro-ai` will get the latest version of tenacity unless there's some weird conflict that we know of that forces them into the problematic 8.4.0. So best to just not do anything here I think and close the PR.

Ah yes, main seems to be fixed now as well!

PR #540, opened 2024-06-21
[Docs] Fix warnings and suggestions from `vale`

huong-li-nguyen:
## Description
- Fix all warnings and suggestion from `vale` that were not visible until an error was introduced

---

@stichbury - returning to your comment on why certain things were not raised on Vale in the last PRs. I just quickly checked, and there are two relevant factors:

- Previously, `e.g.` was considered an error, but `e.g` was not -> I've made it stricter in this PR now
- It seems like `vale` as a pre-commit hook will always return "passed" as long as there is no new error introduced (3 different levels: error, warning, suggestion). 
 
This means that over time, `vale` probably raised a couple of suggestions and warnings, but they were never shown to us unless we simultaneously introduced a `vale` error. To find all of the warnings/suggestions, I intentionally added an error (adding `e.g. this` in any of the docs) and then ran the linter. Only then were all of the warnings/suggestions visible. I've fixed them all in this PR now, but it's essential to be aware of that limitation. 

@stichbury - I couldn't find a quick way of configuring that warnings/suggestions are also errors for us. Do you have an idea? Otherwise, it's good to be aware of the restriction above 🔝  It might be worth considering switching to [github actions for vale](https://github.com/errata-ai/vale-action). It seems like there is a configuration option [fail_on_error](https://github.com/errata-ai/vale-action/issues/26)  that might does that, but one would have to test it out, because I am not sure if it behaves similarly to the pre-commit hook. So I'll leave that for someone to test out next sprint 👍 

## Notice

- [x] I acknowledge and agree that by checking this box and clicking "Submit Pull Request":

### DIFF
--- a/.vale/styles/Microsoft/Foreign.yml
+++ b/.vale/styles/Microsoft/Foreign.yml
@@ -7,7 +7,7 @@ nonword: true
 action:
   name: replace
 swap:
-  '\b(?:eg|e\.g\.)[\s,]': for example
-  '\b(?:ie|i\.e\.)[\s,]': that is
+  '\b(?:eg|e\.g\.|e.g)[\s,]': for example
+  '\b(?:ie|i\.e\.|i.e)[\s,]': that is
   '\b(?:viz\.)[\s,]': namely
   '\b(?:ergo)[\s,]': therefore
--- a/.vale/styles/Microsoft/Foreign.yml
+++ b/.vale/styles/Microsoft/Foreign.yml
@@ -7,7 +7,7 @@ nonword: true
 action:
   name: replace
 swap:
-  '\b(?:eg|e\.g\.|e.g)[\s,]': for example
-  '\b(?:ie|i\.e\.|i.e)[\s,]': that is
+  '\b(?:eg|e\.g\.|e\.g)[\s,]': for example
+  '\b(?:ie|i\.e\.|i\.e)[\s,]': that is
   '\b(?:viz\.)[\s,]': namely
   '\b(?:ergo)[\s,]': therefore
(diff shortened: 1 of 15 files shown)


.vale/styles/Microsoft/Foreign.yml
```diff
@@ -7,7 +7,7 @@ nonword: true
 action:
   name: replace
 swap:
-  '\b(?:eg|e\.g\.)[\s,]': for example
-  '\b(?:ie|i\.e\.)[\s,]': that is
+  '\b(?:eg|e\.g\.|e.g)[\s,]': for example
+  '\b(?:ie|i\.e\.|i.e)[\s,]': that is
```
petar-qb, line 11:
```suggestion
  '\b(?:eg|e\.g\.|e\.g)[\s,]': for example
  '\b(?:ie|i\.e\.|i\.e)[\s,]': that is
```

petar-qb (commented):
🧹 😍

stichbury:
Aha! That makes sense, thanks for working out what was happening and resolving those issues.

> I couldn't find a quick way of configuring that warnings/suggestions are also errors for us. Do you have an idea? 

We could just make the `level` setting of the Vale styles that are important into errors rather than warnings/suggestions? It's not non-trivial but it's relatively easy to do for the main culprits that we occur regularly such as spelling, capitalisation and abbreviation. 

Of course, by doing that, it then means that all reports have to be fixed even where they are valid cases (this may mean simply switching vale on and off with comments around the section in question). 

Having a way to see all suggestions clearly at all times, and then decide whether to act on them is the ideal. I'll discuss with @antonymilne in the next sprint as we have a ticket to look at this. For now though, you've cleared up what was there, and that's awesome. Thank you 🙏

stichbury (approved):
Nice modifications to make Vale happy, thank you! 🏆

petar-qb (approved):

PR #547, opened 2024-06-25
[Feat] Improve validation of CapturedCallable

antonymilne:
## Description

~@petar-qb @huong-li-nguyen @maxschulz-COL would be great if one of you could do a quick 10 mins manual test just to check it works as expected. I've done pretty thorough testing and have lots of unit tests but would like to be extra sure I haven't broken something that was working well before (e.g. features dashboard). Just one person checking it in addition to me should be fine.~ Done, thank you @huong-li-nguyen! 

### Improved error messages

**Before**: CapturedCallable mode mismatch. Expected graph but got table.
**Now** (small improvement): CapturedCallable was defined with @capture('table') rather than @capture('graph') and so is not compatible with the model.

**Before**: You must provide a valid CapturedCallable object. If you are using a plotly express figure, ensure that you are using `import vizro.plotly.express as px`. If you are using a table figure, make sure you are using `from vizro.tables import dash_data_table`. If you are using a custom figure or action, that your function uses the @capture decorator.
**Now** (big improvement): Invalid CapturedCallable. Supply a function imported from vizro.plotly.express or defined with decorator @capture('graph')

### Nice refactor
https://github.com/mckinsey/vizro/pull/340 introduced `mode` to `CapturedCallable` to ensure that e.g. you don't accidentally supply a dash_data_table to AgGrid. This was done using a reusable validator.

While working on this I realised things would be much cleaner and we can generate better error messages if we did the validation _inside `CapturedCallable` validators itself_, so I moved it. The control flow and responsibility of `CapturedCallable`'s validation is now much clearer. The only `CaptureCallable` related validator that hasn't moved to `CaptureCallable` and so remains duplicated across `Figure/Graph/Table/AgGrid` is`_process_callable_data_frame` which would take more thought before moving so I haven't touched.

## Notice


vizro-core/changelog.d/20240621_132255_antony.milne_improve_captured_callable_validation.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Improve validation error messages for `CapturedCallable` ([#547](https://github.com/mckinsey/vizro/pull/547)).
```
antonymilne, line 22:
```suggestion
- Improve validation error messages for `CapturedCallable`. ([#547](https://github.com/mckinsey/vizro/pull/547))
```

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -32,7 +32,7 @@ class Action(VizroBaseModel):
 
     """
 
-    function: CapturedCallable = Field(..., import_path=vizro.actions)
+    function: CapturedCallable = Field(..., import_path=vizro.actions, mode="action")
```
antonymilne, line 35: This `mode` mirrors `type` for the models that have `type` available.

vizro-core/src/vizro/models/types.py
```diff
@@ -203,19 +206,43 @@ def _parse_json(
             raise ValueError(f"_target_={function_name} cannot be imported from {import_path.__name__}.") from exc
 
         # All the other items in figure are the keyword arguments to pass into function.
-        function_kwargs = callable_config
+        function_kwargs = captured_callable_config
 
         # It would seem natural to return cls(function, **function_kwargs) here, but the function is already decorated
         # with @capture, and so that would return a nested CapturedCallable.
-        captured_callable = function(**function_kwargs)
```
antonymilne, line 210: Having this logic duplicated between here and lines 177 onwards was always very unsatisfactory. Now I've detangled the validation into multiple validators it can be removed.

vizro-core/tests/unit/vizro/models/_components/test_ag_grid.py
```diff
@@ -47,17 +49,25 @@ def test_mandatory_figure_missing(self):
         with pytest.raises(ValidationError, match="field required"):
             vm.AgGrid()
 
-    def test_wrong_captured_callable(self, standard_dash_table):
-        with pytest.raises(ValidationError, match="CapturedCallable mode mismatch"):
-            vm.AgGrid(figure=standard_dash_table)
-
-    def test_failed_ag_grid_with_no_captured_callable(self, standard_go_chart):
-        with pytest.raises(ValidationError, match="must provide a valid CapturedCallable object"):
+    def test_captured_callable_invalid(self, standard_go_chart):
```
antonymilne, line 52: These tests for each model aren't strictly required now we have tests in `test_types.py` but I thought safest to leave them here and have somewhat duplicated tests, since these are better integration tests than the ones in `test_types`.

huong-li-nguyen:
Hello, first of all - this all sounds amazing! 👏 🚀 

I've been manually testing all the demo dashboards (features/demo/kpi) and their yaml version if it exists and all works fine. I've also been playing around with the dev example and find a couple of behaviours confusing. There are two cases where I would have expected error messages, but there are currently none. I might also not have the right understanding here, so just to clarify:

**Case 1: Example with no capture call** - this currently runs through without any errors, but the update_layout has no effect. Here I would have expected that it throws an error here that I need to use the @capture("graph")? 

```
def f(data_frame):
    """Blah"""
    return px.bar(data_frame, "Actual").update_layout(title="x")


page = vm.Page(
    title="KPI Indicators",
    components=[
        vm.Graph(
            figure=f(df_kpi),
        ),
        vm.Figure(figure=kpi_card(data_frame=df_kpi, value_column="Actual", title="KPI with value")),
    ],
)
```


**Case 2: Example with no function at all** - this currently runs through without any errors, but the update_layout has no effect. Here I would have expected that it throws an error here that I need to use the @capture("graph")?  I think this is something users might commonly try. Actually we did have a user question on this and back then they though all the update_layout calls are not compatible with Vizro, so this might be something we do have to fix?

```
page = vm.Page(
    title="KPI Indicators",
    components=[
        vm.Graph(
            figure=px.bar(df_kpi, "Actual").update_layout(title="x"),
        ),
        vm.Figure(figure=kpi_card(data_frame=df_kpi, value_column="Actual", title="KPI with value")),
    ],
)
```

vizro-core/tests/unit/vizro/models/_components/test_ag_grid.py
huong-li-nguyen: At some point when we think about subclassing AgGrid, Table and Graph from Figure, we could probably kill all these duplicated tests and just test for these in Figure right?

antonymilne: Probably yes, but I feel like there is also value in having good integration tests that run more "real" examples. Otherwise our tests are probably a bit too theoretical and will test the underlying mechanism of `Figure` but not whether it actually works in practice with `AgGrid` and there's a bit of a gap in testing coverage where things could conceivably go wrong and not be caught.

So if we remove them then we should make sure that there's good coverage of these models in our integration tests and/or QA tests.

antonymilne:
Thank you for trying it all out! The two cases you raise are great ones to understand 😀 

First of all, your two cases are really exactly the same. When there's no `@capture` decorator, calling `f(df_kpi)` is not doing anything other than producing `px.bar(data_frame, "Actual").update_layout(title="x")`. There's nothing magical happening here - it's just calling a normal Python function and getting the value. It doesn't matter where that value comes from.

What happens next depends on what your `px` import is:
* if it's `plotly.express` then the value supplied to `figure` will have type `go.Figure` and this should raise an error telling you to use `vizro.plotly.express` instead
* if it's `vizro.plotly.express` then the value supplied to `figure` will have type `_DashboardReadyFigure` and no error is raised

(please do check the above works for you as I expect it to depending on what your `px` import is)

When `px` is `vizro.plotly.express`, what happens is:
* `px.bar` produces a `_DashboardReadyFigure` that is a subclass of `go.Figure` and has a `_captured_callable` attribute that contains the thing the dashboard actually needs to function
* `update_layout` applies to the `go.Figure` part of this but does nothing to the `_captured_callable` attribute
* hence if you put this code in a Jupyter notebook it will show the graph with updated layout as you would expect
* but if you put it into a `vm.Graph` model the new `CapturedCallable._extract_from_attribute` extracts the `_captured_callable` attribute that doesn't have the updated layout attached in any way, so it doesn't work as expected in a dashboard

This is indeed very confusing and exactly the point about post-figure updates we discussed in https://github.com/McK-Internal/vizro-internal/issues/937 and that meeting a couple of weeks ago. Ideally we would still solve this case or raise a good error message but it's not a super quick and easy fix so we didn't do anything about it yet.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -37,15 +37,14 @@ class Graph(VizroBaseModel):
     """
 
     type: Literal["graph"] = "graph"
-    figure: CapturedCallable = Field(..., import_path=px)
+    figure: CapturedCallable = Field(..., import_path=px, mode="graph")
```
petar-qb, line 40: Should we add the `description` here as we did for the other Vizro figure-like object?

vizro-core/src/vizro/models/types.py
petar-qb, line 1: Add `capture("figure")` in the exception message at the line 357.

vizro-core/src/vizro/models/types.py
```diff
@@ -153,44 +153,47 @@ def _arguments(self):
...
+        # captured_callable could be _SupportsCapturedCallable, CapturedCallable, dictionary from JSON/YAML or
+        # invalid type at this point. Begin by parsing it from JSON/YAML if applicable:
+        yield cls._parse_json
+        # At this point captured_callable is _SupportsCapturedCallable, CapturedCallable or invalid type. Next extract
+        # it from _SupportsCapturedCallable if applicable:
+        yield cls._extract_from_attribute
+        # At this point captured_callable is CapturedCallable or invalid type. Check it is in fact CapturedCallable
+        # and do final checks:
+        yield cls._check_type
+
```
petar-qb, line 174: It's much cleaner now 🥇.

petar-qb (approved):
Nicely done 🧹 👍

huong-li-nguyen:
> This is indeed very confusing and exactly the point about post-figure updates we discussed in [McK-Internal/vizro-internal#937](https://github.com/McK-Internal/vizro-internal/issues/937) and that meeting a couple of weeks ago. Ideally we would still solve this case or raise a good error message but it's not a super quick and easy fix so we didn't do anything about it yet.

Got it - thanks for clarifying! 

Yeah, I see. Let's pick this up next sprint then!

huong-li-nguyen (approved):
LGTM 👍 Thanks for doing this! 🧹

vizro-core/src/vizro/models/_components/graph.py
antonymilne: Yes, these were very inconsistent before actually! I've just fixed it: [d452ee4](https://github.com/mckinsey/vizro/pull/547/commits/d452ee475f2cd050b1269280efc647fb1b1eeb6e).

vizro-core/src/vizro/models/types.py
antonymilne: Done in [d452ee4](https://github.com/mckinsey/vizro/pull/547/commits/d452ee475f2cd050b1269280efc647fb1b1eeb6e).

Issue #557, opened 2024-07-01
Parameters as dictionary of label-value mappings not working in custom chart when "ALL" selected


GiulioMorinaQB:
### Description

When using a custom figure and setting options for a parameter using a dictionary, when "ALL" is selected no data is displayed.

### Which package?

vizro

### Package version

0.1.18

### Python version

3.9

### OS

MacOS

### How to Reproduce

```
import pandas as pd
from vizro import Vizro
import vizro.plotly.express as px
import vizro.models as vm
from vizro.models.types import capture

df_stocks = px.data.stocks(datetimes=True)

df_stocks_long = pd.melt(
    df_stocks,
    id_vars='date',
    value_vars=['GOOG', 'AAPL', 'AMZN', 'FB', 'NFLX', 'MSFT'],
    var_name='stocks',
    value_name='value'
)

@capture("graph")
def vizro_plot(data_frame, stocks_selected, **kwargs):
    return px.line(data_frame[data_frame["stocks"].isin(stocks_selected)], **kwargs)

df_stocks_long['value'] = df_stocks_long['value'].round(3)

page = vm.Page(
    title="My first page",
    components=[
        vm.Graph(id="my_graph", figure=vizro_plot(data_frame=df_stocks_long, stocks_selected=list(df_stocks_long["stocks"].unique()), x="date", y="value", color="stocks")),
    ],
    controls=[
        vm.Parameter(
            targets=["my_graph.stocks_selected"],
            selector=vm.Dropdown(
                options=[{'label': s, 'value': s} for s in df_stocks_long["stocks"].unique()],
            ),
        ),
        vm.Filter(column="value"),
        vm.Filter(column="date"),
    ],
)

dashboard = vm.Dashboard(pages=[page])

Vizro().build(dashboard).run()
```

### Output

<img width="1962" alt="Screenshot 2024-07-01 at 15 36 41" src="https://github.com/mckinsey/vizro/assets/51332198/06f77b14-2f18-4864-b1e9-b8d2719b93e8">

### Code of Conduct

- [X] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

lingyielia:
Hi @GiulioMorinaQB , thanks for raising the bug report! We are investigating it and will come back to you once we find the root cause. Meanwhile, you can config your parameter and custom chart in two ways:
Approach 1:
```
import pandas as pd
from vizro import Vizro
[...]

PR #562, opened 2024-07-03
[Bug] Fix inconsistent return types in selector values

huong-li-nguyen:
## Description
Previously if you've specified the options of a selector as a `List[Dict]` inside a `Parameter`, the value was returned as a List[Dict] when "ALL" was selected (introduced by us), but would return a List of values only when selecting other values from the multi selector (dash default). This could lead to issues when using custom charts, as it's not immediately transparent for the user what the return type is, see https://github.com/mckinsey/vizro/issues/557. 

- Fixes inconsistent return types in selectors by always returning a List (similar to Dash)

FYI: @lingyielia @petar-qb @antonymilne 

## Notice

  - I certify that (a) this contribution is my original creation and/or (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
  - I have not directly or indirectly referenced individuals, products or companies in any commits.

### DIFF
--- a/vizro-core/changelog.d/20240703_091955_huong_li_nguyen_fix_selector_return_type.md
+++ b/vizro-core/changelog.d/20240703_091955_huong_li_nguyen_fix_selector_return_type.md
@@ -37,7 +37,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 ### Fixed
 
-- Ensure that selectors always return a list of values. ([#562](https://github.com/mckinsey/vizro/pull/562))
+- Ensure that categorical selectors always return a list of values. ([#562](https://github.com/mckinsey/vizro/pull/562))
 
 <!--
 ### Security
--- a/vizro-core/changelog.d/20240703_091955_huong_li_nguyen_fix_selector_return_type.md
+++ b/vizro-core/changelog.d/20240703_091955_huong_li_nguyen_fix_selector_return_type.md
@@ -1,0 +1,47 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Ensure that selectors always return a list of values. ([#561](https://github.com/mckinsey/vizro/pull/561))
+
+<!--
+### Security
+
+- A bullet item for the Security category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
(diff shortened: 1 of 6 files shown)


vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -138,12 +138,19 @@ def _get_parametrized_config(target: ModelID, ctd_parameters: List[CallbackTrigg
...
-            selector_value = selector.options
+
+            # Even if options are provided as List[Dict], the Dash component only returns a List of values.
+            # So we need to ensure that we always return a List only as well to provide consistent types.
+            if all(isinstance(i, dict) for i in selector.options):
+                selector_value = [value["value"] for value in selector.options]
+            else:
+                selector_value = selector.options
+
         selector_value = _validate_selector_value_none(selector_value)
```
lingyielia, line 154: what are the possible types of `selector_value` into function `_validate_selector_value_none`? Would it only be `NONE_OPTION` or `list` now?

vizro-core/examples/_dev/app.py
```diff
@@ -1,64 +1,49 @@
...
-    value_format: str = "{value}",
-    agg_func: str = "sum",
-    title: Optional[str] = None,
-    icon: Optional[str] = None,
-) -> dbc.Card:  # (2)!
-    """Creates a custom KPI card."""
-    title = title or f"{agg_func} {value_column}".title()
-    value = data_frame[value_column].agg(agg_func)
+@capture("graph")
+def vizro_plot(data_frame, stocks_selected, **kwargs):
```
lingyielia, line 20: what would be the best way to inform users that in this case, the `stocks_selected` should be a list? Because from user perspective, when they do 
```
options=[{"label": s, "value": s} for s in df_stocks_long["stocks"].unique()]
```
It's not intuitive to think `stocks_selected` should be a list right?

lingyielia (approved):
Thanks Li! I agree with should follow Dash default in this case. Approved with 2 comments.

vizro-core/examples/_dev/app.py
huong-li-nguyen: I don't think we need to inform them - given that we stick with Dash default return type now, there should be no confusion hopefully. Other than that, there is always our logging message when you run the app in DEBUG mode, where you can also see the inputs/outputs of the actions. And if that doesn't help one can always check the input/outputs inside the custom chart function in DEBUG mode :)

Screenshot 2024-07-03 at 18 28 12

vizro-core/src/vizro/actions/_actions_utils.py
huong-li-nguyen: No, it can also be a single value. The typing of that function is actually correct now, so  

`def _validate_selector_value_none(value: Union[SingleValueType, MultiValueType]) -> ValidatedNoneValueType:`

where:

```
SingleValueType = Union[StrictBool, float, str, date]
MultiValueType = Union[List[StrictBool], List[float], List[str], List[date]]
```

vizro-core/examples/_dev/app.py
huong-li-nguyen: I've changed the depth level of the logging messages now, so it's visible there:

Screenshot 2024-07-03 at 18 46 22

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -138,12 +138,19 @@ def _get_parametrized_config(target: ModelID, ctd_parameters: List[CallbackTrigg
...
+        selector_value = ctd["value"]
+
         if hasattr(selector_value, "__iter__") and ALL_OPTION in selector_value:  # type: ignore[operator]
             selector: SelectorType = model_manager[ctd["id"]]
-            selector_value = selector.options
+
+            # Even if options are provided as List[Dict], the Dash component only returns a List of values.
+            # So we need to ensure that we always return a List only as well to provide consistent types.
+            if all(isinstance(i, dict) for i in selector.options):
+                selector_value = [value["value"] for value in selector.options]
```
antonymilne, line 150:
```suggestion
            if all(isinstance(option, dict) for option in selector.options):
                selector_value = [option["value"] for option in selector.options]
```

vizro-core/changelog.d/20240703_091955_huong_li_nguyen_fix_selector_return_type.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Ensure that selectors always return a list of values. ([#562](https://github.com/mckinsey/vizro/pull/562))
```
antonymilne, line 40:
```suggestion
- Ensure that categorical selectors always return a list of values. ([#562](https://github.com/mckinsey/vizro/pull/562))
```

If I understand correctly, this is really only relevant for dropdown and checklist?

antonymilne (approved):
Amazing work all round @GiulioMorinaQB for the great bug report and @lingyielia and @huong-li-nguyen for the very speedy temporary and permanent fix! 💯 

Just to check: @huong-li-nguyen  does the new test fail before your fix?

huong-li-nguyen:
> Just to check: @huong-li-nguyen does the new test fail before your fix?

Yes, of course! That's how I checked that the test was working 😄

PR #563, opened 2024-07-03
[Docs] Update issue templates

stichbury:
## Description
Maybe not exactly docs, more like a comms issue, to make it simpler to report bugs, request features and ask questions.

I've addressed #1098 from our internal issues queue to edit the issues templates and simplify them a little.

## Notice

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -6,8 +6,8 @@ body:
   - type: markdown
     attributes:
       value: >-
-        Hi there! Thank you so much for taking the time to fill in this bug report. 
-        As we have limited triage capacity, we'd appreciate it if you could search on Vizro's 
+        Hi there! Thank you so much for taking the time to fill in this bug report.
+        As we have limited triage capacity, we'd appreciate it if you could search on Vizro's
         existing issues prior to submission to avoid any duplicates. Thanks!
 
   - type: textarea
@@ -22,7 +22,7 @@ body:
     attributes:
       label: Expected behavior
       description: >-
-        A clear and concise description of what you expected to happen. 
+        A clear and concise description of what you expected to happen.
   - type: dropdown
     attributes:
       label: Which package?
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -6,12 +6,9 @@ body:
   - type: markdown
     attributes:
       value: >-
-        Hi there! Thank you so much for taking the time to fill in this bug form. It will be
-        a valuable contribution to Vizro.
-
-        We'd appreciate it if you could search on Vizro's existing issues prior to filing
-        a bug report to avoid any duplicate tickets, as we have limited maintainer capacity to triage
-        them. Thanks!
+        Hi there! Thank you so much for taking the time to fill in this bug report. 
+        As we have limited triage capacity, we'd appreciate it if you could search on Vizro's 
+        existing issues prior to submission to avoid any duplicates. Thanks!
 
   - type: textarea
     attributes:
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/general-question.yml
```diff
@@ -27,27 +24,12 @@ body:
...
       label: Which package?
       options:
         - vizro
         - vizro-ai
-  - type: input
-    attributes:
-      label: Package version
-  - type: input
-    attributes:
-      label: Python version
```
huong-li-nguyen, line 47: Do we ask for the package and python elsewhere? I think this might be useful to keep?

huong-li-nguyen (approved):
LGTM 👍 Have a minor comment, but ultimately don't mind if you remove it or not. But we do ask for the python and package version normally when we try to debug :)

stichbury:
> LGTM 👍 Have a minor comment, but ultimately don't mind if you remove it or not. But we do ask for the python and package version normally when we try to debug :)

I feel like we ask too much, but I do understand it's useful to have this information (though we haven't made it a requirement in the past). How about I include it as something they should tell us, so it's still a suggested thing to include, but not in a separate field? I'll make the commit, see what you think @huong-li-nguyen

.github/ISSUE_TEMPLATE/bug-report.yml
antonymilne, line 1: I would completely get rid of the following questions too:
* Expected behavior 
* Python version
* OS
* Output

antonymilne: And let's move "Which package" and "Version" to the top of the form also I think.

.github/ISSUE_TEMPLATE/general-question.yml
```diff
@@ -25,29 +22,14 @@ body:
...
       label: Code/Examples
       description: >-
-        Please post any supporting code around your question here.
-
-  - type: textarea
-    attributes:
-      label: Other information
-      description: >-
-        Please provide any attempts on answering the question, or any related information you already found here.
+        Please post any supporting code around your question here. Don't forget to include the Python version and package version, if it's relevant.
```
antonymilne, line 25:
```suggestion
        Please post any supporting code around your question here. 
```

Personally I agree with your original removal of this 🙂

.github/ISSUE_TEMPLATE/general-question.yml
antonymilne: I don't think it's relevant for a general question, just for bug reports (and even then we don't really care about Python version).

antonymilne (approved):
Thanks for doing this, it looks a lot better! I suggest we go even more drastic with this - see comments.

Re: package and version - I think this is only relevant on the bug report so would only include it there. Python version and OS have never been relevant AFAIK so not worth mentioning at all - we can always follow up asking someone what those are if we suspect that might be the problem.

maxschulz-COL (approved):
LGTM! I would wait merging this until MOnday for when we have the py.cafe tile - then I would actively refer to py.cafe, and suggest people post there code problem via py.cafe if that is relevant

Then we can just click on the link and see the issue!

maxschulz-COL:
> LGTM! I would wait merging this until MOnday for when we have the py.cafe tile - then I would actively refer to py.cafe, and suggest people post there code problem via py.cafe if that is relevant
> 
> Then we can just click on the link and see the issue!

@stichbury We now have the py.cafe tile - so maybe we can include that?

.github/ISSUE_TEMPLATE/bug-report.yml
stichbury: Done!

stichbury:
@maxschulz-COL I've added in a link to py.cafe in a couple of places. Please see what you think!

maxschulz-COL:
> @maxschulz-COL I've added in a link to py.cafe in a couple of places. Please see what you think!

Yes I think it's great, let's go!

Issue #567, opened 2024-07-04
Custom action targeting custom graph inputs / graph function parameters


szt-ableton:
### Question

Hey team,
I have a question regarding the use of custom actions.
I would like to be able to click on a cell on my AgGrid element, and for the value of the click to update a parameter in the function that builds my graph.
However, I am not sure how to target a parameter within a function that builds the Graph component.

Below is an example - where I am able to control the text in a card element ( as in a standard example published by you). 
Instead, I would like to control specific inputs to my function. To be more clear, basically, I would like this custom action to work as a parameter - being able to change the question input to my custom graph.

How is it possible to target custom graph inputs through custom actions? If that is not possible, is it possible to target the selected parameter somehow?

Thank you so much for the amazing work on this tool, and for the help in advance!

### Code/Examples

import vizro.models as vm
import vizro.plotly.express as px
from vizro import Vizro
from vizro.actions import filter_interaction
from vizro.models.types import capture
import pandas as pd
from vizro.tables import dash_ag_grid
import vizro.plotly.express as px

import json  # Testing

data = pd.read_csv('./develop/lsurvey.csv')
questions = pd.DataFrame(data['question'].drop_duplicates())
segments = pd.DataFrame(list(data[data['question_type']=='single_choice']['question'].unique()) + ['none'], columns = ['segment'])

df = px.data.iris()

@capture("action")
def my_custom_action(clickedcell: dict):
    """Custom action."""
    # text = json.dumps(clickedcell)

    # return lambda: f"You clicked on cell with value: {text}"  # Return an update function
    return (
        json.dumps(clickedcell["value"])
        if clickedcell
        else "Click on a point on the above graph(func)."
    )

@capture("graph")
def build_selected_figure(data_frame, question, segmentation):
    fig=px.scatter_matrix(
                     data_frame = data_frame
        , dimensions=["sepal_length", "sepal_width", "petal_length", "petal_width"]
        , color="species"
        , title = question
        ),
        
    return fig
    
page = vm.Page(
    title="Example of a custom action with UI inputs and outputs",
    path="custom_actions",
    id="custom_actions",
    layout=vm.Layout(
        grid=[[0, 1], [2, 2]],
        row_gap="25px",
    ),
    components=[
        vm.AgGrid(
            id="cell-selection-simple-click-callback",
            figure=dash_ag_grid(
                id="cell-selection-simple-click-callback-grid",
                data_frame=data,
                rowData=data.to_dict("records"),
[...]

PR #574, opened 2024-07-08
[Feat] Update `optionHeight` of `vm.Dropdown` dynamically

huong-li-nguyen:
## Description
- Update `optionHeight` of `vm.Dropdown` dynamically based on character length

## Screenshot
Screenshot 2024-07-08 at 18 59 17
Screenshot 2024-07-08 at 18 59 08
Screenshot 2024-07-08 at 18 59 13

## Notice

### DIFF
--- a/vizro-core/src/vizro/models/_components/form/dropdown.py
+++ b/vizro-core/src/vizro/models/_components/form/dropdown.py
@@ -71,6 +71,7 @@ class Dropdown(VizroBaseModel):
                     value=self.value if self.value is not None else default_value,
                     multi=self.multi,
                     persistence=True,
+                    optionHeight=32 if all(len(option) <= 37 for option in full_options) else 60,  # noqa: PLR2004
                     persistence_type="session",
                 ),
             ]
--- a/vizro-core/src/vizro/models/_components/form/dropdown.py
+++ b/vizro-core/src/vizro/models/_components/form/dropdown.py
@@ -73,7 +73,7 @@ class Dropdown(VizroBaseModel):
                     multi=self.multi,
                     persistence=True,
                     # 37 is the cut-off character length where the text inside the dropdown starts to wrap
-                    optionHeight=32 + 24 * math.floor(max(len(option) for option in full_options) / 37),
+                    optionHeight=32 + 24 * math.floor(max(len(str(option)) for option in full_options) / 37),
                     persistence_type="session",
                 ),
             ]
--- a/vizro-core/src/vizro/models/_components/form/dropdown.py
+++ b/vizro-core/src/vizro/models/_components/form/dropdown.py
@@ -1,3 +1,4 @@
+import math
 from typing import List, Literal, Optional, Union
 
 from dash import dcc, html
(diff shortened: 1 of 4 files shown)


huong-li-nguyen:
@antonymilne - regarding your question here: https://github.com/mckinsey/vizro/issues/567 - I do like the look better if all options are short or all are long. If it's a mixture of these two, we can only set one `optionHeight` which in this case should be the longer one. I would be fine implementing this, but generally I think our recommendation should be to reduce the length of labels if possible and add any additional information into the page subtitles or descriptions 😄

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -71,6 +71,7 @@ def build(self):
                     value=self.value if self.value is not None else default_value,
                     multi=self.multi,
                     persistence=True,
+                    optionHeight=32 if all(len(option) <= 37 for option in full_options) else 60,  # noqa: PLR2004
```
antonymilne, line 74: Where do these numbers come from?

Fine if 37 is just a rough heuristic about what to treat as a cutoff, but just curious where it comes from. I guess 32 is what we have now, but then why 60 for a double height line rather than 64?

Presumably if someone does a really long option this will suffer from the same problem we currently have - is there anything more general that would handle that too? e.g. we could do the sort of analytic continuity of your step function like this if you see what I mean: https://www.desmos.com/calculator/c5fh81syid. People are unlikely to (or shouldn't be encouraged to) do huge long values, but if it's easy to accommodate a more general case rather than just a single "critical value" of 37 then we might as well.

antonymilne (changes requested):
Thanks very much for trying this out! I agree this looks better and seems moderately useful so it's a pre-emptive approval from my side.

> If it's a mixture of these two, we can only set one optionHeight which in this case should be the longer one. There it doesn't look so nicely but okay-ish. 

Agree, this is looks ok to me.

> I would be fine implementing this, but generally I think our recommendation should be to reduce the length of labels if possible and add any additional information into the page subtitles or descriptions 😄

Agree also - people generally shouldn't be encouraged to do this, but I think it's a reasonable enough request and something we can accommodate fairly easily so I think it's a good quality of life improvement overall.

vizro-core/src/vizro/models/_components/form/dropdown.py
antonymilne, line 74:
```suggestion
                    optionHeight=32 if max(len(option) for option in full_options) <= 37 else 60,  # noqa: PLR2004
```

huong-li-nguyen: Ah yeah, I like that idea! I'll implement that.

To your question - I'll add some comments there as well. 37 is the cut off value where the text starts to wrap on the next line. This is determined by our left-side panel width and dropdown width. If this changes, that number also has to change. I could only determine it manually though.

32 and 60 might still change - I have to take a look at the screenshots later. The reason why it's not 32 and then 64 is padding. The font-size remains the same, so the top and bottom padding should be the same as in the case of only 1-line of text. If you would choose 64, the padding-top and bottom would change. So this is more a design guidance, but I am not super strict about it.

maxschulz-COL (approved):
Lgtm!

vizro-core/src/vizro/models/_components/form/dropdown.py
huong-li-nguyen: In the end, it was 32 and 56 :)

PR #578, opened 2024-07-10
[Feat] Create `vizro.figures.library` to enable usage in pure Dash app

huong-li-nguyen:
## Description
This is the first PR to create a library for our components to be used in a standalone Dash app, independent of the Vizro framework. For context: The key performance indicator (KPI) cards are the first components eligible for use in a pure Dash app. In the future, we might add custom charts and other components as well. This PR lays the foundation for enabling these additions. 

**To ensure KPI components, and future components, work seamlessly outside of Vizro, we made the following adjustments:**

- Created vizro.figures.library to house KPI card Dash components usable outside the Vizro framework.
- Enabled offline use of the Google Material Icons library.
- Shipped CSS and fonts with the component library. Without this, KPI cards will appear unstyled and fail to render Google Material Icons correctly. Vizro KPI cards used in a standalone Dash app **will not look the same as within Vizro** but will adopt the look of the provided Bootstrap theme.
- Updated the Vizro-Bootstrap theme and modified the KPI Cards to look acceptable across all Bootstrap themes (https://github.com/McK-Private/vizro-bootstrap/pull/30)

Detailed code comments from @antonymilne explain these changes further (see changed files)

## Screenshot
**Vizro KPI Cards inside Vizro or with vizro-bootstrap (not public yet)**
Screenshot 2024-08-16 at 11 47 10

**Vizro KPI Cards in pure dash app with other bootstrap themes:**

**dbc.themes.BOOTSTRAP**
Screenshot 2024-08-16 at 11 50 01

**dbc.themes.SANDSTONE**
Screenshot 2024-08-16 at 11 48 22

**dbc.themes.CYBORG**
Screenshot 2024-08-16 at 11 49 10

## Notice


vizro-core/src/vizro/figures/kpi_cards.py
```diff
@@ -1,144 +1,8 @@
...
-            ),
-            html.Span(
-                reference_format.format(value=value, reference=reference, delta=delta, delta_relative=delta_relative)
-            ),
-        ],
-        className="color-pos" if delta > 0 else "color-neg" if delta < 0 else "",
-    )
-    return dbc.Card([header, body, footer], className="card-kpi")
+kpi_card = capture("figure")(kpi_card)
+kpi_card_reference = capture("figure")(kpi_card_reference)
```
huong-li-nguyen, line 8: @antonymilne - do you think it looks really weird if I actually do it the other way around? Currently we have the pure function code live in figure.undecorated.kpi_cards and in figure.kpi_cards we just wrap the function as above.

However, we could also have it the other way around. So let the pure function code live in figure.kpi_cards and in figure.undecorated.kpi_cards, we do this instead:

```
kpi_card = kpi_card.__wrapped__
kpi_card_reference = kpi_card.__wrapped__
```

I know it's less intuitive but the current implementation messes up our API reference as it can't read it 😬

vizro-core/examples/_dev/app.py
```diff
@@ -1,38 +1,53 @@
...
-    + " is one common species you can select in the iris dataset is one common species you can select in the iris data."
+df_kpi = pd.DataFrame({"Actual": [100, 200, 700], "Reference": [100, 300, 500], "Category": ["A", "B", "C"]})
+
+# Add single CSS file figures.css or
+vizro_bootstrap = (
+    "https://raw.githubusercontent.com/mckinsey/vizro/main/vizro-core/src/vizro/static/css/vizro-bootstrap.min.css"
 )
+vizro_css = "https://raw.githubusercontent.com/mckinsey/vizro/main/vizro-core/src/vizro/static/css/figures.css"
+vizro_tokens = "https://raw.githubusercontent.com/mckinsey/vizro/main/vizro-core/src/vizro/static/css/token_names.css"
+vizro_variables = "https://raw.githubusercontent.com/mckinsey/vizro/main/vizro-core/src/vizro/static/css/variables.css"
```
antonymilne, line 15:
```suggestion
base = f"https://cdn.jsdelivr.net/gh/mckinsey/vizro@0.1.18/vizro-core/src/vizro/static/css/"
vizro_bootstrap = base + "vizro-bootstrap.min.css"
vizro_css = base + "figures.min.css"
vizro_tokens = base + "token_names.min.css"
vizro_variables = base + "variables.min.css"
```

The reason this didn't work is because raw github doesn't work well as a CDN - let's us jsDelivr for that (same as bootstrap uses). Note the `.min` files get created automatically, so it doesn't matter that these don't exist in our repo.

If we do things this way then we should just put all the required CSS files together into some list so you'd just do `external_stylesheets=vizro.themes` or something.

With all that said, I wonder whether we should actually do this differently and bundle the CSS files with vizro itself [like this](https://community.plotly.com/t/introducing-agridable-effortlessly-format-tabular-data-in-your-dash-app/84815/6). That would mean you don't need to add anything to your external stylesheets in order to use these components (just like when we use dmc or AgGrid components for example).

In fact I wonder whether we should just be serving all our vizro static assets that way even within a Vizro app (not just to use our components in a pure Dash app) 🤔 Let us try to understand this more and see what the pros/cons are...

vizro-core/src/vizro/figures/kpi_cards.py
antonymilne: Fine by me 👍 I'm surprised the API docs can't handle this though, would be curious to understand why 🤔 Maybe it's because of a name clash and it would work if you imported the unwrapped ones explicitly as `unwrapped_kpi_card`?

antonymilne:
> Potential candidates for the naming of the module:
> 
> undecorated
> unwrapped
> unaltered
> raw
> plain
> pure
> Open for any other suggestion 😄 I am either for undecorated or unwrapped.

I like unwrapped or raw best.

vizro-core/examples/_dev/app.py
huong-li-nguyen: I'll pause this PR for now, let's discuss tomorrow if you're back 👍

vizro-core/src/vizro/figures/kpi_cards.py
huong-li-nguyen: Hmm.. not sure why to be honest. I've tried to rename it as you've suggested but it didn't work. 

Read through this:
- https://github.com/mkdocstrings/mkdocstrings/issues/162 
- https://mkdocstrings.github.io/troubleshooting/#my-wrapped-function-shows-documentationcode-for-its-wrapper-instead-of-its-own

However, I saw that we do use `functools.wrap` already and still does not work 🤔 I didn't spend much more time on this as the alternative solution above works, but if it's really important to understand and get the other way around working I can take a look again after I've taken a look at the CSS stuff.

antonymilne: Yes indeed, I was expecting `functools.wrap` to make this work automagically so weird that it doesn't.

Don't worry about spending any more time on this - it's not really important. But please leave this unresolved and I might try to figure it out some time, either in this PR or a future one.

vizro-core/src/vizro/static/css/figures.css
```diff
@@ -23,7 +23,7 @@
 }
 
 .card-kpi .card-footer {
-  color: var(--bs-secondary);
+  color: red;
```
antonymilne, line 26:
```suggestion
  color: var(--bs-secondary);
```

vizro-core/src/vizro/__init__.py
```diff
@@ -1,20 +1,36 @@
...
+from dash.development.base_component import Component
 
 from ._vizro import Vizro
 
-ComponentRegistry.registry.add("vizro")
+logging.basicConfig(level=os.getenv("VIZRO_LOG_LEVEL", "WARNING"))
+
+__all__ = ["Vizro"]
+
+__version__ = "0.1.20.dev0"
```
antonymilne, line 12: @huong-li-nguyen You'll need to make the tag `0.1.12.dev0` to test this before releasing with `serve_locally=False`.

antonymilne: Actually tagging is a bad way to do this because it points to a fixed commit... Let's try [4c260c1](https://github.com/mckinsey/vizro/pull/578/commits/4c260c16c38d0ce30f46e0ca00654ba1987dc9fa).

vizro-core/src/vizro/figures/kpi_cards.py
antonymilne: I've decided that we don't to repeat the exact same docs for everything wrapped and unwrapped. Doing things this way round (`capture` and then `__wrapped__`) is fine compared to the opposite direction. The captured version is the more important one to have docstrings, and I've added a note in the API docs to explain what `vizro.figures.library` is.

vizro-core/src/vizro/static/css/figures.css
```diff
@@ -1,50 +1,94 @@
...
+  height: 100%;
+  padding: 1rem;
+}
+
+/* We currently ship this CSS file with the package, so this needs to live here at the moment. */
+@font-face {
+  font-family: "Material Symbols Outlined";
+  font-style: normal;
+  font-weight: 400;
+  src: url("https://fonts.gstatic.com/s/materialsymbolsoutlined/v199/kJF1BvYX7BgnkSrUwT8OhrdQw4oELdPIeeII9v6oDMzByHX9rA6RzaxHMPdY43zj-jCxv3fzvRNU22ZXGJpEpjC_1v-p_4MrImHCIJIZrDCvHOejbd5zrDAt.woff2")
```
antonymilne, line 12: This URL looks kind of weird... Where did you get it from?

antonymilne (commented):
Rough notes on where this stands - I will explain more when I get a chance! But rest assured that 99% of this is now figured out and we just need to execute (I think).
* we should first do https://github.com/mckinsey/vizro/pull/615
* then complete below TODO for this PR

Here's the scheme we should go for (unless there's any strong objections): also considered `vizro.library.figures/charts`. Also might use this when calling one already-captured figure function inside a custom function definition - YES, this would be nicer than current approach. For that the `vizro.figures.library` etc. is better.
```
# from vizro.figures import kpi_card -> captured                from vizro.figures.library  import kpi_card
# from vizro.charts import waterfall -> captured                from vizro.charts.library import waterfall
# from vizro.plotly.express import scatter -> captured          from plotly.express import scatter -> uncaptured
```

TODO in this PR:
- [x] talk about right way to handle fonts - best to have two css files, one that works locally and one with `external_url`? Or just say if want to operate completely offline need to download font file locally, like with dash bootstrap. Then would also need to download extra css file that refers to font though
- [x] make sure our docs are good enough (I think they are already after the note I just added, but maybe we should add something on the figures page too, see what you think)
- [x] figure out why `vizro_bootstrap` doesn't work as @antonymilne though it would (probably fine, I just misunderstood something)
- [x] tidy PR description
- [x] delete roboto if not needed
- [ ] manual testing **with packaged vizro**
  - [ ] Vizro app with `serve_locally=True/False`
  - [ ] Pure dash app with `serve_locally=True/False` and non-vizro bootstrap theme
  - [x] Pure dash app with `serve_locally=True/False` and vizro bootstrap theme

vizro-core/src/vizro/static/css/figures.css
huong-li-nguyen: You sent me the link to this actually 😄 

https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined

antonymilne: Ah ok. Please can you download the woff2 file to our fonts and then make this relative, like ` url("./fonts/something-that's-not-too-gross")`. Together with a comment on where the woff2 file comes from.

Then I will just check something about how the CDN works and either that's solved or we keep as you have it here.

huong-li-nguyen: I've tried to do that here https://github.com/mckinsey/vizro/pull/578/commits/e839bc4ce7b4427e0b1a27ce4130e76674e315fe, but it doesn't seem to read the local font file 🤔 

We probably also need to ship the fonts with the package right? Is that what you mean with you want to check how the CDN works?

vizro-core/src/vizro/__init__.py
```diff
@@ -1,10 +1,51 @@
...
+# even though we don't need the component for anything. _css_dist is automatically served on import of vizro, regardless
+# of whether the Vizro class or any other bits are used.
+class _Dummy(Component):
+    pass
+
+
+# For dev versions, a branch or tag called e.g. 0.1.20.dev0 does not exist and so won't work with the CDN. We point
+# to main instead, but this can be manually overridden to the current feature branch name if required.
+# _git_branch = __version__ if "dev" not in __version__ else "main"
+_git_branch = "tidy/create-module-pure-functions"
```
antonymilne, line 29:
```suggestion
_git_branch = __version__ if "dev" not in __version__ else "main"
```

Must be applied just before merging.

vizro-core/src/vizro/figures/_kpi_cards.py
```diff
@@ -137,7 +142,12 @@ def kpi_card_reference(  # noqa: PLR0913
     delta = value - reference
     delta_relative = delta / reference if reference else np.nan
 
-    header = dbc.CardHeader([html.P(icon, className="material-symbols-outlined") if icon else None, html.H2(title)])
+    header = dbc.CardHeader(
+        [
+            html.P(icon, className="material-symbols-outlined") if icon else None,
+            html.H4(title, className="card-kpi-title"),
```
huong-li-nguyen, line 148: Change to `H4` is required as otherwise the KPI Cards look bad on all other bootstrap themes. Our font-sizes defined by the QB design library do not match up with Bootstrap's heading font-sizes. This should probably change in the future.

vizro-core/examples/scratch_dev/pure-app.py
```diff
@@ -0,0 +1,66 @@
+"""Dev app to try things out."""
```
huong-li-nguyen, line 1: I'll delete this file before merge.

vizro-core/src/vizro/static/css/figures.css
```diff
@@ -1,50 +1,97 @@
...
-
-.card-kpi .card-header .material-symbols-outlined {
-  line-height: 30px;
+  gap: 0.5rem;
 }
 
 .card-kpi .card-body {
   align-items: center;
-  color: var(--text-secondary);
+  color: var(--bs-secondary);
```
huong-li-nguyen, line 64: Important for future reference: In case we do add new components, colors need to be taken from the bootstrap variables, otherwise the new components are not compatible with other bootstrap themes.

huong-li-nguyen:
> * [ ]  manual testing **with packaged vizro**
>   
>   * [ ]  Vizro app with `serve_locally=True/False`
>   * [x]  Pure dash app with `serve_locally=True/False` and non-vizro bootstrap theme
>   * [x]  Pure dash app with `serve_locally=True/False` and vizro bootstrap theme


1. Vizro app with `serve_locally=True` and  **no internet** ✅ 


2. Vizro app with `serve_locally=False` and  **no internet** ✅ 
Screenshot 2024-08-16 at 12 23 08


3. Pure dash app with `serve_locally=True`, non-vizro bootstrap theme and **no internet** ✅ 
Screenshot 2024-08-16 at 12 13 06

4. Pure dash app with `serve_locally=False`, non-vizro bootstrap theme and **internet** ✅ 
Screenshot 2024-08-16 at 12 16 06

vizro-core/examples/scratch_dev/pure-app.py
petar-qb: I've tested several combinations and they worked like a charm.

vizro-core/src/vizro/__init__.py
```diff
@@ -2,15 +2,54 @@
...
-pio.templates["vizro_light"] = light
+# For the below _css_dist to be used by Dash, it must be retrieved by dash.resources.Css.get_all_css(). This means it
+# must be added to dash.development.base_component.ComponentRegistry. The simplest way to do this is to run
+# ComponentRegistry.registry.add("vizro") and this appears to be sufficient for our needs, but it is not documented
+# anywhere. The same function is run (together with some others which we probably don't need) when subclassing
+# Component, thanks to the metaclass ComponentMeta. So we define a dummy component to go through this safer route,
+# even though we don't need the component for anything. _css_dist is automatically served on import of vizro, regardless
+# of whether the Vizro class or any other bits are used.
+class _Dummy(Component):
+    pass
```
petar-qb, line 27: I like this "hack" 😃

vizro-core/src/vizro/static/css/vizro-bootstrap.min.css
petar-qb, line 1: From where these changes come from?

petar-qb (approved):
Wonderful work!

I'm curious: How far are we and do we actually want to support applying any bootstrap theme inside a Vizro application just by propagating the theme to the Vizro external_stylesheets argument?

vizro-core/docs/pages/API-reference/figure-callables.md
```diff
@@ -3,6 +3,11 @@
 API reference for all pre-defined [`CapturedCallable`][vizro.models.types.CapturedCallable] figure functions to be used in the
 [`Figure`][vizro.models.Figure]. The [how-to guide on figures](../user-guides/figure.md) contains more information.
 
+Identical functions can also be used outside the Vizro framework in a pure Dash app by importing from the `vizro.figures.library` namespace:
+```python
+from vizro.figures.library import kpi_card, kpi_card_reference
+```
```
maxschulz-COL, line 9: I would make this a note admonition, otherwise it might even get lost

vizro-core/src/vizro/static/css/vizro-bootstrap.min.css
huong-li-nguyen: From the Github repository I've linked in the PR description :) https://github.com/McK-Private/vizro-bootstrap/pull/30

vizro-core/src/vizro/static/css/figures.css
maxschulz-COL: Shall we document this in our wiki, and potentially also in our code?

huong-li-nguyen:
> I'm curious: How far are we and do we actually want to support applying any bootstrap theme inside a Vizro application just by propagating the theme to the Vizro external_stylesheets argument?

It's definitely on our roadmap - there is a detailed plan written out [here](https://mckinsey.atlassian.net/wiki/spaces/VIZX/pages/737509632/Vizro+Bootstrap+Workplan) and we've started development a while ago if you remember. You can track the progress there as well. However, we've paused it for a while given other priorities.  

The main goal is to create a complete Vizro Bootstrap theme and make it open-source so that anyone using a pure Dash app can use it too. We want all our CSS to be part of this theme, with only a few tweaks for third-party libraries like Dash-Mantine and AG-Grid. Everything else should be converted to Bootstrap. This way, we have one consistent source, making it easier to add other custom components.

vizro-core/src/vizro/__init__.py
```diff
@@ -2,15 +2,54 @@
...
+_css_dist = [
+    {
+        "namespace": "vizro",
+        "relative_package_path": f"{css_file}.css",
+        "external_url": f"{_base_external_url}{css_file}.min.css",
+    }
+    for css_file in _library_css
+]
+
+# Include font file so that figures with icons can be used outside Vizro as pure Dash components.
```
maxschulz-COL, line 46: I don't understand how we serve things with the package on the one hand, but we have also an external URL? Maybe I haven't understood the machanism quite yet. Could we document this better in the code?

maxschulz-COL (approved):
😍  Wow this is super cool! Tried out a few bootstrap themes, and it's really cool to see how the cards can be used in a pure dash app, and how they work well with those themes. Really great work @antonymilne @huong-li-nguyen !! ⭐ 

I have almost no comments - let me know if I should test something more manual, but going through the scratch app, all seemed fine 💪

vizro-core/src/vizro/__init__.py
antonymilne: I'm pleased because this took a lot of poking around in Dash internals to figure out 😅

vizro-core/src/vizro/__init__.py
```diff
@@ -2,15 +2,54 @@
...
+
+
+# For dev versions, a branch or tag called e.g. 0.1.20.dev0 does not exist and so won't work with the CDN. We point
+# to main instead, but this can be manually overridden to the current feature branch name if required.
+# _git_branch = __version__ if "dev" not in __version__ else "main"
+_git_branch = "tidy/create-module-pure-functions"
+_library_css = ["static/css/figures"]
+_base_external_url = f"https://cdn.jsdelivr.net/gh/mckinsey/vizro@{_git_branch}/vizro-core/src/vizro/"
+
+_css_dist = [
```
antonymilne, line 37:
```suggestion
_base_external_url = f"https://cdn.jsdelivr.net/gh/mckinsey/vizro@{_git_branch}/vizro-core/src/vizro/"

# CSS is packaged and accessed using relative_package_path when serve_locally=False (the default) in
# the Dash instantiation. When serve_locally=True then, where defined, external_url will be used instead.
```

vizro-core/src/vizro/__init__.py
antonymilne: Good question 👍 I added a code comment that explains this. FYI we will probably migrate all our CSS/JS to work this way if this seems to work well.

antonymilne (approved):
Looks great, thanks for all your patience and perseverance seeing this through!

vizro-core/src/vizro/static/css/figures.css
huong-li-nguyen: There is a ticket on creating a guide on an "intro to css" where we could add this in 👍 

So i wouldn't add it to this PR for now. All PRs changing CSS would currently go through @nadijagraca, @pruthvip15  and/or me anyway :)

PR #580, opened 2024-07-11
[Feat] Make Vizro app itself a WSGI callable

antonymilne:
## Description

As discussed with @maartenbreddels from py.cafe, the Vizro `app` object itself now behaves as a WSGI callable.

What does this mean in practice? There's two recommended ways of running a Vizro app:

### Development

`Vizro().build(dashboard).run()` and then `python app.py`, which is what we do across our docs. This only works while you're developing but I like recommending it as the first port of call for users because it's simple, quick and easy, like Vizro should be. There's no need to define `app`.

### Deployment

```

# If you also want it to run during development with python app.py you also need this:
if __name__ == "__main__":
    app.run()
```
and then e.g. `gunicorn app:app`. **The key change of this PR is that in this context there's no longer any need to define `server = app.dash.server`** (although that will still work).

@maxschulz-COL with the HF examples let's go for this second pattern.

The integration tests in this repo do something a bit different but that's just due to some technicalities of how they run and so don't show a generally recommended pattern.

## Notice


vizro-core/docs/pages/user-guides/run.md
```diff
@@ -133,6 +132,6 @@ A Vizro app wraps a Dash app, which itself wraps a Flask app. Hence to deploy a
 - [Flask deployment documentation](https://flask.palletsprojects.com/en/2.0.x/deploying/)
 - [Dash deployment documentation](https://dash.plotly.com/deployment)
 
-In particular, `app = Vizro()` exposes the Flask app through `app.dash.server`. As in the [above example with Gunicorn](#gunicorn), this provides the application instance to a [WSGI](https://werkzeug.palletsprojects.com/en/3.0.x/terms/#wsgi) server.
+Internally, `app = Vizro()` contains a Flask app in `app.dash.server`. However, as a convenience, the Vizro `app` itself implements the [WSGI application interface](https://werkzeug.palletsprojects.com/en/3.0.x/terms/#wsgi) as a shortcut to the underlying Flask app. This means that, as in the [above example with Gunicorn](#gunicorn), the Vizro `app` object itself can be directly supplied to the WSGI server.
```
antonymilne, line 135:
```suggestion
Internally, `app = Vizro()` contains a Flask app in `app.dash.server`. However, as a convenience, the Vizro `app` itself implements the [WSGI application interface](https://werkzeug.palletsprojects.com/en/3.0.x/terms/#wsgi) as a shortcut to the underlying Flask app. This means that, as in the [above example with Gunicorn](#gunicorn), the Vizro `app` object can be directly supplied to the WSGI server.
```

antonymilne:
@l0uden is it easy to add a test on the QA repo for this? I think you already do a `gunicorn` deployment?

petar-qb (approved):
I like it 🥇

maxschulz-COL (approved):
Makes sense - thanks for implementing it!

maartenbreddels:
Looks good!

l0uden:
> @l0uden is it easy to add a test on the QA repo for this? I think you already do a `gunicorn` deployment?

added

PR #583, opened 2024-07-12
[Tidy] Various small tidyings

antonymilne:
## Description

* Add some comments and refactor `optionHeight` to use `ceil` rather than `floor` following #574 - just to make it clearer when we look back on it in future
* Update `CODEOWNERS` so we don't need @stichbury to approve all docs changes like https://github.com/mckinsey/vizro/pull/579 (but she still gets notified on all of them) - definitely needed while she's away, and probably a good thing to do longer term
* Tiny update to docs replacing `pie_chart` with `Pie Chart`

## Notice

### DIFF
--- a/vizro-core/src/vizro/models/_components/form/dropdown.py
+++ b/vizro-core/src/vizro/models/_components/form/dropdown.py
@@ -63,6 +63,12 @@ class Dropdown(VizroBaseModel):
     @_log_call
     def build(self):
         full_options, default_value = get_options_and_default(options=self.options, multi=self.multi)
+        # 37 is the cut-off character length where the text inside the dropdown starts to wrap.
+        # We look at the longest option to find number_of_lines it requires. Option height is the same for all options
+        # and needs 24px for each line + 8px padding.
+        number_of_lines = math.ceil(max(len(str(option)) for option in full_options) / 37)
+        option_height = 8 + 24 * number_of_lines
+
         return html.Div(
             children=[
                 dbc.Label(self.title, html_for=self.id) if self.title else None,
@@ -72,8 +78,7 @@ class Dropdown(VizroBaseModel):
                     value=self.value if self.value is not None else default_value,
                     multi=self.multi,
                     persistence=True,
-                    # 37 is the cut-off character length where the text inside the dropdown starts to wrap
-                    optionHeight=32 + 24 * math.floor(max(len(str(option)) for option in full_options) / 37),
+                    optionHeight=option_height,
                     persistence_type="session",
                 ),
             ]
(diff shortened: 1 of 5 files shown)


huong-li-nguyen (approved):
LGTM 👍

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -63,6 +63,12 @@ def validate_multi(cls, multi, values):
     @_log_call
     def build(self):
         full_options, default_value = get_options_and_default(options=self.options, multi=self.multi)
+        # 37 is the cut-off character length where the text inside the dropdown starts to wrap.
+        # We look at the longest option to find number_of_lines it requires. Option height is the same for all options
+        # and needs 24px for each line + 8px padding.
+        number_of_lines = math.ceil(max(len(str(option)) for option in full_options) / 37)
+        option_height = 8 + 24 * number_of_lines
```
antonymilne, line 70: This is the same number as before, just more obvious where it comes from:
```
y = c + mx, with x = floor(z)
x' = ceil(z) = x + 1
So y = c + m(x'-1) = (c-m) + mx'
```
where m = 24, c = 32.

maxschulz-COL (approved):
Lgtm except maybe the maths calculation :P

vizro-core/src/vizro/models/_components/form/dropdown.py
antonymilne: Good catch! I was hoping that the unit tests would fail if I had done something wrong here 😀 

But I'm confused now. It sounds like the maximum number of characters in a line is 36 rather than 37 like I had understood before. Is that right?

If so then actually calculation before this PR is not quite right since for 73 characters it would give 2 lines but should in fact be on 3 lines? So then the correct formula should have 36 rather than 37 as the critical value and would be:
```
        number_of_lines = math.ceil(max(len(str(option)) for option in full_options) / 36)
        option_height = 8 + 24 * number_of_lines
```

huong-li-nguyen: I am getting a bit confused now as well  😄 I think it's because of differences in the `floor` and `ceil` function.

I just know that before it has been working correctly 😄 So with 73 characters it should give 2 lines, with 74 characters it then gives 3 lines. 

As long as the new calculation passes the unit tests, I think it should be fine 😄

antonymilne: I have done some more scientific experimenting and decided that 36 as a cut off was a bit big actually so changed it to 30 - see [34e87fe](https://github.com/mckinsey/vizro/pull/583/commits/34e87fe299b9d0853f49121f42621b91b4e40244).

Let me know if you strongly object 😅

huong-li-nguyen: I didn't take a look how it looks visually, but for me that's fine 👍 I don't care that much when we start wrapping 😅

PR #585, opened 2024-07-16
[Bug] External links for vm.Card now open in the full body

maxschulz-COL:
## Description
Previously when clicking `vm.Card` used as a `NavCard`, ie with `href` enabled, would lead to external links opening in the same tab, but not in a new tab, on py.cafe. This is now fixed.

Adding some more info (thanks @petar-qb ):
- More info on what `target="_blank"` does - [see link](https://www.w3schools.com/tags/att_a_target.asp)
- It looks from the docs that `dbc.NavLink` works out when to format links and when not: https://dash-bootstrap-components.opensource.faculty.ai/docs/components/nav/

P.S. It is debatable whether this is really a bug or a feature, but given that it broke things on py.cafe, I will leave the Bug tag - happy to have it changed though

P.P.S Based on @antonymilne new suggestion I changed the PR title to reflect the new approach - external links now open in the full body

## Notice

### DIFF
--- a/vizro-core/changelog.d/20240716_091858_maximilian_schulz_try_out_link_target.md
+++ b/vizro-core/changelog.d/20240716_091858_maximilian_schulz_try_out_link_target.md
@@ -1,0 +1,47 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Internal and external links for `href` in `vm.Card` now behave correctly on py.cafe ([#579](https://github.com/mckinsey/vizro/pull/579))
+
+<!--
+### Security
+
+- A bullet item for the Security category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
--- a/vizro-core/changelog.d/20240716_091858_maximilian_schulz_try_out_link_target.md
+++ b/vizro-core/changelog.d/20240716_091858_maximilian_schulz_try_out_link_target.md
@@ -37,7 +37,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 ### Fixed
 
-- Internal and external links for `href` in `vm.Card` now behave correctly on py.cafe ([#579](https://github.com/mckinsey/vizro/pull/579))
+- Internal and external links for `href` in `vm.Card` now behave correctly on py.cafe ([#585](https://github.com/mckinsey/vizro/pull/585))
 
 <!--
 ### Security
--- a/vizro-core/changelog.d/20240716_091858_maximilian_schulz_try_out_link_target.md
+++ b/vizro-core/changelog.d/20240716_091858_maximilian_schulz_try_out_link_target.md
@@ -37,7 +37,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 ### Fixed
 
-- Internal and external links for `href` in `vm.Card` now behave correctly on py.cafe ([#585](https://github.com/mckinsey/vizro/pull/585))
+- External `href` links in `vm.Card` now open in a new browser tab. ([#585](https://github.com/mckinsey/vizro/pull/585))
 
 <!--
 ### Security
--- a/vizro-core/changelog.d/20240716_091858_maximilian_schulz_try_out_link_target.md
+++ b/vizro-core/changelog.d/20240716_091858_maximilian_schulz_try_out_link_target.md
@@ -37,7 +37,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 ### Fixed
 
-- External `href` links in `vm.Card` now open in a new browser tab. ([#585](https://github.com/mckinsey/vizro/pull/585))
+- External `href` links in `vm.Card` now open in the full body of the window as opposed to in the same frame as they were clicked. ([#585](https://github.com/mckinsey/vizro/pull/585))
 
 <!--
 ### Security
--- a/vizro-core/src/vizro/models/_components/card.py
+++ b/vizro-core/src/vizro/models/_components/card.py
@@ -39,6 +39,7 @@ class Card(VizroBaseModel):
             dbc.NavLink(
                 children=text,
                 href=get_relative_path(self.href) if self.href.startswith("/") else self.href,
+                target="_blank",
             )
             if self.href
             else text
--- a/vizro-core/src/vizro/models/_components/card.py
+++ b/vizro-core/src/vizro/models/_components/card.py
@@ -39,7 +39,7 @@ class Card(VizroBaseModel):
             dbc.NavLink(
                 children=text,
                 href=get_relative_path(self.href) if self.href.startswith("/") else self.href,
-                target="_blank",
+                target="_top",
             )
             if self.href
             else text
--- a/vizro-core/examples/scratch_dev/app.py
+++ b/vizro-core/examples/scratch_dev/app.py
@@ -1,38 +1,66 @@
-"""Dev app to try things out."""
+"""Example app to show all features of Vizro."""
 
-import numpy as np
+# check out https://github.com/mckinsey/vizro for more info about Vizro
+# and checkout https://vizro.readthedocs.io/en/stable/ for documentation
+
 import vizro.models as vm
 import vizro.plotly.express as px
 from vizro import Vizro
 
 df = px.data.iris()
-df["species_one_long"] = np.where(
-    df["species"] == "setosa", "setosa is one common species you can select in the iris dataset.", df["species"]
-)
-df["species_long"] = df["species"] + " is one common species you can select in the iris dataset."
-df["species_very_long"] = (
-    df["species"]
-    + " is one common species you can select in the iris dataset is one common species you can select in the iris data."
-)
 
 page = vm.Page(
-    title="",
+    title="Vizro on PyCafe",
+    layout=vm.Layout(
+        grid=[[0, 0, 0, 1, 2, 3], [4, 4, 4, 4, 4, 4], [4, 4, 4, 4, 4, 4], [5, 5, 5, 5, 5, 5], [5, 5, 5, 5, 5, 5]],
+        row_min_height="175px",
+    ),
     components=[
-        vm.Graph(
-            id="graph_1",
-            figure=px.scatter(df, title="Title", x="sepal_width", y="sepal_length", color="species"),
+        vm.Card(
+            text="""
+        ### What is Vizro?
+
+        Vizro is a toolkit for creating modular data visualization applications.
+        """
         ),
+        vm.Card(
+            text="""
+                ### Github
+
+                Checkout Vizro's github page.
+            """,
+            href="https://github.com/mckinsey/vizro",
+        ),
+        vm.Card(
+            text="""
+                ### Docs
+
+                Visit the documentation for codes examples, tutorials and API reference.
+            """,
+            href="https://vizro.readthedocs.io/",
+        ),
+        vm.Card(
+            text="""
+                ### Nav Link
+
+                Click this for page 2.
+            """,
+            href="/page2",
+        ),
+        vm.Graph(id="scatter_chart", figure=px.scatter(df, x="sepal_length", y="petal_width", color="species")),
+        vm.Graph(id="hist_chart", figure=px.histogram(df, x="sepal_width", color="species")),
     ],
     controls=[
-        vm.Filter(column="species"),
-        vm.Filter(column="species_long"),
-        vm.Filter(column="species_one_long"),
-        vm.Filter(column="species_very_long"),
+        vm.Filter(column="species", selector=vm.Dropdown(value=["ALL"])),
+        vm.Filter(column="petal_length"),
+        vm.Filter(column="sepal_width"),
     ],
 )
 
+page2 = vm.Page(
+    title="Page2", components=[vm.Graph(id="hist_chart2", figure=px.histogram(df, x="sepal_width", color="species"))]
+)
 
-dashboard = vm.Dashboard(pages=[page])
+dashboard = vm.Dashboard(pages=[page, page2])
 
-if __name__ == "__main__":
-    Vizro().build(dashboard).run()
+Vizro().build(dashboard).run()
--- a/vizro-core/examples/scratch_dev/app.py
+++ b/vizro-core/examples/scratch_dev/app.py
@@ -63,4 +63,5 @@ page2 = vm.Page(
 
 dashboard = vm.Dashboard(pages=[page, page2])
 
-Vizro().build(dashboard).run()
+if __name__ == "__main__":
+    Vizro().build(dashboard).run()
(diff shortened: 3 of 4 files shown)


petar-qb:
I'm sharing this [link](https://www.w3schools.com/tags/att_a_target.asp) with the reviewers in case you're unfamiliar with the context of `target="_blank"` as I haven't been.

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -39,6 +39,7 @@ def build(self):
             dbc.NavLink(
                 children=text,
                 href=get_relative_path(self.href) if self.href.startswith("/") else self.href,
+                target="_blank",
```
petar-qb, line 42: I'm curious how come internal links like `/page-2` at the end (UI client side) end up without `target="_blank"` specified, and the external link ends up with the `target="_blank"`? It looks like we are adding `target="_blank"` here for both external and internal links. 🤔

image

vizro-core/changelog.d/20240716_091858_maximilian_schulz_try_out_link_target.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Internal and external links for `href` in `vm.Card` now behave correctly on py.cafe ([#585](https://github.com/mckinsey/vizro/pull/585))
```
petar-qb, line 40:
```suggestion
- External `href` links in `vm.Card` now open in a new browser tab. ([#585](https://github.com/mckinsey/vizro/pull/585))
```

petar-qb (approved):
I looks and works well 🚀

huong-li-nguyen (commented):
It previously always opened in the same tab no? It's not specific to py.cafe?

I didn't take a look how it works on py.cafe but from a UX perspective I am wondering if we even want to add this to our `Card` model? There are some common guidelines of when "opening in a new tab" vs "opening the same tab" are preferrable, and generally it tends towards "opening in the same tab". See this for context: https://blog.pope.tech/2024/01/02/how-to-make-external-links-accessible/

That being said, in the article it's also mentioned that if links open in a new tab, there should be a visual clue, which is not the case for the NavCard because the actual link is not visible. 

So I am wondering if this should even live in our common `Card` model or if this should actually be a custom component where you extend the `Card` with that argument.

vizro-core/src/vizro/models/_components/card.py
maxschulz-COL: This is von the docs linked in the PR description: 

> target (string; optional): Target attribute to pass on to the link. Only applies to external links.

antonymilne: `dbc.NavLink` only adds `target` for external links - see https://dash-bootstrap-components.opensource.faculty.ai/docs/components/nav/. Note that the detection of whether something is `external_link` itself is found automatically on the dbc side, which is very useful for us.

I was also curious how this worked so found the relevant bit of code yesterday 😀 https://github.com/facultyai/dash-bootstrap-components/blob/c9367e6284823df7118594c5f55e8caf488c6a6d/src/private/Link.js#L84

So basically this is a great bit of design by dbc and makes our lives easy 😀

maxschulz-COL:
> It previously always opened in the same tab, right? It's not specific to py.cafe?

Ah yes, I only tested this on py.cafe, and never locally, but you are correct. Changed the PR title

> I didn't look at how it works on py.cafe but from a UX perspective, I wonder if we even want to add this to our `Card` model. There are some standard guidelines of when "opening in a new tab" vs "opening the same tab" is preferable, and generally, it tends towards "opening in the same tab". See this for context: https://blog.pope.tech/2024/01/02/how-to-make-external-links-accessible/

On py.cafe this does not work, and this was the suggested fix by Maarten and Antony. I just read through the guidelines and they seem pretty strong and mostly convincing! Personally I have never perceived this, but I think the accessibility points are a good point.

> That said, the article also mentions that if links open in a new tab, there should be a visual clue, which is not the case for the NavCard because the actual link is not visible.
> 
> So I am wondering if this should even live in our standard `Card` model or if this should be a custom component where you extend the `Card` with that argument. I currently prefer the second considering above.

Happy to go either way - we certainly cannot make this a custom component for the py.cafe landing snippet because we want that code to be as clean as possible. If we choose not to implement this, then we have to find another solution. Curious to hear what @Joseph-Perkins and @antonymilne think.

@Joseph-Perkins For context: do we want the Card with `href` to open external links in a new browser tab or not? (this is the bug we encountered with your snippet example - but fixing this we this solution causes other accessibility issues)

huong-li-nguyen:
> @Joseph-Perkins For context: do we want the Card with `href` to open external links in a new browser tab or not? (this is the bug we encountered with your snippet example - but fixing this we this solution causes other accessibility issues)

Why don't we add the same footer-banner as in our demo app? That didn't add much code. I think the links to the docs and git repo shouldn't be that prominent as currently in the form of NavCards anyway 😄 

I think that's still clean enough and then you can add the argument directly inside the NavLink below. No need to create a custom component and I think that visually looks better + doesn't require that much code.

```
app = Vizro(assets_folder="./assets").build(dashboard)
app.dash.layout.children.append(
    html.Div(
        [
            dbc.NavLink("Contact us", href="https://github.com/mckinsey/vizro/issues", className="footer-link"),
            dbc.NavLink("GitHub", href="https://github.com/mckinsey/vizro", className="footer-link"),
            dbc.NavLink(
                "Docs", href="https://vizro.readthedocs.io/en/stable/", className="footer-link"
            ),
            html.Div(["Made in ", html.Img(src=get_asset_url("logo.svg"), id="banner", alt="Vizro logo"), "vizro"],
                     className="banner-text"),
        ],
        className="anchor-container"
    )
)
```

antonymilne:
> It previously always opened in the same tab, right? It's not specific to py.cafe?

Yes indeed. It's just the py.cafe is a good example (but not the only one) where opening in the same tab doesn't work well.

> I didn't look at how it works on py.cafe but from a UX perspective, I wonder if we even want to add this to our `Card` model. There are some standard guidelines of when "opening in a new tab" vs "opening the same tab" is preferable, and generally, it tends towards "opening in the same tab". See this for context: https://blog.pope.tech/2024/01/02/how-to-make-external-links-accessible/
> 
> That said, the article also mentions that if links open in a new tab, there should be a visual clue, which is not the case for the NavCard because the actual link is not visible.
> 
> So I am wondering if this should even live in our standard `Card` model or if this should be a custom component where you extend the `Card` with that argument. I currently prefer the second considering above.

Indeed I've read very similar things in the past, and this was exactly why I didn't want to set the [link target in the `dcc.Markdown`](https://dash.plotly.com/dash-core-components/markdown#link-target). In general the correct default behaviour for links is definitely to open them in the same tab.

However, I think for the special (and not very common) case of a `dbc.Card` that does an external link, opening in a new tab is reasonable as a default (not just for py.cafe). Ideally we would have a the little "opens in new tab" icon on the card like you say also, but I'm ok to not do that straight away.

In this case I would prefer to not do as a custom component because this will be part of the py.cafe template code so should be a simple example to get people started and that will just make it look way too complicated. So the only other possibility here is that we add another argument to `vm.Card` that allows a user to specify whether to open external links in new tabs or not. That seems like overkill though (and could itself just be done through a custom component). So overall I think this is a good default, and the only thing I would change is to make it clearer visually that the link opens in a new tab, but we can save that for another time.

wdyt?

huong-li-nguyen:
> wdyt?

See answer above :) Independently of the technicalities, I would prefer to just have the footer and not have these external links as prominent as the NavCard if this should be our pyCafe landing page? And then I think we don't need a custom component and the code it still simple enough?

antonymilne:
The footer banner is an interesting idea and not something I'd thought of here, but I don't think it solves this case (which is relevant for other examples I've seen outside py.cafe also, not just our demos banner). It's still kind of weird custom code that we'd need to put in - especially because we'd need all the CSS as well, not just the Python. To avoid it being weird custom code at all we'd need to create a new "Vizro footer" component that bundles the CSS too, which is not impossible but a lot more effort. Also maybe in future it would be useful to have some `insert_footer` command or even `Footer` model or similar to simplify this beyond just the Vizro banner, but again that would be a much bigger change.

So in terms of visuals, the navcards is good enough for me since it shows off the native use of vizro components, and even if the banner looks a bit nicer I don't think it's worth the extra complexity, especially since we should solve the navcard link opening for other contexts anyway.

BUT your comments have made me go back and think again about there's a different fix here that would avoid opening in a new tab, and actually I think there is! If we set `target="_top"` then this will not open anything in a new tab but instead break out of the iframes, which is really what we want - [py.cafe example](https://py.cafe/snippet/vizro/v1#c=H4sIAKA7lmYAA61XW2_bNhT-K5zyMAdQ5KRbt8GAN3S9rUBaFH3oHqJAoKUjiQhFaiRlRy363_eRkhzZddHuIhuQdHguPBd-5-hjlOuColV0xvKa8jumO8dq51q7Wi4r4epuk-S6WTb5nVCW-uVWfDCaldqwRhtiQpWa8Y2Xeu9XUnXGuCoGZXNdQS4xxAtXU6Fzmwi9JLW0jm8kLYNGkLuGlONOaJXiJ5pWG8cG2VZqJ_uE7ltD1jJuWXufqtLoZmBgI_e4jwPZBk7KILNt5jJJKaoO6ibZu1ZkOTdFvH_KDJVkSOU0yrm-FaqaBK6FI8PlbLMFt3W20dpZZ3ibIXitVnAqWC82-ajGs01KijyPWUUOtiR831LWcld7nc70q1QxXEGo7QuunMiT7dUk-0KQLFJF9zm1jr0KxOfGaLNi7Iy1hlcNXzGlWa63JzQdqTmIzRi0eVh_55Zee_IJxiQb7lnnxINYJnWFQMoQoqJka2QtKbjjiTDCLs49OZccCX2KaC-2TeLv56PXBZVs0wlZLCzJcqL6y9G9gzKELnnNzV2hd2ohirVnSwTyl9eQQt4GiueOEXNVkdGdlX2GHeldVrtGrl9waQkbmVSHtOdaOWQNJhYPK2FLmzx5w7fXQt0drQTZyWyw-Pl6jXJaf5bq4F3i186ZKNn-LcHpMM7ucA4XabRMo3OGANMDwwkLkID-dRplTrdpdMQxc9Nfc2OHK8GOd-KBHHJ1GKSQuDVLI_96ofg2jQ50DmrSCL-9pCHXGRXiGFK-D9k87kig1_2GNzTS_WvYQapaXhGsolbe4mnMghNOEtwOdcq0Ym_7p7ykfQQk74FHawhdh6dZ8iqDyrm5uYzZ8L-K2aOY_XAbs5sfY3bw_xLtccwO_idpt7NkGFRfI1RWk6hqn66rnx8Dz6bdno_3B_hY3zwIj6fkqP58tqDoMNpnZ2fsz5o7JuxwhH-bZ3EIFpY4c1rLO-ECEOfAaedRDie6kxzIjAOLw247LsWHAM-Mt60UeXi2yVznO96KQvahCC4smS3qqCagn6VmA7ouWd5ZpxvxgYqAgxsNZ4AYiiEinQMcXzBf875_eFFFYPT74sWWA4nRXnTht-d7BllRKYauQEZ4mIasA9p5D4iVku4F-kvoSRYw5JtNjNKAcnVByr_6PXy7d8fxnRLlr2_NyTw3L0OPnVvYa5taaEjS95YN7ZiF6vfRKDuD6JjQg00z7ht-AloIOA3QRygT9hQnyogNUBm-MI6mzeWO95btSKK86LtD08Nmj2AjwFYafXUwOBD8r6F5hinhVGDeC4s69YVxMDAMlYsWZFENvGkl2Zi5zmkjAPAhMk_evkJ0xn6OtP5Tv08OMf-rzy_Rl8j0p9z-A2Z9aU_lMLJ6kiX0BKSSbYgjz2Un5-fKxwU5aoUJcfoXftddVeG8lRxRy_VXUg1weonJp_btOI1w6BxGpCyv0ckgw4aJa40pYFxaFGXM7j0rtVxmklSF4QecPWgtOdB2AgH3pFxLbTxrS7kgi4b4Zcu1sO6UWU_XleHNkeFvMDIBuG9SRstjTH4hpPcH8l2j5gpiD4aUoxR9_3lmdBvmlS2XHUFH9OT6Oo1uj305VjfEYorPV5gPfJrvP_TQfXUMbfTZ9Lrw4AK3_O02cIZqW5wnwxS2lztPTKf89BbFkaG_OmHIH0SLL4lQHb-uL5Or5OqXwICR2X9ihAW8AlvfC9pFq9IPXnFEhXDPAxRHK2c6UNqAzxBpewB9QRfby-TRT8klhIc2Hq0-RgYz9jsNNR8nA-ip4JjmiWh1s1_BR4Yl9-VFrPg5DC_YXNL20adb_D7FUSkAI2C-_fQ3PDKrGCoNAAA). This is much more the expected user behaviour so we don't need to worry about visually indicating that the link opens in a new tab any more.

huong-li-nguyen:
> This is much more the expected user behaviour so we don't need to worry about visually indicating that the link opens in a new tab any more.

That sounds like a great solution! 😄 👍

maxschulz-COL:
> BUT your comments have made me go back and think again about there's a different fix here that would avoid opening in a new tab, and actually I think there is! If we set `target="_top"` then this will not open anything in a new tab but instead break out of the iframes, which is really what we want - [py.cafe example](https://py.cafe/snippet/vizro/v1#c=H4sIAKA7lmYAA61XW2_bNhT-K5zyMAdQ5KRbt8GAN3S9rUBaFH3oHqJAoKUjiQhFaiRlRy363_eRkhzZddHuIhuQdHguPBd-5-hjlOuColV0xvKa8jumO8dq51q7Wi4r4epuk-S6WTb5nVCW-uVWfDCaldqwRhtiQpWa8Y2Xeu9XUnXGuCoGZXNdQS4xxAtXU6Fzmwi9JLW0jm8kLYNGkLuGlONOaJXiJ5pWG8cG2VZqJ_uE7ltD1jJuWXufqtLoZmBgI_e4jwPZBk7KILNt5jJJKaoO6ibZu1ZkOTdFvH_KDJVkSOU0yrm-FaqaBK6FI8PlbLMFt3W20dpZZ3ibIXitVnAqWC82-ajGs01KijyPWUUOtiR831LWcld7nc70q1QxXEGo7QuunMiT7dUk-0KQLFJF9zm1jr0KxOfGaLNi7Iy1hlcNXzGlWa63JzQdqTmIzRi0eVh_55Zee_IJxiQb7lnnxINYJnWFQMoQoqJka2QtKbjjiTDCLs49OZccCX2KaC-2TeLv56PXBZVs0wlZLCzJcqL6y9G9gzKELnnNzV2hd2ohirVnSwTyl9eQQt4GiueOEXNVkdGdlX2GHeldVrtGrl9waQkbmVSHtOdaOWQNJhYPK2FLmzx5w7fXQt0drQTZyWyw-Pl6jXJaf5bq4F3i186ZKNn-LcHpMM7ucA4XabRMo3OGANMDwwkLkID-dRplTrdpdMQxc9Nfc2OHK8GOd-KBHHJ1GKSQuDVLI_96ofg2jQ50DmrSCL-9pCHXGRXiGFK-D9k87kig1_2GNzTS_WvYQapaXhGsolbe4mnMghNOEtwOdcq0Ym_7p7ykfQQk74FHawhdh6dZ8iqDyrm5uYzZ8L-K2aOY_XAbs5sfY3bw_xLtccwO_idpt7NkGFRfI1RWk6hqn66rnx8Dz6bdno_3B_hY3zwIj6fkqP58tqDoMNpnZ2fsz5o7JuxwhH-bZ3EIFpY4c1rLO-ECEOfAaedRDie6kxzIjAOLw247LsWHAM-Mt60UeXi2yVznO96KQvahCC4smS3qqCagn6VmA7ouWd5ZpxvxgYqAgxsNZ4AYiiEinQMcXzBf875_eFFFYPT74sWWA4nRXnTht-d7BllRKYauQEZ4mIasA9p5D4iVku4F-kvoSRYw5JtNjNKAcnVByr_6PXy7d8fxnRLlr2_NyTw3L0OPnVvYa5taaEjS95YN7ZiF6vfRKDuD6JjQg00z7ht-AloIOA3QRygT9hQnyogNUBm-MI6mzeWO95btSKK86LtD08Nmj2AjwFYafXUwOBD8r6F5hinhVGDeC4s69YVxMDAMlYsWZFENvGkl2Zi5zmkjAPAhMk_evkJ0xn6OtP5Tv08OMf-rzy_Rl8j0p9z-A2Z9aU_lMLJ6kiX0BKSSbYgjz2Un5-fKxwU5aoUJcfoXftddVeG8lRxRy_VXUg1weonJp_btOI1w6BxGpCyv0ckgw4aJa40pYFxaFGXM7j0rtVxmklSF4QecPWgtOdB2AgH3pFxLbTxrS7kgi4b4Zcu1sO6UWU_XleHNkeFvMDIBuG9SRstjTH4hpPcH8l2j5gpiD4aUoxR9_3lmdBvmlS2XHUFH9OT6Oo1uj305VjfEYorPV5gPfJrvP_TQfXUMbfTZ9Lrw4AK3_O02cIZqW5wnwxS2lztPTKf89BbFkaG_OmHIH0SLL4lQHb-uL5Or5OqXwICR2X9ihAW8AlvfC9pFq9IPXnFEhXDPAxRHK2c6UNqAzxBpewB9QRfby-TRT8klhIc2Hq0-RgYz9jsNNR8nA-ip4JjmiWh1s1_BR4Yl9-VFrPg5DC_YXNL20adb_D7FUSkAI2C-_fQ3PDKrGCoNAAA). This is much more the expected user behaviour so we don't need to worry about visually indicating that the link opens in a new tab any more.

Shall we go with this then?

huong-li-nguyen (approved):
I haven't tested this out on pyCafe so this approval is purely based on the current branch 😄

antonymilne (approved):

Issue #588, opened 2024-07-17
kpi_card throws a nasty Pydantic Type missing error 


objectivitydrx:
### Which package?

vizro

### Package version

0.1.18 and 0.1.19

### Description

Single kpi_card (taken from the vizro demo app) on a page throws type error (see code) 

Minimum failure code attached.   Pydantic called via super().__init__ is complaining about a need for a type.

Note: kpi_cards included in the vizro demo.app do not get called/rendered on any pages.

### How to Reproduce

Running the following with either 0.1.18/0.1.19 

``` python 
import pandas as pd
import vizro.models as vm
from vizro import Vizro
from vizro.figures import kpi_card

# data from the demo app
df_kpi = pd.DataFrame(
    {
        "Actual": [100, 200, 700],
        "Reference": [100, 300, 500],
        "Category": ["A", "B", "C"],
    }
)


home = vm.Page(
    title="value_error.discriminated_union.missing_discriminator",
    components=[
        # vm.Card(text="I am a Card"),
        kpi_card(data_frame=df_kpi, value_column="Actual", title="KPI with value"),
    ],
)


dashboard = vm.Dashboard(pages=[home])

if __name__ == "__main__":
    Vizro().build(dashboard).run(port=8057, debug=True)



```

### Output

File "vtest.py", line 16, in <module>
    home = vm.Page(
  File ".../python3.10/site-packages/vizro/models/_page.py", line 83, in __init__
    super().__init__(**data)
  File ".../python3.10/site-packages/vizro/models/_models_utils.py", line 11, in _wrapper
    return_value = method(self, *args, **kwargs)
  File ".../python3.10/site-packages/vizro/models/_base.py", line 42, in __init__
    super().__init__(**data)
  File ".../python3.10/site-packages/pydantic/v1/main.py", line 341, in __init__
    raise validation_error
pydantic.v1.error_wrappers.ValidationError: 1 validation error for Page
components -> 0
  Discriminator 'type' is missing in value (type=value_error.discriminated_union.missing_discriminator; discriminator_key=type)

### Code of Conduct

- [X] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

huong-li-nguyen:
Hey @objectivitydrx ,

you need to wrap the `kpi_card` function inside our `vm.Figure` model :)

[...]

PR #590, opened 2024-07-18
[Tidy] Improve validation error message if `CapturedCallable` is directly provided

huong-li-nguyen:
## Description
- Improve validation error if CapturedCallable is directly provided as a  `component`
- Context: https://github.com/mckinsey/vizro/issues/588

Thanks to @objectivitydrx for raising the bug and making a suggestion to improve our docs! 📝 🚀 

## Screenshot
Screenshot 2024-07-18 at 15 40 20

## Notice

### DIFF
--- a/vizro-core/changelog.d/20240718_151233_huong_li_nguyen_validation_error_captured.md
+++ b/vizro-core/changelog.d/20240718_151233_huong_li_nguyen_validation_error_captured.md
@@ -39,7 +39,6 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 - Improve validation error message if `CapturedCallable` is directly provided. ([#587](https://github.com/mckinsey/vizro/pull/587))
 
-
 <!--
 ### Security
 
--- a/vizro-core/changelog.d/20240718_151233_huong_li_nguyen_validation_error_captured.md
+++ b/vizro-core/changelog.d/20240718_151233_huong_li_nguyen_validation_error_captured.md
@@ -16,12 +16,11 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 - A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
 
 -->
-<!--
+
 ### Added
 
-- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+- Added validation error message if `CapturedCallable` is directly provided. ([#590](https://github.com/mckinsey/vizro/pull/590))
 
--->
 <!--
 ### Changed
 
@@ -34,11 +33,12 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 - A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
 
 -->
-
+<!--
 ### Fixed
 
-- Improve validation error message if `CapturedCallable` is directly provided. ([#590](https://github.com/mckinsey/vizro/pull/590))
+- A bullet item for the Fixed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
 
+-->
 <!--
 ### Security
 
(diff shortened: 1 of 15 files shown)


vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -16,7 +18,27 @@ def _wrapper(self, *args, **kwargs):
...
+    for component in field:
+        if isinstance(component, (_DashboardReadyFigure, CapturedCallable)):
+            mode = (
+                component._captured_callable._mode if isinstance(component, _DashboardReadyFigure) else component._mode
+            )
+            error_message = mode_to_error.get(mode)
+            if error_message:
+                raise ValueError(error_message)
+
+    return field
```
huong-li-nguyen, line 44: @antonymilne - I was initially very confused why my unit tests passed for `dash_ag_grid, dash_data_table and kpi_card` but not for any `px chart` until I saw that the type of the px.chart is not `CapturedCallable` as for the others but it's _DashboardReadyFigure. Similar to access the mode, it's different for these. 

Is there a reason for this? Would you do the check as I have it or is there a better way to check the mode etc?

antonymilne: Hmmm yes, this was a bit more fiddly than I had anticipated. You did a great job of it - and I have just rewritten it a bit - see these commits:
image

vizro-core/changelog.d/20240718_151233_huong_li_nguyen_validation_error_captured.md
```diff
@@ -0,0 +1,48 @@
...
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
```
maxschulz-COL, line 38: Changed?

vizro-core/src/vizro/models/_components/_form.py
```diff
@@ -12,7 +12,7 @@
 from vizro.models import VizroBaseModel
 from vizro.models._components.form import Checklist, Dropdown, RadioItems, RangeSlider, Slider
 from vizro.models._layout import set_layout
-from vizro.models._models_utils import _log_call, _validate_min_length
+from vizro.models._models_utils import _log_call, check_captured_callable, validate_min_length
```
maxschulz-COL, line 15: What's the difference between a private function from a private module, and a non private function from a private model @antonymilne - ie why does `_log_call` have a `_`?

vizro-core/src/vizro/models/_page.py
```diff
@@ -52,7 +52,10 @@ class Page(VizroBaseModel):
     actions: List[ActionsChain] = []
 
     # Re-used validators
-    _validate_components = validator("components", allow_reuse=True, always=True)(_validate_min_length)
+    _check_captured_callable = validator("components", allow_reuse=True, each_item=True, pre=True)(
+        check_captured_callable
+    )
+    _validate_components_length = validator("components", allow_reuse=True, always=True)(validate_min_length)
```
maxschulz-COL, line 58:
```suggestion
    _validate_min_length = validator("components", allow_reuse=True, always=True)(validate_min_length)
```
To be consistent

maxschulz-COL (approved):
⭐ Really nice work you two ⭐ 

Basically just one question: what was the bug here (rather than the usability and confusing message) - maybe I missed something 🙃

vizro-core/src/vizro/models/_components/_form.py
antonymilne: No difference really. The private function from a private module is just more obvious that it's private because you don't need to look at the import to see that it's private. Probably I would make all of these private with the `_` in front of them for that reason, but it doesn't matter much.

antonymilne:
> ⭐ Really nice work you two ⭐
> 
> Basically just one question: what was the bug here (rather than the usability and confusing message) - maybe I missed something 🙃

Yeah, you're right - there wasn't any bug here, just an improved error message now. So I agree this would go in the changelog better as Changed or maybe even Added (like Added validation for blah blah).

vizro-core/src/vizro/models/_models_utils.py
huong-li-nguyen: I like it! 🚀

vizro-core/src/vizro/models/_components/_form.py
huong-li-nguyen: I would leave as is in this PR for now, because it was driving me a bit crazy that it was inconsistent in this file 😄 I would propose to do a separate PR where we consistently go through each file and implement above then.

PR #592, opened 2024-07-22
[Feat] Add pydantic to python conversion to base model

maxschulz-COL:
## Description
This PR adds a new feature to the `VizroBaseModel` class: the export of the model into a string that would allow re-creating the model in python.

E.g. you can now run
```python
import vizro.models as vm

df = px.data.iris()
graph = vm.Graph(figure=px.bar(df, x="sepal_width", y="sepal_length"))

print(graph._to_python())
```

will give the following (or similar as it has slightly changed):
```python
######## Module Imports ##########
from vizro import Vizro
import vizro.models as vm

########## Data Imports ##########
#####!!! UNCOMMENT BELOW !!!######
# data_manager["dfd9f054-9331-4ec5-ba78-73c7b75e9439"] = ===> Fill in here <===

###### Callable definitions ######


########## Object code ###########
graph = vm.Graph(
    figure=px.bar(
        data_frame="dfd9f054-9331-4ec5-ba78-73c7b75e9439",
        x="sepal_width",
        y="sepal_length",
    )
)
```

The main aim of this is to allow the exporting of dashboards, especially with a hint of VizroAI dashboard generation.

## Current aim of PR review
**EDIT: these questions have now been ressolved and it can be reviewed properly**
The current status of the code is not polished, hence there is room for some refactoring and refinement, but before doing that, and writing docs and unit tests, I would like some feedback on the overall directions:

- Should this be a method of the base model? If not, where else would we want it to sit? --> yes, due to other templating options
- Are we ok with the extra model field? The current only alternative to that is to modify some of pydantics source code (see below) --> fixed
- Are we happy with the `data_manager` handling, ie placeholder code for every required entry? --> yes sadly
- Should the code also handle the imports required for any custom `CapturedCallables`, ie if I use `go` or `List` or anything alike, should this code also import it? --> extra argument
- Shall we use `ruff` as a formatter, or is `black` sufficient? --> done
- We are currently manually removing the `ActionsChain` from the outputted code, and manually re-insert the underlying actions, is that acceptable? --> it is acceptable
- Should there be a mode that exports the data to files, and potentially that also exports the app to a file? This could then change source code to import data from those exported files --> not yet but maybe in future

## How to run this code
- Run the example at `hatch run example` which contains a fairly "chaotic" app, and prints that app to the terminal

## Outstanding improvements
- [x] Better imports sorting
- [x] More selective imports
- [x] Implement fallbacks for not found source code

## Previous iterations and PR explanation
Generally, there are three main functionalities to consider and to combine when thinking of a exporting a model to python code:
1. An apt representation that is representative of code (including the correct model name) that is runnable, ie something like `vm.Graph(figure=...)`
2. A correct representation of the _minimum_ required fields and their values to be configured to obtain back the model that is being exported, ie not everything including action chains, automatic model ids etc
3. Surrounding supporting definitions and imports - like the modules imported, the data_manager settings, and any custom `CapturedCallable` defined.

For 1 the most natural way to go forward is to use pydantics inherent `__repr__` method, however this leaves no control over 2. 

For 2 the most natural way is to use [pydantics powerful export functionality](https://docs.pydantic.dev/1.10/usage/exporting_models/#modeldict) `model.dict()`, where one can exclude unset fields. This however lacks the model name itself. Thus one main tasks was to combine the two. 

The current implementation focusses of using `model.dict(exclude_unset=True)` and adding an extra field to the base model that passes the model name. An implementation without this extra field [was tried out here](https://github.com/mckinsey/vizro/blob/862fc77bd3909aff8ecd4f93ee99bd58a63330e5/vizro-core/src/vizro/models/_base%20copy.py#L136), where we overwrite some of pydantics code.

## Notice


lingyielia (changes requested):
Wow it's so great to see the string output alive! 💯 

Apology that for giving comments without better solutions: 
- The data manager section of string is already powerful and handy in terms of usability. To me it just looks not perfectly streamlined, as it means the user still need to manually edit this portion. (But I know it's a very tricky problem)
- "a mode that exports the data to files", if we do ever want to go this route, is it still belongs to `_base.py`?

vizro-core/src/vizro/models/_utils.py
```diff
@@ -0,0 +1,122 @@
...
+            else:
+                other_content = ", ".join(
+                    f"{key}={_dict_to_python(value, captured_info)}" for key, value in model_data.items()
+                )
+                return f"vm.{model_name}({other_content})"
+        else:
+            return ", ".join(f"{key}={_dict_to_python(value, captured_info)}" for key, value in model_data.items())
+    elif isinstance(model_data, List):
+        return "[" + ", ".join(_dict_to_python(item, captured_info) for item in model_data) + "]"
+    elif isinstance(model_data, CapturedCallable):
```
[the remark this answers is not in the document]
maxschulz-COL: Hm very interesting - I am sure we can get around this because we do have the code for this. I suspect the root cause of this is because it is never imported so that the inspect module cannot find the source.

vizro-core/src/vizro/models/_base.py
```diff
@@ -101,8 +118,60 @@ def _is_discriminated_union(field):
...
+        model_code = f"{str(model_dict.get('model_name','object')).lower()} = " +  _dict_to_python(model_dict, captured_info)
+        import_statements =  STANDARD_IMPORT_PATHS | _get_import_statements(captured_info)
+        data_setting = _get_data_manager_code_strings(captured_info)
+        callable_definitions = _get_callable_code_strings(captured_info)
+
+        concatenated_code = "\n\n".join(
+            [
+                "######## Module Imports ##########\n" + "\n".join((import_statements)),
+                "########## Data Imports ##########\n#####!!! UNCOMMENT BELOW !!!######\n" + "\n".join((data_setting)),
+                "###### Callable definitions ######\n" + ("\n".join(callable_definitions)),
```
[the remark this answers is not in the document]
maxschulz-COL: Yes I was thinking that too - it just looks too ugly without any lines of code below

vizro-core/src/vizro/models/_utils.py
```diff
@@ -0,0 +1,122 @@
...
+    detect_path: str
+    replace_path: str
+    from_import: Callable
+
+
+@dataclass
+class CapturedCallableInfo:
+    name: str
+    module: str
+    args: List[Tuple[str, Any]]
```
antonymilne, line 21: Maybe easier to keep this as a dict or whatever `CapturedCallable._arguments` returns?

antonymilne (commented):
Wow, this looks great! I thought it would be a fun project and it's really nice to see it all coming together ⭐ 🌟 ⭐ Really nice work.

I'll need to spend more time going through `_dict_to_python` and the import replacement, but basically if it seems to generally work then I'm happy with it even if it's not absolutely 100% robust 👍 

There's basically three things I'm not sure about:
1. whether there's any way to hide the `model_name` field - let me take a look at this. Did you find that making it private didn't work?
2. how to handle imports - as discussed today, this is probably the most problematic and fragile part. In reality maybe it works great 99% of the time, which is fine by me. But maybe as per today's chat about how to handle imports that arise in captured callable functions it actually has problems more than 1% of the time? In which case possibly it's worth changing system here
3. is there some way we can replace those randomly generated dataset names with something more readable? There's nothing in the data manager that will tell you where the name comes from, but we could just do [something like this](https://stackoverflow.com/questions/53847404/how-to-check-uuid-validity-in-python) to see if it's randomly generated and then if it is then just rename it to `data_1` etc.?

Also, tests here are going to be very important in order to ensure we don't break stuff when we upgrade to pydantic v2 😬 I wonder if you can do these circularly, like:
```
# dashboard_example.py
import vizro.models as vm
dashboard = vm.Dashboard(pages=[vm.Page(...)])

# in test:
from dashboard_example import dashboard
assert dashboard.to_python() == Path("dashboard_example.py").read_text()
```

antonymilne:
Thanks for the very useful description and guidance about what to review, it's very helpful!

> Should this be a method of the base model? If not, where else would we want it to sit?

I think yes. The only other place I could think of putting it would be a new function `model_to_python(model: VizroBaseModel)` but I don't see any particular advantage in doing that.

> Are we ok with the extra model field? The current only alternative to that is to modify some of pydantics source code (see below)

Not very happy about it but will tolerate it if there's no better solution. This is my only real problem with the current approach. Let me see if I can come up with an alternative, but if I can't then this is ok.

> Are we happy with the data_manager handling, ie placeholder code for every required entry?

Discussed earlier today, but yes.

> Should the code also handle the imports required for any custom CapturedCallables, ie if I use go or List or anything alike, should this code also import it?

Also discussed earlier today and sent you lots of possibly useful links!

> Shall we use ruff as a formatter, or is black sufficient?

Definitely ruff or at least something else that will remove unwanted imports. It doesn't matter if ruff "fails" the linting - we just run it with `--fix` on and take whatever it gives us as a result and that will be fine. It turns out black doesn't actually have an "official" Python API either (I use what looks a lot like a nice Python API but is not officially stable in the changes I made to the chart gallery dashboard). So there's not a big advantage there over doing a subprocess for `ruff`.

ruff should solve the two outstanding TODOs on import sorting and unwanted imports.

> We are currently manually removing the ActionsChain from the outputted code, and manually re-insert the underlying actions, is that acceptable?

Not sure I understand this part of the code - let's discuss. But probably this is fine.

> Should there be a mode that exports the data to files, and potentially that also exports the app to a file? This could then change source code to import data from those exported files

Discussed earlier - I think no need for now, maybe something for the future if it seems useful.

vizro-core/src/vizro/models/_base.py
```diff
@@ -29,6 +39,13 @@ class VizroBaseModel(BaseModel):
         description="ID to identify model. Must be unique throughout the whole dashboard."
         "When no ID is chosen, ID will be automatically generated.",
     )
+    model_name: str = Field("", description="Name of the model.")
+    _private_attr: str = ""
```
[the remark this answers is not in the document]
maxschulz-COL: No it was just used for playing around - good catch though.

As per your above question - no it does not work with it, it basically contradicts the idea of exporting a field that is known..

vizro-core/src/vizro/models/_base.py
```diff
@@ -101,8 +118,60 @@ def _is_discriminated_union(field):
         cls.update_forward_refs(**vm.__dict__.copy())
         new_type.update_forward_refs(**vm.__dict__.copy())
 
+    def to_python(self):
+        model_dict = self.dict(exclude_unset=True)
+        captured_info = []
+        model_code = f"{str(model_dict.get('model_name','object')).lower()} = " +  _dict_to_python(model_dict, captured_info)
```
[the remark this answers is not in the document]
maxschulz-COL: I was thinking that this would be the default, but only `vm.VizroBaseModel` would have the `to_python` method, so it is redundant!

vizro-core/src/vizro/models/_base.py
```diff
@@ -101,8 +118,60 @@ def _is_discriminated_union(field):
...
+        import_statements =  STANDARD_IMPORT_PATHS | _get_import_statements(captured_info)
+        data_setting = _get_data_manager_code_strings(captured_info)
+        callable_definitions = _get_callable_code_strings(captured_info)
+
+        concatenated_code = "\n\n".join(
```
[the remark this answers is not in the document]
maxschulz-COL: Also makes it more flexible in case some of the sections do not apply

vizro-core/src/vizro/models/_base.py
```diff
@@ -101,8 +118,60 @@ def _is_discriminated_union(field):
...
+                "########## Data Imports ##########\n#####!!! UNCOMMENT BELOW !!!######\n" + "\n".join((data_setting)),
+                "###### Callable definitions ######\n" + ("\n".join(callable_definitions)),
+                "########## Object code ###########\n" + model_code,
+            ]
+        )
+
+        # check for formatting: https://mckinsey-hub.slack.com/archives/D03N7KXFXV3/p1721401376669179?thread_ts=1721400781.371419&cid=D03N7KXFXV3
+        #TODO: potentially use ruff over just black
+
+        return format_str(concatenated_code, mode=FileMode(line_length=88))
```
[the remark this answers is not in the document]
maxschulz-COL: Shall we go for the subprocess then?

vizro-core/examples/scratch_dev/app.py
```diff
@@ -57,11 +64,10 @@
...
-page2 = vm.Page(
-    title="Page2", components=[vm.Graph(id="hist_chart2", figure=px.histogram(df, x="sepal_width", color="species"))]
-)
-
 dashboard = vm.Dashboard(pages=[page, page2])
 
 if __name__ == "__main__":
-    Vizro().build(dashboard).run()
+    # print(dashboard.dict(exclude_unset=True, exclude_defaults=True))
+    string, info = dashboard.to_python()
```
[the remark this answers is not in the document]
maxschulz-COL: Yes yes, sorry this was still WIP, I think I changed it in some commit, otherwise will in future

vizro-core/src/vizro/models/_utils.py
```diff
@@ -0,0 +1,122 @@
...
+    from vizro.models.types import CapturedCallable  # TODO: can we get rid of this?
+
+    if captured_info is None:
+        captured_info = []
+
+    if isinstance(model_data, Dict):
+        if MODEL_NAME in model_data:
+            model_name = model_data.pop(MODEL_NAME)
+            if model_name == ACTIONS_CHAIN:
+                action_data = model_data[ACTION]
```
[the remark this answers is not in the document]
maxschulz-COL: My guide is always - if used twice, put it in a constant. But then I felt the other strings looked lonely, so I converted them too :)

vizro-core/src/vizro/models/_utils.py
```diff
@@ -0,0 +1,122 @@
...
+                    return _dict_to_python(action_data, captured_info)
+            else:
+                other_content = ", ".join(
+                    f"{key}={_dict_to_python(value, captured_info)}" for key, value in model_data.items()
+                )
+                return f"vm.{model_name}({other_content})"
+        else:
+            return ", ".join(f"{key}={_dict_to_python(value, captured_info)}" for key, value in model_data.items())
+    elif isinstance(model_data, List):
+        return "[" + ", ".join(_dict_to_python(item, captured_info) for item in model_data) + "]"
```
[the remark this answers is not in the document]
maxschulz-COL: I fiddled with this a lot, and it was quite tricky to get right. I need to remind myself of the details again, but there was a very annoying problem with the string containing an extra set of `'` where it shouldn't. Can take a look again 👍

vizro-core/src/vizro/models/_utils.py
```diff
@@ -0,0 +1,122 @@
...
+                else:
+                    return _dict_to_python(action_data, captured_info)
+            else:
+                other_content = ", ".join(
+                    f"{key}={_dict_to_python(value, captured_info)}" for key, value in model_data.items()
+                )
+                return f"vm.{model_name}({other_content})"
+        else:
+            return ", ".join(f"{key}={_dict_to_python(value, captured_info)}" for key, value in model_data.items())
+    elif isinstance(model_data, List):
```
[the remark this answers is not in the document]
maxschulz-COL: For example when we have the components, they are themselves models in a list

vizro-core/src/vizro/models/types.py
```diff
@@ -245,6 +247,20 @@ def _check_type(cls, captured_callable: CapturedCallable, field: ModelField) ->
 
         return captured_callable
 
+    def __repr__(self):
+        args = ", ".join(f"{key}={value!r}" for key, value in self._arguments.items())
+        x = f"{self._function.__module__}.{self._function.__name__}({args})"
+        return x
+
+    def _repr_clean(self):
```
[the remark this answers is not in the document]
maxschulz-COL: Yes exactly, to get pretty namespace aliases

vizro-core/src/vizro/models/_utils.py
maxschulz-COL: Sure can also do that

vizro-core/src/vizro/models/_base.py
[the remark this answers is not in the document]
antonymilne: Yeah, let's do something that looks roughly like `code = subprocess.run(["ruff", "--fix", "-"], input=input_code, capture_output=True)` and then use `code.stdout`. You'll need to check the subprocess docs and play around to get stuff like str vs. bytes and newlines working and also to make sure that we don't capture anything other than the raw code output (like we don't want the text that says "ruff fixed 10 lines" in the output code, and even if ruff fails to fix the code we still want the code and not a crash).

vizro-core/src/vizro/models/_utils.py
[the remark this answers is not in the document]
maxschulz-COL: This is ressolved now

vizro-core/hatch.toml
```diff
@@ -26,7 +26,8 @@ dependencies = [
   "chromedriver-autoinstaller>=0.6.4",
   "toml",
   "pyyaml",
-  "openpyxl"
+  "openpyxl",
+  "jupyter",
```
maxschulz-COL, line 30: I would pledge to add this, make life so much better

vizro-core/src/vizro/models/types.py
```diff
@@ -245,6 +247,19 @@ def _check_type(cls, captured_callable: CapturedCallable, field: ModelField) ->
 
         return captured_callable
 
+    def __repr__(self):
+        args = ", ".join(f"{key}={value!r}" for key, value in self._arguments.items())
+        x = f"{self._function.__module__}.{self._function.__name__}({args})"
+        return x
```
antonymilne, line 253: These are still the quick and dirty versions and should be tidied:

```suggestion
    def __repr__(self):
        args = ", ".join(f"{key}={value!r}" for key, value in self._arguments.items())
        return f"{self._function.__module__}.{self._function.__name__}({args})"
```

vizro-core/src/vizro/models/types.py
```diff
@@ -245,6 +247,19 @@ def _check_type(cls, captured_callable: CapturedCallable, field: ModelField) ->
...
+
+    def _repr_clean(self):
+        """Alternative __repr__ method with cleaned module paths."""
+        args = ", ".join(f"{key}={value!r}" for key, value in self._arguments.items())
+        module_path = f"{self._function.__module__}"
+        modified_module_path = _clean_module_string(module_path)
+        x = f"{modified_module_path}{self._function.__name__}({args})"
+        return x
```
antonymilne, line 261:
```suggestion
    def __repr_clean__(self):
        """Alternative __repr__ method with cleaned module paths used by to_python."""
        args = ", ".join(f"{key}={value!r}" for key, value in self._arguments.items())
        return f"{_clean_module_path(modified_module_path)}{self._function.__name__}({args})"
```

vizro-core/src/vizro/models/_utils.py
```diff
@@ -0,0 +1,167 @@
...
+            return f"vm.{__vizro_model__}({fields})"
+    elif isinstance(object, dict):
+        fields = ", ".join(f"{field_name}={_dict_to_python(value)}" for field_name, value in object.items())
+        return "{" + fields + "}"
+    elif isinstance(object, list):
+        # Need to do this manually to avoid extra quotation marks that arise when doing repr(List).
+        code_string = ", ".join(_dict_to_python(item) for item in object)
+        return f"[{code_string}]"
+    elif isinstance(object, CapturedCallable):
+        return object._repr_clean()
```
antonymilne, line 102:
```suggestion
        return object.__repr_clean__()
```

vizro-core/src/vizro/models/_utils.py
```diff
@@ -0,0 +1,167 @@
...
+# Note also that these functions find also unintended model_manager additions, a known but accepted limitation.
+def _extract_captured_callable_source() -> Set[str]:
+    from vizro.models.types import CapturedCallable
+
+    captured_callable_sources = set()
+    for model_id in model_manager:
+        for _, value in model_manager[model_id]:
+            if isinstance(value, CapturedCallable) and all(
+                replacement.original not in value._function.__module__ for replacement in REPLACEMENT_STRINGS
+            ):
```
antonymilne, line 135: Just to understand, why is the comparison here substring rather than `==`?

i.e. if we changed `REPLACEMENT_STRINGS` then why would this be wrong?
```suggestion
            if isinstance(value, CapturedCallable) and value._function.__module__ not in REPLACEMENT_STRINGS):
```

vizro-core/src/vizro/models/_utils.py
```diff
@@ -0,0 +1,167 @@
...
+                try:
+                    source = textwrap.dedent(inspect.getsource(value._function))
+                    captured_callable_sources.add(source)
+                except OSError:
+                    pass
```
antonymilne, line 140:
```suggestion
from contextlib import suppress

                with suppress(OSError):
                    # SOME COMMENT EXPLAINING WHY OSError COULD BE RAISED
                    source = textwrap.dedent(inspect.getsource(value._function))
                    captured_callable_sources.add(source)
```

vizro-core/src/vizro/models/_utils.py
```diff
@@ -0,0 +1,167 @@
...
+    return captured_callable_sources
+
+
+def _extract_captured_callable_data_info() -> Set[str]:
+    from vizro.models.types import CapturedCallable
+
+    return {
+        f'# data_manager["{value._arguments["data_frame"]}"] = ===> Fill in here <==='
```
antonymilne, line 148:
```suggestion
        f'# data_manager["{value["data_frame"]}"] = ===> Fill in here <==='
```

vizro-core/src/vizro/models/_utils.py
antonymilne, line 1: Our usual policy is to put helper functions in the same module as where the function is used. Only if it's used in multiple files do we need a dedicated utils file for it.

vizro-core/tests/unit/vizro/models/test_utils.py
```diff
@@ -0,0 +1,248 @@
...
+from vizro.actions import export_data
+from vizro.models._utils import _concatenate_code
+from vizro.models.types import capture
+
+extra_imports_string = "import vizro.models as vm\nimport pandas as pd"
+code_string = "vm.Card(text='Foo')"
+callable_defs_string = """def f():\n  return 'hi'"""
+data_settings_string = """# data_manager["iris"] = ===> Fill in here <==="""
+
+expected_assembled_linted_code = textwrap.dedent(
```
antonymilne, line 16: What do we need the dedent for here?

vizro-core/tests/unit/vizro/models/test_utils.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+####### Data Manager Settings #####
+#######!!! UNCOMMENT BELOW !!!#####
+# from vizro.managers import data_manager
+# data_manager["iris"] = ===> Fill in here <===
+
+
+########### Model code ############
+vm.Card(text="Foo")
```
antonymilne, line 33: I was expecting `model = ` here?

vizro-core/tests/unit/vizro/models/test_utils.py
```diff
@@ -0,0 +1,248 @@
...
+        @capture("graph")
+        def chart_dynamic(data_frame, hover_data: Optional[List[str]] = None):
+            return px.bar(data_frame, x="sepal_width", y="sepal_length", hover_data=hover_data)
+        """
+    )
+
+    local_scope = {}
+    exec(function_string, globals(), local_scope)
+    chart_dynamic = local_scope["chart_dynamic"]
+    return chart_dynamic
```
antonymilne, line 84:
```suggestion
    exec(function_string)
    return locals()["chart_dynamic"]
```

Not 100% sure but I think this will do what you're looking for a bit more simply and conventionally.

antonymilne: Actually I have no idea why we're doing `exec` here at all? 🤔 What am I missing? Why can't we just return the `chart_dynamic` function? We don't need it in a fixture if that's making it more awkward?

antonymilne (commented):
👏 This is awesome! Really great work - I know first hand how fiddly this was to get right, but the results are great ⭐ 🌟 ⭐ Very close to ✅ from me - just a few comments and questions, mainly minor stuff. I didn't actually try out the code but I trust your testing + some other independent reviewer on that.

One thing that would give me some more confidence (e.g. when it comes to making sure this still works the same when we upgrade to pydantic v2) is if we had some more "real" end to end examples that test some more complex cases e.g. whole `vm.Dashboard.to_python`. Currently the tests (which are great) only handle simple toy examples and nothing is actually testing the `_dict_to_python` since the tests look at isolated models and not trees of models with parents and children. The best way to do that would be through just making some more nested `vm.Dashboard` objects and doing `to_python` on them. Happy to not add these before merging though if you can't be bothered.

One thing I realised about the randomly generated data source names which look kind of weird: we might fix this in future by generating a better name when we add to the `data_manager`. This is probably better than trying to rename in the `to_python` stuff here. It's not a priority at all though because the random names are acceptable and won't be used for Vizro-AI anyway which is the main current use case.

vizro-core/hatch.toml
antonymilne: Fine by me

vizro-core/src/vizro/models/_utils.py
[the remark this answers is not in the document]
antonymilne: This still confused me upon re-reading the code fresh a couple of weeks after I first made this comment.

IMO the problem with the constants is that they just introduce an unnecessary indirection since when I read the code I have to click elsewhere to find out what `ACTION` etc. are rather than it just being there directly in the code. Especially when `ACTION` is literally just `"action"`.

Happy to see what a new reviewer thinks though if you disagree 🙂

vizro-core/src/vizro/models/_base.py
```diff
@@ -101,8 +108,116 @@ def _is_discriminated_union(field):
...
+        return _dict
+
+    # This is like pydantic's own __exclude_fields__ but safer to use (it looks like __exclude_fields__ no longer
+    # exists in pydantic v2).
+    # Root validators with pre=True are always included, even when exclude_default=True, and so this is needed
+    # to potentially exclude fields set this way, like Page.id.
+    def __vizro_exclude_fields__(self) -> Optional[Union[Set[str], Mapping[str, Any]]]:
+        return None
+
+    def _to_python(self, extra_imports: Optional[Set[str]] = None, extra_callable_defs: Set[str] = set()) -> str:
```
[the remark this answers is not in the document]
maxschulz-COL: You are completely right - what an oversight 🙈

vizro-core/src/vizro/models/_page.py
```diff
@@ -87,6 +87,9 @@ def __init__(self, **data):
                 f"as the page title. If you have multiple pages with the same title then you must assign a unique id."
             ) from exc
 
+    def __vizro_exclude_fields__(self) -> Set[str]:
```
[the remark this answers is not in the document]
maxschulz-COL: I see, yes I misunderstood that 👍

vizro-core/src/vizro/models/_utils.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+@dataclass
+class PathReplacement:
+    original: str
+    new: str
+
+
+REPLACEMENT_STRINGS = [
+    PathReplacement("plotly.express", "px."),
```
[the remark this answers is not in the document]
maxschulz-COL: It is more convenient in the case of no module string - currently there is no `.` in the overall template, which is good if we define the callable in the same file (as for the captured callables)

vizro-core/examples/scratch_dev/app.py
```diff
@@ -57,11 +60,10 @@
...
-page2 = vm.Page(
-    title="Page2", components=[vm.Graph(id="hist_chart2", figure=px.histogram(df, x="sepal_width", color="species"))]
-)
-
 dashboard = vm.Dashboard(pages=[page, page2])
 
 if __name__ == "__main__":
-    Vizro().build(dashboard).run()
+    # print(dashboard.dict(exclude_unset=True, exclude_defaults=True))
+    string = dashboard._to_python()
```
petar-qb, line 67:
```suggestion
    string = dashboard._to_python(extra_imports={"from dash_ag_grid import AgGrid"})
```

vizro-core/src/vizro/models/_base.py
```diff
@@ -101,8 +108,116 @@ def _is_discriminated_union(field):
...
+        """Converts a Vizro model to the Python code that would create it.
+
+        Args:
+            extra_imports: Extra imports to add to the Python code. Provide as a set of complete import strings.
+            extra_callable_defs: Extra callable definitions to add to the Python code. Provide as a set of complete
+                function definitions.
```
petar-qb, line 135: Should we give examples on how to add `extra_imports` and `extra_callable_defs` here or only in the docs?

vizro-core/src/vizro/models/_base.py
```diff
@@ -101,8 +108,116 @@ def _is_discriminated_union(field):
...
+        # Inject __vizro_model__ into the dictionary - needed just for to_python.
+        _dict["__vizro_model__"] = self.__class__.__name__
+        return _dict
+
+    # This is like pydantic's own __exclude_fields__ but safer to use (it looks like __exclude_fields__ no longer
+    # exists in pydantic v2).
+    # Root validators with pre=True are always included, even when exclude_default=True, and so this is needed
+    # to potentially exclude fields set this way, like Page.id.
+    def __vizro_exclude_fields__(self) -> Optional[Union[Set[str], Mapping[str, Any]]]:
+        return None
```
petar-qb, line 127: If `__vizro_exclude_fields__()` returns None, could we replace (I'm probably misunderstanding something):
```py
        if kwargs.get("exclude") is None:
            kwargs["exclude"] = self.__vizro_exclude_fields__()
```
with just:
```py
kwargs.setdefault("exclude", None)
```

vizro-core/src/vizro/models/_base.py
```diff
@@ -101,8 +108,116 @@ def _is_discriminated_union(field):
...
+        # Imports
+        extra_imports_concat = "\n".join(extra_imports) if extra_imports else None
+
+        # CapturedCallable definitions
+        callable_defs_set = _extract_captured_callable_source() | extra_callable_defs
+        callable_defs_concat = "\n".join(callable_defs_set) if callable_defs_set else None
+
+        # Data Manager
+        data_defs_list = _extract_captured_callable_data_info()
+        data_defs_concat = "\n".join(data_defs_list) if data_defs_list else None
```
petar-qb, line 154:
```suggestion
        data_defs_set = _extract_captured_callable_data_info()
        data_defs_concat = "\n".join(data_defs_set) if data_defs_set else None
```

vizro-core/src/vizro/models/_utils.py
maxschulz-COL: Similar reason as to above, I would like to avoid trying to match exact module paths, like `plotly.express._graph_obj` or similar. It bears a small risk, but will overall be more robust if there is alternative to import paths, but when the main content machtes

vizro-core/src/vizro/models/_base.py
petar-qb, line 154: Minor suggestion: To enhance readability and maintain consistency, consider reordering the sections of the code as follows:

- #Imports
- #CapturedCallable Definitions
- #Data Manager
- #Model

If you agree, please adjust the order of the `_concatenate_code` arguments and the `TO_PYTHON_TEMPLATE.format` arguments accordingly to reflect this structure.
```py
TO_PYTHON_TEMPLATE.format(
        extra_imports=extra_imports if extra_imports else "",
        callable_defs=callable_defs,
        data_settings=data_settings,
        code=code,
)
```

vizro-core/src/vizro/models/_utils.py
maxschulz-COL: I opted for

```py
                try:
                    source = textwrap.dedent(inspect.getsource(value._function))
                    captured_callable_sources.add(source)
                except OSError:
                    # OSError is raised when the source code is not available. This is expected 
                    # for built-in functions or dynamically defined functions (via exec or eval).
                    logging.warning(f"Could not extract source for {value._function}. Definition will not be included.")
                    pass
```

petar-qb (approved):
This is a magnificent piece of work @maxschulz-COL 🎉 🥇 

After addressing my comments and others, this PR will be ready to merge by my opinion.

vizro-core/tests/unit/vizro/models/test_utils.py
maxschulz-COL: It is true it is not necessary, but it was my swiss army knife for code strings, so I put it everywhere where a string was defined, in case we every moved something to an indented file. Will remove if you prefer, no strong optinion

vizro-core/tests/unit/vizro/models/test_utils.py
maxschulz-COL: We want to test the case where the source code does not exist. At first I thought of mocking it, but I think it is even better to test a realistic case. Wdyt?

maxschulz-COL: Applied your changes though, thanks!

vizro-core/src/vizro/models/_base.py
maxschulz-COL: Currently it is a private function, and it is not even added in the docs. @antonymilne do you think we should:
- make it non-private
- add it to the docs
- announce it in release notes

Any combination of the above?

vizro-core/src/vizro/models/_base.py
maxschulz-COL: Good point, implemented the new order

vizro-core/src/vizro/models/_utils.py
```diff
@@ -0,0 +1,167 @@
...
+    linted = subprocess.check_output(
+        ["ruff", "check", "--fix", "--exit-zero", "--silent", "--isolated", "-"], input=code_string, encoding="utf-8"
+    )
+    formatted = subprocess.check_output(
+        ["ruff", "format", "--silent", "--isolated", "-"], input=linted, encoding="utf-8"
+    )
+    return formatted
+
+
+def _clean_module_string(module_string: str) -> str:
```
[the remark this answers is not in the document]
maxschulz-COL: My last comment got lost somehow.. :(

So the idea of the function is essentially the following:
- iterate through the list of replacements, if you match a substring, take this as the module replacement
- if not replacement is found, then return ""


The idea is to match substrings, because (as explained in another comment) we do not want to target precise modules for extra stability, ie something like `plotly.express._graph_obj.abcs`.

And yes, there is only ever one match.

Let the code as is for now

vizro-core/src/vizro/models/_utils.py
maxschulz-COL: Okies - didnt want to clog up the base file, but moved it there now

vizro-core/src/vizro/models/_base.py
```diff
@@ -101,6 +218,72 @@ def _is_discriminated_union(field):
...
 
+    def dict(self, **kwargs):
+        # Overwrite pydantic's own `dict` method to add __vizro_model__ to the dictionary.
+        # To get exclude as an argument is a bit fiddly because this function is called recursively
+        # inside pydantic, which sets exclude=None by default.
+        if kwargs.get("exclude") is None:
+            kwargs["exclude"] = self.__vizro_exclude_fields__()
+        _dict = super().dict(**kwargs)
+        # Inject __vizro_model__ into the dictionary - needed just for to_python.
+        _dict["__vizro_model__"] = self.__class__.__name__
```
lingyielia, line 229: Curious why `__vizro_model__` cannot be directly defined in each Vizro model. 

In vizro-ai, the card create is using `dict` in the flow of `Card` -> `Card.dict()` -> `replace with a readable id` -> `Card.parse_obj(proxy_dict)`, which would fail validation as 
```
ValidationError: 1 validation error for Card
__vizro_model__
  extra fields not permitted (type=value_error.extra)
  ```
  
it can be fixed by adding `proxy_dict.pop("__vizro_model__", None)`, but also wondering how often would this cause error, when a user would like to do ".dict()" then ".parse_obj()" (I guess not too often)

Comment (maxschulz-COL) [reply vizro-core/src/vizro/models/_base.py:229]: This is a good point! And something I had not considered as such. In a previous iteration it did exist as a field, but @antonymilne preferred not to add it as such. Curious to see what he thinks!

Comment (maxschulz-COL) [reply vizro-core/src/vizro/models/_base.py:229]: For our case a couple of additional thoughts:
- we could also use the `exclude` argument of `dict()` as an alternative fix
- why do we go this route again? Is it important to have a readable ID? I'd prefer to not user proxies unless needed (like for the filter control)

File: vizro-core/src/vizro/models/_utils.py (hunk 18 above)
Comment (antonymilne) [reply vizro-core/src/vizro/models/_utils.py]: Nice, looks good 👍

File: vizro-core/src/vizro/models/_utils.py (hunk 17 above)
Comment (antonymilne) [reply vizro-core/src/vizro/models/_utils.py]: Oh ok, I see 👍 Thanks.

File: vizro-core/src/vizro/models/_utils.py (hunk 30 above)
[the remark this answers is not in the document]
Comment (antonymilne) [reply vizro-core/src/vizro/models/_utils.py]: ok, this makes more sense now 🙂 

Please can you add a brief comment explaining the above?

Personally I would still find it clearer as a Python mapping + iteration rather than list of dataclasses + `next`, like this. But up to you.
```
REPLACEMENT_STRINGS = {"plotly.express": "px.", ...}

def _clean_module_string(module_string: str) -> str:
     # EXPLANATORY COMMENT HERE 
     for module_original, module_alias in REPLACEMENT_STRINGS.items():
        if module_original in module_string:
            return module_alias
```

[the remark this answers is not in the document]
Comment (antonymilne) [reply vizro-core/src/vizro/models/_utils.py]: Actually wait, no... I still don't quite understand 😅 Let's have a quick call about it today to get to the bottom of it!

File: vizro-core/tests/unit/vizro/models/test_utils.py (hunk 22 above)
Comment (antonymilne) [reply vizro-core/tests/unit/vizro/models/test_utils.py]: Ohhh I see, so the whole point here is that if you write this out as a "normal" function then it will exist in the source code of test_utils.py and hence you won't be able to test the "code is not available" case. This makes sense now.

The comment in the test is good but if you added something to this fixture just to explain what it's doing then that would be great.

File: vizro-core/src/vizro/models/_base.py (hunk 27 above)
Comment (antonymilne) [reply vizro-core/src/vizro/models/_base.py:247]: For now I would:
* keep it private
* not should it in docs
* not add to changelog
* incorporate @petar-qb's suggestion of adding example to the docstrings because this would be useful anyway

Once it's used in anger by Vizro-AI and we're definitely happy with it let's then make it public and add to docs and changelog. Just in case e.g. we decide that the current API is actually not convenient for Vizro-AI and we want to change it (shouldn't happen because @maxschulz-COL already thought this through, but still there's no big rush to make it public given that Vizro-AI is the primary consumer to begin with).

File: vizro-core/src/vizro/models/_base.py (hunk 31 above)
Comment (antonymilne) [reply vizro-core/src/vizro/models/_base.py:229]: Very good point indeed @lingyielia.

I think the best solution is to make our overriding of `dict` less disruptive like this:
```
        # Dunder properties like __vizro_model__ are ignored even when explicitly in `include` so we need to add this
        # manually.
        if kwargs.get("include") is not None and "__vizro_model__" in kwargs["include"]:
            _dict["__vizro_model__"] = self.__class__.__name__
```

And then in `_to_python` we do `dict(include={"__vizro_model__"})` to explicitly include it.

This way the default `dict` behaviour is left the same as the pydantic default which feels better.

antonymilne (approved):
Awesome work! 👏 

The extra tests looks great.

For me there's just two things to deal with before merging:
* let's talk about the module path replacement because I still don't understand, sorry
* see if my suggested fix for @lingyielia's problem with the `__vizro_model__` being included looks good to you

vizro-core/src/vizro/models/_base.py
antonymilne: P.S. probably worth adding a new test to ensure that the basic `.dict()` functionality remains unaltered i.e. does not include `__vizro_model__`. So we don't accidentally break this in future.

lingyielia: > * Is it important to have a readable ID?

Yes, right now all components (graph, table, and card) have readable IDs, which will be used by filter and layout creation.

lingyielia: I like the idea "basic .dict() functionality remains unaltered" so it's not disruptive in any possible cases 🙌

maxschulz-COL: ressolved now by using global `_PATCH_VIZRO_BASE_MODEL_DICT`

vizro-core/src/vizro/models/_utils.py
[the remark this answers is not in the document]
maxschulz-COL: Ressovled with simplified data structure for REPLACEMENT_STRINGS

Issue #612, opened 2024-08-01
[Bug] Marks are displayed incorrectly in Sliders


huong-li-nguyen:
### Which package?

vizro

### Package version

recent

### Description

Question: How to use the marks argument in vm.Slider? (I want all keys and values in the marks dictionary to be shown)
I use it twice in what appears the same way (cf code), but get different outcomes (cf image).

![Screenshot 2024-07-31 at 17 26 53 (1)](https://github.com/user-attachments/assets/8be06e2e-52e3-458f-bd57-04fb150da566)

```
vm.Parameter(
            targets=["waterfall.iv"],
            selector=vm.Slider(
                type="slider",
                min=0,
                max=1,
                value=MARKET_IV,
                title="Implied volatility",
                step=0.01,
                marks={0: "0%", MARKET_IV: "Market", 1: "100%"},  # <---------------- 0 NOT SHOWN, MIDDLE SHOWN LEFT (MARKET_IV=0.21)
            ),
),
vm.Parameter(
            targets=["waterfall.S"],
            selector=vm.Slider(
                type="slider",
                min=0,
                max=200,
                value=60,
                title="Flat price",
                step=1,
                marks={0: "0", MARKET_FLAT: "Market", 200: "200"},  # <-------------- LEFT MIDDLE AND RIGHT SHOWN AS INTENDED
            ),
),
```

### How to Reproduce

-

### Code of Conduct

- [X] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

Issue #230, opened 2023-12-19
BUG: `value` for slider does not work properly


maxschulz-COL:
The below example does not set the slider default (3) properly. It seems that the custom action `on_page_load` fetches the minimum.

```python
import vizro.models as vm
import vizro.plotly.express as px
from vizro import Vizro
from vizro.models.types import capture


@capture("graph")
def scatter_with_line(data_frame, x, y, color=None, size=None, hline=None):
    fig = px.scatter(data_frame=data_frame, x=x, y=y, color=color, size=size)
    fig.add_hline(y=hline, line_color="gray")
    return fig


page_0 = vm.Page(
    title="Custom chart",
    components=[
        vm.Graph(
            id="enhanced_scatter",
            figure=scatter_with_line(
                x="sepal_length",
                y="sepal_width",
                color="species",
                size="petal_width",
                hline=3,
                data_frame=px.data.iris(),
            ),
        ),
    ],
    controls=[
        vm.Parameter( 
            targets=["enhanced_scatter.hline"],
            selector=vm.Slider(min=2, max=5, step=1, value=3, title="Horizontal line"),
        ),
    ],
)
dashboard = vm.Dashboard(pages=[page_0])

Vizro().build(dashboard).run()
```

PR #613, opened 2024-08-01
[Bug] Fix display of marks in `vm.Slider` and `vm.RangeSlider`

huong-li-nguyen:
## Description
Closes: https://github.com/mckinsey/vizro/issues/612: Previously, our marks were converted to integers, leading to incorrect value displays. This, combined with an existing bug in Dash, caused the marks to be displayed incorrectly.

- Fix typing inside `vm.Slider` and `vm.RangeSlider`
- Added a validator to consistently format floats to integers when possible, thus circumventing this Dash bug: https://github.com/plotly/dash-core-components/issues/159#issuecomment-380581043)

@nadijagraca - I noticed that we previously had the typing `marks: Optional[Dict[float, str]]`, but it was changed in this PR: https://github.com/mckinsey/vizro/issues/230. Do you remember the reason for this change? I couldn't reproduce the original bug that the PR fixes, so I'm uncertain if reverting to the original typing will cause any issues. Could you also take a look? 🤔

## Notice

### DIFF
--- a/vizro-core/src/vizro/models/_components/form/slider.py
+++ b/vizro-core/src/vizro/models/_components/form/slider.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Literal, Optional, Union
+from typing import Dict, List, Literal, Optional
 
 from dash import ClientsideFunction, Input, Output, State, clientside_callback, dcc, html
 
@@ -43,7 +43,7 @@ class Slider(VizroBaseModel):
     min: Optional[float] = Field(None, description="Start value for slider.")
     max: Optional[float] = Field(None, description="End value for slider.")
     step: Optional[float] = Field(None, description="Step-size for marks on slider.")
-    marks: Optional[Dict[int, Union[str, Dict[str, Any]]]] = Field({}, description="Marks to be displayed on slider.")
+    marks: Optional[Dict[float, str]] = Field({}, description="Marks to be displayed on slider.")
     value: Optional[float] = Field(None, description="Default value for slider.")
     title: str = Field("", description="Title to be displayed.")
     actions: List[Action] = []
--- a/vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
+++ b/vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
@@ -229,15 +229,19 @@ class TestRangeSliderInstantiation:
         "marks, expected",
         [
             ({i: str(i) for i in range(0, 10, 5)}, {i: str(i) for i in range(0, 10, 5)}),
-            ({15: 15, 25: 25}, {15.0: "15", 25.0: "25"}),
-            ({"15": 15, "25": 25}, {15.0: "15", 25.0: "25"}),
+            ({15: 15, 25: 25}, {15: "15", 25: "25"}),  # all int
+            ({15.5: 15.5, 25.5: 25.5}, {15.5: "15.5", 25.5: "25.5"}),  # all floats
+            ({15.0: 15, 25.0: 25}, {15: "15", 25: "25"}),  # all floats, but convertible to int
+            ({"15": 15, "25": 25}, {15: "15", 25: "25"}),  # all string
             (None, None),
         ],
     )
     def test_valid_marks(self, marks, expected):
         range_slider = vm.RangeSlider(min=0, max=10, marks=marks)
-
         assert range_slider.marks == expected
+
+        if marks:
+            assert all(type(key) is type(next(iter(expected.keys()))) for key in range_slider.marks)
 
     def test_invalid_marks(self):
         with pytest.raises(ValidationError, match="2 validation errors for RangeSlider"):
--- a/vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
+++ b/vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
@@ -231,7 +231,7 @@ class TestRangeSliderInstantiation:
             ({i: str(i) for i in range(0, 10, 5)}, {i: str(i) for i in range(0, 10, 5)}),
             ({15: 15, 25: 25}, {15: "15", 25: "25"}),  # all int
             ({15.5: 15.5, 25.5: 25.5}, {15.5: "15.5", 25.5: "25.5"}),  # all floats
-            ({15.0: 15, 25.5: 25.5}, {15: "15", 25.5: "25.5"}),  # all floats, but convertible to int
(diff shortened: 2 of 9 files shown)


vizro-core/src/vizro/models/_components/form/slider.py
```diff
@@ -43,7 +43,7 @@ class Slider(VizroBaseModel):
     min: Optional[float] = Field(None, description="Start value for slider.")
     max: Optional[float] = Field(None, description="End value for slider.")
     step: Optional[float] = Field(None, description="Step-size for marks on slider.")
-    marks: Optional[Dict[int, Union[str, Dict[str, Any]]]] = Field({}, description="Marks to be displayed on slider.")
```
huong-li-nguyen, line 46: The correct typing for the value is indeed  `Dict[str, Any]` - however, we've decided to simplify it to `str` as the use case of `Dict[str, Any]` seems to be required only if one wants to style the values. This could also be achieved via CSS

nadijagraca: Yes, there was a discussion at the time, and it was decided to support everything that the dcc range slider supports. Styling with a dictionary is more straightforward than using custom CSS. However, you're right that it's not used often, so I'm okay with simplifying it again.

huong-li-nguyen: How about the change from `int` to `float` again? Am I breaking anything with the reverting change? I tried to run the previous PR example, but I couldn't replicate the bug 🤔

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -229,16 +229,20 @@ def test_validate_step_invalid(self):
...
+            ({"15": 15, "25": 25}, {15: "15", 25: "25"}),  # all string
             (None, None),
         ],
     )
     def test_valid_marks(self, marks, expected):
         range_slider = vm.RangeSlider(min=0, max=10, marks=marks)
-
         assert range_slider.marks == expected
 
+        if marks:
```
antonymilne, line 243: What is this? Why does the above `assert` not suffice?

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -229,16 +229,20 @@ def test_validate_step_invalid(self):
         "marks, expected",
         [
             ({i: str(i) for i in range(0, 10, 5)}, {i: str(i) for i in range(0, 10, 5)}),
-            ({15: 15, 25: 25}, {15.0: "15", 25.0: "25"}),
-            ({"15": 15, "25": 25}, {15.0: "15", 25.0: "25"}),
+            ({15: 15, 25: 25}, {15: "15", 25: "25"}),  # all int
+            ({15.5: 15.5, 25.5: 25.5}, {15.5: "15.5", 25.5: "25.5"}),  # all floats
+            ({15.0: 15, 25.0: 25}, {15: "15", 25: "25"}),  # all floats, but convertible to int
```
antonymilne, line 234: A mixed case might be good here too
```suggestion
            ({15.0: 15, 25.5: 25}, {15: "15", 25.5: "25"}),
```

antonymilne (approved):
Thank you for getting to the bottom of this and fixing it! Super clear dev example also - I love the containers with title which is much better than me getting confused about which row is which 😅 🤦 

Just a couple of things about the tests but otherwise looks great.

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
huong-li-nguyen: I was surprised as well, but if you remove the added assert it doesn't detect a difference between a float and an integer value e.g. this test case should fail: `({15.0: 15, 25.0: 25}, {15.0: "15", 25.0: "25"}), `

It doesn't fail if you don't have the added assert here.

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -229,16 +229,20 @@ def test_validate_step_invalid(self):
...
             (None, None),
         ],
     )
     def test_valid_marks(self, marks, expected):
         range_slider = vm.RangeSlider(min=0, max=10, marks=marks)
-
         assert range_slider.marks == expected
 
+        if marks:
+            assert all(type(key) is type(next(iter(expected.keys()))) for key in range_slider.marks)
```
antonymilne, line 244:
```suggestion
       assert [type(result_key) for result_key in range_slider.marks] == [type(expected_key) for expected_key in expected]
```

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
antonymilne: Ah, I see 👍 This is surprising to me also! TIL `assert 15 == 15.0` passes. https://www.reddit.com/r/learnpython/comments/tsb9vo/how_can_an_int_and_a_float_be_equal/

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
```diff
@@ -231,7 +231,7 @@ def test_validate_step_invalid(self):
             ({i: str(i) for i in range(0, 10, 5)}, {i: str(i) for i in range(0, 10, 5)}),
             ({15: 15, 25: 25}, {15: "15", 25: "25"}),  # all int
             ({15.5: 15.5, 25.5: 25.5}, {15.5: "15.5", 25.5: "25.5"}),  # all floats
-            ({15.0: 15, 25.0: 25}, {15: "15", 25: "25"}),  # all floats, but convertible to int
+            ({15.0: 15, 25.5: 25.5}, {15: "15", 25.5: "25.5"}),  # all floats, but convertible to int
```
antonymilne, line 234: This comment needs updating now 🙂

vizro-core/tests/unit/vizro/models/_components/form/test_range_slider.py
huong-li-nguyen: Thanks!! I like the first 👍 It prints out the types as well.

nadijagraca (approved):
LGTM. 🙌

vizro-core/src/vizro/models/_components/form/slider.py
nadijagraca: I tested it on few different examples and changing to `float` does not break anything. 👍

PR #620, opened 2024-08-06
[POC] Add second batch of charts to gallery

huong-li-nguyen:
## Description

As discussed with @Joseph-Perkins, we'll be adding the content for the top 30 charts. This PR adds content for the second badge of charts:
- [x] Grouped column
- [x] Grouped bar
- [x] Area
- [x] Connected scatter
- [x] Histogram
- [x] Stacked column
- [x] Stacked bar

Run `hatch run example _chart-gallery` to check result.

## Notice


vizro-core/examples/_chart-gallery/pages/_factories.py
```diff
@@ -74,3 +74,39 @@ def butterfly_factory(group: str):
...
+    """Reusable function to create the page content for the column chart with a unique ID."""
+    return vm.Page(
+        id=f"{group}-connected-scatter",
+        path=f"{group}/connected-scatter",
+        title="Connected Scatter",
+        layout=vm.Layout(grid=PAGE_GRID),
+        components=[
+            vm.Card(
+                text="""
+                #### What is a connected scatter chart?
```
huong-li-nguyen, line 89: @stichbury - could you look at all the added markdown text? 🙏

stichbury, line 89:
```
                #### What is a connected scatter chart?
```
Is there a reason you're starting with H4, rather than H1?

vizro-core/examples/_chart-gallery/pages/_factories.py
```diff
@@ -74,3 +74,39 @@ def butterfly_factory(group: str):
...
+
+                A connected scatter chart visualizes two variables using dots, with lines connecting the dots in the
+                order of the data points. One variable is plotted along the x-axis and the other along the y-axis,
+                showing both the relationship and the sequence of the data.
+
+                &nbsp;
```
stichbury, line 95:
```
                &nbsp;
```
Does this render OK? Do you need it, and why?

vizro-core/examples/_chart-gallery/pages/_factories.py
```diff
@@ -74,3 +74,39 @@ def butterfly_factory(group: str):
...
+
+                #### When should I use it?
+
+                Use connected scatter charts to show the relationship between two variables and the sequence of data
```
stichbury, line 99:
```suggestion
                Use connected scatter charts to show the relationship between two variables and a sequence of data
```

vizro-core/examples/_chart-gallery/pages/_factories.py
```diff
@@ -74,3 +74,39 @@ def butterfly_factory(group: str):
...
+                points. They are ideal for paired numerical data, helping to reveal trends and patterns over time or in
+                a specific order. Remember, correlation is not causation, so ensure your audience understands this to
+                avoid misinterpretations.
```
stichbury, line 102:
```suggestion
                avoid misinterpretation.
```

stichbury: Or you could use the word "misrepresentation" I think, but misinterpretation (by the audience) is fine. Maybe you could put a link under "correlation is not causation" to point readers to further info, but I reckon that's unnecessary for all but the most naive reader.

vizro-core/examples/_chart-gallery/pages/_factories.py
huong-li-nguyen: You need it for an extra empty line :) This is how you do it inside the markdown text 👍

vizro-core/examples/_chart-gallery/pages/_factories.py
huong-li-nguyen: It's inside a text card, so from a hierarchy view it's lower than the dashboard and page title (which are both H2). H3 still looked too big, so I opted for H4 😄

stichbury (approved):
All great from my point of view for the markdown in the examples. I think it's amazing that you've included this information as it's educational, but if it becomes too much to maintain, we could link out to a trusted resource to avoid having it.

🌟 🥇

vizro-core/examples/_chart-gallery/pages/_factories.py
stichbury: Ah, perfect. I always use <br /> but maybe I shouldn't!

vizro-core/examples/_chart-gallery/pages/_factories.py
stichbury: I thought it must be something like that ... all good from my perspective.

huong-li-nguyen:
> All great from my point of view for the markdown in the examples. I think it's amazing that you've included this information as it's educational, but if it becomes too much to maintain, we could link out to a trusted resource to avoid having it.
> 
> 🌟 🥇

All good! The majority of the info text has already been created by us and lives on QB protocol page! The rest is quick to generate :)

antonymilne:
Reminder to myself to discuss this re: bar charts and histograms:
> I believe it’s important to demonstrate the aggregation. There isn’t a simpler dataset in px.data, because as soon as you have more than one observation, aggregation becomes necessary; otherwise, the chart appears as shown below, which can be quite confusing. What Plotly does is stack each observation. The simplest dataset would be one we create from scratch with a single observation. However, I wanted to show the aggregation because I remember how confusing it was initially to see charts like this, as it’s not what you would expect. To get a clean chart, you need to add an aggregation in about 80% of the cases :smile:

antonymilne:
Oops, didn't meant to close it, sorry...

antonymilne (commented):
Looks great, I really love this dashboard! <3 

I think some of the examples could be improved a bit to make the concepts clearer but otherwise looks good.

Now to respond to this point from the last PR which is especially relevant now you have a histogram. I said:
> is there a simpler dataset that would work for the bar/column charts? Surely there's something in px.data that doesn't require us to do the aggregations on it?

You said:
> I believe it’s important to demonstrate the aggregation. There isn’t a simpler dataset in px.data, because as soon as you have more than one observation, aggregation becomes necessary; otherwise, the chart appears as shown below, which can be quite confusing. What Plotly does is stack each observation. The simplest dataset would be one we create from scratch with a single observation. However, I wanted to show the aggregation because I remember how confusing it was initially to see charts like this, as it’s not what you would expect. To get a clean chart, you need to add an aggregation in about 80% of the cases 😄

Absolutely agree with the confusion with how plotly does bar charts like this. But I think the example you have for bar/column chart can be done directly as a histogram - in fact it's [exactly this example](https://plotly.com/python/histograms/#histograms-on-categorical-data) in their docs. 

So now that we have histogram as a chart I would suggest changing the bar chart to something that doesn't require aggregations, because the easy and plotly recommended way to achieve that is to just use a histogram instead: https://plotly.com/python/bar-charts/#aggregating-into-single-colored-bars.

huong-li-nguyen:
> Absolutely agree with the confusion with how plotly does bar charts like this. But I think the example you have for bar/column chart can be done directly as a histogram - in fact it's [exactly this example](https://plotly.com/python/histograms/#histograms-on-categorical-data) in their docs.
> 
> So now that we have histogram as a chart I would suggest changing the bar chart to something that doesn't require aggregations, because the easy and plotly recommended way to achieve that is to just use a histogram instead: https://plotly.com/python/bar-charts/#aggregating-into-single-colored-bars.

Oh, I wasn't aware that they recommend using `px.histogram` for this purpose and that it's interchangeable! 🥳 I implemented your suggestion, but I still slightly prefer the previous data aggregation approach as we avoid the need for custom charts, which just looks more complicated 😬

Previously, I aggregated the dataframe and sorted it based on the grouped values because `px.charts` display the bars in the order of the dataframe unless specified otherwise. Since we're not doing that anymore, we need a post-update call to sort the bars in descending order for the `ordered bar` and `ordered column` chart 🫠

For the `bar` and `column` chart, I would normally also update the y-axis titles now (since it now says `sum of total_bill`), but I didn't feel strongly enough about this particular case to create custom charts for it.

WDYT? Would you still prefer the `histogram` approach even if we now have to use custom charts?

vizro-core/examples/_chart-gallery/pages/_factories.py
```diff
@@ -74,3 +74,39 @@ def butterfly_factory(group: str):
...
+        id=f"{group}-connected-scatter",
+        path=f"{group}/connected-scatter",
+        title="Connected Scatter",
+        layout=vm.Layout(grid=PAGE_GRID),
+        components=[
+            vm.Card(
+                text="""
+                #### What is a connected scatter chart?
+
+                A connected scatter chart visualizes two variables using dots, with lines connecting the dots in the
```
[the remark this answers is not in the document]
huong-li-nguyen: I've changed it to `n=1` now 👍

antonymilne (approved):
I've just pushed [a35efdb](https://github.com/mckinsey/vizro/pull/620/commits/a35efdb627bca258c0f4643edeb37c1f57caaab0) which has answers to your questions and some proposals. Basically:
* plain and ordered bar/column - agree that I don't like the histogram approach with the custom chart code for ordering. In fact I think it's a bit weird that these "simple" bar/column examples need histograms at all so I tried an example that doesn't
* paired bar/column chart didn't look right to me so I tried changing number of categories and I think it makes more sense now
 
I just wrote the proposal for the bar charts but if you like them then they should obviously be done for the column equivalents and the examples files too.

Other than that it all looks great I think so happy to give a pre-approval 😀

antonymilne:
Oh, and you're going to hate me for this... But I spotted a couple more places we're missing monospace font. 🙈  The second one might disappear I know but probably worth fixing anyway if these are easy to fix since it might come again in future.

image
image

huong-li-nguyen:
> Oh, and you're going to hate me for this... But I spotted a couple more places we're missing monospace font. 🙈 The second one might disappear I know but probably worth fixing anyway if these are easy to fix since it might come again in future.

No worries, I saw these too haha I was planning to fix these in the end 👍

huong-li-nguyen:
I've left most of your commits as is and applied it to the rest. The only change I've added was to replace the bar chart example with some comparable countries to have some nicer shapes 👍 I took the 5 biggest ones in terms of population.

PR #626, opened 2024-08-09
[Tidy] Improve code style in response models

lingyielia:
## Description

Addressed 2 TODOs in `controls.py`, as per discussion with @maxschulz-COL, to make response models more consistent in styles
1. make “_create_filter_proxy” a callable class
2. instead of requesting the df_name, we should request the target component name, replace df_name with target_component_name, then later retrieve the df_name from the controllable components. This logic is more aligned with the Vizro usage pattern.

Plus 2 minor cleanup in `components.py` and `layout.py`

## Notice


vizro-ai/src/vizro_ai/dashboard/_response_models/controls.py
maxschulz-COL, line 1: I was more imagining something along the following lines:

```python
from typing import Any


try:
    from pydantic.v1 import BaseModel, Field, create_model
except ImportError:  # pragma: no cov
    from pydantic import BaseModel, Field
    
class FooModel(BaseModel):
    foo: str
    bar: int = 123

class ProxyModel:
    def __new__(cls, color) -> Any:
        return create_model(
            'BarModel',
            apple='russet',
            banana=color,
            __base__=FooModel,
        )
        

Proxy = ProxyModel(color="orange")

print(Proxy.__fields__.keys())
print(Proxy.__fields__.values())
```

So in the above toy example, `Proxy` can just act as a normal response model no?

vizro-ai/src/vizro_ai/dashboard/_response_models/controls.py
```diff
@@ -100,31 +92,58 @@ class ControlPlan(BaseModel):
...
+            for component_plan in components_plan
+            if component_plan.component_id in target_controllable
+        }
+
+        if len(df_names) > 1:
+            logger.warning(
+                f"""
+[FALLBACK] Multiple dataframes found in the target components: {df_names}.
+Choose one dataframe to build the filter.
+"""
```
maxschulz-COL, line 115: I think I do not understand this. Should the logic not be:
- model determines which components should be controlled
- we then check if those components can be controlled, and take a subset at worst
- the remaining dfs are what we get the metadata for (and the could be multiple no?)

maxschulz-COL (commented):
Hey, thanks for tackling those already - here are my initial thoughts

PR #632, opened 2024-08-15
[Bug] Dynamically change `title_pad_t` inside `vm.Graph` if subtitle is provided

huong-li-nguyen:
## Description
In the past, including a subtitle with the chart resulted in it being cut off, causing confusion. Two users have already reported this issue internally. To address it, we've been using a post-update call `fig.update_layout(title_pad_t=24)`

I wanted to discuss the possibility of permanently incorporating this solution into our `vm.Graph`. Directly encoding it into the template is challenging due to the way Plotly charts function. Any changes to the template disrupt the alignment between the graph and table titles, which is essential for the dashboard.

## Screenshot
**Before:**
Screenshot 2024-08-15 at 13 21 25

**After:**
Screenshot 2024-08-15 at 13 23 16

## Notice

### DIFF
--- a/vizro-core/src/vizro/models/_components/graph.py
+++ b/vizro-core/src/vizro/models/_components/graph.py
@@ -56,11 +56,12 @@ class Graph(VizroBaseModel):
         kwargs.setdefault("data_frame", data_manager[self["data_frame"]].load())
         fig = self.figure(**kwargs)
 
-        # Reduce top margin if no title is provided
+        # Reduce `margin_t` if no title is provided and `margin_t` is not explicitly set.
         if fig.layout.margin.t is None and fig.layout.title.text is None:
             fig.update_layout(margin_t=24)
 
-        # Add top title padding if subtitle is provided, otherwise it's being cut off
+        # Increase `title_pad_t` if subtitle is provided and `title_pad_t` is not explicitly set.
+        # Otherwise, the title is being cut off.
         if fig.layout.title.pad.t is None and fig.layout.title.text and "<br>" in fig.layout.title.text:
             fig.update_layout(title_pad_t=24)
 
--- a/vizro-core/src/vizro/models/_components/graph.py
+++ b/vizro-core/src/vizro/models/_components/graph.py
@@ -57,9 +57,13 @@ class Graph(VizroBaseModel):
         kwargs.setdefault("data_frame", data_manager[self["data_frame"]].load())
         fig = self.figure(**kwargs)
 
-        # Remove top margin if title is provided
+        # Reduce top margin if no title is provided
         if fig.layout.title.text is None:
             fig.update_layout(margin_t=24)
+
+        # Add top title padding if subtitle is provided, otherwise it's being cut off
+        if fig.layout.title.text and "<br>" in fig.layout.title.text:
+            fig.update_layout(title_pad_t=24)
 
         # Possibly we should enforce that __call__ can only be used within the context of a callback, but it's easy
         # to just swallow up the error here as it doesn't cause any problems.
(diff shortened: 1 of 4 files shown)


vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -57,10 +57,14 @@ def __call__(self, **kwargs):
         kwargs.setdefault("data_frame", data_manager[self["data_frame"]].load())
```
antonymilne, line 57: At a glance, if this works then lgtm.

We should probably let user-specified `margin_t` and `title_pad_t` win out though, e.g. set `fig.layout.margin_t` only if it's not previously set.

huong-li-nguyen: Sounds good! I'll go ahead and make that change. However, I believe this might just be a temporary solution. After giving it more thought, I'm now more convinced that incorporating a header/footer argument inside the `vm.Graph` might be our long-term solution. This includes putting the graph title into its own container 😬 

There are several advantages to having the title/subtitle in its own container actually:

- It would automatically align with all our major components (Table and AgGrid), allowing us to remove these inconsistent padding numbers from the template.
- It would automatically wrap the text, making it more responsive (currently, in Plotly, you have to manually add <br>).
- We could eliminate all these custom overwrites here (including margin_t and title_pad_t).

But let's discuss that another time. 👍

antonymilne: Yeah, makes sense. Just so we don't forget about it, here were my previous thoughts on this: https://github.com/McK-Internal/vizro-internal/issues/142#issuecomment-1784873722

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -56,10 +56,14 @@ def __call__(self, **kwargs):
...
         fig = self.figure(**kwargs)
 
-        # Remove top margin if title is provided
-        if fig.layout.title.text is None:
+        # Reduce top margin if no title is provided
+        if fig.layout.margin.t is None and fig.layout.title.text is None:
```
antonymilne, line 60: Does this work? Shouldn't we be checking `if fig.layout.margin.t != pio.templates["vizro_dark"].margin.t` instead to see if the user altered it? Since our default `margin.t` is something that's not `None`? 🤔

petar-qb (approved):
It looks and works really well 👍

I'm just wondering where the magic number 24 comes from?
Does it come from some "font.title.size" (or something similar) and should it be dynamically set rather than hard-coded?

vizro-core/src/vizro/models/_components/graph.py
huong-li-nguyen: Yes, this works! I just pushed a dev example, so you can see what's happening.

If the user does not set a `margin_t`, this is what you can find:

```
fig.layout.margin.t = None
fig.layout.template.layout.margin.t = 64 (our default)
```

It's going back to the discussion we've had. There are these extra arguments in the top-level of the figure that take precedence. 
Screenshot 2024-08-16 at 10 32 16

```
@capture("graph")
def my_graph(data_frame):
    fig = px.scatter(data_frame, x="sepal_width", y="sepal_length")
    print(fig.layout.margin.t)  # This is None
    print(fig.layout.template.layout.margin.t)  # This is 64 our default
    fig.update_layout(margin_t=0)
    print(fig.layout.margin.t) # This is 0 now
    return fig
```

huong-li-nguyen:
> I'm just wondering where the magic number 24 comes from? Does it come from some "font.title.size" (or something similar) and should it be dynamically set rather than hard-coded?

Not that I can think of 😅 It's paddings/margins that were previously set by Dan, but I don't know the relationship between them. These numbers still come from VizX time. I wouldn't worry too much about hard-coding them, because that's essentially what our template does as well.

nadijagraca (approved):
Nice fix, works really well.

PR #635, opened 2024-08-19
[Tidy] Remove snyk, gitleaks and package-lock.json

antonymilne:
## Description

* Snyk is no longer active for us so removed everything about that
* ~gitleaks is no longer required since we have gitguardian which works better~ **still under investigation**
* package-lock.json seemed unnecessary (it's just pinned versions of JS dev dependencies, which are purely for testing - is that right @petar-qb?) and having it there raised false positive dependency alerts in the past. So it's also gone
* Most excitingly, we are now free of the pins that snyk encouraged us to do but should never have really been there 🎉 FYI @lingyielia ipython now unpinned for vizro-ai

## Notice


.github/workflows/secret-scan.yml
maxschulz-COL, line 1: Have you tested git guardian?

maxschulz-COL (approved):
Ah that feels good 🧹  Thanks for doing it.

Have you tested Git Guardian? Why is it better (I know we get it automatically, just curious!)

.github/workflows/secret-scan.yml
antonymilne: Not explicitly here, but in the past I've actually found it useful on other projects.

In theory it's better because it has protection against [pushing secrets](https://docs.github.com/en/code-security/secret-scanning/introduction/supported-secret-scanning-patterns#supported-secrets) (not sure how tbh) and nice github integration e.g. you can see some results here https://github.com/mckinsey/vizro/security/secret-scanning.

In reality I'm not sure how great it is because my test commit [2a7a0b4](https://github.com/mckinsey/vizro/pull/635/commits/2a7a0b466ef1bb494b409bc5a12ca0c0c1830a65) doesn't seem to have done anything 😅 Even though we definitely have it turned on.

petar-qb:
> package-lock.json seemed unnecessary (it's just pinned versions of JS dev dependencies, which are purely for testing - is that right petar-qb?

Correct. 👍 Do you think we should put the `package-lock.json` in the `.gitignore` file then?

.github/workflows/secret-scan.yml
maxschulz-COL: Hm, that is not ideal then no? I remember when doing gitleaks that it was quite fiddly, but at the very least _i think_ it was working. At least it caught the trial bits that I entered back when testing it.

Should we proceed even if it doesn't catch obvious violations?

maxschulz-COL: Pre-commit hook certainly worked:

maxschulz-COL: Equally works in CI:
image

maxschulz-COL: I'd prefer us getting to the bottom of this before removing something that works reasonably well. Otherwise I am of course all for using the more integrated tool!

maxschulz-COL: P.S. I should note that I actually received an email from GitGuardian about the above secret, so it has found that one at least

antonymilne: Hmm yes, thank you for this. Something definitely doesn't seem to be working quite right. Let me try and figure it out...

antonymilne: Yes indeed. I am in touch with them to try and get to the bottom of it...

antonymilne: Still to be resolved so:
* I'm leaving secret scanning as it is now
* merging the rest of the changes
* we can come back to secret scanning if infosec can work it out for us

vizro-core/pyproject.toml
```diff
@@ -22,11 +22,7 @@ dependencies = [
   "pydantic>=1.10.13",  # must be synced with pre-commit mypy hook manually
   "dash_mantine_components<0.13.0",  # 0.13.0 is not compatible with 0.12,
   "flask_caching>=2",
-  "wrapt>=1",
-  "numpy>=1.22.2",  # not directly required, pinned by Snyk to avoid a vulnerability: https://security.snyk.io/vuln/SNYK-PYTHON-NUMPY-2321970
-  "setuptools>=65.5.1",  # not directly required, pinned by Snyk to avoid a vulnerability: https://security.snyk.io/vuln/SNYK-PYTHON-SETUPTOOLS-3180412
-  "werkzeug>=3.0.1",  # not directly required, pinned by Snyk to avoid a vulnerability: https://security.snyk.io/vuln/SNYK-PYTHON-WERKZEUG-6035177
-  "ruff"
+  "wrapt>=1"
```
antonymilne, line 25:
```suggestion
  "wrapt>=1",
  "ruff"
```

PR #641, opened 2024-08-21
[Feat] Enable dashboard code generation

lingyielia:
## Description
```
result = vizro_ai.dashboard([df1, df2], user_question_2_data, return_elements=True)
print(result.code)
```

Current version of output:
```
############ Imports ##############
import vizro.tables as vt
import vizro.models as vm


####### Function definitions ######
@capture("graph")
def gapminder_scatter_plot(data_frame=None):
    df_copy = data_frame.copy() if data_frame is not None else pd.DataFrame()
    # Selecting relevant columns for the scatter plot
    scatter_df = df_copy[["lifeExp", "gdpPercap", "continent", "country"]]
    # Re-indexing the DataFrame if necessary
    scatter_df.reset_index(drop=True, inplace=True)

    # Create a scatter plot using Plotly Express
    fig = px.scatter(
        scatter_df,
        x="gdpPercap",
        y="lifeExp",
        color="continent",
        hover_name="country",
        title="Life Expectancy vs. GDP per Capita by Country",
        labels={"gdpPercap": "GDP per Capita", "lifeExp": "Life Expectancy"},
        template="plotly_white",
    )

    # Update layout for better aesthetics
    fig.update_layout(
        xaxis_title="GDP per Capita",
        yaxis_title="Life Expectancy",
        legend_title="Continent",
        showlegend=True,
    )

    return fig


####### Data Manager Settings #####
#######!!! UNCOMMENT BELOW !!!#####
# from vizro.managers import data_manager
# data_manager["266f87b0-fe3a-4056-9dc5-7dfaa06a421e"] = ===> Fill in here <===
# data_manager["tech_stock_data"] = ===> Fill in here <===


########### Model code ############
model = vm.Dashboard(
    pages=[
        vm.Page(
            components=[
                vm.AgGrid(
                    id="tech_stock_table",
                    figure=vt.dash_ag_grid(data_frame="tech_stock_data"),
                )
            ],
            title="Tech Stock Data",
            layout=vm.Layout(grid=[[0]]),
            controls=[],
        ),
        vm.Page(
            components=[
                vm.Card(
                    id="gapminder_card_1",
                    type="card",
                    text="The Gapminder dataset provides historical data on countries' development indicators.",
                    href="",
                ),
                vm.Graph(
                    id="gapminder_scatter_plot",
                    figure=gapminder_scatter_plot(
                        data_frame="266f87b0-fe3a-4056-9dc5-7dfaa06a421e"
                    ),
                ),
                vm.Card(
                    id="gapminder_card_2",
                    type="card",
                    text="Data spans from 1952 to 2007 across various countries.",
                    href="",
                ),
            ],
            title="Gapminder Data Visualization",
            layout=vm.Layout(grid=[[0, 1, 1], [2, 1, 1]]),
            controls=[
                vm.Filter(column="continent", targets=["gapminder_scatter_plot"]),
                vm.Filter(column="year", targets=["gapminder_scatter_plot"]),
            ],
        ),
    ],
    title="Tech and Gapminder Data Dashboard",
)
```

## Notice


vizro-ai/src/vizro_ai/plot/components/dataframe_craft.py
```diff
@@ -43,8 +43,7 @@ class DataFrameCraft(BaseModel):
 2. Do not create any new DataFrames; work only with df.
 3. Always make a hard copy of the DataFrame before manipulating it. Important: Do not modify the original DataFrame.
 4. Ensure that any aggregated columns are named appropriately and re-indexed if necessary.
-5. If a visualization is implied by the user's query, only write the necessary DataFrame manipulation
-code for that visualization.
+5. Focus on providing operations related to aggregation, filtering, sorting, grouping, and other transformations.
```
lingyielia, line 46: This is to avoid a common error found in integration test: ""error: matplotlib is required for plotting when the default backend "matplotlib" is selected.""

vizro-ai/src/vizro_ai/dashboard/_response_models/components.py
```diff
@@ -41,22 +41,28 @@ class ComponentPlan(BaseModel):
...
-
-                return vm.Graph(
-                    id=self.component_id,
-                    figure=vizro_ai.plot(
-                        df=all_df_metadata.get_df(self.df_name), user_input=self.component_description
+                result = vizro_ai._run_plot_tasks(
+                    df=all_df_metadata.get_df(self.df_name),
+                    user_input=self.component_description,
+                    chart_name=self.component_id,
+                    max_debug_retry=2,
```
maxschulz-COL, line 55: Is this parameter not something we need to fetch from the `.dashboard` code

vizro-ai/src/vizro_ai/dashboard/utils.py
```diff
@@ -61,3 +63,30 @@ def _register_data(all_df_metadata: AllDfMetadata) -> vm.Dashboard:
 
     for name, metadata in all_df_metadata.all_df_metadata.items():
         data_manager[name] = metadata.df
+
+
+def _dashboard_code(
```
maxschulz-COL, line 68: Do we need a separate function for this?

vizro-ai/src/vizro_ai/dashboard/utils.py
```diff
@@ -61,3 +63,30 @@ def _register_data(all_df_metadata: AllDfMetadata) -> vm.Dashboard:
...
+    dashboard: vm.Dashboard, extra_callable_defs: Optional[Set[str]] = None, extra_imports: Optional[Set[str]] = None
+) -> str:
+    """Get the code for the dashboard."""
+    return dashboard._to_python(extra_callable_defs=extra_callable_defs, extra_imports=extra_imports)
+
+
+def _process_to_set(list_of_list_of_dict):
```
maxschulz-COL, line 75: I would rename the argument so something more descriptive. Can we add typing?

This will not be needed anymore with the refactor, are we accounting for that?

maxschulz-COL (approved):
LGTM, couple of comments

vizro-ai/src/vizro_ai/dashboard/utils.py
antonymilne, line 75: What does this variable look like? It's not at all obvious to me from reading why we end up with a list of list of dicts here. Should we flatten this out before passing it into the function?

It would also be nice to give these variables names that reflect their _use_ rather than their data type also.

vizro-ai/src/vizro_ai/dashboard/utils.py
```diff
@@ -61,3 +63,30 @@ def _register_data(all_df_metadata: AllDfMetadata) -> vm.Dashboard:
...
+    custom_func = set()
+    imports = set()
+    for list_of_dict in list_of_list_of_dict:
+        for d in list_of_dict:
+            for value in d.values():
+                # Extract imports
+                import_match = re.match(r"((?:from|import).*?)\n\n", value, re.DOTALL)
```
antonymilne, line 82: Is this reliable? i.e. are there always `\n\n` separating imports from the rest of the code?

vizro-ai/src/vizro_ai/dashboard/utils.py
```diff
@@ -61,3 +63,30 @@ def _register_data(all_df_metadata: AllDfMetadata) -> vm.Dashboard:
...
+                if import_match:
+                    imports.add(import_match.group(1))
+
+                # Remove leading imports and the last line (any chart creation)
+                value = re.sub(r"^(?:from|import).*?\n\n", "", value, flags=re.DOTALL)  # noqa: PLW2901
+                value = re.sub(r"\n\nfig = .*?\(data_frame=df\)$", "", value)  # noqa: PLW2901
```
antonymilne, line 88: It would be a bit nicer to do this without any more `re` if you can. Something like this:
```
code = code[import_match.end():] # to remove imports
code.rpartition("\n\n") # something like this to remove the last line
```

antonymilne (approved):
Generally looks good, just a few rough bits of implementation that could be improved 👍 So it's an approval because I know you don't want to wait too long on this.

Out of my suggestion I would maybe try to improve `_process_to_set` now and then leave `create` for another time, but up to you!

vizro-ai/src/vizro_ai/dashboard/_response_models/components.py
lingyielia: you mean "max_debug_retry"? I don't find it helpful to expose "max_debug_retry" in .dashboard(). For example, my experience in "_get_pydantic_model", too many retries doesn't seem to help llm converge to the correct answer. Instead it's more likely to cause panic and the response becomes very off-track. But also I don't have through test on this conclusion, happy to discuss further if you found it useful to expose number of retry.

vizro-ai/src/vizro_ai/_vizro_ai.py
```diff
@@ -66,15 +66,18 @@ def _run_plot_tasks(
...
         max_debug_retry: int = 3,
         explain: bool = False,
+        chart_name: Optional[str] = None,
     ) -> PlotOutputs:
         """Task execution."""
         chart_type_pipeline = self.pipeline_manager.chart_type_pipeline
         chart_type = chart_type_pipeline.run(initial_args={"chain_input": user_input, "df": df})
 
         # TODO update to loop through charts for multiple charts creation
+        if chart_name is None:
```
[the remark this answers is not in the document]
lingyielia: When call `.plot()` which runs `self._run_plot_tasks(
            df=df, user_input=user_input, explain=explain, max_debug_retry=max_debug_retry
        )`, there  chart_name is None. Putting chart_name = "custom_chart" in the function signature would be contradictory in this case

vizro-ai/src/vizro_ai/dashboard/_response_models/components.py
```diff
@@ -41,22 +41,28 @@ class ComponentPlan(BaseModel):
         """,
     )
 
-    def create(self, model, all_df_metadata) -> Union[vm.Card, vm.AgGrid, vm.Figure]:
+    def create(self, model, all_df_metadata) -> Tuple[Union[vm.Card, vm.AgGrid, vm.Figure], Optional[str]]:
```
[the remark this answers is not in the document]
lingyielia: > Having an optional second return value that's only relevant in 1 out of 3 cases doesn't feel right. It's ok as a temporary solution but I think we should refactor some time.
> 
> e.g. one solution might be to have independent `create_card`, `create_figure` etc.

Good point. When we say later enable more sophisticated AgGrid, would it more follows the chart creation pattern? i.e., maybe we also need to retrieve code from here.

vizro-ai/src/vizro_ai/dashboard/utils.py
lingyielia: updated

vizro-ai/src/vizro_ai/dashboard/utils.py
lingyielia: So far this is quite consistent. So when llm generate the code, it always has an empty line between imports and the "@capture('graph')".  So in total there are 2 "\n". We should be good at least for OpenAI models, but I don't know whether other llm has the same pattern.
Consider this function will be gone after Max's refactor, let me leave it as is now. But I will try with all current models and check the pattern

PR #644, opened 2024-08-21
[Tidy] Return empty data frame from AgGrid, Table and Figure build methods

petar-qb:
## Description
References: 

- PR link that introduced returning the entire data_frame from the AgGrid/Table build method -> [#439](https://github.com/mckinsey/vizro/pull/439)
- [Reasons](https://github.com/mckinsey/vizro/pull/446) behind the mentioned PR:
  - https://github.com/McK-Internal/vizro-internal/issues/670
  - https://github.com/McK-Internal/vizro-internal/issues/660
  - https://github.com/McK-Internal/vizro-internal/issues/747 

---

#### Short reminder:
There are two phases that happen when the page is loaded:
1. `build` phase -> renders all page components by returning the content made inside the components build method. 
2. `on_page_load` phase -> triggers a process that calculates the final appearance of all targeted figures based on the selected filters/parameters/filter_interactions. 

Even when the big data is handled and showed with different figures, the `build` phase has to be completed `quickly`, so Vizro users don't stuck with the "blank white page". Since the calculation for the final figures look could take a lot of time (e.g. if the data loading process is slow or if big data has to be transported through the network), we should return some light-weight `placeholder components` that will be replaced with the real components afterwards. 

After the build phase, the `on_page_load` mechanism is triggered which immediately puts all the targeted figures in the "loading state" and after some time it returns and renders filtered & parametrised figures to the UI. 

#### How it works now?

Currently the `vm.Graph.build` method works as expected. It returns a `placeholder graph` from the `build` phase which is later replaced with the actual (filtered and parametrised) Graph object (result of the on_page_load mechanism) later. 
However, the `vm.AgGrid.build` and the `vm.Table.build` methods returns objects made of `entire` (non-filtered) data_frame and original `parameters` from the `build phase` which becomes replaced with the actual AgGrid/Table objects (result of the on_page_load mechanism). 

(**We decided to return entire data_frames for ag_grid/tables just to quick fix the bugs mentioned at the beginning of this description.**)

#### What do we want to achieve?
We want to optimise the AgGrid and Table build methods so they `do not` return an entire data_frame in build phase. Except the response time optimisation (which is the most important I guess), we also want to [align](https://github.com/McK-Internal/vizro-internal/issues/747#issue-2262699927) them with the [Graph build method](https://github.com/mckinsey/vizro/blob/3187e0be402b1c58519c423b8b1f0d655701aa79/vizro-core/src/vizro/models/_components/graph.py#L149-L170) that also `does not` return an entire data_frame from the build method. In this process we shouldn't recreate any of the persistence (or similar) bugs we solved with the [#439](https://github.com/mckinsey/vizro/pull/439).

Vizro UX will hugely benefit from this "relieving the initial page loading" process, and this step is crucial for all the upcoming backend changes. 

---

Task acceptance criteria:
- Relieve the `page.build` phase:
  - Ideally by returning a hardcoded placeholder component from the AgGrid/Table build methods (similarly like it's done for the [vm.Graph](https://github.com/mckinsey/vizro/blob/3187e0be402b1c58519c423b8b1f0d655701aa79/vizro-core/src/vizro/models/_components/graph.py#L149-L170))
  - OR by returning a real (fully configured) AgGrid/Table object but with an empty data_frame. (see in the table below why this is not optimal solution. This option is described within the "self.call(data_frame=pd.DataFrame())" column) .
- Do not introduce any persistence (or similar) inconsistencies like:
  - Reseting AG grid resets column widths when reloading the screen (if the columnSize="autoSize"),
  - Column filter persistence,
  - Not sending an empty data_frame into figure functions,
  - Make the `pagination` and `selectedRows` to work smoothly (together with all other possible arguments).
- Optimise the number of figure function calls. Ideally it should be called only once from the `on_page_load` phase, with the filtered data_frame and parameterised function arguments.

---

The table below shows how different objects returned from the vm.AgGrid build method (`columns`) match different acceptance criteria (`rows`):

Returned objects like `dag.AgGrid()` / `html.Div()` are not considered at all as they raise the following error (even with `suppress_callback_exceptions=True` set) -> "A nonexistent object was used in an `State` of a Dash callback...".

|                    | **dag.AgGrid(id=self._input_component_id)** | **self.__call__(data_frame=pd.DataFrame())** | **self.__call__()**         | **html.Div(id=self._input_component_id)** |
|:-----------------------------------------------------:|---------------------------------------------|----------------------------------------------|-----------------------------|-------------------------------------------|
|    Returns a light-weigh component from the build phase   |                      ✅                      |                       ✅                      |              ❌              |                     ✅                     |
|          Keeps column widths on page refresh          |                      ❌                      |                       ❌                      | *❌ (we thought we fixed it) |                     ✅                     |
|            Keeps column widths on page nav            |                      ❌                      |                       ❌                      | *❌ (we thought we fixed it) |                     ✅                     |
|         Keeps column widths on filter changes         |                      ❌                      |                       ❌                      |              ❌              |   *❌ (probably internal dash-ag-grid bug)|
|               "Column filter" persistence               |                      ✅                      |                       ❌                      |              ✅              |                     ✅                     |
|                "pagination" persistence                |                      ❌                      |                       ✅                      |              ✅              |                     ✅                     |
|                "selectedRows" persistence               |                      ✅                      |                       ❌                      |              ❌              |                     ✅                     |
| Not sending an empty data_frame into figure functions |                      ✅                      |                       ❌                      |              ✅              |                     ✅                     |
|     number of figure function calls - build phase     |                      0 - ✅                      |                       2 - ❌                      |              2 - ❌             |                     0 - ✅                     |
|  number of figure function calls - on_page_load phase |                      1 - ✅                     |                       2 - ❌                      |              2 - ❌            |                     1 - ✅                     |

**Legend:**
 1. `*❌ (we thought we fixed it)` - points out that the https://github.com/McK-Internal/vizro-internal/issues/670 is not actually fully solved. To prove it, run the `scratch_dev` example, select filter [1, 3], and refresh the page.
 2. `*❌ (probably internal dash-ag-grid bug)` - See the issue https://github.com/plotly/dash-ag-grid/issues/318

--- 

### Conclusions:
Based on all the knowledge gathered so far (which is mainly presented in the table above), we will go with the returned object `html.Div(id=self._input_component_id)` option from the vm.AgGrid/vm.Table build methods.    

---

## Notice


vizro-core/src/vizro/models/_components/figure.py
```diff
@@ -52,11 +52,9 @@ def __getitem__(self, arg_name: str):
     @_log_call
     def build(self):
         return dcc.Loading(
-            # Optimally, we would like to provide id=self.id directly here such that we can target the CSS
-            # of the children via ID as well, but the `id` doesn't seem  to be passed on to the loading component.
-            # I've raised an issue on dash here: https://github.com/plotly/dash/issues/2878
-            # In the meantime, we are adding an extra html.div here.
-            html.Div(self.__call__(), id=self.id, className="figure-container"),
```
petar-qb, line 59: @huong-li-nguyen I found that [this PR](https://github.com/plotly/dash/pull/2888) is merged, but didn't succeed to move the `self.id` to the `dcc.Loading` component without messing up its appearance.

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -210,6 +210,11 @@ def _get_modified_page_figures(
 ) -> Dict[str, Any]:
     targets = targets or []
 
+    # This only exists to simulate a delay in the on_page_load or filter actions.
+    # from time import sleep
+
+    # sleep(1.5)
+
```
petar-qb, line 217: Remove before merging.

vizro-core/src/vizro/models/_components/figure.py
huong-li-nguyen: The PR was merged on Jun 18th, but they haven't done a release since then, which is why the bug is still in there 😬 It will be officially fixed in the next Dash release. So for now, I would leave the code comments there.

petar-qb: Ah, I missed that they haven't released it yet.. Thanks 😃

vizro-core/tests/unit/vizro/tables/test_dash_ag_grid.py
```diff
@@ -50,76 +48,20 @@ def test_dash_ag_grid(self):
         )
         # skipping only dashGridOptions as this is mostly our defaults for data formats, and would crowd the tests
 
-
-class TestCustomDashAgGrid:
```
petar-qb, line 54: `class TestCustomDashAgGrid` tests are restructured into:
 - the `test_custom_dash_ag_grid` test in this file and
 - two new tests under the `class TestDunderMethodsAgGrid` inside the `test_ag_grid.py`.
 
Same for `class TestCustomDashDataTable` tests.

vizro-core/src/vizro/models/_components/figure.py
antonymilne: Unresolving this just to remind myself of something...

vizro-core/src/vizro/_vizro.py
```diff
@@ -32,7 +32,13 @@ def __init__(self, **kwargs):
...
 
         """
-        self.dash = dash.Dash(**kwargs, use_pages=True, pages_folder="", title="Vizro")
+        self.dash = dash.Dash(
+            **kwargs,
+            pages_folder="",
+            suppress_callback_exceptions=True,
+            title="Vizro",
+            use_pages=True,
+        )
```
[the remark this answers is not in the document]
antonymilne: > Should we allow users to overwrite these arguments?

No, because they are all (apart from `title`) essential to the functioning of Vizro. And the `title="Vizro"` here gets overwritten in the `build` method so its effect is very minor.

> If not, should we at least explain them somewhere in the docs?

Also no, because no user will care 😅

The only possible exception to this is the `suppress_callback_exceptions` where we're possibly suppressing too much, but let's just note that in a comment (see suggestion).

vizro-core/changelog.d/20240821_165515_petar_pejovic_return_empty_data_frame_from_tables_build.md
```diff
@@ -0,0 +1,47 @@
...
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Return an empty data on initial build of AgGrid, Table and Figure objects. ([#644](https://github.com/mckinsey/vizro/pull/644))
+- Reduce number of figure function calls for AgGrid, Table and Figure objects. ([#644](https://github.com/mckinsey/vizro/pull/644))
```
antonymilne, line 29: Let's try and make this a bit more user-facing and less implementation-specific. What is the actual impact of this change for users? Basically improved performance.

```suggestion
- Improve page loading time load for `AgGrid`, `Table` and `Figure`. ([#644](https://github.com/mckinsey/vizro/pull/644))
```

antonymilne (approved):
👏 👏 👏 Really awesome work - great investigation and write-up. I didn't actually test it but I trust your manual tests + if @maxschulz-COL can give it a quick spin!

I would love to get to the bottom of the question I had about why this doesn't trigger the `filter_interaction` since as you know it influences other things like the dynamic controls. But if it all works then happy to merge this without fully understanding it!

---

Just for other reviewers and to have it all in one place, there's a couple of things worth thinking about here.

Firstly, consider the case when you build a dashboard using dynamic data and it's running for 5 years. The data could have completely changed between the time you start the dashboard (run `pre_build` and then first run `build`) and the time you visit it (re-run `build`). Hence any kind of pre-calculated AgGrid or Graph or whatever might be completely irrelevant and _should not_ be cached between pre-build and build time.

Secondly, operations that run when you go to a page with an AgGrid are basically:

1. data loading
2. backend running of dag.AgGrid
3. network time
4. frontend rendering

(plus some overhead with server handling routing etc.)

Previously I had always assumed that (1) would be the bottleneck here. This was wrong since in @maxschulz-COL and @huong-li-nguyen examples the network time was actually very significant. So In actual fact:
* 1 is quick for static data, potentially slow for dynamic data
* 3 is quick for small data, potentially slow for large data
* 2 and 4 are quick

What this means is that in order for `build` methods to me fast they need to avoid both 1 and 3. What that boils down to is a very important rule that **`build` methods must not load data**.

I'm slightly sad that that the more we work on this, the less "useful content" is in `AgGrid` and similar models and the more it just becomes the same as `Figure` + the "action info" type stuff. My original intention was that the `build` method would be a bit more "real" in encoding the actual Dash component... but I am 100% in agreement now that this is the right way to do it. So it's great that we now have a good solution and a very clear direction and it's obvious what the right split between `build` vs. `__call__` is, which has been an open question for a loooong time. Definitely we're making good progress here! 💯

vizro-core/src/vizro/models/_components/figure.py
antonymilne: ok, all good 😀

vizro-core/src/vizro/tables/_dash_ag_grid.py
```diff
@@ -57,6 +57,9 @@ def dash_ag_grid(data_frame: pd.DataFrame, **kwargs) -> dag.AgGrid:
         >>> vm.Page(title="Page", components=[vm.AgGrid(figure=dash_ag_grid(...))])
 
     """
+    if "columnDefs" in kwargs:
```
maxschulz-COL, line 60: @petar-qb This is a massive speedup as well - if you agree It hink we should implement it, and also for the Dash datatable. Together with your improvements, this will speed up most table implementations

maxschulz-COL (approved):
⭐ 🕵️ Fantastic investigation and great implementation!!

I reviewed the code in detail, and I like it, well done! I have _not_ tested the entire table you showed, I take it you did that plenty, but trust that this is therefore the best solution. Giving it a short spin, this is quite nice!

I added another optiimzation that I could find, it is quite significant. If you agree, then let's add that one as well 💪 

After that consideration, good to merge from my side

vizro-core/src/vizro/tables/_dash_ag_grid.py
petar-qb: @maxschulz-COL Thanks for the optimisation! For the "Page_3" in the example, the server response time is decreased from `9.5s` -> `2.5s`. 

However, after spending some time investigating the `columnDefs` for AgGrid I found a few more ways to define its elements:
1.     {"valueGetter": {"function": "params.data.sepal_length"}, "headerName": "Col_4"}
2.     {
        'headerName': 'Petal',
        'children': [
            {'field': "petal_width", 'headerName': 'Width'},
            {'field': "petal_length", 'headerName': 'Length'},
        ]
    },
3. And like this -> https://dash.plotly.com/dash-ag-grid/column-definitions#accessing-nested-data-values  
4.     {"field": "petal_length", "headerName": "Col_1"},
       {"field": "petal_length", "headerName": "Col_2"},

To accommodate all these combinations, we'd need to implement a complex if-else structure combined with recursion (for nested fields) to handle the search mechanism (searching for the final set of columns for the `rowData`).

Even then, we don't have a perfect solutions because there are some use-cases when the `rowData` has to contain an entire data_frame (no matter what is in the `columnDefs`). For example, if users want to implement a `custom filter interaction`, where the target is filtered by a `hidden column` when the source agGrid row data is clicked. With this optimisation, this `hidden column` would be preemptively deleted and callback would raise an exception. 

So, I'm not convinced that adding this logic to `dash_ag_grid` or `dash_data_table` figure functions by default is the best approach. Instead, Vizro users can filter the data_frame before passing it to the data_manager, which could lead to even better performance improvements.

petar-qb:
@antonymilne 
> I'm slightly sad that that the more we work on this, the less "useful content" is in AgGrid and similar models and the more it just becomes the same as Figure + the "action info" type stuff

Just to say that absence of custom code across different models makes me really happy. This enhances the genericity of the codebase, making it more maintainable, easier to understand and easier to customise from the users' side.

The "actions_info" is an idea I got earlier this year while documenting ActionsV2 process and is placed on the Miro board as the first ticket on the AV2 Timeline.

vizro-core/src/vizro/tables/_dash_ag_grid.py
maxschulz-COL: Hm very good points! This is a tricky one, because I would guess that many people would provide a DF, then pick some columns with `columnDefs` and be done with it.

The problem with your suggested solution is that it does not cover a case that might also occur frequently: minimizing data loading from source, ie wanting to have one central DF loaded, but also customizing the table shown nonetheless.

Overall I think you are correct though, and we should not do what I proposed. Do you think we should work with docs and/or even some warning?

For the simple case I provided, maybe we could raise a warning that this may unnecessarily slow down the server response time or something similar?

petar-qb: I've had an idea for a long time for working with big data in Vizro (and I'm already gathering and writing down good practices). I think that big-data handling in Vizro should get a whole page in the documentation, where this case will be one of the many items. 

What do you think? Do we have to emphasise anything just now, or we can wait until the "big-data handling in Vizro" doc page become added?

maxschulz-COL: If it is not too far away, then it's ok to now have it now haha. But it should come soon!

maxschulz-COL (approved):

vizro-core/src/vizro/tables/_dash_ag_grid.py
petar-qb: Here's the [Issue](https://github.com/McK-Internal/vizro-internal/issues/1230) where I referenced this discussion and your commit along with some other recommendations we should add.

petar-qb:
@antonymilne 
> I would love to get to the bottom of the question I had about why this doesn't trigger the filter_interaction since as you know it influences other things like the dynamic controls. But if it all works then happy to merge this without fully understanding it!

After the on_page_load action returns figures on the page, the [trigger_to_global_store](https://github.com/mckinsey/vizro/blob/208af57e6139f478d6f963af4b4e26c7cca19c0f/vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js#L1-L3) is triggered. This client-side callback is triggered as expected according to -> https://dash.plotly.com/advanced-callbacks#prevent-callback-execution-upon-initial-component-render, because the callback `Input` is generated by the on_page_load and its `Output` is a global local store (that already exists in the client side of the app).

This should cause the [gateway](https://github.com/mckinsey/vizro/blob/208af57e6139f478d6f963af4b4e26c7cca19c0f/vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js#L5-L77) client-side callback to be triggered as well, since the `trigger_to_global_store` Output is the Input of the `gateway` callback (which would cause filter_interaction to be triggered). 🤔 However, I didn't succeed to figure out why the gateway callback is not triggered (as it should be), and wasn't able to trigger the gateway callback even when I removed all the `prevent_initial_call=True` parameters.

antonymilne:
> After the on_page_load action returns figures on the page, the [trigger_to_global_store](https://github.com/mckinsey/vizro/blob/208af57e6139f478d6f963af4b4e26c7cca19c0f/vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js#L1-L3) is triggered. This client-side callback is triggered as expected according to -> https://dash.plotly.com/advanced-callbacks#prevent-callback-execution-upon-initial-component-render, because the callback `Input` is generated by the on_page_load and its `Output` is a global local store (that already exists in the client side of the app).
> 
> This should cause the [gateway](https://github.com/mckinsey/vizro/blob/208af57e6139f478d6f963af4b4e26c7cca19c0f/vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js#L5-L77) client-side callback to be triggered as well, since the `trigger_to_global_store` Output is the Input of the `gateway` callback (which would cause filter_interaction to be triggered). 🤔 However, I didn't succeed to figure out why the gateway callback is not triggered (as it should be), and wasn't able to trigger the gateway callback even when I removed all the `prevent_initial_call=True` parameters.

Thanks for looking into it. I wonder if it's related to this issue? https://github.com/McK-Internal/vizro-internal/issues/1105#issue-2378867831

No need to solve it now, but let's make a note of it so we don't lose track of it.

petar-qb:
> I wonder if it's related to this issue?

Great point! It seems like we're dealing with a similar scenario here.

My main goal is to prevent any action from triggering another `implicitly` (which seems like it's the case right now). Instead, all actions should be `explicitly` defined using action chains. Given that, I wouldn't suggest altering the current behaviour. However, I'd like to dedicate more time to investigating this further and making the action loop more robust. We can address this in a separate PR when we find more time.

PR #645, opened 2024-08-22
[Tidy] Consolidate themes to optimise for usage outside dashboard

huong-li-nguyen:
## Description
For context on why we do this: https://github.com/McK-Internal/vizro-internal/issues/1208

- Optimise vizro templates for charts (so usage of charts outside of dashboards e.g. notebook)
- Add dashboard specific overwrites as discussed to `Graph.__call__`

## Screenshot
**Before (in notebook):**
before

**After (in notebook):**
after

Charts used inside dashboard should not visually change
## Notice


vizro-core/tests/unit/vizro/actions/test_filter_interaction_action.py
```diff
@@ -101,7 +101,7 @@ def target_scatter_filtered_continent(request, gapminder_2007, scatter_params):
     if country:
         data = data[data["country"].isin([country])]
 
-    return px.scatter(data, **scatter_params).update_layout(margin_t=24)
+    return px.scatter(data, **scatter_params).update_layout(margin_t=24, title_pad_l=0, title_pad_r=0, margin_l=24)
```
huong-li-nguyen, line 104: @petar-qb @antonymilne  - can you guys remind me again why we need this everywhere in the unit tests for the actions? 😬 

If we add more overwrites to optimise the charts inside the dashboard, this might get lengthier as well. Not sure if there is a better way of doing this?

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -137,6 +129,28 @@ def _filter_interaction(
 
         return data_frame
 
+    @staticmethod
+    def _optimise_fig_layout_for_dashboard(fig):
```
huong-li-nguyen, line 133: @antonymilne and I decided to put it inside the graph.__call__ for now. If at some point there will be more overwrites such that it's difficult to keep track, we can create separate theme and merge it to the existing vizro_dark/vizro_light.

vizro-core/tests/unit/vizro/actions/test_filter_interaction_action.py
huong-li-nguyen: For example, I am wondering whether we even need to compare the entire fig. The template and custom dashboard overwrites are covered in other unit tests as well. So wouldn't it suffice if we just compare` fig[data]` as this is the relevant part for the actions unit tests? Or would that strip off too much?

I am also fine leaving it as is and we can get back to it if it indeed becomes lengthier 😄

petar-qb: Actions like filter/parameter/filter_interaction always return "called" figure objects (like `vm.Graph.__call__()`) . 

As we apply style changes to the `actual` vm.Graph inside the `__call__()` method, we have to apply the same changes to the `expected` vm.Graph so we can compare them properly.

petar-qb: > Or would that strip off too much

I think so, especially since the parameters affect the "layout" of the figure.

huong-li-nguyen: Okay! I'll leave it as is then

antonymilne (approved):
Looks good, thanks for doing this!

An unrelated change I'm going to make soon might mean that we refactor this stuff again in a couple of days, but your changes are still definitely a good step in that direction and worth making 👍 Please leave https://github.com/McK-Internal/vizro-internal/issues/1208 open for now though.

vizro-core/tests/unit/vizro/actions/test_filter_interaction_action.py
antonymilne: Agree this is a bit annoying with tests but let's leave it like this for now since a change I make soon might simplify them anyway.

petar-qb (approved):
I like it! 🥇

petar-qb:
> Looks good, thanks for doing this!
> 
> An unrelated change I'm going to make soon might mean that we refactor this stuff again in a couple of days, but your changes are still definitely a good step in that direction and worth making 👍 Please leave [McK-Internal/vizro-internal#1208](https://github.com/McK-Internal/vizro-internal/issues/1208) open for now though.

@antonymilne 
I really like the theme mechanism improvement you mentioned yesterday.
Do you think it is possible to do all the customisations from `_optimise_fig_layout_for_dashboard` in a similar way as you want to process the theme?

PR #647, opened 2024-08-23
[QA] Vizro ai dashboard score tests

l0uden:
## Description
Here I've tried to write score tests for dashboard creation. 
Reports can be found in the artefacts for the tests: https://github.com/mckinsey/vizro/actions/runs/10526889888

Now it is very straightforward, but also efficient in my opinion.

Let's discuss!

## Notice


maxschulz-COL:
Great start Alexey! ⭐ And I think it goes 100% in the right direction!

A couple of loose thoughts right from the start, no definite comments, just to bounce with you:
- could the reports be visible on click? It is a bit cumbersome to download them?
- could we establish a more streamlined prompt and expectation structure, something that can ideally be easily extended
  - I am picturing something like: test writer chooses DFs plus prompt and fills in a certain data structure about the expected results
  - Then tests iterate through all such defined tests for all different models, and output in single and aggregated form
  - This is essentially what you have just less manual and more extensible
  - Ideally the criteria that are evaluated are the same for every requested dashboard, so that we could compare easy and more difficult dashboards
- Introduce tags for the individual tests, so that we know whether this is a easy, medium or hard test
- It would be amazing if this could run under `hatch run test-integration`, so that it is easy to be used in local dev

Does that make sense, happy to chat

l0uden:
> Great start Alexey! ⭐ And I think it goes 100% in the right direction!
> 
> A couple of loose thoughts right from the start, no definite comments, just to bounce with you:
> 
> * could the reports be visible on click? It is a bit cumbersome to download them?
> * could we establish a more streamlined prompt and expectation structure, something that can ideally be easily extended
>   
>   * I am picturing something like: test writer chooses DFs plus prompt and fills in a certain data structure about the expected results
>   * Then tests iterate through all such defined tests for all different models, and output in single and aggregated form
>   * This is essentially what you have just less manual and more extensible
>   * Ideally the criteria that are evaluated are the same for every requested dashboard, so that we could compare easy and more difficult dashboards
> * Introduce tags for the individual tests, so that we know whether this is a easy, medium or hard test
> * It would be amazing if this could run under `hatch run test-integration`, so that it is easy to be used in local dev
> 
> Does that make sense, happy to chat

Thanks Max!

> * could the reports be visible on click? It is a bit cumbersome to download them?

Unfortunately no, but here's what I've found from GitHub developer. Anyway it was 4 years ago and no ETA...
https://github.com/actions/upload-artifact/issues/14#issuecomment-620728238
https://github.com/actions/upload-artifact/issues/14#issuecomment-1856520972

> * could we establish a more streamlined prompt and expectation structure, something that can ideally be easily extended

My thoughts about it were that we should change the report and number of score steps according to the different prompts or changes to them. So if you're adding one more page to the prompt you just add one more check and change the report. I don't think we will do it very often. So there's no any difficult score rating under it yet. I'm open to do it more flexible in future like you've described

> * Introduce tags for the individual tests, so that we know whether this is a easy, medium or hard test

I can easily mark them using `@pytest.mark` to even run them separately

> * It would be amazing if this could run under `hatch run test-integration`, so that it is easy to be used in local dev

You can run it like this already

maxschulz-COL (approved):
Approve from my side 👍 

Let's iterate on the feedback in next PRs.

How are the artifacts then created when you run it with `hatch run test-integration`?

l0uden:
> Approve from my side 👍
> 
> Let's iterate on the feedback in next PRs.
> 
> How are the artifacts then created when you run it with `hatch run test-integration`?

The artifact for the local run will be only one file with the report right now (in future there will as many reports as tests) under vizro-ai folder.

vizro-ai/tests/integration/test_dashboard.py
```diff
@@ -21,10 +22,15 @@ def setup_test_environment():
         chromedriver_autoinstaller.install()
 
 
+@pytest.mark.parametrize(
+    "model_name",
+    ["gpt-3.5-turbo"],
+    ids=["gpt-3.5"],
+)
```
lingyielia, line 29: why I seems to receive 2 sets of test result in the output file? Is it because the "append" mode of CSV? In that case, does it mean it will keep appending infinitely on the runner machine?
Screenshot 2024-08-27 at 17 22 27

vizro-ai/tests/integration/test_dashboard.py
```diff
@@ -51,3 +57,81 @@ def test_simple_dashboard(dash_duo):
...
+        controls_num_1,
+        controls_num_2,
+        components_types_1,
+        components_types_2,
+        controls_types_2,
+    ]
+    score = sum(prescore)
+
+    with open(f"report_model_{model_name}.csv", "a", newline="") as csvfile:
+        writer = csv.writer(csvfile, delimiter=",", quotechar="|", quoting=csv.QUOTE_MINIMAL)
```
lingyielia, line 124: why do we need "," and "|" in the csv? In this case, the result when open up from excel is not readable.

Maybe only keep the ","? something like:
```
    with open(f"report_model_{model_name}.csv", "a", newline="") as csvfile:
        writer = csv.writer(csvfile)
        writer.writerow(["Description", "Score"])
        writer.writerow(["Pages exists", str(pages)])
        writer.writerow(["Correct pages number", str(pages_num)])
        writer.writerow(["Correct grid first page", str(grid_1)])
        writer.writerow(["Correct grid second page", str(grid_2)])
        writer.writerow(["Correct components number first page", str(components_num_1)])
        writer.writerow(["Correct components number second page", str(components_num_2)])
        writer.writerow(["Correct controls number first page", str(controls_num_1)])
        writer.writerow(["Correct controls number second page", str(controls_num_2)])
        writer.writerow(["Correct components types first page", str(components_types_1)])
        writer.writerow(["Correct components types second page", str(components_types_2)])
        writer.writerow(["Correct controls types second page", str(controls_types_2)])
        writer.writerow(["Total", f"'{score}/{len(prescore)}'"])
```
to get rid of "|"

vizro-ai/tests/integration/test_dashboard.py
```diff
@@ -51,3 +57,81 @@ def test_simple_dashboard(dash_duo):
...
+        writer.writerow([f"Correct grid first page, {grid_1}"]),
+        writer.writerow([f"Correct grid second page, {grid_2}"]),
+        writer.writerow([f"Correct components number first page, {components_num_1}"]),
+        writer.writerow([f"Correct components number second page, {components_num_2}"]),
+        writer.writerow([f"Correct controls number first page, {controls_num_1}"]),
+        writer.writerow([f"Correct controls number second page, {controls_num_2}"]),
+        writer.writerow([f"Correct components types first page, {components_types_1}"]),
+        writer.writerow([f"Correct components types second page, {components_types_2}"]),
+        writer.writerow([f"Correct controls types second page, {controls_types_2}"]),
+        writer.writerow([f"Total, {score}/{len(prescore)}"]),
```
lingyielia, line 137:
```suggestion
        writer.writerow(["Total", f"{(score / len(prescore)):.4f}"])
```
Maybe use float between 0.0000 and 1.0000? So the final scores are more comparable across test sets (say the "easy" test has 11 items, and the "difficult" test has 40 items)

lingyielia (changes requested):
This is so cool 😎 
For designing the different levels of difficulties, happy to join a PS to craft the prompt together!

vizro-ai/tests/integration/test_dashboard.py
```diff
@@ -51,3 +57,81 @@ def test_simple_dashboard(dash_duo):
...
+        components_num_2,
+        controls_num_1,
+        controls_num_2,
+        components_types_1,
+        components_types_2,
+        controls_types_2,
+    ]
+    score = sum(prescore)
+
+    with open(f"report_model_{model_name}.csv", "a", newline="") as csvfile:
```
lingyielia, line 123: Ideally if all results can be centralized in a folder under tests/integration, it would help keep local env clean. Currently all result CSVs are under vizro-ai

vizro-ai/tests/integration/test_dashboard.py
l0uden: That is because on the CI we have two runs in one job. One for the vizro from pypi and one for the vizro from the branch.

vizro-ai/tests/integration/test_dashboard.py
l0uden: Done

vizro-ai/tests/integration/test_dashboard.py
l0uden: Nice suggestion, thanks! Changed

vizro-ai/tests/integration/test_dashboard.py
l0uden: Done

lingyielia (approved):
Looks great to me! 👏 
Could you double check one more thing before merge? Could you try run this test locally one more time, and then do "hatch run lint". Last time when I reviewed it, the linting failed because of those generated data files (CSVs). 

Otherwise it looks great!

vizro-ai/tests/integration/test_dashboard.py
maxschulz-COL: Could we label the different reports?

vizro-ai/tests/integration/test_dashboard.py
maxschulz-COL: Yes, could we add them even in a report folder?

maxschulz-COL:
> Looks great to me! 👏 Could you double check one more thing before merge? Could you try run this test locally one more time, and then do "hatch run lint". Last time when I reviewed it, the linting failed because of those generated data files (CSVs).
> 
> Also on local, if I don't delete the test output CSVs, each time I run it, the results get appended. If that's the intended behavior when runs locally, maybe we shall add a column of ID (e.g., timestamp) so in the output file we can know which run is the latest, and we can filter one dimension (e.g., Correct components number first page) to compare performance across different runs. wdyt?
> 
> Otherwise it looks great!

Good points, agree with them!

- @l0uden We should definitely resolve the linting issue, I encountered that too.
- Let's give the different reports a time-stamp and a type (ie pypi and local)

Otherwise happy to merge. Can you schedule a PS next week that we discuss how to build on top of this? I have a few ideas about the report structure that differ from the current setup, but nothing too complicated 💪

l0uden:
@lingyielia and @maxschulz-COL , thanks for the ideas.
I've put reports into separate folder, added timestamp and vizro type to the report.
And resolved the linting issue :)

PR #648, opened 2024-08-23
[Demo] Replace KPI cards with dynamic figure functions

huong-li-nguyen:
## Description
- Replace static KPI cards with KPI card functions
- Introduce FlexContainer and optimise layout
- Update README and list out all resources
- Reduce font-size for KPI card functions
- Replace area chart with unstacked area chart incl. two traces

After this PR is merged, I will remove the entire demo in a separate PR (so we have the full commit history in this repository) and then the demo will only live in huggingface afterwards.

**TO DO:**

- [x] Update screenshots in QA repo

## Notice


vizro-core/src/vizro/static/css/figures.css
```diff
@@ -39,7 +39,7 @@ Note the relative url works with both serve_locally=True and serve_locally=False
 /* This should be replaced, when the bootstrap theme is finalized. Currently our font tokens don't match with the
  bootstrap font tokens, leading to deviations which we need to overwrite. */
 .card-kpi-title {
-  font-size: 1.5rem;
+  font-size: 1.2rem;
```
huong-li-nguyen, line 42: I've reduced the title for all KPI cards, as they previously had the same font-size as the page-title, which was too big

vizro-core/examples/kpi/utils/_charts.py
```diff
@@ -2,44 +2,32 @@
...
-                ),
-            ],
-            className="kpi-card-ref",
+        """Returns a flex container."""
+        return html.Div(
+            id=self.id, children=[component.build() for component in self.components], className=self.classname
         )
 
 
+vm.Container.add_type("components", FlexContainer)
```
nadijagraca, line 28: This can be removed as the type is added in the app.py as well.

nadijagraca (approved):
This looks amazing. Great job. 🙌 

@lingyielia should we advise CNX team to use new KPI cards instead of the hack with bar chart and traces?

vizro-core/examples/kpi/app.py
```diff
@@ -2,113 +2,128 @@
...
+        vm.Figure(
+            figure=kpi_card_reference(
+                df_kpi_cards,
+                value_column="Consumer disputed_2019",
+                reference_column="Consumer disputed_2018",
+                title="Consumer disputed",
+                value_format="{value:.1f}%",
+                reference_format="{delta:+.1f}pp vs. 2018 ({reference:.1f}%)",
+                icon="sentiment_dissatisfied",
+            )
```
petar-qb, line 76: Do you get a white square from the screenshot for the "Consumer disrupted" KPI card?

image

vizro-core/examples/kpi/utils/_helper.py
```diff
@@ -38,18 +40,89 @@ def clean_data_and_add_columns(data: pd.DataFrame):
...
     data["Region"] = data["State"].map(REGION_MAPPING)
     data["Company response"] = np.where(
         data["Company response - detailed"].str.contains("Closed"), "Closed", data["Company response - detailed"]
     )
     data["Company response - Closed"] = np.where(
         data["Company response - detailed"].str.contains("Closed"), data["Company response - detailed"], "Not closed"
     )
+
+    # Filter 2018 and 2019 only
+    data = data[(data["Year"].isin(["2018", "2019"]))]
```
petar-qb, line 60: The page loading process will be faster when we merge this PR -> https://github.com/mckinsey/vizro/pull/644. 

Therefore, it is likely that other years can be included in the table if it brings some important information.

petar-qb (approved):
Awesome work as usual! 👍

vizro-core/examples/kpi/app.py
huong-li-nguyen: Yes, unfortunately we can't get rid of this. It's the native scrolling bar corner 😬

vizro-core/examples/kpi/utils/_helper.py
huong-li-nguyen: I thought so, but I think for this demo it doesn't matter anymore :) I am also only adding two years as otherwise the homepage will loose a lot of horizontal space that would be reserved for a single filter 😅  For the next demos, for sure though!

vizro-core/examples/kpi/app.py
petar-qb: I succeeded to get rid of this with:
```py
kpi_banner = FlexContainer(
    id="kpi_banner_flex_container",
    components=[
    ...
)
```

and custom css:

```css
#kpi_banner_flex_container{
  overflow: hidden;
}
```

huong-li-nguyen: Yes, but unfortunately it doesn't function properly anymore then.

The only reason why I put it into a FlexContainer is that this functions as a KPI banner, meaning if you reduce the width of the screen, the cards will take a minimum width but there will be scrolling horizontally. This ensures that the cards are always well visible and do not look too squeezed.

However, if you turn off the overflow, they will be just cut-off, and parts of the cards will not visible.

petar-qb: I see it now! thanks for the explanation 😃

PR #649, opened 2024-08-23
[Tidy] Remove graph server side theme setting

antonymilne:
## Description

https://github.com/mckinsey/vizro/pull/523 moved the theme selector's update of graph template to the clientside, which was a huge improvement. However, there was still a bit of server-side theme updating hidden in `Graph.__call__` which implicitly relied on the clientside selected theme. The purpose of this was just to avoid an annoying flickering of the graph when the selected theme was not the same as the default theme.

I've now come up with a better solution so that _all the server-side theme settings can be removed_.

* This might seem like a minor optimisation, but actually it's another incremental step towards actions v2 since the "controls determine components" paradigm is getting cleaner now we can say that the theme selector _does not_ any more act like a control, since its behaviour is entirely clientside
* Changing theme during an animation still works as before
* Performance compared to the old solution is actually slightly worse unfortunately, but it's a tiny difference and I think worth it. 
  * Theme switching performance is identical to before since it's always been purely clientside
  * On page load and on applying controls: we've exchanged serverside `fig.layout.template` for an extra client-side callback that runs when the graphs are rendered. So server load and times are very slightly reduced and clientside is very slightly increased
  * Very rough profiling shows that for a page with n = 64 graphs (which is really _a lot_ and would probably cause lots of other problems), server response on my computer is maybe 0.5 seconds improved but the additional clientside callback adds ~2 seconds
  * For any kind of realistic page with a sensible number of graphs (say 16) the difference is really not noticeable as the clientside callback runs so quickly (100ms or so)

@petar-qb please could you give it a quick test? I've already tested pretty carefully but just to be sure. I've been using the very handy example app you wrote before in https://github.com/mckinsey/vizro/pull/523 🙂 

## Notice


vizro-core/src/vizro/static/js/models/dashboard.js
```diff
@@ -13,24 +13,22 @@ export function _update_ag_grid_theme(theme_selector_checked) {
...
+      template: vizro_themes[theme_to_apply],
+    },
+  };
 
-  // Adjust `layout` property for the Plotly graph element
-  Plotly.relayout(plotly_graph, { template: theme_to_apply });
-
-  return dash_clientside.no_update;
+  // {} resets the figure.style to undo the {"visibility": "hidden"} from set_props in Graph.__call__.
+  return [updated_figure, {}];
```
petar-qb, line 31: Is it possible for a custom style to be overwritten by an empty {} figure.style?

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -57,28 +58,16 @@ def __call__(self, **kwargs):
...
         # to just swallow up the error here as it doesn't cause any problems.
-        try:
-            # At the moment theme_selector is always present so this if statement is redundant, but possibly in
-            # future we'll have callbacks that do Graph.__call__() without theme_selector set.
-            if "theme_selector" in ctx.args_grouping.get("external", {}):
-                theme_selector_checked = ctx.args_grouping["external"]["theme_selector"]["value"]
-                fig.layout.template = "vizro_light" if theme_selector_checked else "vizro_dark"
-        except MissingCallbackContextException:
-            logger.info("fig.update_layout called outside of callback context.")
+        with suppress(MissingCallbackContextException):
```
petar-qb, line 63: Does set_props throw MissingCallbackContextException if the ctx doesn't exist?

vizro-core/src/vizro/static/js/models/dashboard.js
```diff
@@ -13,24 +13,21 @@ export function _update_ag_grid_theme(theme_selector_checked) {
...
-  const theme_to_apply = theme_selector_checked
-    ? vizro_themes["light"]
-    : vizro_themes["dark"];
+  const theme_to_apply = theme_selector_checked ? "vizro_light" : "vizro_dark";
 
-  // Find the Plotly graph element in the HTML document
-  const plotly_graph = document
-    .getElementById(graph_id)
-    .querySelector(".js-plotly-plot");
+  const updated_figure = {
```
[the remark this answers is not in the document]
petar-qb: If performance is similar for 64 charts then I'm totally okay with this approach 👍

vizro-core/changelog.d/20240827_130318_antony.milne_remove_graph_server_side_theme_setting.md
petar-qb, line 1: Have you tested this solution in Jupyter notebook?

petar-qb (approved):
Awesome refactoring! I like getting rid of ctx theme from the server side (even though we introduce a small `{"visibility": "hidden"}` hack).

Thanks for fixing unit tests as well (I know how "not optimised" they actually are) 😄

I'm also fine with additional rendering time since it's quite negligible 👍

vizro-core/src/vizro/models/_components/graph.py
antonymilne: Yes:
```
dash.exceptions.MissingCallbackContextException: dash.callback_context.set_props is only available from a callback!
```

vizro-core/changelog.d/20240827_130318_antony.milne_remove_graph_server_side_theme_setting.md
antonymilne: No, but let me do so! Good thinking.

antonymilne: Tested and all is good

huong-li-nguyen (approved):
This looks great! 🚀  The unit tests look so much cleaner now! 🧹 Great to get rid of all the template specific stuff there. 

I think the failing screenshot test is a false positive. At least I checked the screenshot and it seems to be fine actually? Maybe just re-run and see if it passes.

vizro-core/src/vizro/static/js/models/dashboard.js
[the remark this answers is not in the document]
huong-li-nguyen: I hope no one will ever create a dashboard with 64 charts 😆

PR #658, opened 2024-08-28
[Tidy] Add UV as installer, include pip in default env

maxschulz-COL:
## Notice

### DIFF
--- a/vizro-core/hatch.toml
+++ b/vizro-core/hatch.toml
@@ -17,6 +17,7 @@ scripts = {add = "scriv create --add", collect = ["scriv collect --add", "- hatc
 
 [envs.default]
 dependencies = [
+  "pip",
   "devtools[pygments]",
   "coverage[toml]>=6.5",
   "pytest",
@@ -29,6 +30,7 @@ dependencies = [
   "openpyxl",
   "jupyter"
 ]
+installer = "uv"
 
 [envs.default.scripts]
 example = "hatch run examples:example {args:scratch_dev}"  # shortcut script to underlying example environment script.
--- a/vizro-ai/hatch.toml
+++ b/vizro-ai/hatch.toml
@@ -10,6 +10,7 @@ scripts = {add = "scriv create --add", collect = ["scriv collect --add", "- hatc
 
 [envs.default]
 dependencies = [
+  "pip",
   "devtools[pygments]",
   "coverage[toml]>=6.5",
   "pytest",
@@ -24,6 +25,7 @@ dependencies = [
   "chromedriver-autoinstaller>=0.6.4",
   "urllib3<2.0.0"  # helps to resolve bug with urllib3 timeout from vizro-ai integration tests: https://bugs.launchpad.net/python-jenkins/+bug/2018567
 ]
+installer = "uv"
 
 [envs.default.env-vars]
 VIZRO_AI_LOG_LEVEL = "DEBUG"


vizro-core/hatch.toml
```diff
@@ -14,9 +14,11 @@ matrix.python.features = [
 dependencies = ["scriv"]
 detached = true
 scripts = {add = "scriv create --add", collect = ["scriv collect --add", "- hatch run lint:lint --files=CHANGELOG.md"]}
+installer = "uv"
 
 [envs.default]
 dependencies = [
+  "pip",
```
antonymilne, line 21: Why do we need this? I tried removing it and creating a fresh environment already had `pip` installed.

antonymilne (approved):
🚀 🚀 🚀 Thank you for doing this! Efficient development with ephemeral environments and fast CI has always been of vital important to me, and this is a HUGE boost. On CI it's shortened the environment setup from ~50s to ~7s. If we can get those QA tests faster @l0uden then this will be awesome.

When I first tried this out with `hatch run example` I was very disappointed that it was still running very slowly, and then I realised it's because we have some environments that don't inherit from `default` and so you need to set `installer` for them too. I did this in [99a8655](https://github.com/mckinsey/vizro/pull/658/commits/99a865591109fb5cb91fa8a809fb216e59fc00af) and all works well.

Also people need to upgrade to v1.10.0 for this to be useful. I decided to completely uninstall it (from pipx) and install the [new recommended GUI route](https://hatch.pypa.io/1.12/install/). It took about 30 seconds and all works great. So if people struggled before to setup hatch and did it in a conda environment or with pip or similar then I'd recommend they uninstall and do it this new approach which is way nicer. Let's post about this on the team channel when this is merged.

I must say that, reading through the new hatch blog posts and learning more about this, I'm reminded here of just how great hatch is 😀 So any astral alternative will have to be really really amazing to make me want to switch (which is definitely possible, since astral make incredible stuff too, but I'm really happy that I went for hatch 18 months ago).

maxschulz-COL:
> When I first tried this out with `hatch run example` I was very disappointed that it was still running very slowly, and then I realised it's because we have some environments that don't inherit from `default` and so you need to set `installer` for them too. I did this in [99a8655](https://github.com/mckinsey/vizro/pull/658/commits/99a865591109fb5cb91fa8a809fb216e59fc00af) and all works well.

Awesome, thanks! Weird though why it occured for `hatch run example`, does that not use the default example?


> Also people need to upgrade to v1.10.0 for this to be useful. I decided to completely uninstall it (from pipx) and install the [new recommended GUI route](https://hatch.pypa.io/1.12/install/). It took about 30 seconds and all works great. So if people struggled before to setup hatch and did it in a conda environment or with pip or similar then I'd recommend they uninstall and do it this new approach which is way nicer. Let's post about this on the team channel when this is merged.

I updated with homebrew, that also works fine 💪 

> I must say that, reading through the new hatch blog posts and learning more about this, I'm reminded here of just how great hatch is 😀 So any astral alternative will have to be really really amazing to make me want to switch (which is definitely possible, since astral make incredible stuff too, but I'm really happy that I went for hatch 18 months ago).

Do you mean this blog: https://hatch.pypa.io/1.10/blog/2024/04/18/hatch-v1100/#visual-studio-code?

The link I posted I wanted to mention also at TL, seems like VS Code supports Hatch environments now, but for me that has not worked so well yet, so would love to discuss with a fellow VS Code user 😅

maxschulz-COL:
@antonymilne Final question, did you try this _when UV was not installed_? I would like to know what happens? If you haven't tried, I might uninstall `hatch` to try so myself!

vizro-core/hatch.toml
antonymilne: Hmm no, this worked fine for me straight away. Doing `hatch env prune` and `hatch run pip freeze` works without the explicit `pip` dependency.

antonymilne:
> Awesome, thanks! Weird though why it occured for `hatch run example`, does that not use the default environment?

No, I made a new `examples` environment to isolate it a couple of weeks ago just to keep things tidy when we added `black` as a dependency for the chart gallery.

> Do you mean this blog: https://hatch.pypa.io/1.10/blog/2024/04/18/hatch-v1100/#visual-studio-code?
> 
> The link I posted I wanted to mention also at TL, seems like VS Code supports Hatch environments now, but for me that has not worked so well yet, so would love to discuss with a fellow VS Code user 😅

Haha no, unfortunately you're on your own in VSCode still until I have time to give it another bash 😅 I just meant the general progress that's been made over the last 12 months, e.g. distributing Python itself, the work on pyapp, etc.

> @antonymilne Final question, did you try this when UV was not installed? I would like to know what happens? If you haven't tried, I might uninstall uv to try so myself!

Yes! I did it without uv installed and all worked great.

vizro-core/hatch.toml
antonymilne: I guess the source of truth here will be whether it runs on CI without the dependency listed?

antonymilne: Ah, I think you are right right here and `pip` doesn't come with the environment indeed. Looks like my computer was just picking up a global `pip`.

That said, why do we actually want `pip`? Is it just for the `pip freeze` steps in CI?

maxschulz-COL: That and I do find myself sometimes using it when I am not sure why my `hatch` environment atm is actually at. So I think that is quite a natural command

antonymilne: Cool. Over time maybe we should try and remove this to become fully `uv`-native but doesn't matter really.

I guess you might need to add `pip` as a dependency to some of the isolated environments also btw.

petar-qb (approved):
This is absolutely stunning. After I updated `hatch` and installed `uv`, running:
```
hatch env prune
hatch run example
```

on the `tidy/add_uv_as_installer` branch took me 10x less time than running the same two command on the `main` branch. It takes `~7 min` for the main and `~40 s` for the feature branch. 
I'll definitely continue investigating `uv`. 

The PR looks nice! However, this PR has to be approved by @lingyielia or @Joseph-Perkins (codeowners) too as I can see.

Joseph-Perkins (approved):
approved

PR #661, opened 2024-08-28
[Docs] Fix examples section of several API references

maxschulz-COL:
## Description
Missing `:` does not render docs examples properly

## Notice


petar-qb:
@maxschulz-COL your changes are overwritten by `pre-commit-ci`.

maxschulz-COL:
> @maxschulz-COL your changes are overwritten by `pre-commit-ci`.

What a bummer...

vizro-core/tests/tests_utils/asserts.py
```diff
@@ -27,7 +27,12 @@ def assert_component_equal(left, right, *, keys_to_strip=None):
     If keys_to_strip is set to STRIP_ALL then only the type and namespace of component
     will be compared, similar to doing isinstance.
 
-    Examples
+    Args:
+        left: Dash component to compare.
+        right: Dash component to compare.
+        keys_to_strip: Keys to strip from the component dictionary before comparison.
```
huong-li-nguyen, line 33: If we add the args description, might be worth moving the doc description on keys_to_strip to the actual args description 👍 
```suggestion
        left: Built Dash component.
        right: Expected Dash component.
        keys_to_strip: Keys to strip from the component dictionary before comparison.  If keys_to_strip is set to STRIP_ALL then only the type and namespace of component will be compared, similar to doing isinstance.
```

vizro-core/src/vizro/tables/_dash_ag_grid.py
```diff
@@ -50,7 +50,11 @@
 def dash_ag_grid(data_frame: pd.DataFrame, **kwargs) -> dag.AgGrid:
     """Implementation of `dash_ag_grid.AgGrid` with sensible defaults to be used in [`AgGrid`][vizro.models.AgGrid].
 
-    Examples
+    Args:
+        data_frame: DataFrame containing the data to be displayed.
+        kwargs: Additional keyword arguments to be passed to the `dash_ag_grid.AgGrid` component.
+
```
huong-li-nguyen, line 56: Shall we add a  Returns as well? To be consistent with e.g. the kpi_card functions and their docstrings. Same for dash_data_table.

```
Returns:
        A Dash Bootstrap Components card (`dbc.Card`) containing the formatted KPI value and reference.

```

huong-li-nguyen (approved):
All good from my side 👍 Seems like the docs don't like the typing for the kwargs argument though 😅 Hope we can fix it easily 🤞

vizro-core/tests/tests_utils/asserts.py
petar-qb: I like Li's suggestion. Just remove the double space character between "... comparison." and "If keys_to_strip ..." (which would probably be handled by linter)

petar-qb (approved):
After applying Li's suggestion, this is going to be ready for merge by my opinion 👍

vizro-core/tests/tests_utils/asserts.py
maxschulz-COL: Great suggestion. I did not do the suggestion on left and right though, because I think that is too specific. While it is the general convention, the function also could handle the other way.

maxschulz-COL:
Just for reference, the initial behaviour was actually a bug: https://github.com/astral-sh/ruff/issues/13139

This PR fixes those issues though, but will be goof to remember for posterity

PR #665, opened 2024-08-29
[Feat] Make chart `bg-color` transparent inside dashboard

huong-li-nguyen:
## Description
@antonymilne - Your PR https://github.com/mckinsey/vizro/pull/662 arrived at the perfect time 🥳

I'm currently working on a demo dashboard and realized that this change is not only something I need but also broadly beneficial.

At present, if users want to change the background colors of charts, they have to update the template or configure it for each chart in the post update. Although previous PRs have made this process easier, users still have to hardcode color values, limiting them to a single background color for all charts inside the dashboard.

What users—and I—actually need is for the chart background to be transparent (if used inside the dashboard). This would allow users to flexibly color the container in which the charts are placed, enabling the charts to inherit the background color of their parent container.

**Note: Charts used outside the dashboard will look as normal with the bg-color applied** 

FYI @Joseph-Perkins

## Screenshot

Screenshot 2024-08-29 at 12 12 15

## Notice


antonymilne (approved):

Joseph-Perkins (approved):
approved

maxschulz-COL:
Quickly checking something

maxschulz-COL (changes requested):
Cool thing in principle, but are we happy with the following consequence (which I think would stem from this change)?

Downloaded charts via the plotly chart bar (thing above charts) are now quite ugly (see below screenshot):

image

If yes, then of course let's go :), but I find this behaviour a bit strange 🤔

antonymilne:
Hmmm this is a great point @maxschulz-COL. Having a download to image button feels like nice plotly functionality that's weird to turn off, but also I've never actually had any reason to use it rather than just doing a print screen so I don't know if anyone cares...

Overall I have zero preference at all between transparent background + turn off download button vs. leave how it is now. So up to you guys.

If we do go for the transparent background then let's make this bar transparent too - maybe `modebar_bgcolor` I think?
image

huong-li-nguyen:
> Hmmm this is a great point @maxschulz-COL. Having a download to image button feels like nice plotly functionality that's weird to turn off, but also I've never actually had any reason to use it rather than just doing a print screen so I don't know if anyone cares...
> 
> Overall I have zero preference at all between transparent background + turn off download button vs. leave how it is now. So up to you guys.

I am not in a rush to merge this, as I can also define it individually for each chart. Then, it's just a matter of custom charts. We can also leave it for now and come back to it when more people request it 👍  I am also curious how many people actually use the mode bar at all 🤔. They always take up space and are not really intended for mobile.

Issue #673, opened 2024-09-02
Some template arguments don't seem to be properly passed on


huong-li-nguyen:
We have this specification in our template_dark and template_light:

```
    # CHART TYPES
    template_dark.data.bar = [
        go.Bar(
            marker={"line": {"color": template_dark["layout"]["paper_bgcolor"], "width": 1}},
            textfont={"color": COLORS["WHITE_55"]},
        )
    ]
```

However, the width doesn't actually seem to be taken on by any of our bar charts, so I am not even sure the specifications inside template_XX.data.XXX work properly. 

[...]

PR #679, opened 2024-09-04
[Tidy] Remove redundant template chart configurations

huong-li-nguyen:
## Description
- Remove redundant template configurations (e.g. go.Table, bar specific configurations)

Note: Based on this issue https://github.com/mckinsey/vizro/issues/673 I've initially raised, it's important to understand  that `template.data.XXX` impacts both go.Figures and px charts. However, we configured it incorrectly, which is why our specified bar width didn't take effect. However, while creating this PR, it became clear that we should not set a default bar width: https://github.com/mckinsey/vizro/pull/679#issuecomment-2329614756

Another point to note is that px charts can override certain settings from the template, such as the` legend_title_text`. From my observations, they also override the titles and values of the x/y axes, the legend title and the overall color sequence. This means that any specifications within the template unfortunately don't apply. Consequently, I've removed the configuration for the legend title since it had no effect (by default, it's turned off for go.Figures).

## Notice


vizro-core/src/vizro/_themes/_templates/template_dark.py
```diff
@@ -71,31 +71,7 @@ def create_template_dark() -> Template:
...
 
     # CHART TYPES
-    template_dark.data.bar = [
-        go.Bar(
-            marker={"line": {"color": template_dark["layout"]["paper_bgcolor"], "width": 1}},
-            textfont={"color": COLORS["WHITE_55"]},
-        )
-    ]
-
-    template_dark.data.table = [
```
huong-li-nguyen, line 81: Removed the go.Table as this was always just intended as a temporary solution until we had the dash_data_table and/or dash_ag_grid in place

vizro-core/src/vizro/_themes/_templates/common_values.py
```diff
@@ -40,7 +40,6 @@ def create_template_common():
         legend_y=-0.20,
         legend_title_font_family=primary_font,
         legend_title_font_size=14,
-        legend_title_text="",
```
huong-li-nguyen, line 43: Doesn't have any effect, therefore removing here

petar-qb (approved):
I like it! 🧹

petar-qb:
For reviewers: 

This is the example from the `main` branch:
image

And this is from the feature branch:
image

huong-li-nguyen:
Seems like that change on the bar width doesn't work well on the `px.bar` if the data is not aggregated. If I can't get both working, I might have to undo the bar width change 😢 , but the template removal will stay

Screenshot 2024-09-04 at 18 16 41

huong-li-nguyen:
Okay, I went into the go.Bar source code and it seems like we shouldn't set the bar.width inside the template. What previously worked fine for a categorical x-axis, doesn't work anymore for a numerical x-axis, and it's because the width is defined in "position axis units" so needs to vary depending on the x-axis scale. Fixing it inside would therefore only look good on a categorical axis, but not on a numeric one. Will therefore remove the setting of the bar width, but the rest stays 👍 

Screenshot 2024-09-04 at 18 21 56

nadijagraca (approved):
LGTM. 👍

PR #730, opened 2024-09-23
[Tidy] Make vale pre-commit hook run in verbose mode

antonymilne:
## Description

Currently we don't see vale suggestions/warnings unless an error is also raised. This PR changes that so that suggestions and warnings are shown even when no error is raised. The linting job will still pass even with these suggestions and warnings shown, but we'll be more aware of them now.

@stichbury I've tried this locally but would be great if you could try also just so it works as you'd like. Just run `hatch run lint` as per normal and you should see the suggestions and warnings even when there's no errors.

## Notice


huong-li-nguyen (approved):
Works perfectly 👍 🚀

stichbury (approved):
LGTM, and it's good incentive to get those suggestions and warnings "fixed"

PR #738, opened 2024-09-23
[Tidy] contribution flow

antonymilne:
## Notice

### DIFF
--- a/vizro-core/changelog.d/20240923_113528_antony.milne_contribution_flow.md
+++ b/vizro-core/changelog.d/20240923_113528_antony.milne_contribution_flow.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Deprecated
+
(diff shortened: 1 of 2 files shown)


vizro-core/src/vizro/models/_base.py
```diff
@@ -278,14 +278,15 @@ def _to_python(
 
             >>> import vizro.models as vm
             >>> card = vm.Card(text="Hello, world!")
-            >>> print(card._to_python())
+            >>> card._to_python()
```
maxschulz-COL, line 281: isnt it nicer to use `print` to see the result?

maxschulz-COL (approved):
LGTM! Cool stuff! For me the GUi installation did not work after remove the homebrew one, but after reverting to homebrew, all is good with my initial setup :)

vizro-core/src/vizro/models/_base.py
antonymilne: No, because if you actually do this on the REPL then it will show you the result anyway. Just like doing it in Jupyter. That's the most natural way to "read" a docstring example rather than including the print in the example itself.

huong-li-nguyen (approved):
LGTM 👍

.github/workflows/checks-vizro-ai.yml
```diff
@@ -36,10 +36,10 @@ jobs:
           python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Install Hatch
-        run: pip install --upgrade hatch
+        run: pip install hatch
 
-      - name: List dependencies
-        run: hatch run all.py${{ env.PYTHON_VERSION }}:pip freeze
+      - name: Show dependency tree
+        run: hatch run pip tree
```
petar-qb, line 42: Maybe a stupid question but: "What's `pip tree`?"

vizro-core/src/vizro/models/_base.py
maxschulz-COL: But it shows it without the formatting, e.g. linebreaks `\n` no?

.github/workflows/checks-vizro-ai.yml
antonymilne: It's the `uv` equivalent of `pipdeptree`. Basically it gives not just a flat list of dependencies but also the chain showing where they come from. This is very useful in debugging CI failures when one of your transitive dependencies has changed and broken something and you're trying to work out why you're using it. Like this:
```
selenium v4.2.0
├── urllib3 v1.26.20
├── trio v0.26.2
│   ├── attrs v24.2.0
│   ├── sortedcontainers v2.4.0
│   ├── idna v3.10
│   ├── outcome v1.3.0.post0
│   │   └── attrs v24.2.0
│   └── sniffio v1.3.1
└── trio-websocket v0.11.1
    ├── trio v0.26.2 (*)
    └── wsproto v1.2.0
        └── h11 v0.14.0
```

vizro-core/src/vizro/models/_base.py
antonymilne: Ah ok yes, let me revert this then 👍

PR #755, opened 2024-09-30
[Tidy] Remove font assignment to elements in vizro-bootstrap

huong-li-nguyen:
## Description
- Update HF links to show app in full height/width view
- Removal of font-styling for elements `p, label, ul, li, blockquote and a`- based on this PR: https://github.com/McK-Private/vizro-bootstrap/pull/32

- These previously defined properties now lead to differences in the screenshots - decided to remove them and update all screenshots given that the QB design library doesn't assign any styling to elements either: https://github.com/mckinsey/vizro-qa/pull/122

      ```
      letter-spacing: -0.112px;
      line-height: 16px;
      ```

**Result: Elements will have no font-styling attached and will inherit anything their parent has.** 

Would like a review mostly from @pruthvip15 and then another codeowner so we can merge 👍 

## Notice


vizro-core/examples/visual-vocabulary/assets/css/custom.css
```diff
@@ -73,10 +73,6 @@ img[src*="#chart-icon"] {
   margin: 0;
 }
 
-.intro-text p a {
```
huong-li-nguyen, line 76: @lingyielia @nadijagraca or @maxschulz-COL  - this can be removed from the vizro-ai custom.css as well after this merge. I think you've guys used it somewhere?

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
```diff
@@ -9,11 +9,6 @@ but do not want to take over to `vizro-bootstrap` as these settings might not be
   margin-bottom: 0;
 }
 
-/* TO DO: Check if this should live in vizro-bootstrap */
-.card p:last-of-type {
```
huong-li-nguyen, line 13: Shouldn't live here, but instead be defined in a custom.css in an assets folder

pruthvip15 (approved):

README.md
```diff
@@ -112,7 +112,7 @@ Our visual vocabulary dashboard helps you to select and create various types of
 each chart type, and offers sample Python code to create these charts with [Plotly](https://plotly.com/python/) and
 embed them into a Vizro dashboard.
 
-<a href="https://huggingface.co/spaces/vizro/demo-visual-vocabulary/">
+<a href="https://vizro-demo-visual-vocabulary.hf.space">
```
maxschulz-COL, line 115: I am always of two minds when deciding which view to take. The other view also has advantages, as you can like it easier, it leads more naturally to our other demos etc. I think it could make sense to leave it that way?

maxschulz-COL (approved):

README.md
huong-li-nguyen: I wasn't a fan of that view because I prefer the 100% height option for any dashboard-demo actually 😬. Our dashboards are typically optimized for that view, so with the HF upper tab bar it sometimes results in unnecessary scrolling.

I do see the value of your second point, but since we reference our general examples gallery just above this section and this section is primarily to showcase the visual vocabulary, I would prefer the view that is actually optimized for this dashboard.

PR #795, opened 2024-10-07
[Demo] Add diverging stacked bar chart to visual-vocabulary

hxe00570:
## Description
Issue -> https://github.com/mckinsey/vizro/issues/707

- Adds a diverging stacked bar chart (positive and negative values only) - meaning this diverging bar chart function is not suitable for variations with a neutral category group: https://jscharting.com/examples/chart-types/bar/diverging-stacked/
- Plotly resource: https://community.plotly.com/t/need-help-in-making-diverging-stacked-bar-charts/34023

## Screenshot
Screenshot 2024-10-16 at 16 13 58

## Notice

- [X] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":


huong-li-nguyen:
@hxe00570 - could you provide me write access to your forked repository? :)

huong-li-nguyen:
> @huong-li-nguyen invite sent! I do not see where to grant permissions but I assume once you accept the invite I can do so.

I just accepted :)

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -233,3 +233,70 @@ def dumbbell(data_frame: pd.DataFrame, x: str, y: str, color: str) -> go.Figure:
...
     return fig
+
+
+@capture("graph")
+def diverging_stacked_bar(
+    data_frame,
+    y: str,
+    category_pos: List[str],
+    category_neg: List[str],
+    color_discrete_map: Optional[Dict[str, str]] = None,
```
huong-li-nguyen, line 244: @antonymilne - could I get your opinion on the API? I've kept it to the minimum, so decided against customising labels etc. Also because it would take me too long to style this properly. 

`data_frame`, `y` and `color_discrete_map` - I've kept similar to the plotly express yntax
`category_pos` and `category_neg` - These are the ones I've introduced. I didn't like any of the existing plotly express argument names as something similar doesn't really exist. `x1` and `x2` don't really cut it for me as a List of columns need to be provided here.

Let me know what you think 🤔

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -233,3 +233,70 @@ def dumbbell(data_frame: pd.DataFrame, x: str, y: str, color: str) -> go.Figure:
...
+) -> go.Figure:
+    """Creates a horizontal diverging stacked bar chart (with positive and negative values only) using Plotly's go.Bar.
+
+    This type of chart is a variant of the standard stacked bar chart, with bars aligned on a central baseline to
+    show both positive and negative values. Each bar is segmented to represent different categories.
```
huong-li-nguyen, line 249: I decided to go for the diverging stacked bar chart (without a neutral group). Enabling the neutral group would be a lot more complex and in general that chart with the neutral group is not really a recommended one given the difficulties to interpret it: https://blog.datawrapper.de/divergingbars/

In general, this is one of those charts which look very fancy, but might not be a good recommended chart. I recommend reading through the article above :)

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -233,3 +233,70 @@ def dumbbell(data_frame: pd.DataFrame, x: str, y: str, color: str) -> go.Figure:
...
+        )
+
+    # Update layout and add central baseline
+    fig.update_layout(barmode="relative")
+    fig.add_vline(x=0, line_width=2, line_color="grey")
+
+    # Update legend order to go from most negative to most positive
+    category_order = category_neg[::-1] + category_pos
+    for i, category in enumerate(category_order):
+        fig.update_traces(legendrank=i, selector=({"name": category}))
```
huong-li-nguyen, line 300: I've tried everything out before, and this was the best I came up with that seemed to work. Working with `traceorder` doesn't work because you then also need to switch up the order of how the traces get generated, which messes up the chart as it needs to go from most negative to most positive. 

`legendrank` is a new argument that seems to fix it, but we need to revert the order for the `category_neg`

vizro-core/examples/visual-vocabulary/pages/examples/diverging_stacked_bar.py
```diff
@@ -0,0 +1,122 @@
...
+       y (str): The name of the categorical column in the data frame to be used for the y-axis (categories)
+       category_pos (List[str]): List of column names in the data frame representing positive values. Columns should be
+            ordered from least to most positive.
+       category_neg (List[str]): List of column names in the DataFrame representing negative values. Columns should be
+            ordered from least to most negative.
+       color_discrete_map: Optional[Dict[str, str]]: A dictionary mapping category names to color strings.
+
+    Returns:
+       go.Figure: A Plotly Figure object representing the horizontal diverging stacked bar chart.
+    """
```
huong-li-nguyen, line 62: I reckon this is a very large docstring 😄 Shall we always add these docstrings to the code examples? Or would it suffice to only add these in the `custom_charts.py` ? The difference is that it would either appear or not in the code examples on the right.

vizro-core/examples/visual-vocabulary/pages/deviation.py
```diff
@@ -51,4 +52,49 @@
...
+    components=[
+        vm.Card(
+            text="""
+
+            #### What is a diverging stacked bar?
+
+            A diverging stacked bar chart is like a stacked bar chart but aligns bars on a central baseline instead of
+            the left or right. It displays positive and negative values, with each bar divided into segments for
+            different categories. This type of chart is commonly used for percentage shares, especially in survey
+            results using Likert scales (e.g., Strongly Disagree, Disagree, Neutral, Agree, Strongly Agree).
```
huong-li-nguyen, line 68:
```suggestion
            results using Likert scales (e.g., Strongly Disagree, Disagree, Agree, Strongly Agree).
```

vizro-core/examples/visual-vocabulary/pages/deviation.py
```diff
@@ -51,4 +52,49 @@
...
+            &nbsp;
+
+            #### When should I use it?
+
+            A diverging stacked bar chart is useful for comparing positive and negative values and showing the
+            composition of each bar. However, use this chart with caution: since none of the segments share a
+            common baseline, direct comparisons can be more challenging. For clearer comparisons, consider using a
+            100% stacked bar chart with a baseline starting from the left or right. For more insights on the potential
+            pitfalls, we recommend reading the article from
+            [Datawrapper on diverging stacked bar charts](https://blog.datawrapper.de/divergingbars/).
```
huong-li-nguyen, line 79: This article is really interesting: https://blog.datawrapper.de/divergingbars/
Give it a read! 📖

huong-li-nguyen:
Thank you @hxe00570 for having a first stab at this! I've changed some of the logic of your original chart and refactored the chart to make reusing the chart function a bit simpler for people. You were already very close to the final solution! 🚀 

I'll iterate on this after getting some reviews but I am sure we will merge this by the end of the week :)

vizro-core/examples/visual-vocabulary/pages/examples/diverging_stacked_bar.py
```diff
@@ -0,0 +1,122 @@
...
+        fig.update_traces(legendrank=i, selector=({"name": category}))
+
+    return fig
+
+
+page = vm.Page(
+    title="Diverging stacked bar",
+    components=[
+        vm.Graph(
+            figure=diverging_stacked_bar(
```
huong-li-nguyen, line 105:
```suggestion
            title="Would you recommend the pastry to your friends?",
            figure=diverging_stacked_bar(
```

vizro-core/examples/visual-vocabulary/pages/deviation.py
```diff
@@ -51,4 +52,49 @@
...
+        """
+        ),
+        vm.Graph(
+            figure=diverging_stacked_bar(
```
huong-li-nguyen, line 83:
```suggestion
            title="Would you recommend the pastry to your friends?",
            figure=diverging_stacked_bar(
```

vizro-core/examples/visual-vocabulary/pages/deviation.py
```diff
@@ -51,4 +52,50 @@
...
+        vm.Card(
+            text="""
+
+            #### What is a diverging stacked bar?
+
+            A diverging stacked bar chart is like a stacked bar chart but aligns bars on a central baseline instead of
+            the left or right. It displays positive and negative values, with each bar divided into segments for
+            different categories. This type of chart is commonly used for percentage shares, especially in survey
+            results using Likert scales (e.g., Strongly Disagree, Disagree, Neutral, Agree, Strongly Agree).
+
```
huong-li-nguyen, line 69:
```suggestion
            results using Likert scales (e.g., Strongly Disagree, Disagree, Agree, Strongly Agree).

```

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -233,3 +233,70 @@ def dumbbell(data_frame: pd.DataFrame, x: str, y: str, color: str) -> go.Figure:
...
+
+    Inspired by: https://community.plotly.com/t/need-help-in-making-diverging-stacked-bar-charts/34023
+
+    Args:
+       data_frame (pd.DataFrame): The data frame for the chart.
+       y (str): The name of the categorical column in the data frame to be used for the y-axis (categories)
+       category_pos (List[str]): List of column names in the data frame representing positive values. Columns should be
+            ordered from least to most positive.
+       category_neg (List[str]): List of column names in the DataFrame representing negative values. Columns should be
+            ordered from least to most negative.
```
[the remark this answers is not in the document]
petar-qb: I don't understand why columns should be ordered from least to most positive (and similar to negative values), and also what exactly that means. When I changed the example configuration like below (I just switched the "category_pos" order of columns - and `color_discrete_map` additionally), I got the chart like this below: 

```py
                category_pos=["Strongly Agree", "Agree"],
                category_neg=["Disagree", "Strongly Disagree"],
                color_discrete_map={
                    "Agree": "#1a85ff",
                    "Strongly Agree": "#70a1ff",
                    "Disagree": "#ff5584",
                    "Strongly Disagree": "#d41159",
                },
```

image

vizro-core/examples/visual-vocabulary/pages/examples/diverging_stacked_bar.py
petar-qb: I really think we should leave docstrings in the examples.

petar-qb (approved):
Awesome example and its API added 🚀. This PR looks really good.

vizro-core/examples/visual-vocabulary/pages/examples/diverging_stacked_bar.py
huong-li-nguyen: Yes, I agree. I think users also don't actually access `custom_charts.py` via the UI but only the code copy clipboard, so it probably needs to be added to both right now 👍

vizro-core/examples/visual-vocabulary/custom_charts.py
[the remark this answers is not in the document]
huong-li-nguyen: In a diverging stacked bar chart, it's really important to sort the colors in a specific order. This is because the chart draws the sections one after the other starting from the middle. In your example, the right side starts with "Strongly Agree" and then "Agree," which feels off because the Likert scale naturally goes from "Strongly Disagree" to "Disagree," then "Agree," and finally "Strongly Agree."

When the natural order isn't followed, like in your current setup, it makes the chart harder to read and understand. The colors aren't just random categories where the order doesn't matter; in a diverging stacked bar chart, the order is key. It always shows responses from least to most extreme on both the positive and negative sides.

So, it's important to arrange the categories from least extreme to most extreme to make sure the chart is clear and easy to interpret :)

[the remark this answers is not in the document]
huong-li-nguyen: Yes, exactly. It's a data viz best practice guidance :) I don't want to enforce it via checks etc. , because in the end it's still their chart.

vizro-core/examples/visual-vocabulary/custom_charts.py
antonymilne: tbh I'm not a big fan of `category_pos` and `cateogry_neg` but after playing around with the ordering of legends and traces myself for a couple of hours I absolutely understand why you did it 😅 I have an alternative proposal here - let's discuss this afternoon.

vizro-core/examples/visual-vocabulary/custom_charts.py
antonymilne: Very good article indeed! Maybe at some point in future we could have one with the neutral group separated off like they suggest, but definitely seems like there's a strong case to be made for just not doing this kind of plot at all 👍

antonymilne (approved):
I have a bit of a different way of doing this but I think we should merge this version as it is and then come back to it later 🙂 

Thank you for all your hard work on this @hxe00570 and @huong-li-nguyen 🙏 I know it wasn't an easy chart to do!

Issue #732, opened 2024-09-23
[POC] Investigate whether we can get rid of outer container with className to use vizro-bootstrap


huong-li-nguyen:
Currently, if someone creates a pure Dash app and wants to use the vizro-bootstrap CSS file, they have to wrap their Dash app content inside an outer div with the className=vizro_light or vizro_dark. 

If not provided, it does not take on the styling of our Vizro bootstrap stylesheet. Take this minimal example: https://github.com/mckinsey/vizro/blob/poc/add-example-bootstrap/vizro-core/examples/scratch_dev/app.py#L54-L55

@pruthvip15 - could you investigate the following?
- **Why do we need to provide this outer container with className vizro_dark/vizro_light?** I have a rough understanding here, but would be great if we could get to the bottom of it. I assume that this is how the theme switch in bootstrap works e.g. if you take a look at [this example](https://getbootstrap.com/docs/5.3/customize/color-modes/) from the bootstrap docs, they also always wrap it inside an outer container called "light" or "dark". Otherwise it cannot find the variables and relevant scss files. 
- **Is there any way how we could remove the requirement to define an outer container with classname?** So basically the goal is to be able to remove L54-L55 if possible and optimally, everything should work as expected.

@pruthvip15 - you can take the branch and dev example from above :)

I think because we define this in our bootstrap theme, we need this outer container, so I am not sure if there is any way how this can be read in without having to specify this out container

```
.vizro_dark, [data-bs-theme="dark"] {
...

}
```

pruthvip15:
Hey Yes, We can do this.

If we assume the default theme is dark, then we can change the above to
```
:root, .vizro_dark, [data-bs-theme="dark"] {
...

}
```
And this will ensure that by default the dark mode will be applied.

If the user wants to put light mode, then he has to explicitly wrap the container in .vizro_light or  [data-bs-theme="dark"]

I have made a small example here: https://codepen.io/pruthvip15/pen/mdNJXpm

In the above example, we can see that be default the app is in dark mode, if for a inner container the user wants light mode, he can add the respective class to it

huong-li-nguyen:
Ahh, that's great to know! @antonymilne - would above suffice? Then people only have to specify vizro_light if they want the light theme.

antonymilne:
Thanks for this @pruthvip15. I think this sounds good but will need a slight modification to be exactly right on Vizro - let me just add some more notes and next steps here.

[...]

PR #804, opened 2024-10-10
[Tidy] Remove `vizro_dark` and `vizro_light` as `classNames`

huong-li-nguyen:
## Description
- Closes: https://github.com/mckinsey/vizro/issues/732
- Removes all occurrences of `vizro_dark` and `vizro_light` as `classNames` (this means that we still use `vizro_dark/light` to reference our chart templates or as arguments to the dashboard.theme, however, it is not used anywhere as a className anymore as we use `data-bs-theme` for that instead)
- Connected PR on vizro-bootstrap that created the new `vizro.bootstrap.min.css` file: https://github.com/McK-Private/vizro-bootstrap/pull/33
- Dark theme is the default for Bootstrap

Would like a review from @antonymilne  and @pruthvip15 🙏 

## Notice


vizro-core/src/vizro/__init__.py
```diff
@@ -2,7 +2,7 @@
 import os
 
 import plotly.io as pio
-from dash.development.base_component import Component, ComponentRegistry
```
huong-li-nguyen, line 5: @antonymilne - this didn't seem to be used. Wondering why the linter didn't remove it by itself. Let me know if we need to import this

vizro-core/src/vizro/static/js/models/dashboard.js
```diff
@@ -3,7 +3,7 @@ function update_dashboard_theme(theme_selector_checked) {
     "data-bs-theme",
     theme_selector_checked ? "light" : "dark",
   );
-  return theme_selector_checked ? "vizro_light" : "vizro_dark";
+  return window.dash_clientside.no_update;
```
petar-qb, line 6: I guess we can remove `return` line from here all together with the `dash.Output` from the `update_dashboard_theme` `clientside_callback` in the `Dashboard.build`.  
Reference -> https://github.com/plotly/dash/pull/2867

vizro-core/src/vizro/static/css/variables.css
```diff
@@ -96,7 +96,7 @@
   --text-code-literal: #6bb5fd;
 }
 
-.vizro_light {
+[data-bs-theme="light"] {
```
pruthvip15, line 99: do we need to add :root here?

vizro-core/src/vizro/static/js/models/dashboard.js
huong-li-nguyen: Done! :) Thank you!

vizro-core/src/vizro/static/css/variables.css
huong-li-nguyen: I just added it in, just in case 👍

huong-li-nguyen: I added it in to try, but the layout was wrong then, because our default is actually the dark theme. So if we add root here, it will have a light theme while the charts are dark.

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -161,8 +161,11 @@
   display: none;
 }
 
-#logo-dark,
-#logo-light {
+[data-bs-theme="dark"] #logo-dark {
+  display: inline;
+}
+
+[data-bs-theme="light"] #logo-light {
```
pruthvip15, line 168: not sure if this is relevant for this PR.

Can we refactor the styling for the #logo, display:inline can go to line 157.

Why do we need
```
#logo-dark,
#logo-light {
  display: none;
}
```

vizro-core/src/vizro/static/css/variables.css
pruthvip15: you meant to say default is light ?

vizro-core/src/vizro/static/css/layout.css
huong-li-nguyen: I thought the same before but then remembered again why we have this here 😄 If you change it to above, then the hack of switching between a dark/light logo will not work anymore. The current code ensures that the light logo is only visible when the light theme is activated and the other way around. We don't want to always hide them.

vizro-core/src/vizro/static/css/variables.css
huong-li-nguyen: Vizro's default is the dark theme 😬 Vizro-bootstrap's default is the light theme as this seems to be the more common way of doing it in other bootstrap themes.

petar-qb (approved):
I really like this change!

vizro-core/src/vizro/__init__.py
antonymilne: Hmm yes, no need to import this any more. I wonder why it didn't get removed by `ruff`.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -170,7 +170,6 @@ def build(self):
                 ActionLoop._create_app_callbacks(),
```
antonymilne, line 170: The `Output("dashboard-container", "className")` in this file shouldn't be there any more.

vizro-core/src/vizro/static/js/models/dashboard.js
```diff
@@ -3,7 +3,6 @@ function update_dashboard_theme(theme_selector_checked) {
     "data-bs-theme",
```
antonymilne, line 3: FYI @petar-qb this is an example of a callback that we always want to run even when it's not explicitly triggered. Otherwise we end up with no `data-bs-theme` specified. Currently this works well but it wouldn't if this callback was not triggered automatically unless we set `:root` to be dark theme as well (which might happen).

antonymilne (approved):
Looks much nicer, thank you! I just left a comment on the other PR but it doesn't make a difference for Vizro framework.

vizro-core/src/vizro/static/css/layout.css
huong-li-nguyen: Will wait with merging until we've decided on the default

vizro-core/src/vizro/static/css/variables.css
huong-li-nguyen: Will wait with merging until we've decided on the default

vizro-core/src/vizro/static/js/models/dashboard.js
huong-li-nguyen: > The `Output("dashboard-container", "className")` in this file shouldn't be there any more.

@antonymilne @petar-qb  - If I remove the Output as suggested above, this callback doesn't get triggered anymore and we end up with no `data-bs-theme` specified. The layout doesn't match anymore because of that. 

I assume if we would set` :root` to be the dark theme, this issue doesn't appear anymore, but if we stick with the light theme - how can I always trigger this callback if we simultaneously want to remove the `Output("dashboard-container", "className")`? I guess I would have to add some placeholder Output there in which case I would just leave the current Output in?

vizro-core/src/vizro/models/_dashboard.py
huong-li-nguyen: We currently seem to need it such that the callback is triggered, otherwise the layouts don't match, see: https://github.com/mckinsey/vizro/pull/804#discussion_r1799135660

vizro-core/src/vizro/static/js/models/dashboard.js
petar-qb: This is strange, and probably a bug from the dash side. However, I suggest to leave the Output then (or to target some other dummy component, but not mandatory).

antonymilne: Oh that's weird. I guess it's either to do with that complicated Dash scheme that I never understand about which callbacks get triggered depending on whether the output is already on the page or just a bug like @petar-qb said.

Let's put some current `Output` there with a comment that it doesn't do anything but is needed to run the callback and put `return window.dash_clientside.no_update` in the callback.

Setting `:root` here won't help us regardless of what colour it is because if you've specified `dashboard.theme = <not the default :root theme>` then we need the `update_dashboard_theme` to run to set it correctly. So this callback has to always run whatever we do.

vizro-core/src/vizro/static/css/variables.css
huong-li-nguyen: Dark will be set as the default

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,49 +1,60 @@
...
+                                icon="payment",
+                                title="KPI Card III",
+                            )
+                        ),
+                    ]
+                ),
+            ],
         ),
     ],
+    # **{"data-bs-theme": "light"}
```
huong-li-nguyen, line 56: @antonymilne - I just tested these out. 

**1. Use of vizro bootstrap - default:** If you don't specify any `data-bs-theme`, the default will be dark as discussed. It looks a bit off because it's on a white background, but that's expected as we didn't provide any `bg-color` to the general body container in our `vizro-bootstrap` yet.

Screenshot 2024-10-15 at 09 16 39

**2. Use of vizro bootstrap - light theme** This is how it looks like if you uncomment this line and add `**{"data-bs-theme": "light"}`:
Screenshot 2024-10-15 at 09 15 37

**3. Usage of any other bootstrap theme:** I just checked the dbc.themes.BOOTSTRAP but it looks as expected and default to their default theme.

antonymilne: I think that's perfect, and nothing more to test. Thank you!

If it turns out in the future that adding `bg-color` to match the dark default is annoying or awkward for some reason then I'm happy to turn our default back to the light theme btw. My preference for default = dark is just a weak one.

PR #808, opened 2024-10-14
[Tidy] Complete dumbbell chart example

huong-li-nguyen:
## Description
- Follow-up PR to finish off dumbbell chart

## Notice


vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -193,3 +193,26 @@ def radar(data_frame, **kwargs) -> go.Figure:
     fig = px.line_polar(data_frame, **kwargs)
     fig.update_traces(fill="toself")
     return fig
+
+
+@capture("graph")
+def dumbbell(data_frame, x, y, color):
+    # Add two dots to plot
```
huong-li-nguyen, line 200: TODO: Docstrings to be added

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -193,3 +193,44 @@ def radar(data_frame, **kwargs) -> go.Figure:
...
+        measure (List[str]): List specifying the type of each bar, can be "relative", "total", or "absolute".
+
+    Returns:
+        go.Figure: A Plotly Figure object representing the Waterfall chart.
+
+    For additional parameters and customization options, see the Plotly documentation:
+    https://plotly.com/python/reference/waterfall/
+
+    """
+    # Add two dots to plot
```
huong-li-nguyen, line 218: Currently just a placeholder, i'll update that in a sec

huong-li-nguyen:
@antonymilne - you mentioned you had some comments - feel free to take a look already :) I still need to update the docstrings but you can already take a look

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -193,3 +193,43 @@ def radar(data_frame, **kwargs) -> go.Figure:
...
+        df_sub = data_frame[data_frame[y] == i]
+        fig.add_shape(
+            type="line",
+            layer="below",
+            y0=df_sub[y].values[0],
+            x0=df_sub[x].values[0],
+            y1=df_sub[y].values[1],
+            x1=df_sub[x].values[1],
+            line_color="grey",
+        )
```
antonymilne, line 231: You should never really need to follow a "loop through masked sub-dataframes" pattern in pandas. Using `groupby` works more nicely:

```suggestion
    for y_value, group in data_frame.groupby(y):
        fig.add_shape(
            type="line",
            layer="below",
            y0=y_value,
            y1=y_value,
            x0=group[x].min(),
            x1=group[x].max(),
            line_color="grey",
        )
```

vizro-core/examples/visual-vocabulary/pages/distribution.py
```diff
@@ -79,58 +80,63 @@
 
 butterfly = butterfly_factory("distribution")
```
antonymilne, line 82: This is a very confusing diff (not your fault). Is it just switching the order of `dumbell` and `histogram`?

petar-qb (approved):
LGTM 😃

vizro-core/examples/visual-vocabulary/pages/distribution.py
```diff
@@ -79,58 +80,63 @@
...
+            than bar charts, making it easier to compare groups. Common uses include comparing groups, such as showing
+            differences in performance metrics across various categories. Colors can be used to emphasize the direction
+            of changes or to distinguish between categories.
         """
         ),
-        vm.Graph(figure=px.histogram(tips, x="total_bill")),
-        make_code_clipboard_from_py_file("histogram.py"),
+        vm.Graph(
+            figure=dumbbell(
+                tips.groupby(["day", "sex"]).agg({"tip": "sum"}).reset_index(), y="day", x="tip", color="sex"
```
antonymilne, line 134: I think we should keep data-munging to a bare minimum in this dashboard. Currently we don't do any groupby aggregations elsewhere, so ideally we'd have an example here that doesn't need it too.

e.g. maybe this? I didn't try it out so not sure how it looks. The query is made much more complex which I don't like, but overall I think it's better since then a reader doesn't think they need to do groupby in order to use this plot. What matters is they have a dataframe of the expected schema to use the plot. (Possibly we should add more advanced examples in future to explain the sort of data wrangling with groupby that might be needed to achieve that required schema but I would't do that now)
```
gapminder().query("year.isin([1952, 2007]) and continent == 'Americas'"), y="country", x="lifeExp", color="year"
```

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -193,3 +193,43 @@ def radar(data_frame, **kwargs) -> go.Figure:
...
+
+    # Increase size of dots
+    fig.update_traces(marker_size=12)
```
antonymilne, line 234: The line currently feels a bit thin to me but not a strong feeling. I guess lollipop and this plot should have the same line thickness, whatever that is.

antonymilne (approved):
Thank you for pushing this to completion! Just a few comments but do merge when you're happy with it 👍

vizro-core/examples/visual-vocabulary/pages/distribution.py
huong-li-nguyen: Hmm.. your example would fit better in the `time` category then 🤔 which I agree, the dumbbell can also be used for that, but I think for the distribution example a categorical column would be better suited (similar to how we do it for the bar charts)

huong-li-nguyen: I can also just start a new data set from scratch 😄

vizro-core/examples/visual-vocabulary/custom_charts.py
huong-li-nguyen: I made it thicker 👍 The Lollipop is not merged, but i took the same numbers now

vizro-core/examples/visual-vocabulary/pages/distribution.py
huong-li-nguyen: Not sure to be honest 😄 The only thing I changed was to add the dumbbell to the bottom of the page, as I always add new charts to the bottom.

vizro-core/examples/visual-vocabulary/custom_charts.py
huong-li-nguyen: Perfect, thanks!

vizro-core/examples/visual-vocabulary/pages/distribution.py
huong-li-nguyen: I've replaced the data set with a dataset I created from scratch. I still need to think whether we actually want to use the dumbbell for the `time` dimension as well, in which case your example is perfect. However, I think there are reasons why the dumbbell is not a suggested chart in the FT visual vocabulary for the time dimension, but instead they use the slope chart: https://github.com/Financial-Times/chart-doctor/blob/main/visual-vocabulary/poster.png

But let me think about it, and we might add a version for the time dimension as well. For now I would merge without 👍

PR #629, opened 2024-08-13
[Docs] READY: First set of changes for demo->gallery transition

stichbury:
## Description
**[Built docs are here](https://vizro--629.org.readthedocs.build/en/629/)**

Some changes to copy where we currently mention "Demo" or point to vizro.mckinsey.com because the target is changing from a single gapminder example demo to a gallery of linked examples.

Max made a good suggestion:

> So concretely one the [examples page](https://vizro--629.org.readthedocs.build/en/629/pages/examples/examples/): could we not make the examples 	tab itself link to [vizro.mckinsey.com](http://vizro.mckinsey.com/) ? 

> What we would lose is the links to these examples: https://vizro--569.org.readthedocs.build/en/569/pages/examples/your-examples/ Any idea how we would bring those in? Maybe we can also add them to [vizro.mckinsey.com](http://vizro.mckinsey.com/)?

I like this idea as I think @antonymilne has also mentioned that the user examples get a bit lost in that second page within the examples section of the docs. We actually could still have the page in the docs to list the examples (move it into explanation perhaps, since longer term it should move away into an "awesome" list on GitHub like the one Kedro has).

Reviewer question: Shall I do this (here of in a separate PR)? 

1. Swap the example tab from linking to docs to linking straight to the gallery at vizro.mckinsey.com
2. Move any text we want to keep on the [examples page](https://vizro--629.org.readthedocs.build/en/629/pages/examples/examples/) and related [user examples](https://vizro--629.org.readthedocs.build/en/629/pages/examples/your-examples/) elsewhere in docs -- explanation section perhaps
3. Later still, move user examples off to an awesome list as originally planned (will find the issue)
4. 👯 👯 👯 

## Notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

  - I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
  - I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
  - I certify that the use of this contribution as authorized by the Apache 2.0 license does not violate the intellectual property rights of anyone else.
  - I have not referenced individuals, products or companies in any commits, directly or indirectly.
  - I have not added data or restricted code in any commits, directly or indirectly.

PR #814, opened 2024-10-17
[Docs] Remove examples section and point through to gallery

stichbury:
## Description

**[Build version of PR is here](https://vizro--814.org.readthedocs.build/en/814/)**

Discussed here: https://github.com/mckinsey/vizro/pull/629
(Internal ticket #1205

* Moved relevant useful text from the examples page to the FAQ section.
* Updated the front page to point through to the gallery, and updated the information architecture. (It would be good if the menu opened the gallery in a separate tab/window if possible -- I'm not sure how to do this with mkdocs)
* Moved `your-examples.md` to the explanation section
* Added @huong-li-nguyen's Charming Data video to the "Your examples" page

## Notice


huong-li-nguyen (approved):
LGTM 👍

stichbury:
@antonymilne so this is an example of where it's flagging that I need a changelog despite only modifying docs.

huong-li-nguyen:
> @antonymilne so this is an example of where it's flagging that I need a changelog despite only modifying docs.

I think it's because you change the `mkdocs.yml`, which is outside the `docs` folder. Every change inside the `docs` folder doesn't require a changelog file, but I agree with you - we should also ignore the `mkdocs.yml` if possible 🤔

vizro-core/docs/pages/explanation/faq.md
```diff
@@ -34,7 +34,17 @@ We do not consider frontend changes (such as changing the appearance of a compon
...
+
+Within Vizro's GitHub repository, the [`examples/dev` folder](https://github.com/mckinsey/vizro/tree/main/vizro-core/examples/dev) of `vizro-core` contains example code using the pydantic model. If you have any problems running the example code, please [raise an issue](https://github.com/mckinsey/vizro/issues) on the Vizro repository.
+
+!!! note
+
+    The pydantic model should be your preferred method of dashboard configuration, but you can also [define a dashboard with YAML, JSON, or as a Python dictionary](../user-guides/dashboard.md). To guide you, the `examples/dev` folder also contains an alternative `yaml_version` of the gapminder example, which offers the same example as the pydantic model but via YAML configuration.
+
+### Examples from Vizro users
+
+We maintain a separate, curated page of [videos, blog posts, and examples of Vizro usage from our community](your-examples.md).
```
antonymilne, line 47:
```suggestion
We also maintain a separate, curated page of [videos, blog posts, and examples of Vizro usage from our community](your-examples.md).
```

antonymilne: Now that we have HuggingFace up and running, these dev examples are really intended only for developers, and they're explained in the contributing guide.

Things which I think might still be worth highlighting here or elsewhere in the documentation, in this PR or ini another one. wdyt @maxschulz-COL?
1. The HuggingFace gallery allows you to look at code as well as the dashboards
2. Swap our HF features dashboard into the YAML version of it and link to that as a complete example of a YAML dashboard in https://vizro.readthedocs.io/en/latest/pages/user-guides/dashboard/
3. Link in https://vizro.readthedocs.io/en/latest/pages/user-guides/run/#jupyter to the Jupyter notebook `dev/example` because I think that's the only Juypyter notebook example we have? Or maybe just delete that example entirely since we never use it and it doesn't add anything to the docs really, since we already have a copy and pastable example there of a simple dashboard

antonymilne (commented):
Thank you for doing this! Looks good, but I have a few comments and questions that I'd like @maxschulz-COL to consider before we merge.

For the link opening in a new window vs. current one: I think this is fine how it is, since generally speaking we should prefer to open in the same tab. On the homepage we could replace the markdown link with an HTML one that includes "opens in new page" icon, but I don't think this would look good on the top bar of the navigation. So tbh I'd just leave it how you have it.

As for the changelog being needed: as you and Li said, this is because you modified mkdocs.yml that lives outside the docs. We could put this in so it's treated as part of the docs, but now that I think about it again I actually think we should change the logic so that a changelog is required for only changes to `src`. So docs and everything else outside `src` would all be excluded rather than `mkdocs.yml` being a special exclusion. wdyt @maxschulz-COL>? My reasoning is:
1. changelogs are really only required when there's a change to the package that affects users, which necessarily means changing code in `src`
2. in case there's some edge cases (e.g. dropping a Python version), you can just manually add a changelog

Overall I think this will be a slightly smoother process since a changelog would only be demanded when it might be needed.

maxschulz-COL:
>  So docs and everything else outside `src` would all be excluded rather than `mkdocs.yml` being a special exclusion. wdyt @maxschulz-COL>?

I think that is fine, let's go for that

vizro-core/docs/pages/explanation/your-examples.md
```diff
@@ -18,6 +18,7 @@ If you have something you'd like us to include on the list, or spot something th
     * [Build Python Data Apps with Vizro Dash](https://www.youtube.com/watch?v=wmQ6_GZ0zSk).
     * [Introduction to Vizro Actions - Plotly Dash](https://www.youtube.com/watch?v=bom-9275Cic&t=8s).
     * [Use Dash AG Grid within a Vizro app](https://www.youtube.com/watch?v=YvtVcXwQw0E).
+    * [Visual Vocabulary Dashboard with Vizro Plotly](https://www.youtube.com/watch?v=OZNAokBKT-M).
```
maxschulz-COL, line 21:
```suggestion
    * [Visual Vocabulary Dashboard showcasing charting options in Vizro and Plotly](https://www.youtube.com/watch?v=OZNAokBKT-M).
```

maxschulz-COL (approved):
Lgtm! I like it more now, highlights our examples page more, and hyou can still get to user examples quickly

vizro-core/docs/pages/explanation/faq.md
maxschulz-COL: 1. Agree, where do we want to mention?
2. Sure can do, I'll create a ticket: https://github.com/orgs/McK-Internal/projects/15/views/9
3. I think we need that section, because it allows us to explain the problems with the managers! But we can delete the source code. @stichbury wanna do that quickly?

Wdyt guys?

antonymilne: 1. I'd probably mention it here, something like this:
```
For a gallery of examples of Vizro in action, take a look at [vizro.mckinsey.com](https://vizro.mckinsey.com). For dashboards in our the [HuggingFace gallery](https://huggingface.co/vizro), you can see the app code by clicking the "Files" tab in the top banner.

We also maintain a separate, curated page of [videos, blog posts, and examples of Vizro usage from our community](your-examples.md).
```

2. Thanks!
3. Agree we want the section _and_ we actually also want the source code because it shows the `jupyter_mode` argument and makes it easy for a user to copy and paste into a Jupyter notebook. Sorry I wasn't clear before. The thing I think we should consider deleting is the `example/dev/app.ipynb` example itself, not the docs.

maxschulz-COL: Yes let's delete the code in `example/dev/app.ipynb` but not the docs example.

antonymilne (approved):

vizro-core/docs/pages/explanation/faq.md
stichbury: Thanks both, I'll put that in now.

PR #817, opened 2024-10-18
[Tidy] Remove gitleaks

antonymilne:
## Description

After quite a bit of discussion and [testing](https://github.com/mckinsey/vizro/pull/801), I have decided we are good to remove gitleaks after all. Currently we have GitHub advanced security and Gitguardian working. Gitguardian might be switched off because GitHub advanced security seems to be a superset of its features (e.g. it will even stop you putting credentials in an edit done directly in the GitHub UI).

The reason that test credentials didn't get detected by these tools before is because their threshold for what constitutes a real risky credential is higher than gitleaks. e.g. they only care about AWS keys and IDs if they are provided together, they check that the length and format of these is real to avoid flagging false positives, etc. All the security people at the Firm seem happy that these tools are sensitive enough and that they are working correctly, so I'm proceeding with the tidy up and removing it.

## Notice


maxschulz-COL (approved):
SOunds good to me, thanks for checking and testing. I think that was important after all :)

stichbury (approved):
🏆 Approved from docs perspective, which is somewhat secondary but have my ✅ anyway!

PR #821, opened 2024-10-22
[Demo] Use plotly.express for more charts in the visual vocabulary

antonymilne:
## Description

Various small bits of tidying but the main change is that I've rewritten `butterfly`, `column_and_line`, `dumbbell` and `diverging_stacked_bar`. I already discussed this with @huong-li-nguyen before she went on holiday so I think this should be fine to merge already and then she can take another look when she's back.

Appearance of the graphs is almost exactly the same in all the examples, so no new screenshots here.

Here's our preferred scheme:
* base plot on `px` where possible
* pass through `**kwargs` for all arguments apart from `data_frame` (which you need to specify due to `capture` decorator)
* write code to be as generic as possible i.e. not hardcoding x vs. y or assuming the structure of `data_frame` using `groupby` etc.

Now these plots are much more generic they are suitable for potential addition to `vizro.charts` rather than being specific examples.

The main advantage of this is that we get lots of the nice parts of `px` for free, e.g.
* we can handle long and wide format data
* you can pass arguments like `pattern_shape` or `log_x` etc. and they will work (some arguments like facets probably won't fully, but the vast majority of them will)
* you get the nice hover data, handling of automatic colour schemes, other nice touches etc. from `px`

> [!NOTE]
> (up for discussion, maybe will add to docs if it seems like good advice). In general with Vizro dashboards I recommend always starting off by basing your plot on `px`. Writing something fully generic takes quite a while and is generally not worth the effort if it's a custom chart for a single dashboard, but it's nice to get the hover data and familiar `px` API for free.

## Notice


vizro-core/examples/visual-vocabulary/pages/deviation.py
```diff
@@ -80,18 +80,14 @@
         """
         ),
         vm.Graph(
-            title="Would you recommend the pastry to your friends?",
+            title="I would recommend this pastry to my friends",
+            # TODO: when it's possible to update a figure without making a whole new function, remove the - signs from
+            # the x-axis (and maybe hover labels). It's not possible to do this dynamically in diverging_stacked_bar.
```
antonymilne, line 85:
```suggestion
```

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -13,40 +13,37 @@
...
     A butterfly chart is a type of bar chart where two sets of bars are displayed back-to-back, often used to compare
     two sets of data.
 
     Args:
-        data_frame (pd.DataFrame): The data source for the chart.
-        x1 (str): The name of the column in the data frame for the first set of bars (negative values).
-        x2 (str): The name of the column in the data frame for the second set of bars (positive values).
-        y (str): The name of the column in the data frame for the y-axis (categories).
+        data_frame (pd.DataFrame): The data frame for the chart. Can be long form or wide form.
+            See https://plotly.com/python/wide-form/.
```
petar-qb, line 24: Why do you emphasise long vs wide form for these four added examples?

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -65,80 +62,87 @@ def sankey(data_frame: pd.DataFrame, source: str, target: str, value: str, label
...
+    line = px.line(
+        data_frame,
+        x=x,
+        y=y_line,
+        markers=True,
+        color_discrete_sequence=fig.layout.template.layout.colorway[len(bar.data) :],
     )
+    for trace in line.data:
+        fig.add_trace(trace, secondary_y=True)
+    fig.update_layout(yaxis2={"tickmode": "sync", "overlaying": "y", "title": line.layout.yaxis.title})
```
petar-qb, line 127: Examples are more readable (at least for me) when empty lines are added between code logical parts. Ignore this if you disagree. 
```suggestion
    line = px.line(
        data_frame,
        x=x,
        y=y_line,
        markers=True,
        color_discrete_sequence=fig.layout.template.layout.colorway[len(bar.data) :],
    )

    for trace in line.data:
        fig.add_trace(trace, secondary_y=True)

    fig.update_layout(yaxis2={"tickmode": "sync", "overlaying": "y", "title": line.layout.yaxis.title})
```

vizro-core/examples/visual-vocabulary/pages/time.py
```diff
@@ -70,8 +70,8 @@
         vm.Graph(
             figure=categorical_column(
                 gapminder.query("country == 'Nigeria' and year > 1970"),
-                y="gdpPercap",
                 x="year",
+                y="gdpPercap",
```
petar-qb, line 74: This reminds me of an initiative I had about a year ago to standardise the order of `x` and `y` arguments in all the examples in our repository. Every time I see `y` before `x`, it just hurts my eyes! 😅

petar-qb (approved):
I like this PR and the initiatives from the description. 🎨 
I left just a few small comments, but overall the PR looks 🥇!

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -65,80 +63,87 @@ def sankey(data_frame: pd.DataFrame, source: str, target: str, value: str, label
...
     (y-axis 2), sharing the same x-axis. The y-axes for the bar and line graphs are synchronized and overlaid.
 
     Args:
         data_frame (pd.DataFrame): The data source for the chart.
-        x (str): The column name to be used for the x-axis.
-        y_column (str): The column name to be used for the y-axis 1, representing the column chart.
-        y_line (str): The column name to be used for the y-axis 2, representing the line chart.
+        x (str): Either a name of a column in data_frame, or a pandas Series or array_like object.
+        y_column (str): Either a name of a column in data_frame, or a pandas Series or array_like object.
+        y_line (str): Either a name of a column in data_frame, or a pandas Series or array_like object.
```
maxschulz-COL, line 107:
```suggestion
        x: Either a name of a column in data_frame, or a pandas Series or array_like object.
        y_column: Either a name of a column in data_frame, or a pandas Series or array_like object.
        y_line: Either a name of a column in data_frame, or a pandas Series or array_like object.
```

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -65,80 +63,87 @@ def sankey(data_frame: pd.DataFrame, source: str, target: str, value: str, label
...
 
 
 @capture("graph")
-def column_and_line(data_frame: pd.DataFrame, x: str, y_column: str, y_line: str) -> go.Figure:
+def column_and_line(
+    data_frame: pd.DataFrame,
+    x: Union[str, pd.Series, list[str], list[pd.Series]],
+    y_column: Union[str, pd.Series, list[str], list[pd.Series]],
+    y_line: Union[str, pd.Series, list[str], list[pd.Series]],
+) -> go.Figure:
```
maxschulz-COL, line 97: This is where this approach hits a limit, and we cannot pass kwargs twice - an issue we also encountered in the charts of old, just mentioning this again, as we would start to have inconsistencies between the charts etc

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -65,80 +63,87 @@ def sankey(data_frame: pd.DataFrame, source: str, target: str, value: str, label
...
-        data_frame,
-        x=x,
-        y=y,
-    )
+def categorical_column(data_frame: pd.DataFrame, **kwargs) -> go.Figure:
+    """Creates a column chart where the x-axis values are converted to category type.
+
+    Args:
+        data_frame (pd.DataFrame): The data frame for the chart. Can be long form or wide form.
+            See https://plotly.com/python/wide-form/.
```
maxschulz-COL, line 139:
```suggestion
        data_frame: The DataFrame for the chart. Can be long form or wide form.
            See https://plotly.com/python/wide-form/.
```
Was it possible to add wide and long form on the butterfly chart? If yes we should make the docs sync

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -65,80 +63,87 @@ def sankey(data_frame: pd.DataFrame, source: str, target: str, value: str, label
...
+        **kwargs: Keyword arguments to pass into px.bar (e.g. x, y, labels).
```
maxschulz-COL, line 140: Would add a link here - will not mention again

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -13,40 +13,38 @@
 # it can get out of sync. But probably we don't want the docstrings in the short code snippet.
 # Ultimately these charts will probably move to vizro.charts anyway.
 @capture("graph")
-def butterfly(data_frame: pd.DataFrame, x1: str, x2: str, y: str) -> go.Figure:
+def butterfly(data_frame: pd.DataFrame, **kwargs) -> go.Figure:
     """Creates a custom butterfly chart using Plotly's go.Figure.
```
maxschulz-COL, line 17: Should this not mention `px` now?

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -158,36 +163,31 @@ def waterfall(data_frame: pd.DataFrame, x: str, y: str, measure: list[str]) -> g
...
     )
-    fig.update_layout(showlegend=False)
-    return fig
 
 
 @capture("graph")
-def radar(data_frame, **kwargs) -> go.Figure:
-    """Creates a radar chart using Plotly's `line_polar`.
+def radar(data_frame: pd.DataFrame, **kwargs) -> go.Figure:
+    """Creates a radar chart using Plotly's px.line_polar.
```
maxschulz-COL, line 180: Would maybe go over the one-liners of all charts. Reason: sometimes we say what the underlying framework is (Plotly, px, go) sometimes we just mentioned the chart type. Would unify this a bit

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -158,36 +163,31 @@ def waterfall(data_frame: pd.DataFrame, x: str, y: str, measure: list[str]) -> g
...
 
     A radar chart is a type of data visualization in which there are three or more
     variables represented on axes that originate from the same central point.
 
     Args:
-        data_frame (pd.DataFrame): The data source for the chart.
-        **kwargs: Keyword arguments that can be passed into Plotly's line_polar (i.e. r, theta, etc.)
+        data_frame (pd.DataFrame): The data frame for the chart.
```
maxschulz-COL, line 186:
```suggestion
        data_frame: The DataFrame for the chart.
```

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -196,107 +196,106 @@ def radar(data_frame, **kwargs) -> go.Figure:
...
 
     A dumbbell plot is a type of dot plot where the points, displaying different groups, are connected with a straight
     line. They are ideal for illustrating differences or gaps between two points.
 
     Args:
-        data_frame (pd.DataFrame): The data source for the chart.
-        x (str): Column name in `data_frame` for x-axis values.
-        y (str): Column name in `data_frame` for y-axis values.
-        color (str): Column name in `data_frame` used for coloring the markers.
+        data_frame (pd.DataFrame): The data frame for the chart. Can be long form or wide form.
```
maxschulz-COL, line 206:
```suggestion
        data_frame: The DataFrame for the chart. Can be long form or wide form.
```

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -196,107 +196,106 @@ def radar(data_frame, **kwargs) -> go.Figure:
...
 
     Args:
-       data_frame (pd.DataFrame): The data frame for the chart.
-       y (str): The name of the categorical column in the data frame to be used for the y-axis (categories)
-       category_pos (list[str]): List of column names in the data frame representing positive values. Columns should be
-            ordered from least to most positive.
-       category_neg (list[str]): List of column names in the DataFrame representing negative values. Columns should be
-            ordered from least to most negative.
-       color_discrete_map: Optional[dict[str, str]]: A dictionary mapping category names to color strings.
+        data_frame (pd.DataFrame): The data frame for the chart. Can be long form or wide form.
```
maxschulz-COL, line 254:
```suggestion
        data_frame: The DataFrame for the chart. Can be long form or wide form.
```

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -196,107 +196,106 @@ def radar(data_frame, **kwargs) -> go.Figure:
...
-                marker_color=color_discrete_map.get(column, None) if color_discrete_map else None,
-            )
-        )
-
-    # Update layout and add central baseline
-    fig.update_layout(barmode="relative")
-    fig.add_vline(x=0, line_width=2, line_color="grey")
+    fig = px.bar(data_frame, **kwargs)
+
+    # Fix legend position according to the order of traces. This ensures that "Strongly disagree" comes before
```
maxschulz-COL, line 263: I trust that this works ;)

maxschulz-COL (approved):
🚀  I like it a lot! This seems like a great step towards `vizro.charts` 🚀 

I have mostly nitty gritty tidying suggestions which i mentioned on source code, but not the examples. Hopefully they are easily implemented by just clicking accept. The most important one is to keep a consistent docstring format for types.

As for the comments on function description standardization, I think they are not important, but I thought to mention them. Feel free to ignore them of course!

vizro-core/examples/visual-vocabulary/custom_charts.py
antonymilne: * it's not something that would be immediately obvious from the examples given. Because we only have one example for each chart we only show the `data_frame` that's in either long or wide form, so you might think that your data has to be in that format
* it's like a super power of plotly express that this is handled so well so something that we should boast about 😀

vizro-core/examples/visual-vocabulary/custom_charts.py
antonymilne: Generally agree with you 💯  But here there's a bit of a tradeoff with how big we make the code snippet for the dashboard code markdown cell since if it's too long then it required too much scrolling for the dashboard user, which is ultimately more important than what's readable for us in an IDE. Let me see how this one renders and work out what's best overall!

vizro-core/examples/visual-vocabulary/pages/time.py
antonymilne: Haha yes, I generally agree 😀

vizro-core/examples/visual-vocabulary/custom_charts.py
antonymilne: Yes, let me change that!

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -13,40 +13,38 @@
...
     two sets of data.
 
     Args:
-        data_frame (pd.DataFrame): The data source for the chart.
-        x1 (str): The name of the column in the data frame for the first set of bars (negative values).
-        x2 (str): The name of the column in the data frame for the second set of bars (positive values).
-        y (str): The name of the column in the data frame for the y-axis (categories).
+        data_frame (pd.DataFrame): The data frame for the chart. Can be long form or wide form.
+            See https://plotly.com/python/wide-form/.
+        **kwargs: Keyword arguments to pass into px.bar (e.g. x, y, labels).
```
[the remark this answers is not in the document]
antonymilne: I'm responding to this one in a bigger overall comment just so it's all in one place for future reference.

vizro-core/examples/visual-vocabulary/custom_charts.py
antonymilne: Yes, I spent a long timed thinking about this plot and found it very awkward for exactly that reason! As I saw it the options were:
* declare one chart to be the "primary" one and pass `**kwargs` through to that. But here the column and line charts are really on equal standing so it would be an arbitrary choice which was the primary chart
* pass the same `**kwargs` through to both the column and line charts. This limits `kwargs` to only work with arguments that are common between both and always passes the same value to both. I'm not sure in practice if that is fine or makes it useless
* have arguments `column_kwargs` and `line_kwargs`. I _think_ this is actually ok (it's not so different from how other things are handled in plotly) but didn't think it was worth doing here since I'm not 100% sure

vizro-core/examples/visual-vocabulary/custom_charts.py
antonymilne: I think yes, but let me check!

antonymilne: Yes indeed it is.

vizro-core/examples/visual-vocabulary/custom_charts.py
antonymilne: ok, it still looks good with the extra space, so done 👍

antonymilne:
Thanks for the reviews @petar-qb and @maxschulz-COL!

Definitely good points on the docstrings @maxschulz-COL. I just spent a while going through all of them (not only the ones I changed) to tidy them up so all should be good there. I would have expected `pydocstyle` (through `ruff`) to have picked up that we shouldn't have type hints in the docstrings though - do you think it's working ok? Or maybe we need `pydoclint` or something else to do that?

---

## Notes on `vizro.charts` in response to comments and for future reference

This PR is still kind of MVP so I decided to keep the docstrings short and sweet for now, so they don't mention conditions like two groups or which arguments might not work. If and when we move this stuff to `vizro.charts` I would tighten up on those things, but at the moment the level of documentation here is comparable to the other custom charts which also have lots of implicit assumptions, so I'll leave it like this for now.

Definitely let's discuss again before adding anything to `vizro.charts` - I'm actually not in a huge hurry to do that. I agree the `**kwargs` approach has some problems, and maybe overall using `make_figure` would actually be better for some of these after all.

@maxschulz-COL raised a great point in the team learning session that maybe we actually _don't_ want the visual vocabulary snippets to show this weird looking generic code and would instead be better showing a specific code snippet that's easier for a copy and paste user to understand and tweak. But if we do end up moving this code (or similar) to `vizro.charts` then we'd just simplify all those examples to do `from vizro.charts import butterfly` etc. anyway. So now I'm in two minds whether I've actually improved the dashboard or whether most of these changes should be reverted and saved for possible future additions to `vizro.charts` 😅 I'll leave like this for now anyway because there's some follow up changes I'd like to get finished off in another PR.

If/when it comes to adding stuff to `vizro.charts` then it's not totally clear to me at the moment which of these would be good candidates. Should all plots in visual vocabulary be possible in a single function call (either `vizro.plotly.express` or `vizro.charts`)? e.g. this chart:
```
@capture("graph")
def radar(data_frame: pd.DataFrame, **kwargs) -> go.Figure:
    fig = px.line_polar(data_frame, **kwargs)
    fig.update_traces(fill="toself")
    return fig
```

In future a Vizro user should be able to do this without defining a custom chart function at all so we won't need a function for it. So the only thing a `vizro.charts` version would add is (a) our opinion to fill in the traces and (b) one-to-one corresponds with visual vocabulary dashboard. It's not immediately clear to me how custom a custom chart needs to be before it's worth adding to `vizro.charts`.

No need to resolve these questions now, but just putting them here to keep track of it in one place.

PR #825, opened 2024-10-25
PR test, do not merge

maartenbreddels:
Just to see what happens when someone who is not a member of vizro opens a PR. Do we get the PyCafe message?


maxschulz-COL:
> Just to see what happens when someone who is not a member of vizro opens a PR. Do we get the PyCafe message?

Ah you need to change something in `vizro-core` :) Otherwise there would be not so much point I think

maartenbreddels:
```
github.GithubException.GithubException: 403 {"message": "Resource not accessible by integration", "documentation_url": 
```
This is what I was expecting, and the `workflow_run` trick we describe in https://py.cafe/docs/use-cases/developers#create-a-status-link solves that.

maxschulz-COL:
Closed as ressolved

PR #845, opened 2024-10-30
[Demo] Open ViViVo links outside iframe

antonymilne:
## Description

@Joseph-Perkins noticed that if you open up [ViViVo on HF in an iframe](https://huggingface.co/spaces/vizro/demo-visual-vocabulary) then the links to github don't work. The reason for this is
https://stackoverflow.com/questions/7649361/github-gist-page-wont-load-in-iframe. The solution is to open these links with ~`target="_top"` to escape the iframe~ `target="_blank"`.

## Notice


petar-qb (approved):

antonymilne:
Surprisingly this fix doesn't actually work. So it's blocked pending https://discuss.huggingface.co/t/allow-navigation-outside-iframe/114755/2.

huong-li-nguyen (approved):

vizro-core/examples/visual-vocabulary/custom_components.py
```diff
@@ -50,7 +50,9 @@ class Markdown(vm.VizroBaseModel):
 
     def build(self):
         """Returns a markdown component with an optional classname."""
-        return dcc.Markdown(id=self.id, children=self.text, dangerously_allow_html=False, className=self.classname)
+        return dcc.Markdown(
+            id=self.id, children=self.text, dangerously_allow_html=False, className=self.classname, link_target="_top"
+        )
```
antonymilne, line 55:
```suggestion
        return dcc.Markdown(
            id=self.id, children=self.text, dangerously_allow_html=False, className=self.classname, link_target="_blank"
        )
```

antonymilne:
I've raised the issue on the HF discord also but realistically don't expect them to change anything here, so until we hear otherwise the solution is just to open all these links in a new tab with `target="_blank"`. It's not great for usability but unfortunately it's the only way to get the links to open from inside the iframe.

FYI @nadijagraca @huong-li-nguyen this is what I mentioned yesterday at standup.

antonymilne:
HF have said they won't change the settings here so we must continue to open them in a new tab.

PR #850, opened 2024-11-04
[Tidy] Prepare for dynamic filters, part 1 of 2

antonymilne:
## Description

Refactor filter.py to prepare for dynamic filters. This solves several long-standing TODOs in this file.
* data loading has been isolated to a single place rather than repeated. This improves performance during `pre_build` (useful but not critical) and when updating filters during dashboard runtime (not yet implemented, but performance here _is_ critical)
* `_set` methods all rewritten and simplified (no more loops needed)
* isolated methods that will be required during runtime - these are now called in `__call__` (not yet used anywhere but will be soon)
* validation is slightly improved to handle some cases we didn't handle before - see the new tests to understand what these are
* the newly introduced TODOs will be short-term ones and will go once we've finished implementing dynamic filters

@petar-qb I recommend reviewing as follows:
* look at commits [9449c73](https://github.com/mckinsey/vizro/pull/850/commits/9449c73aff8706de14eb1c9d51b6aa2194bf61d8) and [b37f38f](https://github.com/mckinsey/vizro/pull/850/commits/b37f38f2c2994ca5823b7241bb96759db684dde6)
* read through the whole filter.py file as it now stands
* do some manual testing

Still TODO:
- [x] write tests

## Notice


vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -95,96 +96,40 @@ def check_target_present(cls, target):
...
-
-        if is_numeric_dtype(data_frame[self.column]):
-            self._column_type = "numerical"
-        elif is_datetime64_any_dtype(data_frame[self.column]):
-            self._column_type = "temporal"
+        if self.targets:
+            targeted_data = self._validate_targeted_data(targets=self.targets)
         else:
-            self._column_type = "categorical"
+            # If targets aren't explicitly provided then try to target all figures on the page. In this case we don't
```
antonymilne, line 102: The diff here is quite hard to follow, but basically the key flow is:
* in `pre_build` we now find `targeted_data` once and then everything follows from that
* `targeted_data` is a dataframe which has column names = target and data corresponding to just column's data in the targeted dataframe. No need to store the whole loaded dataframe since we're just interested in the one column from each targeted dataframe
* finding `min/max/options` can be found directly from the dataframe without any reloading of data needed (and also without any looping)

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -199,3 +144,98 @@ def _set_actions(self):
                     function=_filter(filter_column=self.column, targets=self.targets, filter_function=filter_function),
                 )
             ]
+
+    def __call__(self, **kwargs):
```
antonymilne, line 148: This isn't used anywhere yet but will be soon! As with `Figure`, we use `build` for build time and then `__call__`  during dashboard runtime.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -199,3 +144,98 @@ def _set_actions(self):
...
+        #  1. target. In worst case scenario this is needed but can lead to unnecessary repeated data loading.
+        #  2. data_source_name. No repeated data loading but won't work when applying data_frame parameters at runtime.
+        #  3. target + data_frame parameters keyword-argument pairs. This is the correct key to use at runtime.
+        # For now we follow scheme 2 for data loading (due to set() below) and 1 for the returned targeted_data
+        # pd.DataFrame, i.e. a separate column for each target even if some data is repeated.
+        # TODO: when this works with data_frame parameters load() will need to take arguments and the structures here
+        #  might change a bit.
+        target_to_data_source_name = {target: model_manager[target]["data_frame"] for target in targets}
+        data_source_name_to_data = {
+            data_source_name: data_manager[data_source_name].load()
```
antonymilne, line 188: This is the one place in this file where data is loaded.

vizro-core/tests/unit/vizro/models/_action/test_action.py
antonymilne, line 1: All changes to existing tests are just small rewordings and similar, nothing of interest. It's the new tests that are the interesting ones here.

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -150,7 +150,7 @@ def managers_one_page_without_graphs_one_button():
     vm.Page(
         id="test_page",
         title="Test page",
-        components=[vm.Graph(figure=px.scatter(data_frame=pd.DataFrame(columns=["A"]), x="A", y="A"))],
+        components=[vm.Graph(figure=px.scatter(data_frame=pd.DataFrame(data={"A": [1], "B": [2]}), x="A", y="B"))],
```
antonymilne, line 153: This has changed because an empty dataframe with a filter now eagerly raises an error message, which it wouldn't have done before (I think something else probably would have broken when viewing the dashboard before, just we didn't raise an error message in advance).

maxschulz-COL:
Do you guys want any further reviews?

antonymilne:
> Do you guys want any further reviews?

Let's say @petar-qb is the lead reviewer but a quick second pair of eyes would be good thank you!

petar-qb (approved):
This is the ground breaking refactoring that will and optimise and facilitate the implementation of the dynamic filters!! 

The `targeted_data` format is just the master piece 🎨 

I left a few minor comments, but other than that, all good! (tests are also okay, so no comments for that part)

vizro-core/hatch.toml
```diff
@@ -111,7 +111,9 @@ VIZRO_LOG_LEVEL = "DEBUG"
 extra-dependencies = [
   "pydantic==1.10.16",
   "dash==2.17.1",
-  "plotly==5.12.0"
+  "plotly==5.12.0",
+  "pandas==2.0.0",
+  "numpy==1.23.0"  # Need numpy<2 to work with pandas==2.0.0. See https://stackoverflow.com/questions/78634235/.
```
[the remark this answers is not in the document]
antonymilne: It was a somewhat arbitrary choice but I deliberately went for something that's relatively old rather than tight against the `numpy<2` bound. 1.26.4 is from February 2024 while 1.23.0 is from June 2022. With these lower bounds tests we should be testing against something that's old rather than recent.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -199,3 +144,102 @@ def _set_actions(self):
...
+            return "numerical"
+        elif is_temporal.all():
+            return "temporal"
+        elif is_categorical.all():
+            return "categorical"
+        else:
+            raise ValueError(
+                f"Inconsistent types detected in column {self.column}. This column must have the same type for all "
+                "targets."
+            )
```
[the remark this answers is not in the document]
antonymilne: Hmm yes, this is true but I think we should keep it strict. In these cases you should make sure that your datetimes are set as categorical (like strings rather than datetime type).

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -95,96 +96,40 @@ def check_target_present(cls, target):
...
-            options = set()
-            for target_id in self.targets:
-                data_source_name = model_manager[target_id]["data_frame"]
-                data_frame = data_manager[data_source_name].load()
-                options |= set(data_frame[self.column])
-
-            self.selector.options = sorted(options)
+            _min, _max = self._get_min_max(targeted_data)
+            self.selector.min = self.selector.min or _min
+            self.selector.max = self.selector.max or _max
```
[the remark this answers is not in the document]
antonymilne: Very good point, you definitely caught some bugs here! 🙏 I will change this and adds tests for these cases.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -199,3 +144,102 @@ def _set_actions(self):
...
+                "change type while the dashboard is running."
+            )
+
+        # TODO: when implement dynamic, will need to do something with this e.g. pass to selector.__call__.
+        # if isinstance(self.selector, SELECTORS["numerical"] + SELECTORS["temporal"]):
+        #     options = self._get_options(targeted_data)
+        # else:
+        #     # Categorical selector.
+        #     _min, _max = self._get_min_max(targeted_data)
+
```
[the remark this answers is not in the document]
antonymilne: Yes indeed. I will try and do this with my refactoring of `action_utils.py` because I think I have a good idea of how to do it now having rewritten filter.py.

vizro-core/src/vizro/models/_controls/filter.py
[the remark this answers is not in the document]
antonymilne: Done here: [983940b](https://github.com/mckinsey/vizro/pull/850/commits/983940b636fc53c60d093dde982b242f4438e3cf).

I don't think we want to do the same for `self.selector.options` because the default value there is `[]` rather than `None`, and having a selector with no options doesn't actually make sense (unlike a slider with `min=0`, which does make sense)?

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -108,31 +110,57 @@ def build(self):
...
+                # targets were manually specified so it's not ok the column isn't there. If targets were not specified
+                # then it's fine, we just skip this target, and error is not raised.
+                raise ValueError(f"Selected column {self.column} not found in dataframe for {target}.")
+
+        if not target_to_series:
+            raise ValueError(f"Selected column {self.column} not found in any dataframe on this page.")
+
+        self.targets = list(target_to_series)
+        # COMMEnt. will have repeats
+        self._targeted_data = pd.DataFrame.from_dict(target_to_series)
```
maxschulz-COL: I don't understand this line, you create a DF that itself has DFs inside it?

maxschulz-COL: I assume this somewhat goes like: check all columns, append them all to a big series, then do checks for dtype and alter min max etc?

vizro-core/src/vizro/models/_controls/filter.py
maxschulz-COL: I think this somewhat answer my question I had on `set_targets` function

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -97,107 +96,140 @@ def check_target_present(cls, target):
...
             data_frame = data_source_name_to_data[data_source_name]
 
             if self.column in data_frame.columns:
-                target_to_series[target] = data_frame
-            elif self.targets:
-                # targets were manually specified so it's not ok the column isn't there. If targets were not specified
-                # then it's fine, we just skip this target, and error is not raised.
+                # reset_index so that when we make a DataFrame out of all these pd.Series pandas doesn't try to align
+                # the columns by index.
+                target_to_series[target] = data_frame[self.column].reset_index(drop=True)
```
maxschulz-COL, line 202: Ah now I understand :) Ignore previous comment

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
maxschulz-COL, line 1: General comment: do we not also need to test the filters during runtime when the dynamic data changes? I saw some error messages above that indicate that

maxschulz-COL (approved):
As promised, glanced over it and it looks good to me

vizro-core/src/vizro/models/_controls/filter.py
maxschulz-COL: I saw later that this was clearer in a later commit

Issue #834, opened 2024-10-28
404 page appears unstyled


antonymilne:
Our 404 page has lost its styling and doesn't look good...
![image](https://github.com/user-attachments/assets/573ca603-9ccb-47d6-a87f-aa2de1b399f1)

Almost certainly caused by https://github.com/mckinsey/vizro/pull/804. Note that the theme selector callback doesn't run on this page. Just adding `data-bs-theme="dark"` in devtools fixes it so I'm sure this is easy to fix, but I'm curious why it doesn't work already given that we set `:root` to be the same as that?

PR #865, opened 2024-11-12
[Bug] Fix theme flickering and 404 page

huong-li-nguyen:
## Description
- Fixes https://github.com/mckinsey/vizro/issues/866 and https://github.com/mckinsey/vizro/issues/834

> Antony:
> Almost certainly caused by https://github.com/mckinsey/vizro/pull/804. Note that the theme selector callback doesn't > run on this page. Just adding data-bs-theme="dark" in devtools fixes it so I'm sure this is easy to fix, but I'm curious > why it doesn't work already given that we set :root to be the same as that?

The reason why this didn't work, despite setting the root, is due to several factors:

**1. No `data-bs-theme` attribute**: The background color of the 404 page was set using `background: var(--surfaces-bg-03);`. This CSS variable is only available when a `data-bs-theme` is assigned. Since this wasn't the case, the variable couldn't be found and it defaulted to `bs-body-bg`. This issue has been addressed in this PR ✅.

**2. Incomplete Bootstrap Styling:** While we styled selected components in our Bootstrap theme, we didn't add styling for the body, etc. Therefore, even if the root is set to match `data-bs-theme`, it has no effect because `bs-body-bg` defaults to the original Bootstrap color. This has been fixed in [this PR](https://github.com/McK-Private/vizro-bootstrap/pull/35) ✅ .

**3. Theme Order in Bootstrap:** To rely on root and its variables, we needed to change the order of the Bootstrap themes in `vizro-bootstrap.min.css`. Currently, our Vizro Bootstrap theme is listed first, followed by the general Bootstrap theme. As a result, the valid `bs-body-bg` value is taken from the general Bootstrap theme instead of ours. This issue has also been fixed in [this PR](https://github.com/McK-Private/vizro-bootstrap/pull/35) ✅ , but I would like @pruthvip15 to double-check if the changed order is correct.

An alternative and potentially better visual solution would be to assign the `data-bs-theme` attribute directly to the 404 page and possibly the loading page as well. This would prevent inconsistencies where the 404 and loading pages might have a different overall theme than the selected one (e.g., if the default is light). Is that possible?

## Notice


vizro-core/src/vizro/static/css/vizro-bootstrap.min.css
```diff
@@ -1,5 +1,5 @@
...al-400-inverted: #44a8a6;--sequentialTeal-500: #2e9190;--sequentialTeal-500-inverted: #2e9190;--sequentialTeal-600: #1b7b7a;--sequentialTeal-600-inverted: #1b7b7a;--sequentialTeal-700: #0c6565;--sequentialTeal-700-inverted: #0c6565;--sequentialTeal-800: #025050;--sequentialTeal-800-inverted: #025050;--sequentialTeal-900: #003b3c;--sequentialTeal-900-inverted: #003b3c;--sequentialRed-100: #f8d6da;--sequentialRed-100-inverted: #f8d6da;--sequentialRed-200: #fcb6ba;--sequentialRed-200-inverted: #fcb6ba;--sequentialRed-300: #f8989b;--sequentialRed-300-inverted: #f8989b;--sequentialRed-400: #ed7b7f;--sequentialRed-400-inverted: #ed7b7f;--sequentialRed-500: #dd6065;--sequentialRed-500-inverted: #dd6065;--sequentialRed-600: #c9474c;--sequentialRed-600-inverted: #c9474c;--sequentialRed-700: #b22f36;--sequentialRed-700-inverted: #b22f36;--sequentialRed-800: #981822;--sequentialRed-800-inverted: #981822;--sequentialRed-900: #7d000f;--sequentialRed-900-inverted: #7d000f;--surfaces-bg-card: #F5F6F6;--surfaces-bg-card-inverted: #232632;--bs-primary-rgb: #141721e0;--bs-secondary-color: #14172199;--bs-nav-link-color: #14172199;--bs-form-valid-color: #26bf56ff;--bs-form-invalid-color: #f03b3aff}/*!
```
huong-li-nguyen, line 1: The changed order inside [this PR](https://github.com/McK-Private/vizro-bootstrap/pull/35) needs to be reviewed by @pruthvip15 first, afterwards this PR can be merged.

petar-qb:
I like the changes and it would be even better if the theme of the 404 page matched the one currently selected (eg light theme).

I have one general question: Do we still need the loading page?

As I can remember, the loading page was added to avoid a blank white screen when opening a page that visualises big data.
And, after [this PR](https://github.com/mckinsey/vizro/pull/644) is merged, there should not be any blank white screen problems anymore. I can't remember all those stuff clearly, hence I don't have strong opinion on that.

antonymilne:
Thanks for fixing this so quickly! 🙏 

> An alternative and potentially better visual solution would be to assign the data-bs-theme attribute directly to the 404 page and possibly the loading page as well. This would prevent inconsistencies where the 404 and loading pages might have a different overall theme than the selected one (e.g., if the default is light). Is that possible?

Yeah, this is kind of what I was expecting the fix to be. It's possible a few ways but remains to be seen how effective they are...

Now I look at the code again probably the best solution, if it's possible, would be to do this in `Vizro.build`:
```
self.dash.index_str = self.dash.index_str.replace("<html>", f"<html data-bs-theme={dashboard.theme}")
```

Does that work?

huong-li-nguyen:
> Thanks for fixing this so quickly! 🙏
> 
> > An alternative and potentially better visual solution would be to assign the data-bs-theme attribute directly to the 404 page and possibly the loading page as well. This would prevent inconsistencies where the 404 and loading pages might have a different overall theme than the selected one (e.g., if the default is light). Is that possible?
> 
> Yeah, this is kind of what I was expecting the fix to be. It's possible a few ways but remains to be seen how effective they are...
> 
> Now I look at the code again probably the best solution, if it's possible, would be to do this in `Vizro.build`:
> 
> ```
> self.dash.index_str = self.dash.index_str.replace("<html>", f"<html data-bs-theme={dashboard.theme}")
> ```

This works to fix the original issues, but I think we would have to add the bootstrap changes anyway :) However, it doesn't fix the visual inconsistency, that the 404 page and loading page might have different themes if the light theme is the default. This can be fixed if we also add the toggle-switch to the 404 page and the loading page. It's possible for the 404 page, but I don't see a way for the loading-page.

maxschulz-COL:
> > this PR
> 
> The first point depends on whether we can set the data-bs-theme attribute for both the loading-page and the 404 page. I'll try out what @antonymilne wrote below in a sec! :)
> 
> On the loading-page, from what I remember @maxschulz-COL added this is, because there was a use case with an internal team where the general dashboard took a bit longer to load and the default white page with the loading icon was too ugly 😄 I think this is still the case, hence would keep the page. But correct me if I am wrong @maxschulz-COL

@petar-qb @huong-li-nguyen 
The internal team is currently not working on this anymore, so the specific case may have vanished. But if I remember correctly, this fixed two things:
- avoid blank screen on slow loading pages (should ideally not be the case anymore after taking out the data from the page build)
- avoid blank screen before dash renderer has even kicked in - this is not really related to the app itself, and since our default theme is dark we made that dark as well, so that there would be not change from white to dark

For the second reason I am not sure if anything changed in this PR so that we would not need it anymore. So if the page is dark before dash renderer kicks even without the loading page we can get rid of it, otherwise I would keep it!

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -302,14 +302,23 @@ def _make_page_layout(self, page: Page, **kwargs):
...
-    @staticmethod
-    def _make_page_404_layout():
+    def _make_page_404_layout(self):
         # The svg file is available through the _dash-component-suites/vizro route, as used in Dash's
         # _relative_url_path, but that feels too private to access directly. Hence read the file in directly rather
         # than referring to its path.
         error_404_svg = base64.b64encode((VIZRO_ASSETS_PATH / "images/error_404.svg").read_bytes()).decode("utf-8")
         return html.Div(
             [
+                html.Div(
```
huong-li-nguyen, line 312: I added the theme switch to the 404 page. It's kind of hidden and looks as part of the image - I kind of like it :D 

This at least gets rid of the theme flickering when switching from a light theme to the 404 page. The inconsistency in themes is still present for the loading page, but I guess that's what we always accepted.

vizro-core/src/vizro/static/css/vizro-bootstrap.min.css
huong-li-nguyen, line 1: In theory, the original issues are fixed by `self.dash.index_string = self.dash.index_string.replace("<html>", "<html data-bs-theme='dark'")` as @antonymilne  suggested. However, I would still add the current bootstrap changes in, as the default bg colour does not match exactly. The bootstrap theme changes ensure the right bg coloring and fixes some specificity issues as well.

vizro-core/src/vizro/_vizro.py
```diff
@@ -108,9 +108,10 @@ def build(self, dashboard: Dashboard):
 
         # Note that model instantiation and pre_build are independent of Dash.
         self._pre_build()
-
         self.dash.layout = dashboard.build()
 
+        # Add data-bs-theme attribute for 404 error page
```
antonymilne, line 113:
```suggestion
        # Add data-bs-theme attribute that is always present, even for pages without theme selector, i.e. the Dash "Loading..." screen.
```

Please can we put this new bit of code directly after these lines instead? I think it fits better in there.
```
if dashboard.title:
            self.dash.title = dashboard.title
```

vizro-core/src/vizro/models/_dashboard.py
antonymilne: A comment here would be good about why we have a theme selector the 404 page.

vizro-core/src/vizro/_vizro.py
```diff
@@ -108,9 +108,10 @@ def build(self, dashboard: Dashboard):
...
+        self.dash.index_string = self.dash.index_string.replace("<html>", "<html data-bs-theme='dark'")
```
antonymilne, line 114:
```suggestion
        bootstrap_theme = self.dashboard.theme.removeprefix("vizro_")
        self.dash.index_string = self.dash.index_string.replace("<html>", f"<html data-bs-theme='{bootstrap_theme}'")
```

The reason this solution is good is because it hopefully works for both dark and light themes 🙂

vizro-core/src/vizro/static/css/loading.css
```diff
@@ -1,22 +1,18 @@
...
+The dash-loading Div is present when Dash is initially loading, before the layout is built
+The dash-loading-callback Div is present when Dash has loaded, but the layout is still building
+
+Note that the dash-loading-callback Div is present until all elements are loaded, but as soon as the initial page
+elements (before on-page-load) are rendered, it gets pushed outside the viewable area, hence the spinner is not visible
+which is good, as we have individual loading spinners for elements. At the moment, we are not using this class.
+
+This creates a dark background in situations where neither dash-loading nor the Vizro app are displayed */
 html {
   background: rgba(20, 23, 33, 1);
```
antonymilne, line 12: Is this still needed? Now the `html` should always have `data-bs-theme` which should set the background colour to be the same as the `dashboard.theme` so we should be able to remove this or at least use `background: var(...)`?

vizro-core/src/vizro/static/css/loading.css
```diff
@@ -1,22 +1,18 @@
...
-/* Note that the dash-loading-callback Div is present until all elements are loaded, but as soon as the initial page
-elements (before on-page-load) are rendered, it gets pushed outside the viewable area, hence the spinner is not visible
-which is good, as we have individual loading spinners for elements. At the moment, we are not using this class.
-TODO: If we want to use this class, we need to evaluate if this is the best approach.
- */
-
-/* ._dash-loading-callback, */
 ._dash-loading {
   align-items: center;
   background: rgba(20, 23, 33, 1);
```
antonymilne, line 18: Similarly for this colour - maybe we can now remove it or use `var`.

vizro-core/src/vizro/static/css/loading.css
```diff
@@ -28,9 +24,6 @@ TODO: If we want to use this class, we need to evaluate if this is the best appr
   width: 100%;
 }
 
-/* Loading spinner */
-
-/* ._dash-loading-callback::after, */
 ._dash-loading::after {
   animation: spinner-border 0.75s linear infinite;
   border: 0.5rem solid lightgrey;
```
antonymilne, line 29: Very minor, but potentially the `lightgrey` could come from our colour scheme too.

vizro-core/src/vizro/static/css/loading.css
antonymilne: Or maybe we don't even need the spinner here at all now? See overall comment.

antonymilne (approved):
Approved subject to various changes 🙂 Thank you for fixing this!

In response to @petar-qb and @maxschulz-COL points about the loading spinner:
* the significance of the spinner is that it shows instead of the Dash "Loading..." screen. Before we customised this, _even for quickly loading apps_ this was always present before and with a fixed white background which caused a slight flicker
* we still need to customise this page from the Dash default somehow to avoid the white flash. A rotating spinner isn't really needed any more since the screen can no longer show for a long time, but having the right coloured background and not looking too gross is important
* the best we can easily do here is just set the background colour to match `dashboard.theme`. This won't take account of the user-selected theme but will reflect the dashboard developer's preferred theme. Getting it to match the user-selected theme is probably possible but a lot harder and not worth the effort.
* once we have a matching background colour, so long as the "Loading..." page doesn't show for long (hopefully it never does any more) then it doesn't really matter what it looks like. Either the loading spinner or just a blank page is fine by me

vizro-core/src/vizro/static/css/loading.css
petar-qb: I also think that we don't need the spinner anymore and that we can stick with the coloured background only.

petar-qb:
I agree with the Antony's [comment](https://github.com/mckinsey/vizro/pull/865#pullrequestreview-2432581320). I would also get rid of the loading spinner from the loading-page and I'd with stick with the coloured page only.

Currently, the loading page background colour is always dark. If we can align it with the user's selection (theme-selector) easily then it would be the perfect solution. If there's no way to do so easily, then setting the loading-page background colour to the `Dashboard.theme` would also be a good enough solution in my opinion. 

Thanks for handling 404 page, it is really a slick solution 👍

maxschulz-COL:
> I agree with the Antony's [comment](https://github.com/mckinsey/vizro/pull/865#pullrequestreview-2432581320). I would also get rid of the loading spinner from the loading-page and I'd with stick with the coloured page only.
> 
> Currently, the loading page background colour is always dark. If we can align it with the user's selection (theme-selector) easily then it would be the perfect solution. If there's no way to do so easily, then setting the loading-page background colour to the `Dashboard.theme` would also be a good enough solution in my opinion.
> 
> Thanks for handling 404 page, it is really a slick solution 👍

I would just note that I see the spinner quite often (albeit shortly), basically on any app that I open. I like it (but I am not married to any design). I feel having a spinner shows a level of user-reassurance that things are happening, even if it always is short to show. Having the background match `dashboard.theme` is of course a great addition.

But also not too married to any decision, just my 5 cents here :)

vizro-core/src/vizro/static/css/loading.css
huong-li-nguyen: I've removed it

vizro-core/src/vizro/static/css/loading.css
huong-li-nguyen: I removed it 👍

huong-li-nguyen:
> I would just note that I see the spinner quite often (albeit shortly), basically on any app that I open. I like it (but I am not married to any design). I feel having a spinner shows a level of user-reassurance that things are happening, even if it always is short to show. Having the background match `dashboard.theme` is of course a great addition.
> 
> But also not too married to any decision, just my 5 cents here :)

I tried refreshing the page and using some apps that take longer, and I decided to get rid of the spinner. It was just showing up for such a short time that it didn't seem worth it. I agree, it's always good to let users know something is happening, but in this case, the spinner was more distracting because it flashed on and off so quickly.

For almost all of the page refreshes, there was always the chart spinner visible, and afterwards came the loading spinner. Thus, it showed two different screens in such a short amount of time, so getting rid of one of the spinners is actually nicer! For the initial app load, I agree it would be nice, but given that you cannot differentiate between these two cases, I would just leave it 👍

vizro-core/src/vizro/static/css/vizro-bootstrap.min.css
huong-li-nguyen: Merged ✅

petar-qb (approved):
Great work!!

Do you mind opening a ticket to match the background colour of the loading-page with the currently selected theme colour, just to make sure we don't forget this very small improvement that would complete these UX enhancements. If you think it's irrelevant, I'm totally fine with that.

huong-li-nguyen:
I think it's not worth doing - the UX improvement compared to the added complexity in code to enable this is probably not worth it 😄 Taking on the dashboard.theme is a good enough improvement for me 👍

PR #868, opened 2024-11-12
[Demo] Add banner to visual vocabulary dashboard

nadijagraca:
## Description
Add banner to visual vocabulary dashboard

## Screenshot
Screenshot 2024-11-12 at 13 49 00

## Notice


huong-li-nguyen:
Please don't merge yet. I'm trying out different designs and will push later :)

vizro-core/examples/visual-vocabulary/app.py
```diff
@@ -111,17 +111,11 @@ def make_navlink(chart_group: ChartGroup) -> vm.NavLink:
...
-        [
-            html.Div(
-                [
-                    "Made using ",
-                    html.Img(src=get_asset_url("logo.svg"), id="banner", alt="Vizro logo"),
-                    dbc.NavLink("vizro", href="https://github.com/mckinsey/vizro", target="_blank", external_link=True),
-                ],
-                className="anchor-div",
-            ),
-        ],
```
huong-li-nguyen: @nadijagraca - there were lots of redundant html.divs, probably from some legacy code. It's always good to simplify where possible. Putting the dbc.NavLink around the entire container also ensures that the entire container is reactive, not only parts of the container :)

vizro-core/examples/visual-vocabulary/assets/css/custom.css
```diff
@@ -92,20 +92,26 @@ img[src*="#chart-icon"] {
 }
 
 .anchor-container {
-  background: #060a17;
```
huong-li-nguyen: @nadijagraca - there was previously no color set and the background was hard-coded. This normally prevents it from looking good on a dark and light theme. So if possible, always use the CSS variables at hand, such that it's suitable for both themes 👍

vizro-core/examples/visual-vocabulary/assets/css/custom.css
```diff
@@ -92,20 +92,26 @@ img[src*="#chart-icon"] {
...
-  height: 2rem;
-  padding: 4px;
-  place-content: baseline center;
+  font-size: 0.8rem;
+  font-weight: 500;
+  height: 24px;
+  padding: 0 12px;
   position: fixed;
-  width: 100%;
+  right: 0;
```
huong-li-nguyen, line 106: It looked a bit too prominent for me - I just placed it in the bottom right corner.

huong-li-nguyen (approved):

petar-qb (approved):
I really like how the banner looks even for "window" screens! 🎨

Have you pushed the latest version of the banner implementation to all of our hugging-face apps and do we have any other apps where we should do the same?

vizro-core/examples/visual-vocabulary/app.py
```diff
@@ -108,6 +110,15 @@ def make_navlink(chart_group: ChartGroup) -> vm.NavLink:
 )
 
 app = Vizro().build(dashboard)
+app.dash.layout.children.append(
+    dbc.NavLink(
+        ["Made with ", html.Img(src=get_asset_url("logo.svg"), id="banner", alt="Vizro logo"), "vizro"],
+        href="https://github.com/mckinsey/vizro",
+        target="_blank",
+        external_link=True,
```
antonymilne, line 118: Not important but FYI this isn't really necessary since it's set automatically when `href` is an absolute URL.
```suggestion
```

vizro-core/examples/visual-vocabulary/assets/css/custom.css
```diff
@@ -90,3 +90,28 @@ img[src*="#chart-icon"] {
 .intro-text ul:last-of-type {
   margin-bottom: 0;
 }
+
+.anchor-container {
+  align-items: center;
+  background: var(--text-primary);
+  border-top-left-radius: 8px;
```
antonymilne, line 97: 😱 A rounded border?!

antonymilne (approved):
Love it, this looks really really nice! ⭐ 💯

maxschulz-COL:
My 5 cents :) : given this looks nice and is very non-intrusive, is it time to rethink whether this should be added permanenty to all apps? (similar to streamlit)

vizro-core/examples/visual-vocabulary/assets/css/custom.css
antonymilne: And one other thing I noticed. Do we really want the banner to block out the whole horizontal space rather than just floating on top of the underlying content?
image

antonymilne:
> My 5 cents :) : given this looks nice and is very non-intrusive, is it time to rethink whether this should be added permanenty to all apps? (similar to streamlit)

IMO yes. My questions would be:
* should we have an option to turn it off (`Dashboard.watermark = False`)?
* should we save it for vizro 0.2.0 on the grounds that it's a somewhat disruptive/unexpected change? If we have an option to turn it off then I think probably not worth delaying.

vizro-core/examples/visual-vocabulary/assets/css/custom.css
huong-li-nguyen: Yes 😄 I didn't like the border as is and for the banner we can make an exception 👍 

It's not the banner that blocks the whole horizontal space. The banner floats just on top of the underlying content. That horizontal space has always been there and is part of our initial layout design. It's coming from the right-side container and its padding :)

maxschulz-COL:
> * should we have an option to turn it off (`Dashboard.watermark = False`)?
Not an easy question to answer given where we will move, hence answer to next question
> * should we save it for vizro 0.2.0 on the grounds that it's a somewhat disruptive/unexpected change? If we have an option to turn it off then I think probably not worth delaying.
Generally would have said no because FE design changes are no breaking changes, and I think we should adhere to that, but given that the above question is actually more far reaching we should think about it a bit more, thus waiting for 0.2.0 is probably not a bad idea 

@Joseph-Perkins FYI

huong-li-nguyen:
I discussed this with @Joseph-Perkins already and for now, we will only have it in our public demos. There are several design considerations we would need to discuss first, which we can discuss offline 👍

PR #872, opened 2024-11-13
[QA] Component library tests

l0uden:
## Description
Moved component library tests from vizro-qa

## Notice


.github/workflows/test-component-library-vizro-core.yml
```diff
@@ -0,0 +1,66 @@
+name: Integration tests Component Library
```
antonymilne, line 1: Now we have two sorts of integration tests in this repo the naming is going to get confusing. Let's discuss what to call your tests. Maybe e2e tests?

.github/workflows/test-component-library-vizro-core.yml
```diff
@@ -0,0 +1,66 @@
...
+    branches:
+      - main
+
+env:
+  PYTHONUNBUFFERED: 1
+  FORCE_COLOR: 1
+
+jobs:
+  test-component-library-vizro-core:
+    name: test-component-library-vizro-core
```
antonymilne, line 20: I like how this name matches our usual scheme 👍 You shouldn't need to specify `name` though given that there's no matrix entry for this.

Let's discuss overall naming but I think e2e is probably good.

```suggestion
  test-e2e-component-library-vizro-core:
```

.github/workflows/test-component-library-vizro-core.yml
```diff
@@ -0,0 +1,66 @@
...
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches:
+      - main
+
+env:
+  PYTHONUNBUFFERED: 1
+  FORCE_COLOR: 1
```
antonymilne, line 16:
```suggestion
  FORCE_COLOR: 1
  PYTHON_VERSION: "3.12"
```

Just to match how we do it elsewhere. This makes it easier to bump Python version everywhere.

.github/workflows/test-component-library-vizro-core.yml
```diff
@@ -0,0 +1,66 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python 3.12
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.12
```
antonymilne, line 30:
```suggestion
      - name: Set up Python ${{ env.PYTHON_VERSION }}
        uses: actions/setup-python@v5
        with:
          python-version: ${{ env.PYTHON_VERSION }}
```

vizro-core/hatch.toml
```diff
@@ -118,6 +119,14 @@ extra-dependencies = [
 features = ["kedro"]
 python = "3.9"
 
+[envs.tests]
```
antonymilne, line 122: Can we just do this in our normal `envs.default` instead?

.github/workflows/test-component-library-vizro-core.yml
```diff
@@ -0,0 +1,66 @@
...
+
+      - name: Install Hatch
+        run: pip install hatch
+
+      - name: Show dependency tree
+        run: hatch run tests:pip tree
+
+      - name: Run integration tests
+        run: hatch run tests:test-component-library
```
antonymilne, line 39:
```suggestion
      - name: Run e2e component library tests
        run: hatch run test-e2e-component-library
```

antonymilne: This assumes we get rid of the `tests` environment.

vizro-core/tests/__init__.py
antonymilne, line 1: Given that we use `--import-mode=importlib` for pytest, this file should almost definitely not be required.

vizro-core/tests/component_library/test_component_library.py
```diff
@@ -0,0 +1,93 @@
+# ruff: noqa: F403, F405
```
antonymilne, line 1: What are these ignores?

If justified then I'm happy to add more to `[tool.ruff.lint.per-file-ignores]` that works across all `tests` files, but so far we haven't ignored there's so would be good to understand why we need to now.

vizro-core/tests/component_library/test_component_library.py
```diff
@@ -0,0 +1,93 @@
...
+        reference_column="Reference",
+        title="KPI ref. with icon",
+        icon="shopping_cart",
+    ),
+]
+
+
+@pytest.mark.filterwarnings("ignore:HTTPResponse.getheader():DeprecationWarning")
+@pytest.mark.filterwarnings("ignore::pytest.PytestUnhandledThreadExceptionWarning")
+@pytest.mark.filterwarnings("ignore:unclosed file:ResourceWarning")
```
antonymilne, line 74: These look the same as our ignores in `test_examples`. What is it exactly that provokes them? Is it using `dash_duo`?

Fine to leave here for now, but if we're going to be ignoring them in lots of places then let's come up with a better way to do so. rather than copying and pasting the same three filters every time.

Are lots of your other tests also going to use `dash_duo` or just these?

vizro-core/tests/helpers/common.py
antonymilne, line 1: Let's put this in `test_utils/asserts.py`. Helper functions can have initial `_` and then there's the main `assert` function.

vizro-core/tests/helpers/common.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def compare_images(browserdriver, base_image, test_image_name):
+    """Comparison logic and diff files creation."""
+    browserdriver.save_screenshot(f"{test_image_name}_branch.png")
+    original = cv2.imread(f"tests/screenshots/{base_image}")
+    new = cv2.imread(f"{test_image_name}_branch.png")
+    try:
+        comparison_logic(original, new)
+        subprocess.call(f"rm -rf {test_image_name}_branch.png", shell=True)
```
antonymilne, line 40: `subprocess.call` should be replaced with `subprocess.run` since Python 3.5 or so.

But even better can we do this without calling `subprocess`? Generally we prefer to use pathlib wherever possible e.g. `Path(...).unlink()`.

vizro-core/tests/helpers/common.py
```diff
@@ -0,0 +1,48 @@
...
+    except (AssertionError, AttributeError) as exp:
+        subprocess.call(f"cp {test_image_name}_branch.png {base_image}", shell=True)
```
antonymilne, line 42: Let's use `shutil.copy`.

vizro-core/tests/helpers/common.py
```diff
@@ -0,0 +1,48 @@
...
+    (t_var, thresh) = cv2.threshold(dilated, 3, 255, cv2.THRESH_BINARY)
+    cnts = cv2.findContours(thresh, cv2.RETR_LIST, cv2.CHAIN_APPROX_SIMPLE)
+    cnts = imutils.grab_contours(cnts)
+    for contour in cnts:
+        (x, y, width, height) = cv2.boundingRect(contour)
+        cv2.rectangle(new, (x, y), (x + width, y + height), (0, 255, 0), 2)
+    return new
+
+
+def compare_images(browserdriver, base_image, test_image_name):
```
antonymilne, line 33:
```suggestion
def assert_image_equal(browserdriver, base_image, test_image_name):
```

Please can we try to rewrite this function a bit to raise `AssertionError` rather than pure `Exception`? And to surround only minimal bits of code in `try/except`. Currently I'm not exactly what lines of code could raise which exception. e.g. something like this:
```
try:
    compare_images(original_new)
except AssertionError as exc:
    raise AssertionError("Images are different.") from exc
finally:
    # cleanup files
```

I'm not sure when `AttributeError` could be raised?

vizro-core/tests/helpers/common.py
```diff
@@ -0,0 +1,48 @@
+import subprocess
+
+import cv2
+import imutils
+from hamcrest import assert_that, equal_to
+
+
+def comparison_logic(original_image, new_image):
```
antonymilne, line 8:
```suggestion
def _compare_images(original_image, new_image):
```

vizro-core/tests/helpers/common.py
```diff
@@ -0,0 +1,48 @@
...
+    """Comparison process."""
+    difference = cv2.subtract(original_image, new_image)
+    blue, green, red = cv2.split(difference)
+    assert_that(cv2.countNonZero(blue), equal_to(0), reason="Blue channel is different")
+    assert_that(cv2.countNonZero(green), equal_to(0), reason="Green channel is different")
+    assert_that(cv2.countNonZero(red), equal_to(0), reason="Red channel is different")
+
+
+def create_image_difference(original, new):
```
antonymilne, line 17:
```suggestion
def _create_image_difference(original, new):
```

vizro-core/tests/component_library/test_component_library.py
```diff
@@ -0,0 +1,93 @@
...
+                    dbc.Row([dbc.Col(kpi_card) for kpi_card in example_reference_cards]),
+                ],
+                gap=4,
+            ),
+        ]
+    )
+    dash_duo.start_server(app)
+    dash_duo.wait_for_page(timeout=20)
+    dash_duo.wait_for_element("div[class='card-kpi card']")
+    compare_images(dash_duo.driver, "base_kpi_comp_lib.png", "tests_kpi_comp_lib")
```
antonymilne, line 92: Not important for now, but maybe something to tweak later: why do we need to pass in file names manually here? Can't you just label them automatically based on the called `node`? I saw you do something similar in the `dash_br` PR.

vizro-core/tests/helpers/common.py
```diff
@@ -0,0 +1,48 @@
...
+    cnts = imutils.grab_contours(cnts)
+    for contour in cnts:
+        (x, y, width, height) = cv2.boundingRect(contour)
+        cv2.rectangle(new, (x, y), (x + width, y + height), (0, 255, 0), 2)
+    return new
+
+
+def compare_images(browserdriver, base_image, test_image_name):
+    """Comparison logic and diff files creation."""
+    browserdriver.save_screenshot(f"{test_image_name}_branch.png")
```
antonymilne, line 35: Does this get saved to a temporary directory? If it's all temporary then no need to cleanup with `rm` files afterwards.

vizro-core/tests/screenshots/base_kpi_comp_lib.png
antonymilne, line 1: Let's discuss directory structure but probably this will move to `tests/e2e/component_library/screenshots`.

antonymilne (commented):
Generally looks good 👍 I have lots of questions and comments just to make sure we get the initial structure and naming right. I'm not so familiar with testing images either so except lots of questions!

In general, as we move code into our main repo I'm going to apply a slightly higher standard of review compared to the vizro-qa repo so there might be a lot of comments! Very happy to discuss anything that's not clear though 🙂

vizro-core/hatch.toml
l0uden: I needed to add 3 additional libraries for the tests, didn't want to add it to default env

antonymilne: That's totally fine, we can definitely add more packages to the default environment. The dependencies there are already test dependencies and they install so quickly now thanks to `uv` that adding more is completely fine.

vizro-core/tests/component_library/test_component_library.py
l0uden: As I can see for now, only this error should be reused across multiple tests:
`@pytest.mark.filterwarnings("ignore:HTTPResponse.getheader():DeprecationWarning")`

I moved it to `pyproject.toml` both from this file and integration tests

vizro-core/tests/helpers/common.py
l0uden: I've rewrote the code to use more specific exceptions.
I can't use finally here, because every block of code under `try` and `except` is different and has its own logic.

l0uden:
@antonymilne , thanks for the very useful review comments! 
I've fixed them all without comments under your reviews, but left answers were it was most important.

vizro-core/tests/tests_utils/e2e_asserts.py
```diff
@@ -0,0 +1,50 @@
...
+def assert_image_equal(browserdriver, test_image_name):
+    """Comparison logic and diff files creation."""
+    base_image_name = f"{test_image_name.replace('test', 'base')}.png"
+    browserdriver.save_screenshot(f"{test_image_name}_branch.png")
+    original = cv2.imread(f"tests/e2e/screenshots/{base_image_name}")
+    new = cv2.imread(f"{test_image_name}_branch.png")
+    try:
+        _compare_images(original, new)
+        Path(f"{test_image_name}_branch.png").unlink()
+    except AssertionError as exp:
```
antonymilne, line 43:
```suggestion
    except AssertionError as exc:
```

Just to match how we usually do it in vizro. Sames applies on on the next `except` too.

vizro-core/tests/tests_utils/e2e_asserts.py
```diff
@@ -0,0 +1,50 @@
...
+        shutil.copy(f"{test_image_name}_branch.png", base_image_name)
+        diff = _create_image_difference(original=new, new=original)
+        cv2.imwrite(f"{test_image_name}_diff_main.png", diff)
+        raise AssertionError("pictures are not the same") from exp
+    except cv2.error as exp:
+        shutil.copy(f"{test_image_name}_branch.png", base_image_name)
+        raise cv2.error("pictures has different sizes") from exp
```
antonymilne, line 50:
```suggestion
        raise AssertionError("pictures has different sizes") from exp
```

Unless there's some need to keep as `cv2.error`?

vizro-core/tests/e2e/test_component_library.py
```diff
@@ -0,0 +1,88 @@
...
+                    dbc.Row([dbc.Col(kpi_card) for kpi_card in example_reference_cards]),
+                ],
+                gap=4,
+            ),
+        ]
+    )
+    dash_duo.start_server(app)
+    dash_duo.wait_for_page(timeout=20)
+    dash_duo.wait_for_element("div[class='card-kpi card']")
+    assert_image_equal(dash_duo.driver, get_test_name)
```
antonymilne, line 87: I know it was my idea to refactor it like this... But I didn't realise exactly what it would look like until you did it, and now it looks a bit weird 😅 `assert_image_equal` really sounds like it should be taking in images rather than `browserdriver` and `testname`.

I suggest breaking this down like this:
```
# remove get_test_name fixture and just use request directly given it's only used once
# this code might be repeated a lot across tests but that's ok, I think it makes sense
result_image_path, expected_image_path  = make_screenshot_paths(request.node.name)
dash_duo.save_screenshot(result_image_path)
assert_image_equal(result_image_path, expected_image_path)
```

```
# e2e_asserts.py
def make_screenshot_paths(test_name):
    # do check this, I might have done it wrong...
    result_image_path = f"{test_image_name}_branch.png"
    expected_image_path = f"tests/e2e/screenshots/{test_image_name.replace('test', 'base')}.png"
    return  result_image_path, expected_image_path

def assert_image_equal(result_image_path, expected_image_path):
    """Comparison logic and diff files creation."""
    result_image = cv2.imread(result_image_path)    
    expected_image = cv2.imread(expected_image_path)
    
    try:
        # basically the same as now with some renaming of variables
```

vizro-core/tests/tests_utils/e2e_asserts.py
```diff
@@ -0,0 +1,50 @@
...
+    base_image_name = f"{test_image_name.replace('test', 'base')}.png"
+    browserdriver.save_screenshot(f"{test_image_name}_branch.png")
+    original = cv2.imread(f"tests/e2e/screenshots/{base_image_name}")
+    new = cv2.imread(f"{test_image_name}_branch.png")
+    try:
+        _compare_images(original, new)
+        Path(f"{test_image_name}_branch.png").unlink()
+    except AssertionError as exp:
+        shutil.copy(f"{test_image_name}_branch.png", base_image_name)
+        diff = _create_image_difference(original=new, new=original)
```
antonymilne, line 45: I guess you meant this??
```suggestion
        diff = _create_image_difference(original=original, new=new)
```

FYI in my suggestions I've renamed things to `result` and `expected` and always done things in that order, just to match how we normally do things in vizro.

antonymilne (approved):
This looks really great, thank you!

I've added a few more minor suggestions. Most important thing is I don't think `assert_image_equal` quite makes sense yet and we should probably break it into two functions. See what you think.

Once you've made any changes based on this round of review let's get @maxschulz-COL to have a final look through and then it's good to merge 💯

vizro-core/tests/component_library/test_component_library.py
antonymilne: Cool, thanks for doing this - looks good.

vizro-core/tests/helpers/common.py
antonymilne: This looks much better now, thank you!

vizro-core/tests/helpers/common.py
antonymilne: Should we just save these files to a temporary directory in the first place? e.g. in my suggested rewrite it could become:
```
def make_screenshot_paths(test_name, tmpdir):
    # result_image_path is something like tempdir / test_name.png
```

Then we can still do the `shutil.copy` as you do now and just no need to have the `Path().unlink()` line?

l0uden: It doesn't matter where the screenshot will be stored. I need to copy screenshots to artefacts folder for failed tests after all tests will be finished. Because I don't want to copy all of the screenshots, but only the ones related to failed tests, I'm deleting the screenshots for passed one right after comparison is done in a positive way. 

The other way can be to copy screenshots for failed tests into separate folder directly after tests is failed, but it will be more lines of code for copying and creating of the new folder (but the line for deleting files after passed tests could be deleted in this case)

vizro-core/tests/tests_utils/e2e_asserts.py
l0uden: cv2.error is different from AssertionError.

cv2.error happens when smth wrong with initial files for comparison, like no original file existed or the files has different sizes for some reason and could not be compared

AssertionError happens if everything is ok with files, but they are really different

vizro-core/tests/e2e/test_component_library.py
l0uden: Did it a bit different, but should be ok

l0uden:
@antonymilne , finished with your latest suggestions

.github/workflows/test-e2e-component-library-vizro-core.yml
```diff
@@ -0,0 +1,67 @@
...
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - name: Install Hatch
+        run: pip install hatch
+
+      - name: Show dependency tree
+        run: hatch run all.py${{ env.PYTHON_VERSION }}:pip tree
+
+      - name: Run e2e component library tests
+        run: hatch run all.py${{ env.PYTHON_VERSION }}:test-e2e-component-library
```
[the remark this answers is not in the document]
antonymilne: Perfect, thanks!

vizro-core/tests/e2e/test_component_library.py
```diff
@@ -0,0 +1,89 @@
...
+                ],
+                gap=4,
+            ),
+        ]
+    )
+    dash_duo.start_server(app)
+    dash_duo.wait_for_page(timeout=20)
+    dash_duo.wait_for_element("div[class='card-kpi card']")
+    result_image_path, expected_image_path, expected_image_name = make_screenshot_and_paths(dash_duo.driver, request)
+    assert_image_equal(result_image_path, expected_image_path, expected_image_name)
```
antonymilne, line 88:
```suggestion
    result_image_path, expected_image_path = make_screenshot(dash_duo.driver, request.node.name)
    assert_image_equal(result_image_path, expected_image_path)
```

vizro-core/tests/tests_utils/e2e_asserts.py
```diff
@@ -0,0 +1,56 @@
...
+    (t_var, thresh) = cv2.threshold(dilated, 3, 255, cv2.THRESH_BINARY)
+    cnts = cv2.findContours(thresh, cv2.RETR_LIST, cv2.CHAIN_APPROX_SIMPLE)
+    cnts = imutils.grab_contours(cnts)
+    for contour in cnts:
+        (x, y, width, height) = cv2.boundingRect(contour)
+        cv2.rectangle(result_image, (x, y), (x + width, y + height), (0, 255, 0), 2)
+    return result_image
+
+
+def make_screenshot_and_paths(browserdriver, request):
```
antonymilne, line 34:
```suggestion
def make_screenshot_and_paths(browserdriver, request_node_name):
```
and then just use `request_node_name` directly here rather than passing in the whole `request` object.

vizro-core/tests/tests_utils/e2e_asserts.py
```diff
@@ -0,0 +1,56 @@
...
+    expected_image_name = f"{request.node.name.replace('test', 'main')}.png"
+    expected_image_path = f"tests/e2e/screenshots/{expected_image_name}"
+    browserdriver.save_screenshot(result_image_path)
+    return result_image_path, expected_image_path, expected_image_name
+
+
+def assert_image_equal(result_image_path, expected_image_path, expected_image_name):
+    """Comparison logic and diff files creation."""
+    expected_image = cv2.imread(expected_image_path)
+    result_image = cv2.imread(result_image_path)
```
antonymilne, line 45: The only thing I don't like here is passing the `expected_image_name` again. It feels kind of weird and cluttered when it can just easily be calculated again:
```suggestion
def assert_image_equal(result_image_path, expected_image_path):
    """Comparison logic and diff files creation."""
    expected_image = cv2.imread(expected_image_path)
    expected_image_name = Path(expected_image_path).name
    result_image = cv2.imread(result_image_path)
```

antonymilne (approved):
Looks good, just made a small suggestion but otherwise @maxschulz-COL please could you do a quick check through and then feel free to merge!

vizro-core/tests/tests_utils/e2e_asserts.py
l0uden: Fixed

maxschulz-COL (approved):
Looks good to me, thanks also for A to make such a thorough review before. I didn't write down any comment, but I think the `e2e_asserts.py` would benefit from a few inline code comments about what you are doing. Since not everyone will be super familiar with `cv2` etc, it is sometime hard to guess what even the overall strategy is. Maybe a few comments guide a reader on what the code is doing might be helpful?

vizro-core/tests/tests_utils/e2e_asserts.py
```diff
@@ -0,0 +1,56 @@
...
+    try:
+        _compare_images(expected_image, result_image)
+        Path(result_image_path).unlink()
+    except AssertionError as exc:
+        shutil.copy(result_image_path, expected_image_name)
+        diff = _create_image_difference(expected_image=expected_image, result_image=result_image)
+        cv2.imwrite(f"{result_image_path}_difference_from_main.png", diff)
+        raise AssertionError("pictures are not the same") from exc
+    except cv2.error as exc:
```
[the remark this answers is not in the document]
l0uden: it could be `cv2.error` in `try` in this line:
https://github.com/mckinsey/vizro/blob/bbe76f4c96f2041fc32f84c5acc85b2547a65bea/vizro-core/tests/tests_utils/e2e_asserts.py#L11

.github/workflows/test-e2e-component-library-vizro-core.yml
```diff
@@ -0,0 +1,65 @@
...
+        run: pip install hatch
+
+      - name: Show dependency tree
+        run: hatch run pip tree
+
+      - name: Run e2e component library tests
+        run: hatch run test-e2e-component-library
+
+      - name: Copy failed screenshots
+        if: failure()
```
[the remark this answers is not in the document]
l0uden: Great comment! Which helped me to think about separate file for this logic. Right now it is not something necessary, but it will be in the future tests moving.

l0uden:
> Looks good to me, thanks also for A to make such a thorough review before. I didn't write down any comment, but I think the `e2e_asserts.py` would benefit from a few inline code comments about what you are doing. Since not everyone will be super familiar with `cv2` etc, it is sometime hard to guess what even the overall strategy is. Maybe a few comments guide a reader on what the code is doing might be helpful?

Left comments about `cv2` logic

All your comment are now addressed, thanks!

vizro-core/tests/tests_utils/e2e_asserts.py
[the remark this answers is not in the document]
maxschulz-COL: but where is the `try` statement, sorry if I am being think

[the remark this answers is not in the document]
l0uden: here it is:

https://github.com/mckinsey/vizro/blob/bbe76f4c96f2041fc32f84c5acc85b2547a65bea/vizro-core/tests/tests_utils/e2e_asserts.py#L46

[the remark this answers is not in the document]
maxschulz-COL: I see, I had never used a double except myself actually, learned something, thanks :)

PR #787, opened 2024-10-04
[Demo] Add lollipop chart to visual-vocabulary

yonkmanjl:
## Description
Adds lollipop chart example to visual vocabulary.

Resolves #710 .

## Screenshot
image

## Notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

  - I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
  - I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
  - I certify that the use of this contribution as authorized by the Apache 2.0 license does not violate the intellectual property rights of anyone else.
  - I have not referenced individuals, products or companies in any commits, directly or indirectly.
  - I have not added data or restricted code in any commits, directly or indirectly.

PR #874, opened 2024-11-14
[Demo] Add lollipop chart to ViViVo

huong-li-nguyen:
## Description
Based on https://github.com/mckinsey/vizro/pull/787 - thank you again [yonkmanjl](https://github.com/yonkmanjl)! ⭐ 

- Adds refactored code for lollipop chart based on `px.scatter`

## Notice


vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,45 +1,59 @@
...
-            "Pies",
-        ],
-        "Profit Ratio": [-0.10, -0.15, -0.05, 0.10, 0.05, 0.20, 0.15, -0.08, 0.08, -0.12, 0.02, -0.07],
-    }
-)
+from vizro.models.types import capture
+
+
+@capture("graph")
+def lollipop(data_frame: pd.DataFrame, x: str, y: str):
```
huong-li-nguyen, line 12: Old code for reference

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -304,3 +304,53 @@ def diverging_stacked_bar(data_frame: pd.DataFrame, **kwargs) -> go.Figure:
...
+                mode="lines",
+            )
+        )
+
+    fig.update_traces(marker_size=12, line_width=3, line_color=fig.layout.template.layout.colorway[0])
+    fig.update_layout(showlegend=False)
+
+    # These are use-case specific layout updates. Normally not suitable for a chart inside vizro.charts, as these
+    # depend on the data context.
+    fig.update_layout(yaxis_title="")
```
huong-li-nguyen, line 355: This would e.g. not live inside a chart function inside `vizro.charts` but it's something I would like to add to the visual-vocabulary to follow data viz best practices

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -304,3 +304,53 @@ def diverging_stacked_bar(data_frame: pd.DataFrame, **kwargs) -> go.Figure:
...
+    # with both numeric columns. Should we introduce this ourselves? Does this have to be part of this example
+    # or can we hard-code for the visual-vocabulary?
+    for i in range(len(data_frame)):
+        fig.add_trace(
+            go.Scatter(
+                x=[0, x_array[i]],
+                y=[y_array[i], y_array[i]],
+                mode="lines",
+            )
+        )
```
huong-li-nguyen, line 348: This would have to be done more flexibly for any chart inside `vizro.charts`, but even then I am wondering how one could do it? We probably would have to introduce an `orientation` argument ourselves, as it's impossible to know beforehand what the user puts as x and y and which orientation they might want. 

However, for the visual-vocabulary example, hard-coding is fine for me. WDYT?

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -304,3 +304,53 @@ def diverging_stacked_bar(data_frame: pd.DataFrame, **kwargs) -> go.Figure:
...
+    Args:
+        data_frame: DataFrame for the chart. Can be long form or wide form.
+            See https://plotly.com/python/wide-form/.
+        **kwargs: Keyword arguments to pass into px.scatter (e.g. x, y, labels).
+            See https://plotly.com/python-api-reference/generated/plotly.scatter.html.
+
+    Returns:
+        go.Figure: Lollipop chart.
+    """
+    # Should we allow keyword arguments (kwargs) in this context, given the presence of multiple traces?
```
huong-li-nguyen, line 327: @antonymilne - I am still a bit torn which code we should display in the visual-vocabulary here and what the requirements should be. I still prefer any code that can be easily adapted + follows data viz best practices. 

This sometimes interferes with the requirements we might have for any chart that should live in `vizro.charts` as already discussed. I have a couple of open questions here, which we can discuss here or offline.

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -310,3 +310,45 @@ def diverging_stacked_bar(data_frame: pd.DataFrame, **kwargs) -> go.Figure:
...
+        **kwargs: Keyword arguments to pass into px.scatter (e.g. x, y, labels).
+            See https://plotly.com/python-api-reference/generated/plotly.scatter.html.
+
+    Returns:
+        go.Figure: Lollipop chart.
+    """
+    # Unlike the column_and_line chart, where all traces hold equal significance, here the traces differ in importance.
+    # The primary scatter plot is the main trace, while the additional traces merely serve as connecting lines.
+    # Therefore, should we apply the kwargs solely to the main scatter plot, as illustrated below?
+    fig = px.scatter(data_frame, **kwargs)
```
huong-li-nguyen, line 336: One question: In this case, there is one main scatter trace and several subsidiary scatter traces that only add lines. I assume in this case, we only add `kwargs` to the main trace and we don't enable the the `kwargs` for the subsidiary scatter traces?

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -310,3 +310,59 @@ def diverging_stacked_bar(data_frame: pd.DataFrame, **kwargs) -> go.Figure:
...
+        xaxis_showgrid = False
+
+    fig.update_traces(
+        marker_size=12,
+        line_width=3,
+        line_color=fig.layout.template.layout.colorway[0],
+    )
+
+    fig.update_layout(
+        showlegend=False, yaxis_showgrid=yaxis_showgrid, xaxis_showgrid=xaxis_showgrid, yaxis_rangemode="tozero"
```
huong-li-nguyen, line 366: I'm wondering if we should generally include layout customizations that optimize a given chart but might not be applicable depending on the data. For instance, setting `yaxis_rangemode="tozero"` works well for positive data as it removes the gap between the x-axis zeroline and the start of the y-axis, but it's not suitable for negative data.

Additionally, I've just realized that our chart code might not be dynamic enough to handle negative data. Specifically, the calculation code from lines `343-350` does not account for negative values. This might also be the case with other charts, which led me to the question: Does it have to? What's our minimum requirement? 

Personally, I see the visual-vocabulary as a guide for best practices in data visualization. I don't mind if some aspects are a bit hard-coded, as it's meant to be a starting code snippet that users can extend. WDYT?

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -310,3 +310,59 @@ def diverging_stacked_bar(data_frame: pd.DataFrame, **kwargs) -> go.Figure:
...
+                mode="lines",
+            )
+        )
+
+    if orientation == "h":
+        yaxis_showgrid = False
+        xaxis_showgrid = True
+    else:
+        yaxis_showgrid = True
+        xaxis_showgrid = False
```
petar-qb, line 357: I can't remember do we aim for the optimised code or for the readability in vivivo examples? However, there is a half-optimised suggestion.

For the further optimisation we can calculate `x_coords` and `y_coords` with the `x_or_y` technique that's used within many other vivivo examples. (I like "vivivo" abbreviation 😄)

(If you accept the suggestion, apply the same changes in the `examples/lollipop.py` too)

```suggestion
    # Enable for both orientations
    is_horizontal = fig.data[0].orientation == "h"

    x_coords = [[0, x] if is_horizontal else [x, x] for x in fig.data[0]["x"]]
    y_coords = [[y, y] if is_horizontal else [0, y] for y in fig.data[0]["y"]]

    for x, y in zip(x_coords, y_coords):
        fig.add_trace(go.Scatter(x=x, y=y, mode="lines"))

    xaxis_showgrid = is_horizontal
    yaxis_showgrid = not is_horizontal
```

vizro-core/examples/visual-vocabulary/README.md
```diff
@@ -70,7 +70,7 @@ The dashboard is still in development. Below is an overview of the chart types f
...
 | Line                  | ✅     | Time                     | [Line plot with px](https://plotly.com/python/line-charts/)                                                                     | [px.line](https://plotly.com/python-api-reference/generated/plotly.express.line)                                                                                                                |
-| Lollipop              | ❌     | Ranking, Magnitude       |                                                                                                                                 |                                                                                                                                                                                                 |
+| Lollipop              | ✅     | Ranking, Magnitude       | [Lollipop & Dumbbell Charts with Plotly](https://towardsdatascience.com/lollipop-dumbbell-charts-with-plotly-696039d5f85)       | [px.scatter](https://plotly.com/python-api-reference/generated/plotly.express.scatter)                                                                                                          |
```
petar-qb, line 73: Do you think we should reference Medium articles that are available only for the Medium subscribed users? I think users have to pay to read the entire article.

vizro-core/examples/visual-vocabulary/custom_charts.py
```diff
@@ -310,3 +310,59 @@ def diverging_stacked_bar(data_frame: pd.DataFrame, **kwargs) -> go.Figure:
...
+
+
+@capture("graph")
+def lollipop(data_frame: pd.DataFrame, **kwargs):
+    """Creates a lollipop based on px.scatter.
+
+    A lollipop chart is a variation of a bar chart where each data point is represented by a line and a dot at the end
+    to mark the value.
+
+    Inspired by: https://community.plotly.com/t/how-to-make-dumbbell-plots-in-plotly-python/47762
```
petar-qb, line 322: It's a bit weird to me that the `bar` and the `dumbbell` charts referenced, but the `scatter` is used as the "main" chart. Ignore this comment if you disagree 😄

vizro-core/examples/visual-vocabulary/custom_charts.py
petar-qb: >  I assume in this case, we only add kwargs to the main trace and we don't enable the the kwargs for the subsidiary scatter traces?

I think this is totally correct 👍.

vizro-core/examples/scratch_dev/app.py
petar-qb: The new code is much better 🥇

vizro-core/examples/visual-vocabulary/custom_charts.py
huong-li-nguyen: Ah this is great, thanks! ⭐

vizro-core/examples/visual-vocabulary/README.md
huong-li-nguyen: Good question! 

In this case, yes - I think it's fine. This section is primarily to give credit to where the original code came from. For users, our optimized and refactored code snippet is the main one to use. So as long as they can access that, it's fine 👍 

If there are public articles out, that provide the same content, of course we can also prefer these, but in the lollipop example the public available ones did not use the code I used in the end. They use shapes, which I thought was a bit more difficult to understand. So the inspiration for this lollipop code is really coming from the medium article.

vizro-core/examples/visual-vocabulary/custom_charts.py
huong-li-nguyen: The dumbbell reference is wrong! Let me quickly update that 😄 But other than that, I think it's not weird, because you can draw lines and dots with the scatter chart :) And the lollipop is essentially a bar chart alternative.

vizro-core/examples/visual-vocabulary/custom_charts.py
antonymilne: In this particular case I actually think negative values work fine already.

The general point is definitely a good one and I largely agree with what you say here. My ultimate ideal vision would be something like:
* `vizro.charts`:
    * mimics plotly express as much as possible with maybe some small tweaks to formatting. But e.g. doesn't fully declutter unless it's really essential for that sort of chart (like sparkline)
    * as general as possible in handling wide/long form data etc., orientation and common plotly express arguments but it's a best effort attempt so we get maximum value for minimum effort - no need to make it handle absolutely everything
* visual vocabulary:
    * more opinionated, shows best practices
    * can include further modification to vizro.charts e.g. to remove a legend or add `%` label to axes

The question for me is how much effort we should put into achieve this vision and what it's priority should be. e.g. should we just do `vizro.charts` now? It's effort in terms of documentation, tests, etc. for little proven value so not sure it's high priority tbh. But maybe we could do an MVP version that doesn't take too long and is very little effort. Personally I'd still put that below several other things on our roadmap though, but let's see what others think. We should talk to @maxschulz-COL about this in particular.

vizro-core/examples/visual-vocabulary/pages/examples/lollipop.py
```diff
@@ -0,0 +1,40 @@
...
+    )
+
+    fig.update_layout(
+        showlegend=False, yaxis_showgrid=yaxis_showgrid, xaxis_showgrid=xaxis_showgrid, yaxis_rangemode="tozero"
+    )
+    return fig
+
+
+fig = lollipop(
+    data_frame=gapminder.query("year == 2007 and gdpPercap > 36000").sort_values("gdpPercap"),
```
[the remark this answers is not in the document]
huong-li-nguyen: Ah, I didn't catch that this was the new order now. Will keep that in mind 👍

vizro-core/examples/visual-vocabulary/pages/examples/lollipop.py
```diff
@@ -0,0 +1,40 @@
+import pandas as pd
+import plotly.express as px
+import plotly.graph_objects as go
+from vizro.models.types import capture
+
+gapminder = px.data.gapminder()
+
+
+@capture("graph")
+def lollipop(data_frame: pd.DataFrame, **kwargs):
```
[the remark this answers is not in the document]
huong-li-nguyen: I'll take it 😄 I think we are already at a point where parts are harder to understand, but I think it's fine

vizro-core/examples/visual-vocabulary/custom_charts.py
huong-li-nguyen: Let's discuss it on Wednesday, but I briefly discussed it with J. already, and I think there is no need to work on this anytime soon. I am currently also not in strong favor of actually implementing it, but that's a separate question.

petar-qb (approved):
This looks good to go! 🎨

PR #793, opened 2024-10-05
[Demo] Add sparkline chart to visual-vocabulary

ataraexia:
## Description
@huong-li-nguyen - This is my first attempt at making a multi-sparkline chart (issue #783), but fig.update_* functions don't seem to working as intended, so I am unable to disable axes and labels.

Any idea why this would work in normal Plotly (at least according to the linked thread), but not in Vizro?

## Screenshot
Screenshot 2024-10-04 at 6 29 34 PM

PR #875, opened 2024-11-14
[Demo] Add sparkline to ViViVo

huong-li-nguyen:
## Description
Based on https://github.com/mckinsey/vizro/pull/793 - thank you again [ataraexia](https://github.com/ataraexia)! ⭐ 

- Adds code for sparkline chart based on `px.line`

## Screenshot
Screenshot 2024-11-19 at 09 57 18

## Notice


vizro-core/examples/visual-vocabulary/README.md
```diff
@@ -36,80 +36,79 @@ Credits and sources:
 
 The dashboard is still in development. Below is an overview of the chart types for which a completed page is available.
 
-| Chart Type            | Status | Category                 | Credits & sources                                                                                                               | API                                                                                                                                                                                             |
```
huong-li-nguyen, line 39: https://github.com/mckinsey/vizro/tree/demo/add-sparkline-chart/vizro-core/examples/visual-vocabulary

petar-qb (approved):
This looks really good!

vizro-core/examples/visual-vocabulary/pages/time.py
```diff
@@ -240,6 +240,46 @@
...
+                A sparkline chart is a compact line or area chart that displays multiple time series over a continuous
+                period. Without visible axes or labels, they are ideal for embedding within text, tables, or dashboards,
+                highlighting relative movement rather than precise values for a quick visual summary of trends.
+
+                &nbsp;
+
+                #### When should I use it?
+
+                Use sparkline charts to show trends for multiple time series sharing the same Y-axis quantity over the
+                same X-axis time range. They emphasize relative movement rather than precise values. To keep them
```
antonymilne, line 263:
```suggestion
                Use sparkline charts to show trends for multiple time series sharing the same y-axis quantity over the
                same x-axis time range. They emphasize relative movement rather than precise values. To keep them
```

vizro-core/examples/visual-vocabulary/pages/examples/sparkline.py
```diff
@@ -0,0 +1,17 @@
...
+
+@capture("graph")
+def sparkline(data_frame, **kwargs):
+    fig = px.line(data_frame, **kwargs)
+    fig.update_xaxes(visible=False)
+    fig.update_yaxes(visible=False)
+    fig.update_layout(showlegend=False)
+    return fig
+
+
```
antonymilne, line 16: With other charts (butterfly, dumbbells, diverging stacked bar) we have kept the data in wide form to avoid doing `melt` so I would favour doing so here as well unless you disagree strongly.

These two are identical but I prefer the first since it doesn't need any knowledge of "advanced" pandas. Internally when plotting wideform data plotly does exactly the `melt` operation you do manually here so it's kind of unnecessary for us to do it.

```
fig = sparkline(
    stocks, x="date", y=["GOOG", "AMZN", "AAPL"], labels={"variable": "stock"}, facet_row="variable"
)

fig2 = sparkline(stocks_melt, x="date", y="value", color="stock", facet_row="stock")
```

antonymilne (approved):
Looks nice, although actually I found having three charts with `facet_row` kind of confusing:
* are they the same y-axis scale?
* are they on a shared y axis or separate? Like without any context I think from glancing at the graph you could easily misunderstand the plot to be showing that Google stock is always above Amazon which is above Apple

So tbh I'm not sure how "best practice" this is. Maybe we should just show one stock instead of comparing three? Or maybe separate the graphs horizontally with a thin grey line?

No strong feelings though, so if this is commonly how sparklines are done then that's ok.

huong-li-nguyen:
> Looks nice, although actually I found having three charts with `facet_row` kind of confusing:
> 
> * are they the same y-axis scale?
> * are they on a shared y axis or separate? Like without any context I think from glancing at the graph you could easily misunderstand the plot to be showing that Google stock is always above Amazon which is above Apple
> 
> So tbh I'm not sure how "best practice" this is. Maybe we should just show one stock instead of comparing three? Or maybe separate the graphs horizontally with a thin grey line?
> 
> No strong feelings though, so if this is commonly how sparklines are done then that's ok.

Typically, sparklines display multiple traces, so having just a single trace wouldn't be effective in this context. It's not necessary for them to share the same y-axis scale, but they should use the same unit. The focus is less on comparing their magnitudes and more on observing the overall trend across all traces simultaneously.

I appreciate your suggestion to increase the separation. I've now added grey lines to better distinguish between them. 👍

PR #363, opened 2024-03-12
[Tidy] Convert actions to classes

petar-qb:
#### For those who will review this PR:
Here is a small recommendation on how to familiarise yourself with the PR changes in the fastest way and how to review the PR. It is only a recommendation and feel free to skip some of the suggested steps if you are already familiar with them or if you are not interested in them.

1. Don't rush to review the changes right away.
2. Read the entire PR description.
3. Watch the [Team Learning session](https://mckinsey.box.com/s/yj2xa86dwlyvbs27t04ouc5aum84d2p8).
4. Run all examples in `vizro-core/examples/_dev/` folder. They are already sorted alphabetically in the way they should run. For each example: read file docstrings, explore application configuration, and run and play with application UI.
5. Investigate `CapturedActionCallable` → useful Gist that @antonymilne made: https://gist.github.com/antonymilne/174c4c49ace2dae1f1d7674183b8d140
Link to the PS session -> https://mckinsey.box.com/s/fg1sdrpw6emprihhtg19ww4lft2ljma8
7. Open and understand one of the action implementations (e.g. `export_data_action`)
8. Search for `TODO-AV2` and read everything. (especially since some "TODO-AV2" refer to multiple similar pieces of code, but are only written in one place)
9. Start with discussing open questions ("TODO-AV2-OQ")
10. Feel completely free to contribute:
    - Implementing some better architecture solutions,
    - Open any new kind of question.
    - Solving TODOs,
    - Cleaning the code,
    - Applying linting stuff,
    - Fixing/adding tests
    - Literally anything else
11. Start with the standard code reviewing.

#### This PR is in the MVP phase because:
 - There are still some open questions that should be discussed.
 - The code is not completely tidied up.
 - tests, docs and linting is not completed.

#### The most important changes:
- Implementation of class based action using CapturedActionCallable.
- Actions validation and eager input arguments calculation.
- Make `filter_action` and `parameter_action` public,
- Change `_on_page_load` with a public `update_figures` action.
- Expose `Page.actions` so it can be overwritten. 
- `model_manager._get_model_page_id` method improved 

#### Bugfixes:
- Enable any Vizro model to work with actions in the control panel (https://github.com/mckinsey/vizro/issues/376) (`model_manager._get_page_actions_chains` improved)
- `export_data` is not able anymore to export data from other pages anymore (the bug is not spotted earlier and is fixed in this PR),

#### TODO-AV2 legend:
- `TODO-AV2` - I will solve these as the part of this PR,
[...]

PR #880, opened 2024-11-15
[Tidy] Proof of concept replacing ctd for filter and parameter

antonymilne:
## Description

@petar-qb raising this PR just to see if you think it's a good idea as a change. I won't actually implement it until you've done your PR so you don't have to resolve any more conflicts.

I'm finally continuing work on #363 and this seems like a small refactor that will tidy things up some more before the conversion to classes, which will hopefully be the next PR I raise on this topic 🤞

So far this PR is just a proof of concept to show what the change would look like. I haven't fully rolled it out or updated tests yet but the simple demo app works exactly the same as before.

Basically we passed `filters` and `parameters` through `inputs` in the callbacks before but didn't actually use them anywhere and instead looked inside `ctx.args_grouping` to extract all the values and and component ids. The only actual properties used in the `CallbackTriggerDict` are `id` and `value`. We can instead pass through named states in the form `{<id>: State(...)}` to get the `id` passed in automatically so we don't need `ctx.args_grouping` for that. I don't think we'll ever need it for anything else. Here's what's in it to remind you:

```
class CallbackTriggerDict(TypedDict):
    """Represent dash.ctx.args_grouping item. Shortened as 'ctd' in the code.

    Args:
        id: The component ID. If it's a pattern matching ID, it will be a dict.
        property: The component property used in the callback.
        value: The value of the component property at the time the callback was fired.
        str_id: For pattern matching IDs, it's the stringified dict ID without white spaces.
        triggered: A boolean indicating whether this input triggered the callback.

    """

    id: ModelID
    property: Literal["clickData", "value", "n_clicks", "active_cell", "derived_viewport_data"]
    value: Optional[Any]
    str_id: str
    triggered: bool
```

The change here would achieve be half of what you suggest in this comment:
```
        # TODO-AV2-OQ: Consider the following inputs ctx form:
        #  ```
        #  return {
        #      target_1: {'filters': ..., 'parameters': ..., 'filter_interaction': ..., 'theme_selector': ...},
        #      target_2: {'filters': ..., 'parameters': ..., 'filter_interaction': ..., 'theme_selector': ...},
        #  }
        #  ```
        #  Pros:
        #  1. We don't need anymore to send all filter/parameters/filter_interaction inputs to the server
        #  2. Potentially we don't need to dig through the ctx in the _actions_utils.py which decreases the complexity
```

The other half is also captured by my more recent comment:
```
    # TODO: the structure here would be nicer if we could get just the ctds for a single target at one time,
    #  so you could do apply_filters on a target a pass only the ctds relevant for that target.
    #  Consider restructuring ctds to a more convenient form to make this possible.
```

## Notice

### DIFF
--- a/vizro-core/changelog.d/20241115_145305_antony.milne_remove_ctx.md
+++ b/vizro-core/changelog.d/20241115_145305_antony.milne_remove_ctx.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Security
+
+- A bullet item for the Security category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
--- a/vizro-core/examples/scratch_dev/app.py
+++ b/vizro-core/examples/scratch_dev/app.py
@@ -41,6 +41,7 @@ page = vm.Page(
             ),
         ),
     ],
+    controls=[vm.Filter(column="pastry")],
 )
 
 dashboard = vm.Dashboard(pages=[page])
--- a/vizro-core/src/vizro/actions/_actions_utils.py
+++ b/vizro-core/src/vizro/actions/_actions_utils.py
@@ -47,9 +47,7 @@ def _get_component_actions(component) -> list[Action]:
     )
 
 
-def _apply_filter_controls(
-    data_frame: pd.DataFrame, ctds_filters: list[CallbackTriggerDict], target: ModelID
-) -> pd.DataFrame:
+def _apply_filter_controls(data_frame: pd.DataFrame, filters: dict[ModelID, Any], target: ModelID) -> pd.DataFrame:
     """Applies filters from a vm.Filter model in the controls.
 
     Args:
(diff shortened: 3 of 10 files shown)


petar-qb:
I like the initiative! There's a huge room for improvement regarding the `ctds`.  We have to find the best possible format for sending these from the client to the server.

 In your code version, states are sent like:

`dict(component_id_1=State(componenent_id, ...), component_id_2=...)`. 

It's similar to the current `filter_interaction` format. However, for the filter_interaction, we decided to sent it like this
```
list({
        "clickData": State(component_id=self.id, component_property="clickData"),
        "modelID": State(component_id=self.id, component_property="id"),  # required, to determine triggered model
    }, 
    {...},
)
```

The current filter_interaction way enables to propagate multiple different properties from of the same component to the server. So, we should preserve this flexibility I think. 

N.B.
There are two tickets we should keep in mind while thinking about the `ctds` changes:
- https://github.com/McK-Internal/vizro-internal/issues/575
- https://github.com/McK-Internal/vizro-internal/issues/576

antonymilne:
Great point, thanks @petar-qb. You're right that my proposed structure doesn't cater for one component with multiple `State`s.

After thinking about this more, I'm going to close this PR for now but open a ticket so we can revisit another time.

Ultimately I'd like to try and handle controls using pattern-matching callback like this: https://github.com/plotly/dash/issues/3063
```
@callback(Output("graph", "figure"), State({"control_type": "filter", "id": ALL}, "id"), State({"control_type": "filter", "id": ALL}, "value"))
```

and then do `zip(ids, property_values)` to match together the component id and value. When we have controls inside containers imagine you'd address `State({"control_type": "filter", "container": "container_name", "id": ALL}, "id")` or something. This would mean that we can remove a lot of the code that basically handles "find all the controls on this page".

There's only two possible drawbacks with this I think:
* it relies on Dash guaranteeing the order of `ALL` inputs - see https://github.com/plotly/dash/issues/3063. Hopefully that works in future
* we still need to be able to use `id`s for CSS selectors and other callbacks easily (see [1](https://community.plotly.com/t/trouble-linking-dbc-tooltips-to-pattern-matched-objects/41326/3) [2](https://community.plotly.com/t/pattern-matching-callbacks-and-dash-testing/47015). Assuming we do pattern matching just for controls (I don't see any use for it elsewhere anyway) we need the dictionary id to be associated with the selector itself. So we could still put the human-readable filter/parameter model id in some outer wrapper of a selector. We'd potentially need a translation to convert between someone targetting a `target=["filter_id"]` and the more complicated selector id. Hopefully we can avoid that by just putting ids in the right places and building the right innermost dynamically though just like in #879 

As a first step we might adopt the same structure but without pattern-matching ids like this:
```
ids=[State("component1", "id"), State("component1", "id"), State("component2", "id")], 
values=[State("component1", "clickData"), State("component1", "somethingElse"), State("component2", "value")]
```

...but I'm not in a big hurry to do that because it will be heavier changes across the codebase and is not a priority. Let's try and do this only after we've removed `ctds_filter_interaction`.

There's not much point doing a solution like I suggested here with `{<id>: State(...)}` because it _won't_ generalise to pattern-matching in future.

Let me know if it makes sense and I'll make a ticket for it.

petar-qb:
Hey @antonymilne. Thanks for the great investigation!! Pattern-matching looks really promising, and from the top of my head it probably looks like the end goal in dealing with controls as actions inputs. (This solution looks like it could enable -> https://github.com/McK-Internal/vizro-internal/issues/576, but I can't guarantee that right now)

> but I'm not in a big hurry to do that because it will be heavier changes across the codebase and is not a priority. Let's try and do this only after we've removed ctds_filter_interaction.

> There's not much point doing a solution like I suggested here with {<id>: State(...)} because it won't generalise to pattern-matching in future.

I totally agree with these two points. 👍  Let's create a ticket and deal with the `ctds` structure after we get rid of `ctds_filter_interaction`.

antonymilne:
Cool, thanks @petar-qb! Indeed this could solve a lot of things. e.g. also think of cases where a control could be added to a page dynamically. I don't think this would actually be possible without pattern-matching callbacks.

I've made a new issue here: https://github.com/McK-Internal/vizro-internal/issues/1376.

In due course I'll also make an issue to track forthcoming breaking changes to expect in 0.2.0, which currently exists just in rough notes on my computer.

Once we've got the next couple of PRs on filters and actions done let's re-prioritise all the various actions v2 tickets because things have become a lot clearer to me over the last couple of months, and some tickets may no longer be relevant.

Issue #82, opened 2023-09-29
Ability to create control headers with horizontal rulers


huong-li-nguyen:
### What's the problem this feature will solve?

@maxchristlmck raised that he would like to have the ability to create grouping headers for the controls with a horizontal ruler in place, see this draft PR: https://github.com/mckinsey/vizro/pull/81

### Describe the solution you'd like

Ability to add control headers and horizontal rulers

### Alternative Solutions

-

### Additional context

-

### Code of Conduct

- [X] I agree to follow the  [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

huong-li-nguyen:
@maxchristlmck, first of all thanks for raising this [draft PR ](https://github.com/mckinsey/vizro/pull/81) and trying to solve above yourself - much appreciated! 🚀 
I've looked into your draft PR and created an issue to discuss alternative solutions. :)

The above behavior can be achieved without creating two new models as in your current draft PR but by leveraging the existing Card component. It wasn't clear, as CSS was missing for horizontal rulers on our side! I created a quick commit here to show it can look like:  https://github.com/mckinsey/vizro/compare/main...demo/extend_card.

![Screenshot 2023-09-29 at 17 00 01](https://github.com/mckinsey/vizro/assets/90609403/b6947196-f02a-44ea-ac32-733cf03874f9)

@maxchristlmck - would that potentially work for you? The only thing you would have to add as a user on top would be a custom CSS file where you overwrite some of the default properties of the Card.

@antonymilne - The above change would require adding the Card component to the ControlType. What do you think? Both the Card and Button models seem to be quite useful as a ControlType so they can live on the left-side.

maxchristlmck:
Hi Li, your suggestion also works perfectly for me! Happy to go forward with this one :)

antonymilne:
Thank you for very much for the idea and drafting a PR @maxchristlmck! 

I completely understand why this would be useful but do have some qualms about adding `Card` as a `ControlType` @huong-li-nguyen. Probably best to discuss over Zoom since it's quite a complex topic 🙂 

Just to understand @maxchristlmck, could you see yourself wanting to put anything else in the control panel beyond just the title of the group and a horizontal line? e.g. writing a whole paragraph of text in markdown?

huong-li-nguyen:
Hey @maxchristlmck !

[...]

PR #903, opened 2024-11-28
[Feat] Recognise controls outside `page.controls`

antonymilne:
## Description

Following a conversation with @huong-li-nguyen and @maxschulz-COL, this should make the functionality of control groups as in https://github.com/mckinsey/vizro/issues/82 work again - see scratch_dev example.

While I was at it I did some tidying and refactoring to the `model_manager`:
* now there's a function to look up the page of a given model and a function to get all the models of one type on a given page
* we return models rather than model IDs since you can still get `model.id` directly from the model itself and it means we don't need lots of the `model_manager[model_id]` stuff we did before
* some of these changes to actions might not be a good idea and could change back again in work like #363 but let's see
* the `model_manager` changes are an improvement but definitely need more work - hopefully @maxschulz-COL will work on this in the nearish future after we're on pydantic v2

@petar-qb please could you check the failing unit test and see if it's a problem or not? I think there was a bug before in the `model_manager` so don't know if I've fixed that bug or broken something else or if it doesn't matter. Here's the full mesage: https://gist.github.com/antonymilne/7f3a3015772f0c992dd8b1194e6f9489.

## Notice


vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
```diff
@@ -32,21 +32,27 @@ def _get_inputs_of_controls(page: Page, control_type: ControlType) -> list[State
     """Gets list of `States` for selected `control_type` of triggered `Page`."""
     return [
         State(component_id=control.selector.id, component_property=control.selector._input_property)
-        for control in page.controls
-        if isinstance(control, control_type)
+        for control in cast(Iterable[ControlType], model_manager._get_models(control_type, page))
```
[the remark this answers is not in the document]
huong-li-nguyen: Ahh, that's so nice, that we can get all models of a specific type from a selected page as well! 🚀

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,60 +1,70 @@
...
+
+    type: Literal["control_group"] = "control_group"
+    title: str
+    controls: List[ControlType] = []
 
-gapminder_2007 = px.data.gapminder().query("year == 2007")
+    def build(self):
+        return html.Div(
+            [html.H4(self.title), html.Hr()] + [control.build() for control in self.controls],
+            className="control_group_container",
```
huong-li-nguyen, line 23: I don't know if we suggested this anywhere now, but the className is not required anymore, nor is any custom CSS, because the `H4` and the `Hr` element get automatically styled by out vizro-bootstrap theme :) 

```suggestion
```

huong-li-nguyen (approved):
Little refactoring, but big change! 🚀 This is amazing and fixes so many limitations that we've encountered this week.

Understood the main change, and checked all our demos, but I'll leave the action unit tests to @petar-qb :)

vizro-core/examples/scratch_dev/app.py
antonymilne: Ahhh nice, that is very good to hear 👍

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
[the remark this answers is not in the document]
antonymilne: Yes, this has been a TODO for a long time for the `model_manager`. It's still pretty hacky tbh but will definitely improve in future.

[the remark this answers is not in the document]
maxschulz-COL: Does this essentially equate controls and components? Like I could put Graph in controls, and controls into components?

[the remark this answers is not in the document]
antonymilne: I think yes, you could do that. So long as you have the right sort of model it shouldn't matter whether it's in `controls` or `components` or `navigation` or wherever.

vizro-core/src/vizro/models/_components/form/_text_area.py
maxschulz-COL, line 1: [This showcases what I mean](https://py.cafe/snippet/vizro/v1?pycafe-app-view=undefined&pycafe-edit-enable=undefined#c=H4sIAIacSWcEA41Vy27rNhD9FUHZOEBMWHZeDqDFRQoUXRS4i7abq8CgJdrmNU2y5MiJXPTfe4aSnMRAgcgLi_M4c-bB0T957RqVP-Wb4A7ZUZ-Cy_TBu0DZX3yo7HBKKuGNI9MJ9eaDijGTMfNvFyYH4JmkOh4q-w47KMSqdjC3ylIUGxcOYtVGFVba-pbG0H9C8hsLvgZA6o1WMig5-v8BwTecB3ffNdKSrsWxGC2-B32UpL4Rhcr2Px10zEokJBpJUvBxcs2K2kjk-txGcoczs8n57fqpshmeq-x5JJb54LwKpFXMQDGTNWlnURTbZNqSCoOgd-xzXw0-3VMWKYDIB4qTKj9K06oqT4SuUFrxXQZ5UMASsmlW1HkFq6iMqsmFKr85FwEuX3N4z6iyX7G_qEii5uVWgXty38LBcoKkyaiyyn_vso0OEeWBDgi99r2d5Y9ewg8Qfg3S7waI8dENcGItCbxW9U4GOuOMz0Zv26BK9HGwm3Arbz6yGBTZCJC9Mary0qyMslvasayDzCuC7FU3vah2xgU29apGb9GOD7HH95fhv3aWgjOXWZ3LepEZybBVqQaf8xOJd5WPsOMzdqK8aMNk5MHP-M7OqT-NjLu1k6Hpm_TLeJxwSxCc_16SZbr-k2uxbrVpJme_axFa298L_uU3eVB_tzqoA3cQiyTd1bKciULMH5MBj86owNF3tHMWAt-5RjdqepyJ-b0ooDKycy3lT7Y15ibfaKOA-OMFClfv-fiMmiIOnK-yHSg9ZQu1vp-tMaFNoeYPa_m4XDfrzWa5LJZ3s0LNKiutdSRJNVMmEpnavWB5S25j5F6V5VwsRFHZtZH1HqdbUczYYm203atQloVY8rmW9U4ZvWaIXoD7qjcaLjM4zcUcIvQsKppabCZp9IndF4BnldEM_4jKPPSd4MjFI-v4NJXb6TZgwstFAQ_gJ-naOcJGkH764aaAUkoiWdQuqE_KuUj8k3JHB_O_ygPvRfvZGZ0Dq8GA5BqXprzrfVCuiAzATSyG05SLou2WcRNn3VjJJkix37Uo2BQDL3mrcvbAgooidvJWBdcmSiBV2Z_a_pRzdkb-lT3IsG99lJvUoULcQdT5boq9pmzk_cllSMSsijSVsbO1xuwNtbHtwXd8wnTdVlhNnNI5mJe043v8nrA3kviD0mBHleUt0rlnIX_xuAI8FjgP3xLGeUg4gwA1xNZhomgwi9NsAQi97-36ad8ReY7Zh0x3YYrSqJa0YXcMlvAuUnKhE0Q8W7Dle4Y0UwYYEOiDotCl2vO8IEUMXuvJ8dIp75eoNNyifmN9kUqCwslaE_LhMCzoPAA-lRTBEjqd-o6N8dtg0MoFh0_cLy76qwr7k2pBhscDlXvFyNJ76JP2yBvFSOAxmsrm__4H0SEhWH8IAAA)

vizro-ai/hatch.toml
```diff
@@ -29,12 +29,9 @@ dependencies = [
   "langchain-anthropic",
   "pre-commit"
 ]
+env-vars = {VIZRO_AI_LOG_LEVEL = "DEBUG"}
 installer = "uv"
 
-[envs.default.env-vars]
-UV_PRERELEASE = "allow"
-VIZRO_AI_LOG_LEVEL = "DEBUG"
-
```
petar-qb, line 37: I've seen similar changes about UV_PRERELEASE a few times in different PRs. Can you just quickly explain what these changes are supposed to do? 😄

vizro-core/changelog.d/20241128_114150_antony.milne_model_manager.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Custom controls can be nested arbitrarily deep inside `Page.controls`. ([#903](https://github.com/mckinsey/vizro/pull/903))
```
petar-qb, line 28: This PR allows `controls` to live under `Page.components` (even if they are wrapped deep under another component there). Everything works as expected, so would it be worth announcing this update as well?

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
petar-qb, line 35: Would the `_get_inputs_of_controls` be even more flexible if we searched for all `_filter`/`_parameter` Action functions on the page (instead for `vm.Filter`/`vm.Parameter` components)? In that case if a `_filter` action (or its future public counterpart e.g: `filter`) is assigned to the pure `dcc.Dropdown()`, this component would behave exactly like the `vm.Filter` (except the `_dynamic` behaviour and potentially some other.). 

This is more like a theoretical question and I guess its answer mostly depends on how we see Vizro backend in future and it also depends on many other decisions (e.g. how to make a standalone dcc.Dropdown to be dynamic, and many many more). Currently, I can't say what exactly are all the pros/cons of this approach and is any new use-case covered with this approach or another. So, this comment here is just an idea.

petar-qb (approved):
This is a huge improvement! It's so intuitive to use model_manager helper methods now. 🎖️

petar-qb:
@antonymilne 

> please could you check the failing unit test and see if it's a problem or not? I think there was a bug before in the model_manager so don't know if I've fixed that bug or broken something else or if it doesn't matter. Here's the full message:

The test fails because the `_get_action_loop_components` method now returns a list of `dcc.Store` and `html.Div` components in a "random" order , due to how elements are returned from the `model_manager._get_models()`. So, the assertion `assert_component_equal(result, expected)` fails because it expects a specific sequence of elements in the list.

Also, couple of these `dcc.Store` elements have more complex "data" content (e.g. `list[str]`, `dict[str, list[str]`]) that's also randomised now, which complicates refactoring the test. (It's important that this order of elements does not impact the dashboards' functionality.)

After several refactoring attempts I ended with this stupid solution:

```py
    def test_all_action_loop_components(
        self,
        fundamental_components,
        gateway_components,
        action_trigger_components,
        action_trigger_actions_id_component,
        trigger_to_actions_chain_mapper_component,
    ):
        result = _get_action_loop_components()

        # Validate the action_loop_components_div wrapper
        assert_component_equal(result, html.Div(id="action_loop_components_div"), keys_to_strip={"children"})

        # Validate the children of the result component.
        # Since the order of children is not guaranteed due to how elements are retrieved
        # from the model_manager, we perform order-agnostic comparisons.
        #
        # * fundamental_components, gateway_components, and action_trigger_components:
        #   These are dcc.Store objects with simple "data" attributes, so we compare their string
        #   representations directly.
        # * action_trigger_actions_id_component and trigger_to_actions_chain_mapper_component:
        #   These components have more complex attributes, so we compare their `id` and `data`
        #   fields explicitly. "data" is compared as a set to ignore order.

        # Test: fundamental_components + gateway_components + action_trigger_components
        expected_children = fundamental_components + gateway_components + action_trigger_components
        result_children = result.children[:len(expected_children)]
        assert {str(elem) for elem in expected_children} == {str(elem) for elem in result_children}

        # Test: action_trigger_actions_id_component
        index_action_trigger = len(expected_children)
        action_trigger_result = result.children[index_action_trigger]
        assert action_trigger_actions_id_component.id == action_trigger_result.id
        assert set(action_trigger_actions_id_component.data) == set(action_trigger_result.data)

        # Test: trigger_to_actions_chain_mapper_component
        index_trigger_mapper = index_action_trigger + 1
        trigger_mapper_result = result.children[index_trigger_mapper]
        assert trigger_to_actions_chain_mapper_component.id == trigger_mapper_result.id
        assert set(trigger_to_actions_chain_mapper_component.data) == set(trigger_mapper_result.data)

```

vizro-core/src/vizro/actions/_action_loop/_action_loop.py
```diff
@@ -37,5 +41,8 @@ def _build_actions_models():
             List of required components for each `Action` in the `Dashboard` e.g. list[dcc.Download]
 
         """
-        actions = _get_actions_on_registered_pages()
-        return html.Div([action.build() for action in actions], id="app_action_models_components_div", hidden=True)
+        return html.Div(
+            [action.build() for action in cast(Iterable[Action], model_manager._get_models(Action))],
```
petar-qb, line 45: Is this this casting done just for mypy?

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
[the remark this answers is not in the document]
maxschulz-COL: Coming back to this, I think it's amazing. We essentially have the control and component section as fillable canvases, and it is up to the user how to fill it. A little less design guidance, but quite a bit more flexibility ❤️

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -50,90 +50,58 @@ def __iter__(self) -> Generator[ModelID, None, None]:
...
+            if model_type is None or isinstance(model, model_type):
+                yield model
+
+    def _get_model_children(self, model: Model) -> Generator[Model, None, None]:
+        from vizro.models import VizroBaseModel
+
+        if isinstance(model, VizroBaseModel):
+            yield model
+
+        model_fields = ["components", "tabs", "controls", "actions", "selector"]
```
maxschulz-COL, line 76: I would recommend putting this in a constant. I know this is not your style if only used once, but this should be more visible so that later updating becomes clearer

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -50,90 +50,59 @@ def __iter__(self) -> Generator[ModelID, None, None]:
...
-                self._get_model_children(child_model.id, all_model_ids)
-        if hasattr(model, "tabs"):
-            for child_model in model.tabs:
-                self._get_model_children(child_model.id, all_model_ids)
-        return all_model_ids
-
-    # TODO: Consider moving this method in the Dashboard model or some other util file
-    def _get_model_page_id(self, model_id: ModelID) -> ModelID:  # type: ignore[return]
+    # TODO NOW: Make brief comments on how in future it should work to use Page (or maybe Dashboard) as key primitive.
+    def _get_model_children(self, model: Model) -> Generator[Model, None, None]:
```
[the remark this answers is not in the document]
maxschulz-COL: I think this function deserves a docstring

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -102,9 +102,10 @@ def pre_build(self):
         # want to raise an error if the column is not found in a figure's data_frame, it will just be ignored.
         # This is the case when bool(self.targets) is False.
         # Possibly in future this will change (which would be breaking change).
-        proposed_targets = self.targets or model_manager._get_page_model_ids_with_figure(
-            page_id=model_manager._get_model_page_id(model_id=ModelID(str(self.id)))
-        )
+        proposed_targets = self.targets or [
+            model.id
+            for model in model_manager._get_models((Graph, AgGrid, Table, Figure), model_manager._get_model_page(self))
```
maxschulz-COL, line 107: A little similar to before, but would we want to abstract this somewhere and have tuple as a constant, e.g. `FIGURE_MODELS`

maxschulz-COL (approved):
🚀  This will unlock quite a bit quite soon (ie today) 🚀

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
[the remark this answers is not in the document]
antonymilne: I am not sure how much I like it tbh because then the controls vs. components becomes more about page layout rather than functionality. Like almost we are referring to these bits as `vm.Page(sidebar=[vm.Filter(...)], main=[vm.Graph(...)])` etc.? That's [quite similar to what streamlit does](https://docs.streamlit.io/develop/api-reference/layout/st.sidebar) and not necessarily a bad idea but quite different from how we currently do things. I can also see where the extra flexibility is useful here though.

vizro-ai/hatch.toml
antonymilne: So basically `UV_PRERELEASE = "allow"` is like setting `--pre` on `pip install` and means that you get pre-releases of packages, e.g. pydantic sometimes do alpha release that by default you don't get without explicitly allowing them.

In principle it's a good idea to use prereleases in CI because then if a forthcoming release of one of your dependencies breaks your tests you have more notice and can try and fix it in advance of the actual release coming out and affecting users. In reality though I found that a plotly alpha release broke all our tests in a way that will not be the case in their stable release. So testing against prereleases wasn't possible in the end.

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
antonymilne: Good question! I'm actually thinking about this already in #363 right now and not yet sure or how to handle it, so let me keep this unresolved to come back to.

vizro-core/changelog.d/20241128_114150_antony.milne_model_manager.md
petar-qb: I agree we shouldn't rush into announcing something we don't recommend or don't know will it be changed in the future. However, it's good to know it's possible now (and I'm pretty sure it wasn't possible before, but haven't checked that).

vizro-core/src/vizro/actions/_action_loop/_action_loop.py
antonymilne: Yes, `cast` is only ever required for mypy. In these cases actually it feels like mypy should be able to work this out for itself and not need any casting at all, but unfortunately that doesn't work, not sure why.

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
```diff
@@ -504,21 +506,25 @@ def build(self):
 class TestFilterBuild:
     """Tests filter build method."""
 
+    # Use lambda to create test_selector only in the test itself rather than in the parameters, so that it's in the
+    # model_manager for the test.
```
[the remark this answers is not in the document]
antonymilne: Good question 🤔 I just took it out and unit tests still seem to pass. I'm sure they didn't pass before though which is why I made this change. 😕 Let me check again.

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -50,90 +50,58 @@ def __iter__(self) -> Generator[ModelID, None, None]:
...
-            for child_model in model.components:
-                self._get_model_children(child_model.id, all_model_ids)
-        if hasattr(model, "tabs"):
-            for child_model in model.tabs:
-                self._get_model_children(child_model.id, all_model_ids)
-        return all_model_ids
-
-    # TODO: Consider moving this method in the Dashboard model or some other util file
-    def _get_model_page_id(self, model_id: ModelID) -> ModelID:  # type: ignore[return]
+    def _get_models(
```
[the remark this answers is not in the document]
antonymilne: Good idea! I'll make `_get_model_children` into `__get_model_children`.

As for making these methods public: I'm going to wait until @maxschulz-COL and I have reworked the model manager a bit before doing that because I think it's quite likely these will change again.

vizro-core/src/vizro/actions/_callback_mapping/_callback_mapping_utils.py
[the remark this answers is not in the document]
maxschulz-COL: It could still be a case of guiding through argument name, but allowing more than that! Ie you _should_ put your controls in the control argument, and we won't let you even do it elsewise without a little extra work (ie `.add_component`) BUT ....

vizro-core/src/vizro/managers/_model_manager.py
antonymilne: This I have decided not to do because I don't intend to update this list later - instead I'd like it to become completely obsolete as part of `model_manager` refactoring where we can look at _all_ models recursively without listing explicitly the fields we look at. I'll add a TODO to explain this.

vizro-core/src/vizro/models/_controls/filter.py
antonymilne: Good idea! Done, although it was a bit tricky with circular imports.

vizro-core/src/vizro/managers/_model_manager.py
antonymilne: [90d5868](https://github.com/mckinsey/vizro/pull/903/commits/90d586813fc33ffc28f272f1a27e7c7904b72a8d)

vizro-core/src/vizro/managers/_model_manager.py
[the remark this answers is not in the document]
antonymilne: Done in [90d5868](https://github.com/mckinsey/vizro/pull/903/commits/90d586813fc33ffc28f272f1a27e7c7904b72a8d).

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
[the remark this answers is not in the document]
antonymilne: [90d5868](https://github.com/mckinsey/vizro/pull/903/commits/90d586813fc33ffc28f272f1a27e7c7904b72a8d)

vizro-core/src/vizro/managers/_model_manager.py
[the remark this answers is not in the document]
antonymilne: [90d5868](https://github.com/mckinsey/vizro/pull/903/commits/90d586813fc33ffc28f272f1a27e7c7904b72a8d)

PR #907, opened 2024-11-29
[Tidy] Replace `dmc.Switch` with `dbc.Switch`

huong-li-nguyen:
## Description
- Replace `dmc.Switch` with `dbc.Switch`
- Add SASS Code for `dbc.Switch`  -> https://github.com/McK-Private/vizro-bootstrap/pull/40
- Replace static CSS file for `dmc.Switch` with updated `vizro-bootstrap.min.css`
- Replace screenshots and fix xpath  -> https://github.com/mckinsey/vizro-qa/pull/136

## Notice


vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -197,12 +196,11 @@ def _get_page_divs(self, page: Page) -> _PageDivsType:
             else html.H2(id="dashboard-title", hidden=True)
         )
         settings = html.Div(
-            children=dmc.Switch(
-                id="theme_selector",
-                checked=self.theme == "vizro_light",
+            children=dbc.Switch(
+                id="theme-selector",
```
petar-qb, line 200: Maybe we should introduce a convention about usage of `snake_case` and `kebab-case`.

petar-qb (approved):
This looks good! 👍 

If you find it useful, extract the creation of the `dbc.Switch` in a separate helper function (and reuse it in two places) in the `dashboard.py`.
You can also change the variable names in the `dashboard.js` (but also not necessary 😄).

vizro-core/src/vizro/models/_dashboard.py
huong-li-nguyen: I agree! I fear we will always have a mix though 😄 

In general, it's best practice to use dashes instead of underscores for all CSS IDs, classnames etc. Hence, why I changed it here. https://stackoverflow.com/questions/7560813/why-are-dashes-preferred-for-css-selectors-html-attributes

However, we use underscores in our Python code and all our model IDs. Hence, most of our callback logic (inputs/outputs) uses underscores. If you take a look at our HTML doc, some of the HTML divs have a classname provided by dash automatically and they also use underscores/dashes combined 😄

huong-li-nguyen:
> This looks good! 👍
> 
> If you find it useful, extract the creation of the `dbc.Switch` in a separate helper function (and reuse it in two places) in the `dashboard.py`. You can also change the variable names in the `dashboard.js` (but also not necessary 😄).

Thanks for the quick review! 🙏 For now, I think I'll stick with the duplication. Normally, we only need to insert the theme-switch once for all dashboard pages, but the 404 page is an exception since it gets rendered separately and we want it to match. If we come across more cases like the 404 page, we can definitely reconsider. 👍 For now, I find the duplication clearer because having a helper function outside the Dashboard class would make it harder to overwrite, and it doesn't quite qualify as a separate method within the Dashboard class, as it's already part of the output `_get_page_divs` method.

vizro-core/src/vizro/models/_dashboard.py
petar-qb: My personal preference is snake case for one reason only: because I'm used to using it 😄 But, I prefer more to associate all IDs with the same case (no matter which one). However, I don't think we need to do this right away and I consider it a really low priority.

antonymilne: I think it's right to prefer the dashes here to match the CSS container rule 👍 It is indeed inconsistent with some callbacks but I don't think it's possible to achieve full consistency here without changing everything to underscores, which just looks wrong in CSS. So basically I think it's good as it is and we have to live with the inconsistency.

antonymilne (approved):
⭐ 🌟 ⭐ 🌟 ⭐ 🌟 ⭐ So nice to see everything being consolidated into bootstrap here!

Now the only dmc component left is the date picker?

huong-li-nguyen:
> ⭐ 🌟 ⭐ 🌟 ⭐ 🌟 ⭐ So nice to see everything being consolidated into bootstrap here!
> 
> Now the only dmc component left is the date picker?

Yeees!

vizro-core/src/vizro/models/_dashboard.py
petar-qb: @huong-li-nguyen @antonymilne 

> Hence, most of our callback logic (inputs/outputs) uses underscores. 

> It is indeed inconsistent with some callbacks

I'm curious: Is it possible to change all our IDs from underscores to dash-like, even though it requires changing the internal components IDs alongside with places where they are used in the actions/callbacks?

PR #910, opened 2024-12-02
[Tidy] Replace custom classNames with BS utility CSS classes

huong-li-nguyen:
## Description

Our long-term objective is to minimize the use of custom CSS in the static folder, favoring Bootstrap syntax instead. To achieve this, we can replace custom class names with Bootstrap utility CSS classes: https://getbootstrap.com/docs/4.0/utilities/borders/. 

This PR shows one example on how to do so. However, I don't recommend replacing all custom class names with Bootstrap shortcuts, as it could make it harder for users to override styles. 

Here are my current guidelines:

- Use Bootstrap classes instead of custom class names when it's unlikely that users will want to customize the HTML element via CSS, or when there is an alternative method for customization.
- Use custom class names for HTML elements that users are likely to want to customize e.g. all our main containers (page-header, control-panel, right-side, left-side etc.)
- Not related to above: Remove any redundant extra div if you see one

See example here where we could simplify using BS shortcuts:  https://github.com/mckinsey/vizro/pull/910#discussion_r1865476510

## Notice


vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -308,31 +308,18 @@ def _make_page_404_layout(self):
...
-                    ),
-                    id="settings",
+                dbc.Switch(
+                    id="theme-selector",
+                    value=self.theme == "vizro_light",
+                    persistence=True,
+                    persistence_type="session",
                 ),
                 html.Img(src=f"data:image/svg+xml;base64,{error_404_svg}"),
-                html.Div(
```
huong-li-nguyen, line 321: This nesting wasn't necessary anymore, as I changed the whole container to a flex box that is automatically center-aligned.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -308,31 +308,18 @@ def _make_page_404_layout(self):
...
-                        dbc.Button(children="Take me home", href=get_relative_path("/")),
-                    ],
-                    className="error-content-container",
-                ),
+                html.H3("This page could not be found."),
+                html.P("Make sure the URL you entered is correct."),
+                dbc.Button(children="Take me home", href=get_relative_path("/"), className="mt-4"),
             ],
-            className="page-error-container",
+            className="d-flex flex-column align-items-center justify-content-center min-vh-100",
```
huong-li-nguyen, line 322: This is the main change: We would use Bootstrap shortcuts Instead of custom `classNames`.

This approach has the advantage of allowing us to remove more custom CSS from our static folder. However, it also makes it more difficult for users to overwrite styles, as they can no longer target a specific `className` like `page-error-container`. Additionally, these Bootstrap shortcuts apply the CSS properties with an important tag, which makes customization even more difficult.

So, we need to decide on a case-by-case basis whether to implement this change. For example, we want users to easily target and overwrite styles for certain containers (e.g., control panel, left-side, right-side, grid-layout) using `classNames` or `IDs`, so I wouldn't change it to use BS shortcuts.

In this case, I believe users rarely need to overwrite the existing `page-error-container`. If they want to customize the 404 page, they can still modify the `_make_page_404_layout()` method. 

Let me know what you think @antonymilne. I also don't want this to get too complex. So if it's not clear when to use BS shortcuts vs. custom classNames, then I am also fine leaving it as is 😄

nadijagraca (approved):
Looks good to me! 🙌

antonymilne (approved):
Generally agree with everything you said, although I am probably more pro-bootstrap native than what you suggest. My proposed rule of thumb would be something like: use bootstrap shortcuts as much as possible. But if that's not practical then I agree with what you say 👍 

> Use Bootstrap classes instead of custom class names when it's unlikely that users will want to customize the HTML element via CSS, or when there is an alternative method for customization.

Agree.

> Use custom class names for HTML elements that users are likely to want to customize e.g. all our main containers (page-header, control-panel, right-side, left-side etc.)

These are currently done with `id` rather than `className`, so we can potentially combine both our custom CSS and bootstrap and a user can still further customise them?
```
html.Div(id="page-header", classNames="align-items-left ...")
```

Then in our CSS we just provide any extra bits that we need. We wouldn't do any bootstrap overrides here, just provide extra attributes:
```
#page-header {
    color: red; /* Some attribute not specified in bootstrap */
}
```

And if a user wanted to change something they would do:
```
#page-header {
   color: red; /* Some attribute not specified in bootstrap */
   align-items: right !important; /* Override something specified in boostrap */
}
``` 

The only thing that's awkward here is the `!important` that's needed, and maybe if that's already used in bootstrap it doesn't even work well in practice? If that's the case then agree with what you say and I'm fine to just do this like:
```
html.Div(id="page-header")

#page-header {
    align-items: left; /* All the things done by the bootstrap styles put here manually */
    color: red;
}
```

> Not related to above: Remove any redundant extra div if you see one

Always agree with this 👍

vizro-core/src/vizro/models/_dashboard.py
antonymilne: 💯 love this sort of change. If we compare our DOM now compared to 18 months ago the improvements are just HUGE.

vizro-core/src/vizro/models/_dashboard.py
antonymilne: Definitely agree with using bootstrap here.

huong-li-nguyen:
> The only thing that's awkward here is the `!important` that's needed, and maybe if that's already used in bootstrap it doesn't even work well in practice? If that's the case then agree with what you say and I'm fine to just do this like:
> 


Yes, that's why I wasn't more supportive of using native Bootstrap. When you use any of Bootstrap's utility classes, they automatically include the `!important `tag.

For example, if you write: `html.Div(id="page-header", classNames="align-items-left)`

Bootstrap will apply the following CSS:

```
#page-header {
   align-items: left !important;
}
```

This forces the user to also use the` !important` tag in their CSS to override it. Normally, user-defined CSS should take precedence since it is loaded last, but we can't predict how users will utilize vizro and the assets folder. For instance, we had difficulty styling some components of the dcc components because their CSS also used `!important` tags. This made it challenging to override styles due to conflicting orders and as a user you had to configure out the specificity. 

Additionally, our users would need to inspect elements to differentiate between CSS properties from our static folder and Bootstrap's classes with `!important` tags, and their own styles. So I felt that approach would just lead to the user overusing the `!important` which in general is bad practice and would lead to confusion and potential errors 🤔

huong-li-nguyen:
I've already merged this PR, but we can continue discussing our future strategy. I plan to replace all other custom class names with Bootstrap utilities for those elements that users are unlikely to want to customize.

For elements where users might want to apply custom styles, I'll leave the class names unchanged for now, pending our discussion above. 👍

Issue #905, opened 2024-11-28
Try to bump dmc


antonymilne:
Context: https://github.com/mckinsey/vizro/issues/901#issuecomment-2505980839. dmc is now at 0.15. Let's _consider_ unpinning:
1. Higher priority: see if we can get same datepicker behaviour as now but hopefully with fewer bugs (maybe helps with dynamic filter @petar-qb?). Initially we just need to check feasibility of this and whether it's something we can do quickly and easily or whether it's a huge piece of work.
2. Lower priority/separate ticket: consider #318 and how we might like to integrate timepicker into vizro.

After https://github.com/mckinsey/vizro/pull/895 I think the only dmc components we have left are `dmc.Switch` and `dmc.DateRangePicker/dmc.DatePicker`.

@huong-li-nguyen are there plans to replace `dmc.Switch` with `dbc`?

huong-li-nguyen:
Yes, the `dmc.Switch` will be replaced with the `dbc.Checklist` next 👍 To convert the DatePicker could be a bit more complex as the API has changed.

petar-qb:
This issue depends a bit on fixing the dash persistence [bug](https://github.com/McK-Internal/vizro-internal/issues/1389). We can start working on this ticket, but the date_picker persistence will be harder to implement.
P.S. The date_picker persistence already doesn't work.

AnnMarieW:
Hi Vizro Team,

I’m glad to hear that you’re considering upgrading to the latest version of `dash-mantine-components`. As the primary maintainer of DMC, I’m happy to assist if you decide to move forward with this project.

Upgrading from version `0.12` to the latest release does involve significant changes. In large apps, this can be quite time-consuming; however, in Vizro, it’s relatively straightforward since only one component is in use. To help, I’ve prepared a draft PR #924 for you to review.

The current DMC version offers many benefits: enhanced stability, more features, and critical bug fixes, all thanks to its foundation on Mantine V7. You can find more details in our [Roadmap](https://github.com/snehilvj/dash-mantine-components/discussions/377).

We’ve been making rapid progress on DMC, releasing updates approximately every two weeks. You can see recent release announcements in our [GitHub Discussions](https://github.com/snehilvj/dash-mantine-components/discussions?discussions_q=is%3Aopen+Release+0.).

---

### Benefits of Upgrading DMC in Vizro

[...]

PR #924, opened 2024-12-07
[Feat] Update to use dmc 0.15.1

AnnMarieW:
This is a draft PR.  For details, please refer to the discussion in #905

**TO DO:**
- [x] Fix unit tests
- [x] Add new styling to new DatePicker @nadijagraca -> https://github.com/AnnMarieW/vizro/pull/1

## Notice


vizro-core/src/vizro/_vizro.py
```diff
@@ -47,6 +50,9 @@ def __init__(self, **kwargs):
             suppress_callback_exceptions=True,
             title="Vizro",
             use_pages=True,
+            # This is added here temporarily for testing. These stylesheets are required for certain dmc components.
+            # It may be better to include these stylesheets in the css folder.
+            external_stylesheets=dmc.styles.ALL,
         )
```
huong-li-nguyen, line 56: We still need to let users define external_stylesheets like this: `Vizro(external_stylesheets=XXX).build(dashboard).run()`. So, it's better to append the stylesheet to the list instead of setting it here :) 

Let's just add `dmc.styles.DATES` to avoid loading unused stylesheets for now. If users need all `dmc` styles, they can still do it with `Vizro(external_stylesheets=dmc.styles.ALL).build(dashboard).run()`. This way, we only load what we need 👍 

Not sure if below will stay where it is, but for now, let's just add it there.

```suggestion
        )
        
        # Ensure external_stylesheets is a list and append the additional stylesheet
        external_stylesheets = self.dash.config.external_stylesheets
        self.dash.config.external_stylesheets = external_stylesheets if isinstance(external_stylesheets, list) else [
            external_stylesheets]
        self.dash.config.external_stylesheets.append(dmc.styles.DATES)
```

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -27,7 +26,7 @@ class DatePicker(VizroBaseModel):
     [`dmc.DateRangePicker`](https://www.dash-mantine-components.com/components/datepicker#daterangepicker).
 
     Args:
-        type (Literal["date_picker"]): Defaults to `"date_picker"`.
+        type (Literal["default", "range", "multiple"]): Defaults to A single date picker allowing the selection of one date.
```
huong-li-nguyen, line 29: I understand the confusion. This argument isn't the same as the `type` argument inside the `dmc` component. It's actually an argument that all our pydantic models have to differentiate their types. So, we shouldn't change it. If we ever want to introduce the `type` argument from the `dmc` component, we should add it as a new argument with a different name. But for now, let's leave it as is and not introduce it yet 👍 

```suggestion
        type (Literal["date_picker"]): Defaults to `"date_picker"`.
```

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -37,11 +36,13 @@ class DatePicker(VizroBaseModel):
 
     """
 
-    type: Literal["date_picker"] = "date_picker"
+    type: Literal["default", "range", "multiple"] = "default"
```
huong-li-nguyen, line 39:
```suggestion
    type (Literal["date_picker"]): Defaults to `"date_picker"`.
```

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -37,11 +36,13 @@ class DatePicker(VizroBaseModel):
...
     min: Optional[date] = Field(None, description="Start date for date picker.")
     max: Optional[date] = Field(None, description="End date for date picker.")
     value: Optional[Union[list[date], date]] = Field(None, description="Default date for date picker")
     title: str = Field("", description="Title to be displayed.")
+
+    # Could probably delete the `range` arg, but keeping it makes it backwards compatible
```
huong-li-nguyen, line 45: Good point! The `range` argument is definitely redundant if we ever want to introduce the `type` argument of the new `dmc.DatePickerInput` as well. Options would then include: "default", "range", "multiple" in which case this boolean argument doesn't make much sense. 

Let's leave it in for now and discuss when @antonymilne is back 👍

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -71,33 +72,15 @@ def build(self):
...
-
-        date_picker_class = dmc.DateRangePicker if self.range else dmc.DatePicker
 
-        # dropdownPosition must be set to bottom-start as a workaround for issue:
-        # https://github.com/snehilvj/dash-mantine-components/issues/219
-        # clearable must be set to False as a workaround for issue:
-        # https://github.com/snehilvj/dash-mantine-components/issues/212
-        # maxDate must be increased by one day, and later on disabledDates must be set as maxDate + 1 day
-        # as a workaround for issue: https://github.com/snehilvj/dash-mantine-components/issues/230
-        date_picker = date_picker_class(
```
huong-li-nguyen, line 91: Saw that all of these issues have been fixed in the mantine repository @AnnMarieW - this is amazing! 🥳 

@nadijagraca @petar-qb - could you double-check manually if all of the issues we've faced initially are solved now? I remember for one of the bugs, we had to add the datepicker to the right side and then change screen sizes etc. I don't know if there were other scenarios like that which need manual checking. You will know best 👍

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -71,33 +72,15 @@ def build(self):
...
-            maxDate=self.max + datetime.timedelta(days=1) if self.max else None,
+            maxDate=self.max,
             persistence=True,
             persistence_type="session",
-            dropdownPosition="bottom-start",
-            clearable=False,
-            disabledDates=self.max + datetime.timedelta(days=1) if self.max else None,
+            type="range" if self.range else "default",
             className="datepicker",
             **date_range_picker_kwargs,
```
huong-li-nguyen, line 85:
```suggestion
            **date_range_picker_kwargs,
```

Just a comment for myself (no need to push anything yet): If we decide to remove the `range` argument and introduce the `type` argument under a different name instead, this can go and above `date_range_picker_kwargs = {"allowSingleDateInRange": True} if self.range else {}` can be deleted.

huong-li-nguyen (commented):
First, thank you so much, @AnnMarieW, for this PR and your detailed explanations [here](https://github.com/mckinsey/vizro/issues/905)! 🚀 This is incredibly helpful.

We definitely plan to upgrade the `dmc` version, and although we've removed other `dmc` components, we'll likely always keep the DatePicker since it's the best available! 💯 

I've left a few comments, but there's no rush to implement them. We probably won't merge this PR until next year. We were already planning to upgrade `dmc` next year, but we wanted to check feasibility first. Your PR is fantastic for that! Everything seems to work fine, and it looks very feasible and straightforward from a code perspective. Thank you so much! 🙏 

I'll have @nadijagraca and @petar-qb  do some manual checks since they developed the component back then and probably remember better than me all the workarounds we added 😅  @nadijagraca, I'll create a separate ticket, but we need to change our CSS to match the new DatePicker CSS selectors, but that should be straight-forward hopefully!

Overall, I'm confident we can merge this PR next year though 👍 🚀

AnnMarieW:
Hi @huong-li-nguyen 

Thanks for the review and I look forward to working on this more next year!


> need to change our CSS to match the new DatePicker CSS selectors, but that should be straight-forward hopefully!

As far as styling goes --  it's changed quite substantially starting in V0.14.  No need to go into details now, but a couple things to note:

- Setting the  `theme` prop in `MantinePriverder` will make it so that any custom component can also have the Vizro default themes.  For example setting the `theme={"primarColor: "dark"}`  is very close to the Vizro theme and works well in both light and dark modes.  This will reduce the amount of custom CSS needed.  

- by default the calendar dropdown is rendered in a portal which is outside the dom tree.  This means it can't be targeted with the component's `className` prop.  However, it can be styled using  Mantine's  Style API.   More on this later :-)

vizro-core/src/vizro/models/_components/form/date_picker.py
petar-qb: After some testing with various configuration inputs, I can say that all our previous workaround solutions are unnecessary as `dmc.DatePickerInput` handles them all internally 🎉 Here's the example of our workaround solutions.
- maxDate=self.max + datetime.timedelta(days=1) if self.max else None,
- dropdownPosition="bottom-start",
- clearable=False,
- disabledDates=self.max + datetime.timedelta(days=1) if self.max else None,

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -71,33 +72,15 @@ def build(self):
...
             inputs=inputs,
         )
-        # clientside callback is required as a workaround when the date-picker is overflowing its parent container
-        # if there is not enough space. Caused by another workaround for this issue:
-        # https://github.com/snehilvj/dash-mantine-components/issues/219
-        clientside_callback(
-            ClientsideFunction(namespace="date_picker", function_name="update_date_picker_position"),
-            output=Output(self.id, "dropdownPosition"),
-            inputs=Input(self.id, "n_clicks"),
-        )
```
petar-qb, line 81: Feel free to completely remove `update_date_picker_position` from `date_picker.js`. 🚀

vizro-core/src/vizro/models/_components/form/date_picker.py
petar-qb, line 1: Feel free to remove the following code:

```py
        output = [
            Output(self.id, "value"),
            Output(f"{self.id}_input_store", "data"),
        ]
        inputs = [
            Input(self.id, "value"),
            State(f"{self.id}_input_store", "data"),
        ]
        clientside_callback(
            ClientsideFunction(namespace="date_picker", function_name="update_date_picker_values"),
            output=output,
            inputs=inputs,
        )
```

also (below): `dcc.Store(id=f"{self.id}_input_store", storage_type="session", data=init_value),`

and also: `update_date_picker_values` function from `date_picker.js`.

---

After these removals, persistence will work correctly! 😃

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -171,6 +172,11 @@ def build(self):
                 dash.page_container,
             ],
         )
+        return dmc.MantineProvider(
+            layout,
+            # Use the `theme` to style all Mantine components with a Vizro theme. For more info see https://www.dash-mantine-components.com/components/mantineprovider
+            # theme = {...}
+        )
```
petar-qb, line 179:
```suggestion
return dmc.MantineProvider(
    # Use the `theme` to style all Mantine components with a Vizro theme. For more info see https://www.dash-mantine-components.com/components/mantineprovider theme = {...}
    id="dashboard-container",
    children=[
        html.Div(id="vizro_version", children=vizro.__version__, hidden=True),
        dcc.Store(
            id="vizro_themes",
            data={
                "vizro_dark": pio.templates.merge_templates("vizro_dark", dashboard_overrides),
                "vizro_light": pio.templates.merge_templates("vizro_light", dashboard_overrides),
            },
        ),
        ActionLoop._create_app_callbacks(),
        dash.page_container,
    ],
)
```

petar-qb (approved):
@AnnMarieW, thank you for such great work ❤️ I really like `dmc.DatePickerInput`! 

I left a few comments, but this PR is close to being merged in my opinion 🎉

P.S. You can paste the code from the PR description you wrote directly to `vizro-core/examples/scratch_dev/app.py` file, and make it easier for other reviewers to test the change 😃.

vizro-core/pyproject.toml
```diff
@@ -21,7 +21,7 @@ dependencies = [
   "pandas>=2",
   "plotly>=5.12.0",
   "pydantic>=1.10.16",  # must be synced with pre-commit mypy hook manually
-  "dash_mantine_components<0.13.0",  # 0.13.0 is not compatible with 0.12,
+  "dash_mantine_components==0.15.1",  # 0.13.0 is not compatible with 0.12,
```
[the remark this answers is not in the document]
AnnMarieW: With DMC I think it's best to pin an exact version.  We are not yet at a major release (that's coming soon), plus there will be a breaking change when Dash 3.0 comes out that will require an update to DMC.

Whenever this gets merged, I can help with managing dependencies (at least until we get to a major release)

vizro-core/src/vizro/_vizro.py
```diff
@@ -49,6 +52,13 @@ def __init__(self, **kwargs):
             use_pages=True,
         )
 
+        # Ensure external_stylesheets is a list and append the additional stylesheet
+        external_stylesheets = self.dash.config.external_stylesheets
+        self.dash.config.external_stylesheets = (
+            external_stylesheets if isinstance(external_stylesheets, list) else [external_stylesheets]
+        )
```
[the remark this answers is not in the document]
AnnMarieW: The check is helpful for when a user adds a link or links to the `external_stylesheets`. 

 In DMC there are variable to make it easy to add optional stylesheets.  For example,  `dmc.styles.DATES` is a single URL.  Or you can add all the optional stylesheets with `dmc.styles.ALL`  which is a list of links.   This has cause some confusion for DMC users.

vizro-core/src/vizro/_vizro.py
```diff
@@ -17,6 +18,8 @@
 from vizro.managers import data_manager, model_manager
 from vizro.models import Dashboard, Filter
 
+dash._dash_renderer._set_react_version("18.2.0")
+
```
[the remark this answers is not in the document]
petar-qb: Oke 👍 It would be helpful if you can add a comment to consider removing this line when we increase the dash dependency to `>=3.0.0`?

vizro-core/pyproject.toml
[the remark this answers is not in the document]
petar-qb: Sounds great! Thank you 😃

vizro-core/examples/dev/app.py
```diff
@@ -820,13 +820,13 @@ def multiple_cards(data_frame: pd.DataFrame, n_rows: Optional[int] = 1) -> html.
...
-            className="anchor-container",
-        )
+
+   banner = dbc.NavLink(
+        ["Made with ", html.Img(src=get_asset_url("logo.svg"), id="banner", alt="Vizro logo"), "vizro"],
+        href="https://github.com/mckinsey/vizro",
+        target="_blank",
+        className="anchor-container",
     )
+    app.dash.layout.children = [app.dash.layout.children, banner]
```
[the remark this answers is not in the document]
AnnMarieW: Seems to look ok
image

vizro-core/src/vizro/models/_dashboard.py
AnnMarieW: Would it be OK to wait for the refactor here?  We will likely add more props to the `dmc.MantineProvider` so it might be cleaner to keep the `children` as the variable name `layout` as defined right above the `return` statement.  

 But if you prefer it like your  commit suggestion, I'm ok with that too :-)

petar-qb: Ah I see. Then I'm totally okay to keep layout in a separate variable as you originally did.

vizro-core/src/vizro/static/js/models/date_picker.js
petar-qb, line 1: I love seeing this file completely deleted 🚀

vizro-core/src/vizro/_vizro.py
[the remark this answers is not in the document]
huong-li-nguyen: If you define `Vizro(external_stylesheets=dbc.themes.BOOTSTRAP).build(dashboard).run()`, `external_stylesheets `will also be of type `str`, and this code snippet would otherwise fail. So, contrary to what the Dash docs suggest, it's not always a list that is being returned

vizro-core/src/vizro/models/_components/form/date_picker.py
huong-li-nguyen: Ah, perfect! I didn't know he commented on this already. Let's leave it as is then, and @AnnMarieW you could delete this line then: `date_range_picker_kwargs = {"allowSingleDateInRange": True} if self.range else {}`

I don't think it's needed anymore since that is controlled by the `type` argument now instead of `allowSingleDateInRange`

petar-qb: We should always include `allowSingleDateInRange=True,` in the `dmc.DatePickerInput` configuration. It will enable that a single date can be selected when `range=True` (so `type="range"`), and will have no effect when `range=False` (so `type="default"`).

petar-qb:
@AnnMarieW Can you paste the code from the PR description you wrote, directly to `vizro-core/examples/scratch_dev/app.py` file? This would make it easier for other reviewers to test the change 😃.

vizro-core/src/vizro/_vizro.py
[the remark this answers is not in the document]
petar-qb: You're right! Thanks for checking it 🚀

petar-qb:
Unit tests should not be too hard to fix. There are only three tests that are failing:
1. `test_datepicker_build` (2 - for range=False and range=True) (file: `test_date_picker.py`),
2. `test_dashboard_build` (file: `test_dashboard.py`)

You can fix `test_datepicker_build` by changing the `expected_datepicker` to the new form: `dmc.DatePickerInput(...)`. Similarly, for the `test_dashboard_build`, all you need is to wrap the `expected_dashboard_container` inside the `dmc.MantineProvider(expected_dashboard_container)`. 

You can run unit tests with the following command from your terminal (ensure to call it from `.../vizro/vizro-core` location): `hatch run test-unit`.

If you need any help with this, please let us know and we will be happy to jump in for the help. 😃

vizro-core/src/vizro/models/_components/form/date_picker.py
huong-li-nguyen: Ah I see, but then let's specify it directly as ` allowSingleDateInRange=True` instead of adding that line 👍

AnnMarieW: I'm a little confused by this - not sure what props to update :thinking:

AnnMarieW:
[This commit](https://github.com/mckinsey/vizro/pull/924/commits/cfb1c3bc3bb1283eb66f2718a1480a46a948efd8) is to show how to style Mantine componets with a  Vizro style.  

It uses the `theme` prop in the MantineProvider to make the default accent color gray, which works well with VIzro.  This will be helpful for users adding custom Mantine components in the future.

To fine tune the datepicker, it's best to use Mantine's Style API.  I deleted the datepicker.css file because most of the selectors have changed.  Also, it's not possible to style the calendar with the `className` prop because it's rendered in a portal that is outside of the component root and inner elements are not part of the component tree.   We can change it to `withinPortal=False`, but that can cause some issues with the z-Index.  

The ideal way to style the calendar is to use Mantine `Styles API`.  I included an example of the `styles` prop making all the days the same color to remove the default red for weekend days.

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -57,55 +58,23 @@ def build(self):
...
-        date_picker_class = dmc.DateRangePicker if self.range else dmc.DatePicker
-
-        # dropdownPosition must be set to bottom-start as a workaround for issue:
-        # https://github.com/snehilvj/dash-mantine-components/issues/219
-        # clearable must be set to False as a workaround for issue:
-        # https://github.com/snehilvj/dash-mantine-components/issues/212
-        # maxDate must be increased by one day, and later on disabledDates must be set as maxDate + 1 day
-        # as a workaround for issue: https://github.com/snehilvj/dash-mantine-components/issues/230
-        date_picker = date_picker_class(
+        date_picker = dmc.DatePickerInput(
```
petar-qb, line 61:
```suggestion
        date_picker = dmc.DatePickerInput(
            id=self.id,
            minDate=self.min,
            value=init_value,
            maxDate=self.max,
            persistence=True,
            persistence_type="session",
            type="range" if self.range else "default",
            className="datepicker",    
            # removes the default red color for  weekend days
            styles={"day": {"color": "var(--mantine-color-text"}},
            allowSingleDateInRange: True,
        )
```

vizro-core/src/vizro/models/_components/form/date_picker.py
petar-qb: All that is needed is to delete the following line:
`        date_range_picker_kwargs = {"allowSingleDateInRange": True} if self.range else {}`

and insert `allowSingleDateInRange: True` directly to the `dmc.DatePickerInput` configuration like in this [suggestion comment](https://github.com/mckinsey/vizro/pull/924/files#r1888005520).

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -56,7 +56,7 @@ class DatePicker(VizroBaseModel):
 
     def build(self):
         init_value = self.value or ([self.min, self.max] if self.range else self.min)  # type: ignore[list-item]
-        date_range_picker_kwargs = {"allowSingleDateInRange": True} if self.range else {}
+        date_range_picker_kwargs = {}
```
petar-qb, line 59: You can get rid of `date_range_picker_kwargs` variable as it's irrelevant now.

vizro-core/tests/unit/vizro/models/_components/form/test_date_picker.py
```diff
@@ -117,25 +117,25 @@ def test_datepicker_build(self, range, value):
             min="2023-01-01", max="2023-07-01", range=range, value=value, id="datepicker_id", title="Test title"
         ).build()
 
-        date_picker_class = dmc.DateRangePicker if range else dmc.DatePicker
-        additional_kwargs = {"allowSingleDateInRange": True} if range else {}
+        additional_kwargs = {}
```
petar-qb, line 120: You can get rid of `additional_kwargs`.

petar-qb:
@AnnMarieW thanks for writing the unit tests. They work like a charm. 🎉

vizro-core/src/vizro/models/_components/form/date_picker.py
petar-qb: No, it's not needed anymore. This variable previously had a role to propagate `"allowSingleDateInRange": True` to the `dmc.DateRangePicker` (if self.range is True) and not to propagate it to the `dmc.DatePicker` (if self.range is False). Now, we use the new object called `dmc.DatePickerInput` and regardless of `self.range` value, we propagate `"allowSingleDateInRange": True` to this object. That was the only purpose of the `date_range_picker_kwargs` object. 

So, `date_range_picker_kwargs` wasn't there to enable propagating arbitrary arguments through `vm.DatePicker` to the underlying `dmc.DatePickerInput`. We don't allow our users to directly propagate arbitrary arguments to `vm.DatePicker` because of the maintenance reasons. If our users need to customise some other arguments that are not exposed, then they need to create a custom component.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -171,6 +172,11 @@ def build(self):
                 dash.page_container,
             ],
         )
+        return dmc.MantineProvider(
```
antonymilne, line 175: Do we actually need this? I read through the dmc docs and played around and as far as I can tell the only thing that `dmc.MantineProvider` does is set `<head data-mantine-color-scheme=...>`. You already added a line to set this variable in `update_dashboard_theme`, which always runs when the page is first loaded. Hence unless `dmc.MantineProvider` does something in addition to setting `<head data-mantine-color-scheme=...>` this would seem redundant because it just gets overidden straight away by the `update_dashboard_theme` callback.

antonymilne (commented):
Hello @AnnMarieW, great to see you here again! Sorry for the very slow review - I just back after an extended Christmas break this week. Hope you had a wonderful festive period too. And thanks very much @petar-qb and @huong-li-nguyen for all the reviewing and comments so far.

Thank you so much for your comments in #905 and all your work here. I'm super happy to hear that you're now the primary maintainer of dmc and even happier that you raised a PR to do the version bump for us! Even though we have migrated almost everything on vizro to dbc and only use the date picker from dmc now, as you know I am a huge fan of dmc and hope it goes from strength to strength.

FYI I saw that you were running `hatch run examples:example dev`. Actually to run the `dev` example you just need to do `hatch run example dev` and for the `scratch_dev` example (which we usually use for testing) it's just `hatch run example`. I updated our [contributing guidelines](https://vizro.readthedocs.io/en/stable/pages/explanation/contributing/#hatch-run-example) a couple of months ago so hope they're easier to follow now.
 
---

I think we should make this PR ready to review now. I left a few comments. There's a few other things I think we need to deal with but nothing major. Once we've figured these remaining bits out I'd be very happy to approve and get this merged :shipit: 
* @AnnMarieW  [cfb1c3b](https://github.com/mckinsey/vizro/pull/924/commits/cfb1c3bc3bb1283eb66f2718a1480a46a948efd8) - is this meant to remain or was it just to demonstrate something?
* @AnnMarieW until now it's been possible to run vizro entirely offline, but in this PR you've added a dependency to https://unpkg.com/@mantine/dates@7.14.3/styles.css through `dmc.styles.DATES`. Would it be a bad idea for us to just ship this file with vizro? Is the file likely to change much in future? Because then we would have to manually update our file to match.
* @petar-qb so nice to see all those random hacks gone! This is a HUGE impovement 🎉 As things stand after this PR, does the date picker persistence work the same as our other selectors or still not as good?
* @huong-li-nguyen @nadijagraca let's have a quick chat about the best way to handle the CSS

vizro-core/pyproject.toml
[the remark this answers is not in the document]
petar-qb: Before, I was a huge fan of `~=X.Y`, `~=X.Y.Z`, `<=X` notations, but after reading [this article](https://iscinumpy.dev/post/bound-version-constraints/) I changed my mind a bit. Now, I'm not to strict about the silver bullet solution for this topic so I also give a chance to the `>=X.Y.Z` notation.

Actually, for this case, I prefer this (`>=X.Y.Z`) notation because we are not sure that the `dmc==0.16.0` will break the current `dmc.DatePickerInput` (0.15.0 ). I prefer using `~=`, `==` or `<=` notation only if it's much more likely that the new version will break something. Otherwise, let it update, and if something goes wrong (which doesn't happen as ofter as I though before), we can easily pin the lower version until Vizro become adjusted to the breaking change. Guided by those thoughts, I guess we should leave `dash_maintain_components>=0.15.1`, but also pin the dash version to `dash<3.0.0` (as we already did), because it's more likely that dash==3.0.0 will break something in Vizro .

petar-qb:
@antonymilne 
> @petar-qb so nice to see all those random hacks gone! This is a HUGE impovement 🎉 As things stand after this PR, does the date picker persistence work the same as our other selectors or still not as good?

After @AnnMarieW resolved this https://github.com/mckinsey/vizro/pull/924#discussion_r1884279546,  persistence started to work completely correctly.

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -55,57 +56,24 @@ class DatePicker(VizroBaseModel):
...
             persistence=True,
             persistence_type="session",
-            dropdownPosition="bottom-start",
-            clearable=False,
-            disabledDates=self.max + datetime.timedelta(days=1) if self.max else None,
+            type="range" if self.range else "default",
+            allowSingleDateInRange= True,
             className="datepicker",
-            **date_range_picker_kwargs,
+            # removes the default red color for  weekend days
```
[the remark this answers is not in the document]
AnnMarieW: > Would it be better to do dash_mantine_components~=0.15.1 so we get updates like 0.15.2?

Yes, I think that would be OK

> Is setting this react version likely to cause problems with anything else? I don't know whether some Dash components only function with React v16.14.0?


Plotly has been testing with React 18 for over a year.  There hasn't been any issues that I know of.  I was just reminded that Dash 3.0 will be using React 18.3.1, and in that version there will be a deprecation warning in the browser console for any components that use  the `defaultProps` function.  Both `DMC` and `DBC` and  `Dash AG Grid` are on a path to fix this.  I'm not sure about any other community libraries.   The Dash component libraries, `dcc`, `html` and `DataTable` have been updated.  Pinning it to React 18.2.1 would be good until the other libraries are ready.


> Do we actually want this or was it just to demonstrate something? We don't normally do inline styles like this

It would be better if  the styles are applied in the [theme](https://www.dash-mantine-components.com/theme-object) prop in the `MantineProvider`, then it would also apply to any custom DMC  component that was added. 

I'd be happy to join a call to discuss CSS if you like.  Mantine has a different way to style apps and components than Bootstrap, and I could help highlight the differences and leverage the Mantine system to make it easier to maintain going forward.  In the meantime,  here are a few links that might help:

- [Mantine API Overiew](https://www.dash-mantine-components.com/mantine-api)  - great starting point for understanding styling in DMC>=0.14
- [Forum Post on DMC theme builder](https://community.plotly.com/t/dash-mantine-components-theme-builder/89883)  - When I started playing with this app I had an "ah-ha" moment on the power of the Mantine theme object which sets global style defaults.
- [Styles API](https://www.dash-mantine-components.com/styles-api) - how to set component level styles either globally or on individual components.
- The whole "Theming" category in the docs has been updated recently and has lots of great info.

vizro-core/src/vizro/models/_dashboard.py
AnnMarieW: Yes, the `MantineProvider` does more than just set the theme.  It's required as of DMC 14.0 (Mantine 7)  No Mantine component will render if the `MantineProvider` is not included in the app.  More info in [Mantine API Overiew](https://www.dash-mantine-components.com/mantine-api)

AnnMarieW:
@antonymilne 

Welcome back!  I'm happy to be working on another contribution to Vizro :slightly_smiling_face: 
I've answered all the other questions in-line, I think there is only this one question about styling components with a portal remaining:

> please could you point me to where I can read more about this? 

The best description is in the upstream Mantine docs for the [Portal](https://mantine.dev/core/portal/) component.  This component is not available as a separate component in DMC  because it's used as a building block for other components like Modals, Select, Popover, and the dropdown calendar of date picker components.  

The bit about the z-index, I think I read about in an issue on the Mantine Discord.  I'll look for more info.  It may be that it's ok to use this component with or without the portal.  Either way, the bulletproof way to style this component is with the [styles API.](https://www.dash-mantine-components.com/styles-api)

vizro-core/src/vizro/models/_dashboard.py
huong-li-nguyen: Apart from that, it seems like we will need it anyway. I've added the `font-family` to @nadijagraca CSS PR, as otherwise you would have to set it per component. This seemed to be the cleanest way to apply the global font to all dmc components that potentially be added.

```
dmc.MantineProvider(
            layout,
            # Applies to all Mantine components
            theme={"fontFamily": "Inter, sans-serif, Arial, serif"},
        )
```

vizro-core/src/vizro/models/_components/form/date_picker.py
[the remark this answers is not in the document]
huong-li-nguyen: @nadijagraca has opened a PR here: https://github.com/AnnMarieW/vizro/pull/1
From my side, it has all the visual changes we need at the moment. 

@nadijagraca - could you check the links above and see whether we can refactor our CSS some more? :) 
@AnnMarieW - feel free to take a look at that PR as well 👍

petar-qb:
Hey people @AnnMarieW @antonymilne @huong-li-nguyen @nadijagraca 👋.

After the second review, unfortunately, I found two unexpected behaviours in the new `dmc=0.15.0` for the `type="range" dmc.DatePickerInput`component. Both are described in the [PR](https://github.com/AnnMarieW/vizro/pull/2) description.

For easier visibility, I'm posting this PR's description here too (the first issue, that's solved with the linked PR also happens in the `dmc==0.12.0`):

---

If a user selects one of two values in the `type="range"` - `dmc.DatePickerInput`, and then clicks somewhere else on the page, `[null, null]` will be set in the component. This PR (the linked one) enables that the previously selected value (that was correct e.g. `["2025-01-01", "2025-05-05"]`) overwrites the `[null, null]` when this happens. So the bug is solved.

---

There's another problem that I wasn't able to solve with the javascript code or `dmc.DatePickerInput` configuration. So, when user selects the first value for the `type="range"` - `dmc.DatePickerInput`, all callbacks where this component is a `dcc.Input` will be triggered. Component value in that case is for example `["2025-01-01", null]`. Vizro in that case return an empty `data_frame` if this component is used as a `vm.Filter` (so this partially solves the problem).

So, selecting a complete value for this component means two requests to the server which is a bit odd:
1. The first request when only one value is selected (then an empty data_frame is returned for all its filter targets).
2. The second request when the second value is selected (then it works properly).

This problem (that I wasn't able to solve) happens with the `dmc==0.15.0` but did not happen with the `dmc==0.12.0`.
For the `dmc==0.12.0`, callbacks are triggered only after the second value is selected. 

---

My opinion is that we should merge this [PR](https://github.com/AnnMarieW/vizro/pull/2) fix. 
Also, I don't think that the second issue (sending 2 requests to the server) is something that should stop us for merging this PR ("Update to use dmc 0.15.1"). I described it just in case you @AnnMarieW were not aware of this behaviour. Fixing this on the `dmc` side will automatically solve this unexpected behaviour on the Vizro side too.

AnnMarieW:
To address these issues in this comment https://github.com/mckinsey/vizro/pull/924#issuecomment-2598680702 could try setting the following props:

`debounce=True`

**debounce**	(boolean | number; default False): If True, changes to input will be sent back to the Dash server only on enter or when losing focus. If it's False, it will send the value back on every change. If a number, it will not send anything back to the Dash server until the user has stopped typing for that number of milliseconds.

`allowSingleDateInRange=False`

**allowSingleDateInRange**	Determines whether single year can be selected as range, applicable only when type="range".

petar-qb:
> To address these issues in this comment [#924 (comment)](https://github.com/mckinsey/vizro/pull/924#issuecomment-2598680702) could try setting the following props:
> 
> `debounce=True`
> 
> **debounce** (boolean | number; default False): If True, changes to input will be sent back to the Dash server only on enter or when losing focus. If it's False, it will send the value back on every change. If a number, it will not send anything back to the Dash server until the user has stopped typing for that number of milliseconds.
> 
> `allowSingleDateInRange=False`
> 
> **allowSingleDateInRange** Determines whether single year can be selected as range, applicable only when type="range".

Hi @AnnMarieW, I've already tried this combination of properties and the first request when only one value is selected does not really happen 👍. However, after the second value is selected, the user has to click somewhere else on the screen again to trigger the request. This behaviour seemed like a worse UX practice than sending the request twice, so I left `debounce=False` and `allowSingleDateInRange=True`.

I really think we should merge this [PR](https://github.com/AnnMarieW/vizro/pull/2) that fixes the case when the value is cleared (which can happen if one value is selected and then the user clicks somewhere else on the screen). This partially improves the UX.

Also, as I mentioned in https://github.com/mckinsey/vizro/pull/924#issuecomment-2598680702, I think we should merge this [PR](https://github.com/mckinsey/vizro/pull/924) even though two requests are sent to the server. If in a newer version of `dmc` it stops sending the first request after the first value is selected, we will be happy to update the `dmc` version to a newer version in Vizro. 😃

huong-li-nguyen:
> I really think we should merge this [PR](https://github.com/AnnMarieW/vizro/pull/2) that fixes the case when the value is cleared (which can happen if one value is selected and then the user clicks somewhere else on the screen). This partially improves the UX.
> 
> Also, as I mentioned in [#924 (comment)](https://github.com/mckinsey/vizro/pull/924#issuecomment-2598680702), I think we should merge this [PR](https://github.com/mckinsey/vizro/pull/924) even though two requests are sent to the server. If in a newer version of `dmc` it stops sending the first request after the first value is selected, we will be happy to update the `dmc` version to a newer version in Vizro. 😃

On the first point, I actually don't think we need to fix it. I think the UI is fine as is, but let's just take this offline and discuss with @Joseph-Perkins. Either way, we can add that fix after this PR if we decide to do that 👍 

So for me, this PR is ready to merge @AnnMarieW 👍 Let's merge that together with the CSS PR after @antonymilne final approval.

huong-li-nguyen (approved):

vizro-core/examples/dev/app.py
```diff
@@ -820,13 +820,13 @@ def multiple_cards(data_frame: pd.DataFrame, n_rows: Optional[int] = 1) -> html.
...
 if __name__ == "__main__":
     app = Vizro().build(dashboard)
-    app.dash.layout.children.append(
-        dbc.NavLink(
-            ["Made with ", html.Img(src=get_asset_url("logo.svg"), id="banner", alt="Vizro logo"), "vizro"],
-            href="https://github.com/mckinsey/vizro",
-            target="_blank",
-            className="anchor-container",
-        )
+
```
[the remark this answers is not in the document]
antonymilne: Ah interesting, I hadn't thought of that, thanks for the explanation. Let's try and keep `app.dash.layout.children.append` since it's a (somewhat hacky) trick that at least I have done a few times. I'll push a commit to do this together with a comment about why we do this.

vizro-core/src/vizro/models/_dashboard.py
antonymilne: Got it, thanks both!

vizro-core/pyproject.toml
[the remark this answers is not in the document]
antonymilne: Yep, that article is indeed the philosophy that I generally follow too, which is why vizro has never had upper bounds on dependencies _as a general rule_. This has indeed worked well for us so far.

However, this is exactly one of the cases where it's a good idea to add a cap though (as the article states, "If upstream asks users to cap, then I still don’t like it, but it is okay if you want to follow the upstream recommendation.") so given that @AnnMarieW thinks it is a good idea to pin DMC then we should follow her lead. It's unlikely to limit our cross-compatibility with any other packages.

My only question for @AnnMarieW is whether we should do `dash_mantine_components==0.15.1` (the original suggestion) or `dash_mantine_components~=0.15.1` (i.e. `>=0.15.1, <0.16`) so that we get versions `0.15.2` etc. So this basically depends on whether future versions 0.15.x would just be bug fixes or might somehow break our usage of DMC (which is very minor, just this date picker).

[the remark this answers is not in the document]
antonymilne: Oh, I see @AnnMarieW already answered this question elsewhere that future 0.15.x would be good so let's go for `dash_mantine_components~=0.15.1` then.

```suggestion
  "dash_mantine_components~=0.15.1",
```

vizro-core/examples/dev/app.py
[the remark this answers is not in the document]
antonymilne: Done in https://github.com/AnnMarieW/vizro/pull/3.

antonymilne (approved):

vizro-core/pyproject.toml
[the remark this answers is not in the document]
AnnMarieW: @petar-qb 

Glad you pinned dash < 3 because it will definitely break things.  It breaks dmc, dbc and Dash AG Grid.  I recommend pinning all those libraries too until Dash 3.0 has been sorted out.  

Part of the issue is that there have been no breaking changes in Dash for so many years that I for one have been lax about setting an upper bound.  Many libraries including dmc and dbc set a lower bound for Dash but not an upper bound as well.  This is going to cause a huge headache for a while.  :cry:

PR #929, opened 2024-12-13
[Bug] Enable that components can be nested inside basic type structures too

petar-qb:
## Description
This is bugfix made on top of https://github.com/mckinsey/vizro/pull/903. 

This PR enables that model_manager can find components that are nested arbitrarily deep inside lists, dicts or tuples. 

So, for the user's custom component, `vm.Filter` will be found even though it's nested like this:
`CustomComponent.controls: list[list[dict[str, dict[str, vm.Filter]]]]`

See how `scratch/app.py` works from this, but doesn't work from the main branch. The best way to see the problem is to refresh the page and see if the controls are taken into account when the page loads.

This PR does not address the following limitations, as there is currently no requirement to do so:
1. Controls cannot yet be nested within custom component properties that are not included in the predefined `model_fields: ["components", "tabs", "controls", "actions", "selector"]`.

At present, the model_manager is capable of locating deeply nested models (e.g., vm.Filter) only if they are directly or indirectly (e.g., list[list[dict[str, dict[str, vm.Filter]]]]) assigned to a component property that aligns with one of these predefined fields.

## Notice


maxschulz-COL (commented):
Can we write some tests for this case ;)

petar-qb:
> Can we write some tests for this case ;)

I also had this on my mind, but I've decided not to do so because all these util methods are still private. We don't test private much methods because it can easily happen that their structure or API completely changes. And, model_manager will be changed again (maybe even drastically changed) when we introduce `pydantic_v2` (this is also highlighted in the TODO comments).

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -84,28 +84,25 @@ def _get_models(
     def __get_model_children(self, model: Model) -> Generator[Model, None, None]:
         """Iterates through children of `model`.
 
-        Currently looks only through certain fields so might miss some children models.
+        Currently, this method looks only through certain fields so might miss some children models.
```
huong-li-nguyen, line 87:
```suggestion
        Currently, this method looks only through certain fields so might miss some children models.
```

What does that mean that it looks "through certain fields"? I thought it's independent of "components", "tabs", "controls", "actions", "selector" now. Or does it refer to something else?

huong-li-nguyen:
> > Can we write some tests for this case ;)
> 
> I also had this on my mind, but I've decided not to do so because all these util methods are still private. We don't test private methods much because it can easily happen that their structure or API completely changes. And, model_manager will be changed again (maybe even drastically changed) when we introduce `pydantic_v2` (this is also highlighted in the TODO comments).

I agree with @maxschulz-COL that we should include unit tests though. We do test private methods in other models, but for some we have deprioritized it. Ideally, anything that could be unintentionally broken should be tested. We don't need to test every possible level of nesting, but we should cover basic cases, such as the code snippet mentioned in this issue: https://github.com/mckinsey/vizro/issues/82

Is this scenario included in our tests? If not, would you mind adding it? I remember Antony sent out this issue again, but it broke after the action refactoring, and we only noticed when another user requested a similar feature right? I think optimally unit tests should have caught this earlier 😬

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,31 +1,69 @@
...
-    "autoHeaderHeight": True,
-}
+class CustomGroup(vm.VizroBaseModel):
+    """Container to group controls."""
+
+    type: Literal["custom_group"] = "custom_group"
+    controls: List[Tuple[str, List[ControlType]]] = [[]]
+
+    def build(self):
+        return html.Div(
```
huong-li-nguyen, line 19: Amazing, how this works now! 🥳 🍾

petar-qb:
Due to prioritisation and capacity for the current sprint, I will write unit tests and finalise this PR next year.

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,31 +1,69 @@
...
 
-defaultColDef = {
-    "wrapHeaderText": True,
-    "autoHeaderHeight": True,
-}
+class CustomGroup(vm.VizroBaseModel):
+    """Container to group controls."""
+
+    type: Literal["custom_group"] = "custom_group"
+    controls: List[Tuple[str, List[ControlType]]] = [[]]
```
antonymilne, line 16: FYI my preferred pattern for this would be to nest a list of models, each of which does this (a bit like tabs):

```
    title: str
    controls: list[ControlType] = []
```

Or if you want just one model then do it like this:

```
    controls: dict[str, list[ControlType]] = {}
```

As it stands we made the deliberate choice in vizro to avoid `tuple`. We also haven't used the `dict[str, VizroBaseModel]` pattern anywhere but that might change in future.

antonymilne (approved):
Good catch @petar-qb! I have a couple of suggestions but generally looks good.

Re tests: when I wrote this stuff in `model_manager` I intentionally didn't write tests for exactly the reasons that @petar-qb wrote - it didn't feel like a good use of time to write tests for a private method that will likely change a lot in the (hopefully) near future. Now we have these more complicated cases though and want to avoid regressions I think @huong-li-nguyen is right that it would be nice to have some tests now. Let's not spend ages on them though.

Let me know if you need help writing tests @petar-qb because mostly it's my fault for not writing them in the first place I know... If it's easy to test `_get_models` then let's do that instead of `__get_model_children` because it's higher level, and so the tests will hopefully more easily be kept up to date with future changes.

vizro-core/src/vizro/managers/_model_manager.py
petar-qb: I applied your code suggestion @antonymilne  and now it also works with the arbitrary deep `lists` or `dicts`. 🚀

vizro-core/tests/unit/vizro/conftest.py
```diff
@@ -75,11 +75,6 @@ def ag_grid_with_id(gapminder):
     return dash_ag_grid(id="underlying_ag_grid_id", data_frame=gapminder)
 
 
-@pytest.fixture
-def ag_grid_with_id_and_conf(gapminder):
-    return dash_ag_grid(id="underlying_ag_grid_id", data_frame=gapminder)
-
-
```
petar-qb, line 82: Removed as a duplicate of `ag_grid_with_id` (fixture above). The usage of the deleted fixture is replaced with the `ag_grid_with_id` in the `test_ag_grid.py` accordingly.

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,24 +1,50 @@
...
+from typing import List, Literal, Tuple
+from vizro.models.types import ControlType
+from dash import html
+
+
+class CustomGroup(vm.VizroBaseModel):
+    """Container to group controls."""
+
+    type: Literal["custom_group"] = "custom_group"
+    controls: dict[str, List[ControlType]] = {}
```
petar-qb, line 16: Now, all models can now be nested arbitrarily deep within lists or dicts 😃
However, types like `tuple`, `set`, `frozenset` are not supported.

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -84,28 +84,28 @@ def _get_models(
...
 
         if isinstance(model, VizroBaseModel):
             yield model
+        elif isinstance(model, list):
+            for single_model in model:
+                yield from self.__get_model_children(single_model)
+        elif isinstance(model, dict):
+            # We don't look through keys because Vizro models aren't hashable.
+            for single_model in model.values():
+                yield from self.__get_model_children(single_model)
```
petar-qb, line 100: If we want to support searching through types like `tuple`, `set`.. we can adjust the implementation to the following one. (although I don't think it's necessary to do so now):

```suggestion
        from vizro.models import VizroBaseModel
        from collections.abc import Collection, Mapping

        if isinstance(model, VizroBaseModel):
            yield model
        elif isinstance(model, Collection) and not isinstance(model, str):
            for single_model in model:
                yield from self.__get_model_children(single_model)
        elif isinstance(model, Mapping):
            # We don't look through keys because Vizro models aren't hashable.
            for single_model in model.values():
                yield from self.__get_model_children(single_model)
```

vizro-core/tests/unit/vizro/managers/test_model_manager.py
petar-qb, line 1: Implemented unit tests cover only `_get_models` and `_get_model_page` model_manager methods.
For `model_manager.py`, code coverage is increased: `86% -> 99%`.

vizro-core/tests/unit/vizro/managers/test_model_manager.py
```diff
@@ -0,0 +1,208 @@
...
+            # model nested within a list of lists
+            ([[vm.Filter(id="page_1_control_3", column="year")]], "page_1_control_3"),
+            # model nested in a dictionary of dictionaries
+            ({"key_1": {"key_2": vm.Filter(id="page_1_control_4", column="year")}}, "page_1_control_4"),
+            # model nested in a list of dictionaries
+            ([{"key_1": vm.Filter(id="page_1_control_5", column="year")}], "page_1_control_5"),
+            # model nested in a dictionary of lists
+            ({"key_1": [vm.Filter(id="page_1_control_6", column="year")]}, "page_1_control_6"),
+        ],
+    )
```
petar-qb, line 160: Initially, I tried with the following implementation:
```py
@pytest.mark.parametrize(
        "controls",
        [
            (vm.Filter(id="page_1_control_1", column="year")),
            ([vm.Filter(id="page_1_control_1", column="year")]),
            ...
```

Filters across parameter test cases shared the same ID, which was hardcoded to `page_1_control_1` and checked at the end of the test implementation like `assert "page_1_control_1" in result`.

However, this implementation raised DuplicationID exception from the model_manager even though `reset_manager` was triggered between two parameter test cases which is visible with the following command:
`pytest tests/unit/vizro/managers/test_model_manager.py::TestGetModels::test_nested_models --setup-show`

The current implementation looks a bit ugly to me and I would appreciate any suggestions for improvement.

vizro-core/changelog.d/20241213_161527_petar_pejovic_fix_model_manager_get_model_children.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Enable arbitrary deep nesting of custom components. All models can now be nested arbitrarily deep within lists or dicts. ([#929](https://github.com/mckinsey/vizro/pull/929))
```
maxschulz-COL, line 40:
```suggestion
- Enable arbitrarily deep nesting of custom components within lists or dicts. ([#929](https://github.com/mckinsey/vizro/pull/929))
```

vizro-core/tests/unit/vizro/managers/test_model_manager.py
maxschulz-COL: That is strange no? I didn't run the command myself, but isn't that precisely what out `autouse` fixture should avoid

vizro-core/tests/unit/vizro/managers/test_model_manager.py
maxschulz-COL, line 1: Nice tests - very great to read as documentation as well!

maxschulz-COL (approved):
Really cool work @petar-qb , thanks for doing the extra round of tests, I think they really help documenting what's going on

vizro-core/tests/unit/vizro/managers/test_model_manager.py
```diff
@@ -0,0 +1,208 @@
...
+        result = [model.id for model in model_manager._get_models()]
+
+        expected_present = {
+            "page_1_id",
+            "page_1_button_id",
+            "page_1_graph_id",
+            "page_2_id",
+            "page_2_button_id",
+            "page_2_figure_id",
+        }
```
huong-li-nguyen, line 55: Shouldn't we also include the Filter id in here given the configuration above?

```suggestion
    "page_2_filter"
        }
```

vizro-core/tests/unit/vizro/managers/test_model_manager.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+class TestGetModels:
+    """Test _get_models method."""
+
+    def test_model_type_none_page_none(self):
+        """model_type is None | page is None -> return all elements."""
+        result = [model.id for model in model_manager._get_models()]
+
+        expected_present = {
```
huong-li-nguyen, line 48:
```suggestion
        expected = {
```

vizro-core/tests/unit/vizro/managers/test_model_manager.py
```diff
@@ -0,0 +1,208 @@
...
+            "page_1_id",
+            "page_1_button_id",
+            "page_1_graph_id",
+            "page_2_id",
+            "page_2_button_id",
+            "page_2_figure_id",
+        }
+
+        assert expected_present.issubset(result)
```
huong-li-nguyen, line 57: Is there a reason why we check for the subset instead of equality? What else is inside result? 🤔 
```suggestion
        assert expected.issubset(result)
```

vizro-core/tests/unit/vizro/managers/test_model_manager.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def test_model_type_page_none(self):
+        """model_type is vm.Button | page is None -> return all vm.Button from the dashboard."""
+        result = [model.id for model in model_manager._get_models(model_type=vm.Button)]
+
+        expected_present = {"page_1_button_id", "page_2_button_id"}
+        expected_absent = {"page_1_id", "page_1_graph_id", "page_2_id", "page_2_figure_id"}
```
huong-li-nguyen, line 64: Maybe just call it expected and excluded? Only if you also prefer that. Else just leave as is :)
```suggestion
        expected = {"page_1_button_id", "page_2_button_id"}
        excluded = {"page_1_id", "page_1_graph_id", "page_2_id", "page_2_figure_id"}
```

huong-li-nguyen (approved):
Amazing job! 🚀  Really great to read through and understand 👍

vizro-core/tests/unit/vizro/managers/test_model_manager.py
petar-qb: >  What else is inside result?

Layout, Navigation and Dashboard objects are also returned. The real reason why I haven't checked the equality is that we can now extend the `managers_dashboard_two_pages` Dashboard without a need to change all tests.

vizro-core/tests/unit/vizro/managers/test_model_manager.py
petar-qb: Great catch!
I added "page_2_filter" early in the process while I was doing the tests and forgot to remove it 😄. It doesn't contribute to the quality of the tests so I'll just remove it.

vizro-core/src/vizro/managers/_model_manager.py
antonymilne: Nice, it looks good! I'm tempted to make this change right away. I guess we might as well because I don't think it will cause any problems? But this model manager code will change in the nearish future anyway, and there will hopefully be a better way to iterate through composite model field types (list, tuple, etc.). So up to you.

vizro-core/changelog.d/20241213_161527_petar_pejovic_fix_model_manager_get_model_children.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Enable arbitrarily deep nesting of custom components within lists or dicts. ([#929](https://github.com/mckinsey/vizro/pull/929))
```
antonymilne, line 40:
```suggestion
- Enable arbitrarily deep nesting of custom components within lists or dictionaries. ([#929](https://github.com/mckinsey/vizro/pull/929))
```

vizro-core/tests/unit/vizro/managers/test_model_manager.py
petar-qb: Solved with the following implementation:

```py
@pytest.mark.parametrize(
        "get_nested_control",
        [
            # model as a property of a custom model
            (lambda x: x),
            # model inside a list
            (lambda x: [x]),
            # model as a value in a dictionary
            (lambda x: {"key": x}),
            # model nested within a list of lists
            (lambda x: [[x]]),
            # model nested in a dictionary of dictionaries
            (lambda x: {"key_1": {"key_2": x}}),
            # model nested in a list of dictionaries
            (lambda x: [{"key": x}]),
            # model nested in a dictionary of lists
            (lambda x: {"key": [x]}),
        ],
    )
    def test_nested_models(self, page_1, get_nested_control):
        """Model is nested under another model and known property in different ways -> return the model."""

        class ControlGroup(vm.VizroBaseModel):
            controls: Any

        page_1.controls.append(
            ControlGroup(controls=get_nested_control(vm.Filter(id="page_1_control_1", column="year")))
        )

        result = [model.id for model in model_manager._get_models(model_type=vm.Filter, page=page_1)]

        assert "page_1_control_1" in result
```

vizro-core/tests/unit/vizro/managers/test_model_manager.py
```diff
@@ -0,0 +1,209 @@
...
+                components=[
+                    vm.Button(id="page_2_button_id"),
+                    vm.Figure(id="page_2_figure_id", figure=standard_kpi_card),
+                ],
+            ),
+        ]
+    )
+
+
+pytestmark = pytest.mark.usefixtures("managers_dashboard_two_pages")
```
antonymilne, line 37: Why not just `autouse=True` the fixture instead? That will do the same. I would only use this alternative notation that you have if the fixture had been defined in a different file without `autouse=True`.

antonymilne (approved):
Really nice work! Just a few small comments. Thanks for writing all those tests that I had completely neglected to write before. They look perfect 👍

antonymilne:
Thanks for updating the PR description also ⭐ It's great to have an updated accurate description of the changes here 💯 I really appreciate the care and time taken to do that as it would have been easy to ignore or forget!

vizro-core/src/vizro/managers/_model_manager.py
petar-qb: Did it -> https://github.com/mckinsey/vizro/pull/929/commits/4f5e717f7cf8fb9f82e1de81d63e49af73c5aee9 😄 

I added tests for `tuple`, and as vm.Filter is not hashable I wasn't able to add tests for `sets`.
Also, I removed the following test cases as I think are irrelevant: (they pass but I think are irrelevant)
```
            # model nested in a list of dictionaries
            (lambda x: [{"key": x}]),
            # model nested in a dictionary of lists
            (lambda x: {"key": [x]}),
```

PR #931, opened 2024-12-16
[Tidy] Remove redundant CSS and add updated bootstrap file

huong-li-nguyen:
## Description
- Remove `collapse.css` (redundant since we load in the bootstrap theme now)
- Simplify CSS for `table.css` (noticed a lot of complex selectors, I just simplified it)
- Move some CSS for `Card` from bootstrap-theme to `static` folder
- Load in updated bootstrap file based on merged PR: https://github.com/McK-Private/vizro-bootstrap/pull/44
- Updated screenshots in https://github.com/mckinsey/vizro-qa/pull/141 due to:
    - Token for `text-secondary` has changed
    - Removed bottom padding for Dash DataTable
    - Apply `text-primaryHover` color for nav links
## Notice


vizro-core/src/vizro/static/css/collapse.css
```diff
@@ -1,37 +0,0 @@
-@media (prefers-reduced-motion: reduce) {
```
huong-li-nguyen, line 1: I think we've added this back then, because there was no bootstrap theme loaded by default, right? @nadijagraca  All of this shouldn't be required anymore given that there is a bootstrap CSS file loaded now. I checked the example apps and it seems to work fine 👍

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
```diff
@@ -29,6 +29,11 @@ All the HEX values starting with --text-code are taken from the Github code high
 }
 
 /* CARDS */
+.card {
+  height: 100%;
+  width: 100%;
+}
```
huong-li-nguyen, line 35: This was previously defined inside our vizro-bootstrap theme, but I took it out now and decided that it should live inside vizro as an overwrite. Bootstrap cards resize according to their content, so I think we shouldn't change that default behaviour. However, inside vizro, we can set that as a default.

petar-qb: Is there a straightforward way to make the `kpi_card` content dynamically collapse/expand when the app is zoomed in or out? @maxschulz-COL pointed out this issue a few days ago. Specifically, when the app is zoomed out (e.g. to 80%), most of the app's content scales accordingly, but the `kpi_card` content does not adjust as expected.

I tested this with the first kpi_card example from our [documentation](https://vizro.readthedocs.io/en/stable/pages/user-guides/figure/), and the issue persists on this branch as well.

This is not an urgent issue, but if it's an easy enhancement to implement, we should consider addressing it.

vizro-core/src/vizro/static/css/collapse.css
petar-qb: Just to let you know that there is no longer a smooth collapsing transition when the control panel is collapsed/expanded.

vizro-core/src/vizro/static/css/table.css
```diff
@@ -1,93 +1,38 @@
-#right-side .dash-cell {
-  height: 48px !important;
+#dashboard-container .dash-cell {
```
petar-qb, line 1: Why is every selector in this file prefixed with `#dashboard-container`, and we don't do the same in other static/css files? Is the long-term goal to standardise by prefixing all our static/css selectors with #dashboard-container?

vizro-core/src/vizro/static/css/table.css
petar-qb, line 1: Great simplification! I love it 🚀

maxschulz-COL: Wow!

maxschulz-COL (approved):
⭐ 🧹

vizro-core/src/vizro/static/css/table.css
huong-li-nguyen: We do it in other files as well such as the dropdown and aggrid. However, not every file needs this. It's only needed when we have specificity issues with the `dash` components. So you need to have this prefix, such that our CSS is more specific and therefore has precedence :)

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
huong-li-nguyen: In general, we don't optimise for zoom in/zoom out views. Our apps should work well for 100% full page views (which can definitely be improved). There are several reasons for this, but the most important one is font hierarchy 😬 D. has set fixed sizes for fonts and spacings and all our components are scaled according to these. As soon as you zoom in/out, you change the ratio of these, so we don't guarantee that it looks good on all scaled views. 

However, the KPI Card is a separate issue for me. We can definitely tackle that, but I wouldn't do it in this PR as it's not really related. @maxschulz-COL - could you create a separate issue for this and explain what your expected behaviour is? In general the reason why the KPI card doesn't react to your zoom in / zoom out is this line   `font-size: 3.6vh;`. The font-size is dynamic depending on the spacing it gets. You can also hardcode the value e.g. say `font-size: 16px` and then it would change based on the zoom in / zoom out, but currently I am not sure what would look well on **all bootstrap themes**. We would have to try things out 👍

vizro-core/src/vizro/static/css/collapse.css
huong-li-nguyen: Yeah, I noticed, but I didn't think it was that bad for it to need custom static CSS. I rather live with the default collapse behaviour from Bootstrap and have no static CSS 😬

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
maxschulz-COL: Top, I'll write the issue

vizro-core/src/vizro/static/css/table.css
```diff
@@ -1,93 +1,38 @@
...
 
-#right-side
-  .dash-table-container
-  .dash-spreadsheet-container
-  .dash-spreadsheet-inner
-  .column-header-name {
-  font-size: 14px;
+/* Styling for pagination element */
+#dashboard-container .first-page,
+#dashboard-container .previous-page,
```
nadijagraca, line 20: I'm assuming you are repeating `#dashboard-container` for higher specificity? I'm just curious would we achieve the same result without the repetition?

nadijagraca (approved):
Really great work! I love seeing all those selectors removed.  🧹 ✨ 🚀

vizro-core/src/vizro/static/css/collapse.css
petar-qb: @maxschulz-COL @nadijagraca Do you have any thoughts on this? I think we should keep the collapse transition as done in the main branch, bug I'm okay to remove it if you agree.

vizro-core/src/vizro/static/css/table.css
huong-li-nguyen: It would be fine without it. However, if you take a look at the console in the current scratch dev example, you'll see that one dash selector actually has a higher specificity. So if we remove the `dashboard-container` id, the only reason this works is because the current dash selector does not have these properties defined. In this case, it's just safer to add it, so our selector always has higher specificity. 

Screenshot 2024-12-17 at 13 42 08

PR #960, opened 2025-01-20
[Feat] Update to use dmc 0.15.1 Part 2

antonymilne:
## Description

Commits that should have been part of #924 but weren't for technical reasons... The most important of these is the first, which ensures that you can still use vizro offline.

FYI @AnnMarieW 

[Move mantine_dates.css into repo; remove external_stylesheet; delete datepicker.css](https://github.com/AnnMarieW/vizro/commit/7205e850b763620d5c0fd285ba2aba159839b29c) 
[Make layout children a list](https://github.com/AnnMarieW/vizro/commit/0e5c86b4cd48ab4eef71341782525e207017e551)
[Add changelog](https://github.com/AnnMarieW/vizro/commit/ed7c4ea929158c8391fb6da25e64f9df329ccb48)

## Notice


huong-li-nguyen (approved):

vizro-core/src/vizro/static/css/mantine_dates.css
```diff
@@ -0,0 +1,472 @@
+/* Downloaded version of https://unpkg.com/@mantine/dates@7.14.3/styles.css.
+  This is the stylesheet that corresponds to dmc.styles.DATES given at
+  https://github.com/snehilvj/dash-mantine-components/blob/master/dash_mantine_components/styles.py.
+  This ensures vizro can be used offline. */
+.m_468e7eda {
```
huong-li-nguyen, line 5: AM actually mentioned that selectors such as `m_468e7eda` might change. Given that we pull this file manually and put it here, I guess we just need to remember to update the file whenever we update the mantine version as well.

antonymilne: Yeah, I thought the weird selector names here were probably not stable... Let's make sure we update this file whenever we bump our dmc requirement.

vizro-core/src/vizro/_vizro.py
```diff
@@ -53,13 +52,6 @@ def __init__(self, **kwargs):
             use_pages=True,
         )
 
-        # Ensure external_stylesheets is a list and append the additional stylesheet
-        external_stylesheets = self.dash.config.external_stylesheets
-        self.dash.config.external_stylesheets = (
-            external_stylesheets if isinstance(external_stylesheets, list) else [external_stylesheets]
-        )
-        self.dash.config.external_stylesheets.append(dmc.styles.DATES)
-
```
huong-li-nguyen, line 62: Line 56-60 should **not** be removed, as otherwise people cannot add a bootstrap theme for example. 
Also note that we need to do the type checking here. You can test it out by just doing `Vizro(external_stylesheets=dbc.themes.BOOTSTRAP, ....)`. If we don't do the check on the list, that for example wouldn't work anymore.

So only remove  `self.dash.config.external_stylesheets.append(dmc.styles.DATES)`

antonymilne: 🤔 I don't understand. We're not disallowing `Vizro(external_stylesheets=...)` either as string or list by removing these lines. L56-60 were only added in the last PR in order to enable L61, so if we don't need L61 then why would we need the preceding ones now when we didn't before?

huong-li-nguyen: Ahh yeees, I see 😄 I was assuming that this was also not working before `Vizro(external_stylesheets=dbc.themes.BOOTSTRAP, ....`, but it only didn't work because of L61.

Then all good!

PR #961, opened 2025-01-20
[Tidy] Add new style for dropdown checklist components

nadijagraca:
## Description
Adding new styles for checklist and dropdown components. 
New functionality will be implemented in the next PR. 
## Notice


antonymilne:
Not sure if this is ready to review yet but I was excited to took a quick peek already at the example app (didn't look at any code yet)...

This looks awesome, I'm actually super excited by this change even though it seems kind of minor 😀 

The dropdown looks perfect to me. The checklist looks perfect by itself but not sure how well the horizontal line works if you have several controls and navigation there since it's the same separator used to separate the navigation panel out of the controls and used at the bottom of the controls. So it feels like it's breaking up the controls into sections, which it shouldn't. Somehow we maybe need it to be a weaker separator. Maybe a lighter colour or not full width? Or a smaller gap above and below the horizontal line? Or maybe the horizontal line just doesn't work well here and we need some alternative. Would it look weird to put it next to the label "Continent"?
image

I'm not sure what the best way to do it is, happy to defer to you guys and anyone else who has a better idea than me here. Great work anyway!

petar-qb:
I really like the new `dropdown` UI @nadijagraca 🎨 

However, I agree with the @antonymilne https://github.com/mckinsey/vizro/pull/961#issuecomment-2602904677 above. Horizontal line in the checkbox looks a bit confusing especially if stacked with other selectors.

Here's the comparison between the current and the new version. For the new version I changed the `checkbox.css` a bit:
```css
/*
Changes:
- alpha: 0.1 -> 0.05
- margin: 12px -> 8px
- padding 12px -> 8px 
 */
.form-check:nth-child(1) {
  border-bottom: 1px solid rgba(255,255,255,0.05);
  margin-bottom: 8px;
  padding-bottom: 8px;
}
``` 

| Current version            |  New version |
:-------------------------:|:-------------------------:
image | image

In my opinion, the new version looks a little more comfortable because it's cleaner where one selector ends and another begins. However, I don't want to push my opinion here and I'm fine with anything you guys agree on.

huong-li-nguyen:
> In my opinion, the new version looks a little more comfortable because it's cleaner where one selector ends and another begins. However, I don't want to push my opinion here and I'm fine with anything you guys agree on.

I definitely like @petar-qb's suggestion of reducing the gap to 8px and lowering the alpha value 👍. If we want to keep the horizontal divider, that's probably our best option. However, I don't think it will completely solve the issue because the contrast between these dividers and the others in the controls isn't strong enough (even with lowering the alpha value). Increasing the contrast would only work if we use `--border-subtleAlpha02/03` for the other dividers, but then they become too prominent.

But do we even need the horizontal divider? If the goal is to visually distinguish the ALL option, we could simply bold it. I'm not sure how that would look in practice, but we could bold it and maybe even add an icon next to it. For me, just bolding it might be enough though.

Screenshot 2025-01-21 at 10 16 12

The Google Material Icon even has an icon for selecting all 😄 
Screenshot 2025-01-21 at 10 16 52

vizro-core/src/vizro/static/css/dropdown.css
```diff
@@ -151,3 +151,9 @@ wrapper **/
   gap: 4px;
   padding: 4px 8px;
 }
+
+#dashboard-container .VirtualizedSelectOption:nth-child(1):has(> div) {
```
huong-li-nguyen, line 155: Same comment here - is .VirtualizedSelectOption only used inside the Dropdown? 🤔 If yes, then I think as you have it is fine, but if it's potentially used somewhere else as well, it's probably safer to add a className to our dropdown and affect this className only if used inside the dropdown

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -39,6 +39,34 @@ def _calculate_option_height(full_options: OptionsType) -> int:
     return 8 + 24 * number_of_lines
 
 
+def _add_select_all_option(full_options: OptionsType, multi: bool):
+    """Adds div component for select all option."""
+    if not multi:
+        return full_options
```
huong-li-nguyen, line 45:
```suggestion
def _add_select_all_option(full_options: OptionsType):
    """Adds div component for select all option."""
```

We could think about taking this out of the function, such that this function really only has the responsibility of adding the ALL option. 

Then you would need to call the function later on with:

` altered_options = _add_select_all_option(full_options=full_options) if self.multi else full_options`

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -39,6 +39,34 @@ def _calculate_option_height(full_options: OptionsType) -> int:
...
+                altered_options.append(create_select_all_option())
+            else:
+                altered_options.append({"label": option, "value": option})
+        elif isinstance(option, dict):
+            if option.get("value") == "ALL":
+                altered_options.append(create_select_all_option())
+            else:
+                altered_options.append(option)
+
+    return altered_options
```
huong-li-nguyen, line 67: This is quite similar to the code we wrote for the Navigations, where we also had to handle both dictionary and string cases. I really liked our approach back then of first converting everything to a uniform type before applying the logic. I think it made it overall look a bit cleaner - WDYT? :)

Please check below:

```suggestion
def _convert_to_dict(option: Union[str, dict[str, any]]) -> dict[str, any]:
    """Converts a string option to a dictionary format."""
    if isinstance(option, str):
        return {"label": option, "value": option}
    return option

def _add_select_all_option(full_options: OptionsType) -> OptionsType:
    """Adds a 'Select All' option to the list of options."""
    options_dict = [_convert_to_dict(option) for option in full_options]
    select_all_option = {"label": html.Div(["ALL"]), "value": "ALL"}
    return [select_all_option if option.get("value") == "ALL" else option for option in options_dict]
```

huong-li-nguyen (commented):
This looks great! 🚀  Love the visuals. 

Have a few comments and maybe we can iterate on the design with Joe once :)

vizro-core/src/vizro/models/_components/form/dropdown.py
petar-qb: This suggestion is pretty much similar to what I added here -> https://github.com/mckinsey/vizro/pull/961#discussion_r1923356781

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -89,13 +117,14 @@ def validate_multi(cls, multi, values):
     def __call__(self, options):
         full_options, default_value = get_options_and_default(options=options, multi=self.multi)
         option_height = _calculate_option_height(full_options)
+        altered_options = _add_select_all_option(full_options=full_options, multi=self.multi)
```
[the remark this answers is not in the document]
huong-li-nguyen: Ahhh, I actually just suggested something similar hahaha 😄 

I definitely like it! It's something we do for the Navigation code as well, so transforming to a uniform type and then handling that type only. If it's easy to implement, then let's go for it 👍 If it requires lots of changes and you guys wanted to get this PR in first, we can also defer the refactoring to sometime later. Really up to you guys 👍

vizro-core/src/vizro/static/css/dropdown.css
```diff
@@ -151,3 +151,9 @@ wrapper **/
...
+  border-bottom: 1px solid var(--border-subtleAlpha01);
```
huong-li-nguyen, line 156: If it's easy to do - could we add a little 4px gap between the dropdown field and menu (as in our original designs)? 

Screenshot 2025-01-21 at 11 00 50

vizro-core/changelog.d/20250120_173533_nadija_ratkusic_graca_update_dropdown_checklist_css.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
```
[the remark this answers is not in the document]
nadijagraca: My understanding is that we don't mention styling changes here, but if you think it important I can add it.

nadijagraca:
> > In my opinion, the new version looks a little more comfortable because it's cleaner where one selector ends and another begins. However, I don't want to push my opinion here and I'm fine with anything you guys agree on.
> 
> I definitely like @petar-qb's suggestion of reducing the gap to 8px and lowering the alpha value 👍. If we want to keep the horizontal divider, that's probably our best option. However, I don't think it will completely solve the issue because the contrast between these dividers and the others in the controls isn't strong enough (even with lowering the alpha value). Increasing the contrast would only work if we use `--border-subtleAlpha02/03` for the other dividers, but then they become too prominent.
> 
> But do we even need the horizontal divider? If the goal is to visually distinguish the ALL option, we could simply bold it. I'm not sure how that would look in practice, but we could bold it and maybe even add an icon next to it. For me, just bolding it might be enough though.
> 
> Screenshot 2025-01-21 at 10 16 12
> 
> The Google Material Icon even has an icon for selecting all 😄 Screenshot 2025-01-21 at 10 16 52

The designs were taken from qb-design-library for both [checklist](http://qb-design-library.s3-website.eu-west-2.amazonaws.com/?path=/docs/components-checkboxgroup--docs) and [dropdown](http://qb-design-library.s3-website.eu-west-2.amazonaws.com/?path=/docs/components-selectinput--docs). I did not implement all the design for dropdown as some would require significant changes, and some are not feasible.

vizro-core/changelog.d/20250120_173533_nadija_ratkusic_graca_update_dropdown_checklist_css.md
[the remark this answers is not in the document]
petar-qb: To be honest I forgot when do we mention styling changes (if ever) and when we don't. @huong-li-nguyen will know the answer 😄.

vizro-core/src/vizro/models/_components/form/dropdown.py
nadijagraca: I'll take this approach as it does not require changing tests for other selectors that use `get_options_and_default`, but we can refactor in the next iteration when adding functionality if this is preferred approach.

huong-li-nguyen:
> The designs were taken from qb-design-library for both [checklist](http://qb-design-library.s3-website.eu-west-2.amazonaws.com/?path=/docs/components-checkboxgroup--docs) and [dropdown](http://qb-design-library.s3-website.eu-west-2.amazonaws.com/?path=/docs/components-selectinput--docs). I did not implement all the design for dropdown as some would require significant changes, and some are not feasible.

Yes, it's definitely not a mistake on your part! This is one of those instances where it makes sense to deviate from the QB design library. The components in the library were primarily designed as standalones, rather than in conjunction with the control panel and its dividers. So I think, it makes sense for us to deviate a bit.

As for the dropdown, I think it looks fantastic as it is! There's no need to make any changes, especially since most of them aren't feasible as you mentioned. The only thing that would be nice to add is the gap, and hopefully, that won't be too difficult to implement! 🤞

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -39,32 +39,18 @@ def _calculate_option_height(full_options: OptionsType) -> int:
...
 
 
-def _add_select_all_option(full_options: OptionsType, multi: bool):
-    """Adds div component for select all option."""
-    if not multi:
-        return full_options
+def _convert_to_dict(option: Union[str, dict[str, str]]) -> dict[str, Union[str, html.Div]]:
+    """Converts a string option to a dictionary format."""
+    if isinstance(option, str):
+        return {"label": option, "value": option}
```
petar-qb, line 45: Does it work for integers, booleans, None? What happens if you set this control: `vm.Filter(column="year", selector=vm.Dropdown())`?

If it doesn't work, try with similar implementation that's suggested here (by just moving useful parts of the `get_options_and_default` to the `_add_select_all_option `) -> https://github.com/mckinsey/vizro/pull/961#discussion_r1923356781

vizro-core/changelog.d/20250120_173533_nadija_ratkusic_graca_update_dropdown_checklist_css.md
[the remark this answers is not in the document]
antonymilne: I wouldn't bother mentioning it since it's a pretty small style change. Let's save the changelog for when we change the actual "Select All" functionality.

huong-li-nguyen (approved):
Got my pre-approval pending decision from @Joseph-Perkins 👍

vizro-core/src/vizro/models/_components/form/_form_utils.py
```diff
@@ -7,7 +7,9 @@
 from vizro.models.types import MultiValueType, OptionsType, SingleValueType
 
 
-def get_options_and_default(options: OptionsType, multi: bool = False):
+def get_options_and_default(
+    options: OptionsType, multi: bool = False
+) -> tuple[OptionsType, Union[SingleValueType, MultiValueType]]:
```
petar-qb, line 12:
```suggestion
) -> tuple[OptionsType, SingleValueType]:
```

petar-qb, line 12: Sorry for the wrong suggestion earlier, but I think that this should be the return type. Let's try and see is there any `mypy` errors.
```suggestion
) -> tuple[OptionsType, SingleValueType]:
```

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -39,6 +39,17 @@ def _calculate_option_height(full_options: OptionsType) -> int:
     return 8 + 24 * number_of_lines
 
 
+def _add_select_all_option(full_options: OptionsType) -> OptionsType:
+    """Adds a 'Select All' option to the list of options."""
+    options_dict = [
+        cast(OptionsDictType, {"label": option, "value": option}) if not isinstance(option, dict) else option
+        for option in full_options
+    ]
```
petar-qb, line 47: Let's add this TODO and then resolve the [conversation](https://github.com/mckinsey/vizro/pull/961#discussion_r1923356781). 
```suggestion
        # TODO: Move option to dictionary conversion within `get_options_and_default` function as here: https://github.com/mckinsey/vizro/pull/961#discussion_r1923356781 
        options_dict = [
        cast(OptionsDictType, {"label": option, "value": option}) if not isinstance(option, dict) else option
        for option in full_options
    ]
```

petar-qb (approved):
Awesome work @nadijagraca! 🎨 🎉

huong-li-nguyen (approved):
@l0uden @nadijagraca - better to wait for the update of screenshots after this PR has been merged: https://github.com/mckinsey/vizro/pull/963

Otherwise we would have to redo it again 😬 

Pre-approval pending Joe's decision

vizro-core/src/vizro/static/css/dropdown.css
nadijagraca: I was not able to make this work.

huong-li-nguyen: I'll push the change :)

PR #967, opened 2025-01-23
[Tidy] Refactor Vizro chart template generation

huong-li-nguyen:
## Description

- Add script to read in bootstrap variables and automatically create chart templates in json format
- Remove unit tests for theme and replace with schema check whether Plotly figure templates are up-to-date
- Remove `template_dark.py` and `template_light.py`
- Output `vizro_dark.json` and `vizro_light.json`
- Tidy and remove color variables

Visually, nothing has changed. Under the hood, a few things have changed that **make our Plotly template generation a lot more stable**:

- The Plotly templates are now generated via a script that pulls the variables automatically from the bootstrap theme. This ensures that it's always up-to-date with new token values.
- Added a CI step to check for changes in the Plotly template (this is a lot stronger than the unit tests we had for the theme previously, as it checks for the actual json schema)
- Previously, we generated the themes on the run, so everytime a dashboard was run. This was quite inefficient actually, given that our Plotly templates do not change. Now, the Plotly templates are static and stored inside `vizro_dark.json` and `vizro_light.json` and these will be read in at run-time. To update them, one would have to run `hatch run templates`
- These `vizro_light.json` and `vizro_dark.json` will also be re-used for better integration of our bootstrap theme as soon as it's published (more on that later)

## Notice


vizro-core/src/vizro/_themes/common_values.py
```diff
@@ -5,6 +5,9 @@
 from vizro._themes._color_values import COLORS
 
 
+# LQ: I kind of like that this is all in a separate file. Makes it easier to read for me.
```
huong-li-nguyen, line 8: @antonymilne - I left this untouched for now, but wondering whether it should stay here or instead be included inside the json generation as well, such that we have one place. I do like that it's separate though, because it clearly shows what's common in both themes, while the rest is what's different in both themes.

vizro-core/src/vizro/_themes/create_chart_templates.py
```diff
@@ -0,0 +1,120 @@
...
+
+THEMES_FOLDER = Path(__file__).resolve().parent
+CSS_FILE = THEMES_FOLDER.parent / "static" / "css" / "vizro-bootstrap.min.css"
+VARIABLES = ["--bs-primary", "--bs-secondary", "--bs-tertiary-color", "--bs-border-color", "--bs-body-bg"]
+
+
+def _extract_last_two_occurrences(variable: str, content):
+    """Extracts the last two occurrences of a variable from the CSS content.
+
+    Within the `vizro-bootstrap.min.css` file, variables appear multiple times: initially from the default Bootstrap
```
huong-li-nguyen, line 19: As discussed, I know this seems a bit random looking for the last 2 occurences, but I couldn't find a more stable way that seems to work. Looking under :root, [data-bs-theme=dark] gets really messy and doesn't really work. Also it will for sure always be the last two occurrences due to specificity.

vizro-core/src/vizro/_themes/vizro_dark.json
```diff
@@ -0,0 +1,278 @@
+{
```
huong-li-nguyen, line 1: No need to review this - this is automatically generated

vizro-core/src/vizro/_themes/vizro_light.json
```diff
@@ -0,0 +1,278 @@
+{
```
huong-li-nguyen, line 1: No need to review this - this is automatically generated

vizro-core/src/vizro/_themes/create_chart_templates.py
```diff
@@ -0,0 +1,120 @@
...
+    """Generate the Python file content from the extracted values."""
+    FONT_COLOR_PRIMARY = extracted_values["BS-PRIMARY"]
+    BG_COLOR = extracted_values["BS-BODY-BG"]
+    FONT_COLOR_SECONDARY = extracted_values["BS-SECONDARY"]
+    GRID_COLOR = extracted_values["BS-BORDER-COLOR"]
+    AXIS_COLOR = extracted_values["BS-TERTIARY-COLOR"]
+
+    # Apply common values
+    template = create_template_common()
+    template.layout.font.color = FONT_COLOR_PRIMARY
```
huong-li-nguyen, line 58: @antonymilne - we discussed transforming this into a dict format at some point, and during the dev process I did try it out, but I didn't like it 😬 It seemed less readable for me, so I kept it as is, but let me know what you think

vizro-core/tests/unit/vizro/themes/test_theme.py
```diff
@@ -1,42 +0,0 @@
-"""Tests for templates."""
```
huong-li-nguyen, line 1: This will be replaced with a json schema check

huong-li-nguyen, line 1: Not sure what's better - removing these unit tests, that don't really test anything + lowering test coverage or leaving test coverage and these tests untouched 😄

huong-li-nguyen, line 1: Removed these tests and lowered test-coverage because of that. These tests didn't really test anything and the schema test + screenshot tests cover more than whatever was in here.

vizro-core/hatch.toml
```diff
@@ -54,6 +54,8 @@ pypath = "python -c 'import sys; print(sys.executable)'"
 # Only run pre-commit hooks when schema is generated, not when it's checked. This keeps the check fast in CI.
 schema = ["python schemas/generate.py", '- pre-commit run --files="schemas/$(hatch version).json" > /dev/null']
 schema-check = ["python schemas/generate.py --check"]
+templates = ["python src/vizro/_themes/create_chart_templates.py"]
```
antonymilne, line 57: This might need a linting step to run after it just like in the `schema` job if the json files generated by your script need linting.

Alternatively if they don't need linting then we can remove the `templates-check` script entirely and just do it as `hatch run templates --check` and set `templates = "python ...{args}`. This would have been how I did the `schema` script - the only reason it's split into two hatch scripts is because the `templates` one needs linting whereas the `templates-check` one doesn't (and doing the unnecessary linting step is actually quite slow in CI).

vizro-core/src/vizro/__init__.py
```diff
@@ -1,16 +1,19 @@
...
 
 from ._constants import VIZRO_ASSETS_PATH
-from ._themes import dark, light
 from ._vizro import Vizro, _make_resource_spec
 
 logging.basicConfig(level=os.getenv("VIZRO_LOG_LEVEL", "WARNING"))
-pio.templates["vizro_dark"] = dark
-pio.templates["vizro_light"] = light
+
+base_path = Path(__file__).resolve().parent / "_themes"
```
antonymilne, line 14:
```suggestion
base_path = Path(__file__).parent / "_themes"
```
I don't think `resolve` should be needed here.

vizro-core/src/vizro/_themes/__init__.py
```diff
@@ -1,6 +1 @@
 """Init."""
```
antonymilne, line 1:
```suggestion

```
This is not a useful docstring 😅

vizro-core/src/vizro/_themes/create_chart_templates.py
antonymilne, line 1: Very minor: I'd call this `create_plotly_templates.py`.

vizro-core/src/vizro/_themes/create_chart_templates.py
```diff
@@ -0,0 +1,137 @@
...
+from pathlib import Path
+from typing import Optional
+
+from plotly import graph_objects as go
+from plotly.utils import PlotlyJSONEncoder
+
+from vizro._themes._color_values import COLORS
+from vizro._themes.common_values import create_template_common
+
+THEMES_FOLDER = Path(__file__).resolve().parent
```
antonymilne, line 14:
```suggestion
THEMES_FOLDER = Path(__file__).parent
```
probably

vizro-core/src/vizro/_themes/create_chart_templates.py
```diff
@@ -0,0 +1,137 @@
...
+CSS_FILE = THEMES_FOLDER.parent / "static" / "css" / "vizro-bootstrap.min.css"
```
antonymilne, line 15: Also no need to split each part out as a separate string - you can just do `x / "static/css/vizro-bootstrap.min.css"` and so long as `x` is a pathlib path then it will work.

vizro-core/src/vizro/_themes/create_chart_templates.py
```diff
@@ -0,0 +1,137 @@
...
+
+def _extract_last_two_occurrences(variable: str, content: str) -> tuple[Optional[str], Optional[str]]:
+    """Extracts the last two occurrences of a variable from the CSS content.
+
+    Within the `vizro-bootstrap.min.css` file, variables appear multiple times: initially from the default Bootstrap
+    values, followed by the dark theme, and lastly the light theme. We are interested in the final two occurrences,
+    as these represent the values for our dark and light themes.
+    """
+    pattern = re.compile(rf"{variable}:\s*([^;]+);")
+    matches = pattern.findall(content)
```
antonymilne, line 27:
```suggestion
    matches = re.findall(rf"{variable}:\s*([^;]+);", content)
```

vizro-core/src/vizro/_themes/create_chart_templates.py
antonymilne: Yeah, I think this is fine 👍 I couldn't find a much better way of doing it. If you wanted to be sure then you could do something like:
* `content.split("data-bs-theme=dark")` and then find only first match after that
* `content.split("data-bs-theme=light")` and then find only first match after that

Up to you, I'm fine with how it is because we only ever run this manually and if something goes drastically wrong then it will be obvious from the resulting json.

vizro-core/src/vizro/_themes/create_chart_templates.py
```diff
@@ -0,0 +1,137 @@
...
+
+def extract_bs_variables_from_css_file(
+    variables: list[str], css_file_path: Path
+) -> tuple[dict[str, Optional[str]], dict[str, Optional[str]]]:
+    """Extract the last two occurrences for each variable in the CSS file."""
+    extracted_dark = {}
+    extracted_light = {}
+
+    with open(css_file_path) as css_file:
+        css_content = css_file.read()
```
antonymilne, line 42:
```suggestion
        css_content = css_file.read_text()
```
Pathlib makes this sort of thing much easier.

vizro-core/src/vizro/_themes/create_chart_templates.py
```diff
@@ -0,0 +1,137 @@
...
+    """Extracts the last two occurrences of a variable from the CSS content.
+
+    Within the `vizro-bootstrap.min.css` file, variables appear multiple times: initially from the default Bootstrap
+    values, followed by the dark theme, and lastly the light theme. We are interested in the final two occurrences,
+    as these represent the values for our dark and light themes.
+    """
+    pattern = re.compile(rf"{variable}:\s*([^;]+);")
+    matches = pattern.findall(content)
+    if len(matches) >= 2:  # noqa: PLR2004
+        return matches[-2].strip(), matches[-1].strip()
```
antonymilne, line 29: Maybe move this `strip` into `extract_bs_variables_from_css_file` where the other bits of cleaning are done.

vizro-core/src/vizro/_themes/create_chart_templates.py
```diff
@@ -0,0 +1,137 @@
...
+) -> tuple[dict[str, Optional[str]], dict[str, Optional[str]]]:
+    """Extract the last two occurrences for each variable in the CSS file."""
+    extracted_dark = {}
+    extracted_light = {}
+
+    with open(css_file_path) as css_file:
+        css_content = css_file.read()
+
+    for variable in variables:
+        dark_values, light_values = _extract_last_two_occurrences(variable, css_content)
```
antonymilne, line 45: Should just be `dark/light_value` singular here?

vizro-core/src/vizro/_themes/create_chart_templates.py
antonymilne: For me it looks a bit strange compared to to the code in `common_values.py` so I would see what it looks like using that same style. But no strong feelings, up to you. At least I would probably refactor so that you just set `layout`  rather than repeating `template.layout` every line.

vizro-core/src/vizro/_themes/create_chart_templates.py
```diff
@@ -0,0 +1,137 @@
...
+    return template.to_plotly_json()
+
+
+if __name__ == "__main__":
+    extracted_dark, extracted_light = extract_bs_variables_from_css_file(VARIABLES, CSS_FILE)
+    template_dark = generate_json_template(extracted_dark)
+    template_light = generate_json_template(extracted_light)
+
+    parser = argparse.ArgumentParser(description="Generate JSON chart templates.")
+    parser.add_argument("--check", help="check chart templates are up to date", action="store_true")
```
antonymilne, line 125:
```suggestion
    parser = argparse.ArgumentParser(description="Generate JSON plotly templates.")
    parser.add_argument("--check", help="check plotly templates are up to date", action="store_true")
```

vizro-core/src/vizro/_themes/create_chart_templates.py
```diff
@@ -0,0 +1,137 @@
...
+        go.Waterfall(
+            decreasing={"marker": {"color": COLORS["DISCRETE_10"][1]}},
+            increasing={"marker": {"color": COLORS["DISCRETE_10"][0]}},
+            totals={"marker": {"color": "grey"}},
+            textfont_color=FONT_COLOR_PRIMARY,
+            textposition="outside",
+            connector={"line": {"color": AXIS_COLOR, "width": 1}},
+        )
+    ]
+    return template.to_plotly_json()
```
antonymilne, line 116:
```suggestion
    return template
```
I feel like this is unnecessary given the `json.dumps` in `__main__`?

vizro-core/tests/unit/vizro/_themes/test_create_chart_template.py
```diff
@@ -0,0 +1,61 @@
...
+    }
+    expected_light = {
+        "BS-PRIMARY": "#976fd1",
+        "BS-SECONDARY": "#444fff",
+        "BS-TERTIARY": None,
+    }
+
+    mock_path = Path("/fake/path/to/css/vizro-bootstrap.css")
+    with patch("builtins.open", mock_open(read_data=css_content)):
+        result_dark, result_light = extract_bs_variables_from_css_file(variables, mock_path)
```
antonymilne, line 58: This would be more easily unit testable if you changed the code a little bit so that `extract_bs_variables_from_css_file` became `extract_bs_variables_from_css` (no file) and just takes in a string. Then in the real call you can do `extract_bs_variables_from_css(Path(...).read_text())` and in this test you could just pass in `css_content` directly without needing to do any mocking.

vizro-core/src/vizro/_themes/create_chart_templates.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+if __name__ == "__main__":
+    extracted_dark, extracted_light = extract_bs_variables_from_css_file(VARIABLES, CSS_FILE)
```
antonymilne, line 120:
```suggestion
    extracted_dark, extracted_light = extract_bs_variables_from_css(VARIABLES, CSS_FILE.read_text())
```
Will be explained why below in comment on tests.

vizro-core/src/vizro/_themes/common_values.py
antonymilne: No strong feelings, completely up to you.

I would maybe change the file name though to something that makes sense alongside `_colors.py`. Like `_common_template.py` or something.

vizro-core/src/vizro/_themes/_color_values.py
antonymilne, line 1: Now that these are only ever used in the template generation script I have a very minor preference to put them inside a function `get_colors` like is done with `get_template_common` function so that it's done lazily when we choose to manually run the script and not on every `import vizro`.

Also very minor preference to just call this file `_colors.py`.

antonymilne (approved):
Looks awesome 🙌 Lots of small comments but definitely it's pretty much as I was expecting so happy to approve in advance and feel free to merge when it's done as far as I'm concerned.

vizro-core/src/vizro/_themes/create_chart_templates.py
```diff
@@ -0,0 +1,137 @@
...
+    Within the `vizro-bootstrap.min.css` file, variables appear multiple times: initially from the default Bootstrap
+    values, followed by the dark theme, and lastly the light theme. We are interested in the final two occurrences,
+    as these represent the values for our dark and light themes.
+    """
+    pattern = re.compile(rf"{variable}:\s*([^;]+);")
+    matches = pattern.findall(content)
+    if len(matches) >= 2:  # noqa: PLR2004
+        return matches[-2].strip(), matches[-1].strip()
+
+    return None, None
```
antonymilne, line 31: The only thing I didn't understand in this PR is what happens here if there's `None, None`. It's covered in tests but what happens in the json? Does this happen in reality? Why would we want these empty values in the json rather than skipping them completely if the variable isn't found?

vizro-core/tests/unit/vizro/_themes/test_create_chart_template.py
huong-li-nguyen: Ah, that's a great idea! 💡

vizro-core/src/vizro/_themes/create_chart_templates.py
huong-li-nguyen: Actually the strip doesn't do anything - let me just remove it 👍

vizro-core/src/vizro/_themes/create_chart_templates.py
huong-li-nguyen: You're right! We can skip writing to the JSON if the variable can't be found. It still needs to return None, None to avoid errors when the variable isn't found (e.g., due to typos). However, there's no need to write it back into the JSON, so we can skip that step 👍

vizro-core/src/vizro/_themes/__init__.py
huong-li-nguyen: Yeah, that was legacy 😄 I've removed the file now

vizro-core/src/vizro/_themes/create_chart_templates.py
huong-li-nguyen: Ah, got it! I thought you wanted the nested dict structure. I just consolidated the two files using the dot notation 👍

vizro-core/hatch.toml
huong-li-nguyen: Unfortunately, they still do need linting in the end😬 Both of them gave the same result before because of the return of `template.to_plotly_json()`, but I removed the part now as well, as we need the linting in any case 👍

vizro-core/changelog.d/20250123_140015_huong_li_nguyen_refactor_template_generation.md
```diff
@@ -0,0 +1,48 @@
+<!--
```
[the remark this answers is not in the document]
huong-li-nguyen: No, I didn't add that change in this PR yet. I wan't to wait for the next token update before we make this public 👍

vizro-core/src/vizro/_themes/generate_plotly_templates.py
```diff
@@ -0,0 +1,130 @@
...
+    BG_COLOR = extracted_values["BS-BODY-BG"]
+    FONT_COLOR_SECONDARY = extracted_values["BS-SECONDARY"]
+    GRID_COLOR = extracted_values["BS-BORDER-COLOR"]
+    AXIS_COLOR = extracted_values["BS-TERTIARY-COLOR"]
+
+    # Apply common values
+    COLORS = get_colors()
+    template = create_template_common()
+    layout = template.layout
+    layout.update(
```
lingyielia, line 61: This refactor looks great, it's much cleaner than the dictionary assignment earlier

vizro-core/src/vizro/_themes/generate_plotly_templates.py
```diff
@@ -0,0 +1,130 @@
...
+    values, followed by the dark theme, and lastly the light theme. We are interested in the final two occurrences,
+    as these represent the values for our dark and light themes.
+    """
+    matches = re.findall(rf"{variable}:\s*([^;]+);", css_content)
+    if len(matches) >= 2:  # noqa: PLR2004
+        return matches[-2], matches[-1]
+    return None, None
+
+
+def extract_bs_variables_from_css(variables: list[str], css_content: str) -> tuple[dict[str, str], dict[str, str]]:
```
lingyielia, line 34: Does this function only works with `vizro-bootstrap.min.css`? Wondering how general this could be and how does it work if users like to try with other bootstrap. Could a user fork the repo and link to other bootstrap css, and `generate_json_template` helps generate the *_dark.json and  *_light.json?

lingyielia (approved):
LGTM!

vizro-core/src/vizro/_themes/generate_plotly_templates.py
huong-li-nguyen: This depends on the customizations, whether they offer both dark and light themes, and what their default theme is. We needed a custom script because we have many custom changes. In our CSS file, the default Bootstrap theme loads first, followed by our custom changes. This causes elements like [data-bs-theme=light] to appear twice. If they have minimal customization and only change some variables, the logic will be different.

Our default theme is dark, which is different from most other Bootstrap themes. So, if someone else has a different default (like light), they'll need to adjust the function to output the light theme first, as the order in the CSS file changes.

For those with a custom Bootstrap theme that only has a few changes (like just some variables), AM script is better suited for this! It's built on Bootswatch themes and handles generic color calculations and assignments: https://github.com/AnnMarieW/dash-bootstrap-templates/blob/1feb51177d72048f5448925c870a94d44e596409/_create_templates.py#L262

Issue #971, opened 2025-01-28
Issue Displaying Certain Icons in KPI Components


fpeucelle:
### Which package?

vizro

### Package version

0.1.31

### Description

Good morning,  
I would like to report an issue with the visualization of some icons in the KPIs. I tried to use "folder_check_2" from "Material Design," but I couldn't get it to display correctly. However, I was able to successfully display "create_new_folder."  

Thank you very much, and congratulations on the excellent project!  

### How to Reproduce

[PyCafe example](https://py.cafe/app/fpeucelle/vizro-kpi-cards-tips-data)

```py
import vizro.models as vm
import vizro.plotly.express as px
from vizro import Vizro
from vizro.figures import kpi_card

tips = px.data.tips

# Create a layout with five rows and four columns. The KPI card is positioned in the first cell, while the remaining cells are empty.
page = vm.Page(
    title="KPI card",
    layout=vm.Layout(grid=[[0, 0, -1, -1]] + [[-1, -1, -1, -1]] * 2),
    components=[
        vm.Figure(
            figure=kpi_card( # For more information, refer to the API reference for kpi_card
                data_frame=tips,
                value_column="tip",
                value_format="${value:.2f}",
                icon="folder_check_2",
                title="KPI card I",
            )
        )
    ],
    controls=[vm.Filter(column="day", selector=vm.RadioItems())],
)

dashboard = vm.Dashboard(pages=[page])
Vizro().build(dashboard).run()

```

### Output

![Image](https://github.com/user-attachments/assets/b553c772-b569-4261-9622-4990322afec3)

### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

huong-li-nguyen:
This is now fixed in the latest release @fpeucelle : https://github.com/mckinsey/vizro/releases/tag/0.1.32 :)

PR #973, opened 2025-01-29
[Tidy] Pull material icons on a monthly basis

huong-li-nguyen:
## Description
- Add a script that pulls the Google material font that we need for our icons and stores it inside our static folder
- Create a GitHub Action that runs the font download script on a monthly basis (Google material icons seem to be automatically updated every month: https://github.com/google/material-design-icons/commits/master/variablefont)

**Context**
One of our users raised this issue: https://github.com/mckinsey/vizro/issues/971 which we already fixed, but to prevent similar issues, we should update these icons automatically on a monthly basis. 

**Testing**
I've tested it by setting the base to be my current branch and on pull request to main and it seems to work. So hopefully, it should work fine after merge 🤞 
- https://github.com/mckinsey/vizro/actions/runs/13050618575/job/36409856086?pr=973
- https://github.com/mckinsey/vizro/pull/978

I've now changed the ref branch to main and changed it to run on schedule instead of pull request to main.

## Notice


.github/workflows/update-icons.yml
```diff
@@ -0,0 +1,42 @@
...
+        run: |
+          hatch run update-icons
+      - name: Create PR if change detected
+        uses: peter-evans/create-pull-request@v5
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: Update Material Icons
+          branch: bot/update-material-icons
+          title: "[Bot] Update Material Icons"
+          body: "Automated update of Material Icons"
```
huong-li-nguyen, line 40: From what I recall I can't test whether this GitHub action works until after merge right? 
@l0uden @maxschulz-COL - in this case, is there anything obvious I should change? You both have written similar actions in the past, so would love to get your input! 👍

l0uden: Unfortunately you can't run it anywhere, because `peter-evans/create-pull-request@v5 is not allowed to be used in mckinsey/vizro`.

https://github.com/mckinsey/vizro/actions/runs/13031136981

huong-li-nguyen: Ahhhh, that's annoying 😅

vizro-core/tools/update_material_icons.py
```diff
@@ -0,0 +1,31 @@
...
+
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+logger = logging.getLogger(__name__)
+
+
+def download_font(url, path):
+    """Downloads the Material Design Icons font from Google Fonts and saves it to the local static folder."""
+    try:
+        response = requests.get(url, timeout=TIMEOUT)
+        if response.status_code == STATUS_OK:
```
maxschulz-COL, line 21:
```suggestion
        if response.status_code == 200:
```
Will it ever change?

.github/workflows/update-icons.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Update Material font
+        run: |
+          hatch run update-icons
+      - name: Check for changes
+        run: |
+          git config user.email "145135826+vizro-svc@users.noreply.github.com"
+          git config user.name "Vizro Team"
+          git diff --exit-code || echo "has_changes=true" >> $GITHUB_ENV
+      - name: Delete old branch if exists
+        if: env.has_changes == 'true'
```
maxschulz-COL, line 37: I think it would be cleaner if we pulled this one out if possible as we are using it on every step after. You could have an overall next job that is conditional on this.

You could then nicely see in CI that first the check runs, and if positive, the update runs.

maxschulz-COL (approved):
Cool! 2 small suggestions, but overall looks good! Thanks!

vizro-core/tools/update_material_icons.py
huong-li-nguyen: No, but I changed this because, otherwise the linter complains that it's a magic number and it wants it stored in a variable 😄

maxschulz-COL: lol, ok fair enough

.github/workflows/update-icons.yml
huong-li-nguyen: Ahh, got it!

vizro-core/tools/update_material_icons.py
```diff
@@ -0,0 +1,31 @@
...
+            path.write_bytes(response.content)
+            logger.info(f"✍️ Material font downloaded and saved to {path}")
+        else:
+            logger.error(f"❌ Failed to download Material font. Status code: {response.status_code}")
+    except requests.RequestException:
+        logger.error("❌ Failed to download Material font due to time out.")
```
petar-qb, line 27: Should we raise an exception here to get a better look at the logs of what happened? Time out is not the only reason it can fail.

petar-qb (approved):
Amazing PR (as usual)! 🏅

vizro-core/tools/update_material_icons.py
huong-li-nguyen: Sure 👍

PR #982, opened 2025-02-03
[Dev] Fix github action on updating material icons

huong-li-nguyen:
## Description
The current Github action on updating the material icons didn't run through successfully due to the [last commit in the previous PR](https://github.com/mckinsey/vizro/pull/973/commits/56046c2d0ac85f46572cab7cb6122832dfdc12ec) to separate out the jobs, as a few things had to be rewritten.

This is fixed in this PR now.

Tested here:
- https://github.com/mckinsey/vizro/actions/runs/13110480404/job/36573095644?pr=982

## Notice


petar-qb (commented):
I'm not a wizard in github actions so this might be a stupid question: What would be the reason to separate this into two jobs instead of doing everything in one job?

huong-li-nguyen:
> I'm not a wizard in github actions so this might be a stupid question: What would be the reason to separate this into two jobs instead of doing everything in one job?

@maxschulz-COL preferred having it separate as you don't need to replicate the if statements then, and it's a cleaner separation if/when the first step fails: https://github.com/mckinsey/vizro/pull/973#discussion_r1936856145

I am actually indifferent 😄 Having it in one has the advantage that you don't need to install hatch again, but then again, it just takes a few seconds, so I don't really mind.

petar-qb:
> @maxschulz-COL preferred having it separate as you don't need to replicate the if statements then, and it's a cleaner separation if/when the first step fails: https://github.com/mckinsey/vizro/pull/973#discussion_r1936856145

Ah I see now.. The current code is much prettier than the old one that handles all the process in a single job -> https://github.com/mckinsey/vizro/pull/973/files/c9402be055b4e74d8809292f3793759a3491ba1c

> I am actually indifferent 😄 Having it in one has the advantage that you don't need to install hatch again, but then again, it 
just takes a few seconds, so I don't really mind.

I really like this version with two jobs. It's much cleaner to read and maintain 👍

petar-qb (approved):

antonymilne (approved):
Thanks for writing this script! I had a similar idea before but never got around to actually doing it.

I actually have a couple of small changes to propose on https://github.com/mckinsey/vizro/pull/973/ but will make a separate PR for it. Maybe I will extend it to download the dmc DATES.css file also, not sure. Do we have any other files that we're manually downloading and might go out of date?

Also agreed with the split into two jobs even if it means repeating a step 👍

Issue #871, opened 2024-11-13
Does Google Gemini support Vizro-AI?


richizo:
### Question

I'm trying to follow the example given on [Vizro-AI homepage](https://vizro.readthedocs.io/projects/vizro-ai/en/vizro-ai-0.3.2/pages/tutorials/quickstart/), but instead of using the **OpenAI model** I switched to the **Google Gemini**.

But I'm getting the error bellow.

**Error**:

```
ChatGoogleGenerativeAIError:** Invalid argument provided to Gemini: 400 * GenerateContentRequest.tools[0].function_declarations[0].parameters.properties[imports].items: missing field.
```

**Stacktrace**:

```
Traceback (most recent call last):
  File "/home/ubuntu/Projects/DataAnalysis/graphaito/.venv/lib/python3.10/site-packages/streamlit/runtime/scriptrunner/exec_code.py", line 88, in exec_func_with_error_handling
    result = func()
  File "/home/ubuntu/Projects/DataAnalysis/graphaito/.venv/lib/python3.10/site-packages/streamlit/runtime/scriptrunner/script_runner.py", line 579, in code_to_exec
    exec(code, module.__dict__)
  File "/home/ubuntu/Projects/DataAnalysis/graphaito/graphaito.py", line 42, in <module>
    fig = vizro_ai.plot(
  File "/home/ubuntu/Projects/DataAnalysis/graphaito/.venv/lib/python3.10/site-packages/vizro_ai/_vizro_ai.py", line 34, in wrapper
    return func(*args, **kwargs)
  File "/home/ubuntu/Projects/DataAnalysis/graphaito/.venv/lib/python3.10/site-packages/vizro_ai/_vizro_ai.py", line 87, in plot
    response = _get_pydantic_model(
  File "/home/ubuntu/Projects/DataAnalysis/graphaito/.venv/lib/python3.10/site-packages/vizro_ai/dashboard/_pydantic_output.py", line 84, in _get_pydantic_model
    res = pydantic_llm.invoke(message_content)
  File "/home/ubuntu/Projects/DataAnalysis/graphaito/.venv/lib/python3.10/site-packages/langchain_core/runnables/base.py", line 2879, in invoke
    input = context.run(step.invoke, input, config)
  File "/home/ubuntu/Projects/DataAnalysis/graphaito/.venv/lib/python3.10/site-packages/langchain_core/runnables/base.py", line 5093, in invoke
    return self.bound.invoke(
[...]

PR #986, opened 2025-02-04
[Fix] Temporarily fix the gemini structured output error

lingyielia:
## Description

Provide a temp fix to enable gemini models before the proper dependencies (pydantic v2 and langchain 0.3) are in place.

related to: https://github.com/mckinsey/vizro/issues/871

## Screenshot

Screenshot 2025-02-05 at 17 50 17

## Notice


vizro-ai/src/vizro_ai/dashboard/_pydantic_output.py
```diff
@@ -79,13 +99,35 @@ def _get_pydantic_model(
...
+                pydantic_llm = prompt | llm_model.with_structured_output(schema)
+                res = pydantic_llm.invoke(message_content)
+
+                # Handle case where response is a list, which is the case for Gemini models
+                if isinstance(res, list) and len(res) > 0:
+                    res = res[0]
+                    if isinstance(res, dict) and "args" in res:
+                        res = res["args"]
+
+                return response_model.parse_obj(res)
```
huong-li-nguyen, line 123: This is only google response specific right? Maybe move it inside a separate function `_handle_google_llm_response` to keep the overall `_get_pydantic_model` function cleaner.

```suggestion
            if "google" in llm_model.__class__.__module__.lower():
                from langchain_core.utils.function_calling import convert_to_openai_function

                schema = convert_to_openai_function(response_model)
                schema = _handle_gemini_model(schema)

                pydantic_llm = prompt | llm_model.with_structured_output(schema)
                res = pydantic_llm.invoke(message_content)

                # Handle case where response is a list, which is the case for Gemini models
                if isinstance(res, list) and len(res) > 0:
                    res = res[0]
                    if isinstance(res, dict) and "args" in res:
                        res = res["args"]

                return response_model.parse_obj(res)
```

vizro-ai/src/vizro_ai/dashboard/_pydantic_output.py
```diff
@@ -79,13 +99,35 @@ def _get_pydantic_model(
...
+                from langchain_core.utils.function_calling import convert_to_openai_function
+
+                schema = convert_to_openai_function(response_model)
+                schema = _handle_gemini_model(schema)
+
+                pydantic_llm = prompt | llm_model.with_structured_output(schema)
+                res = pydantic_llm.invoke(message_content)
+
+                # Handle case where response is a list, which is the case for Gemini models
+                if isinstance(res, list) and len(res) > 0:
```
huong-li-nguyen, line 118:
```suggestion
                if isinstance(res, list) and res:
```

vizro-ai/src/vizro_ai/dashboard/_pydantic_output.py
```diff
@@ -79,13 +99,35 @@ def _get_pydantic_model(
...
+                    res = res[0]
+                    if isinstance(res, dict) and "args" in res:
+                        res = res["args"]
```
huong-li-nguyen, line 121: Is there a case where res is a dict but `args` is not in there? What should happen then? Below would e.g. just return None, while the current snippet would return res[0] I guess? Depending on what it should handle you could rewrite :) 

```
if isinstance(res, dict):
    res = res.get("args")
```

huong-li-nguyen (commented):
I haven't tested it out in terms of running the model, but I just looked at the code. Minor comments, but looks good to me 👍 Thanks for fixing! 🚀

vizro-ai/changelog.d/20250203_203402_lingyi_zhang_gemini_api_error.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Added temporary compatibility fix for Gemini models in `_pydantic_output.py` until Pydantic v2 implementation (#issue_number). ([#986](https://github.com/mckinsey/vizro/pull/986))
```
maxschulz-COL, line 40:
```suggestion
- Added temporary compatibility fix for Gemini models in `_pydantic_output.py`. ([#986](https://github.com/mckinsey/vizro/pull/986))
```
Even simpler

vizro-ai/src/vizro_ai/dashboard/_pydantic_output.py
```diff
@@ -63,6 +63,26 @@ def _create_message_content(
     return message_content
 
 
+def _handle_gemini_model(schema: dict) -> dict:
+    """Handle schema modifications specific to Gemini models."""
```
maxschulz-COL, line 67: What are the schema modifications needed? What is the [0] element we are updating to?

maxschulz-COL (changes requested):
Looks overall good, thanks for digging so deep, that must not have been fun... 💯 ⭐ 

I cannot get this to run though, so before merging, let's please check together why I cannot get this to work with Google - might be an issue on my side, but I just want to be sure.

maxschulz-COL (approved):

vizro-ai/src/vizro_ai/dashboard/_pydantic_output.py
lingyielia: The API return schema of one vendor is quite consistent, so the parsing was written in a somewhat linear way. `res.get("args")` would work too, which means the `response_model.parse_obj(res)` will throw an validation error I guess

huong-li-nguyen (approved):

vizro-ai/src/vizro_ai/dashboard/_pydantic_output.py
```diff
@@ -63,6 +63,26 @@ def _create_message_content(
...
+    if isinstance(schema, dict) and "parameters" in schema:
```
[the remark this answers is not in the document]
lingyielia: No, it's specific to the OpenAI API schema

e.g.,
```
{
    "name": "get_weather",
    "description": "Fetches the weather in the given location",
    "strict": true,
    "parameters": {
        "type": "object",
        "properties": {
            "location": {
                "type": "string",
                "description": "The location to get the weather for"
            },
            "unit": {
                "type": "string",
                "description": "The unit to return the temperature in",
                "enum": ["F", "C"]
            }
        },
        "additionalProperties": false,
        "required": ["location", "unit"]
    }
}
```

vizro-ai/src/vizro_ai/dashboard/_pydantic_output.py
lingyielia: It's supposed to tackle nested schema in dashboard creation (Selector in Filter). I think I didn't handled it correctly still. Since the .plot has simpler schema and can be parsed correctly just by using `convert_to_openai_function`, I will drop the uncertain portion and enable .plot only in this pr.

Issue #976, opened 2025-01-29
Add option to customize highlighting of positive and negative changes in the KPI card


fpeucelle:
### Which package?

None

### What's the problem this feature will solve?

The KPI card component shows variations between the value and the reference, but does not allow customizing how to highlight positive and negative changes based on their desirability. Depending on the indicator, a positive change can be desirable (like in growth or efficiency) or undesirable (like in debt or losses). It would be useful to have the option to configure this behavior, while keeping the current format as the default.

### Describe the solution you'd like

Add parameters to the kpi_card_reference function to allow configuring whether positive changes are desirable and negative changes are undesirable (default behavior), or vice versa. This should apply to colors, icons, and text in the card footer.

This will allow customization of how changes are displayed based on their context, while keeping the default behavior, but offering flexibility for other use cases.

src/vizro/figures/_kpi_cards.py
positive_is_desirable = True  # If True, positive changes are desirable
negative_is_desirable = False  # If True, negative changes are desirable

if delta_relative > 0:
    if positive_is_desirable:
        footer_class = "color-pos-highlight"
        footer_text = f"Improvement: {delta_relative:.2%}"
    else:
        footer_class = "color-neg-highlight"
        footer_text = f"Increase: {delta_relative:.2%}"
else:
    if negative_is_desirable:
        footer_class = "color-pos-highlight"
        footer_text = f"Improvement: {delta_relative:.2%}"
    else:
        footer_class = "color-neg-highlight"
        footer_text = f"Decrease: {delta_relative:.2%}"

### Code of Conduct

- [x] I agree to follow the  [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

huong-li-nguyen:
Hello @fpeucelle,

I'm really happy to see you stress-testing our KPI functions! When we developed these functions, we did think about adding an option to change the coloring of the delta. At the time, we decided to keep the API of the KPI card functions as simple as possible and see how many people would want this feature. You're actually the first to request it, which is awesome! I'll keep this issue open so others can up-vote it too.

In the meantime, you can achieve this customization already through custom figures. This is the guide on [how to create a custom KPI card](https://vizro.readthedocs.io/en/stable/pages/user-guides/custom-figures/#custom-kpi-card). If you follow the guide, your customization could look like this: https://py.cafe/huong-li-nguyen/vizro-custom-kpi

This will then achieve your desired result 🌟 

[...]

PR #995, opened 2025-02-05
[Feat] Add `reverse_color` argument to `kpi_card_reference`

huong-li-nguyen:
## Description
When we initially created the KPI card functions, we were already thinking about incorporating the `reverse_color` argument. Back then we decided to keep the API small and see whether it's even requested.

It has been recently requested and is actually fairly simple to implement, so we've decided to introduce the argument now. See https://github.com/mckinsey/vizro/issues/976 for context.

- Enable `reverse_color` argument
- Update docs
- Update unit tests
- Update e2e component test in pure Dash app
- Update kpi screenshot tests in qa repo: https://github.com/mckinsey/vizro-qa/pull/160

## Notice


vizro-core/src/vizro/figures/_kpi_cards.py
```diff
@@ -129,6 +135,11 @@ def kpi_card_reference(  # noqa: PLR0913
         >>> vm.Page(title="Page", components=[vm.Figure(figure=kpi_card_reference(...))])
 
     """
+
+    def _get_footer_class(delta: float, reverse_color: bool) -> str:
+        pos_color, neg_color = ("color-neg", "color-pos") if reverse_color else ("color-pos", "color-neg")
+        return pos_color if delta > 0 else neg_color if delta < 0 else ""
+
```
petar-qb, line 142: This makes the function just a little cleaner, but that's totally up to you. I agree with both versions.
```suggestion

    def _get_footer_class(delta: float, reverse_color: bool) -> str:
        if delta == 0:
            return ""
            
        pos_color, neg_color = ("color-neg", "color-pos") if reverse_color else ("color-pos", "color-neg")
        return pos_color if delta > 0 else neg_color

```

petar-qb (approved):
I enjoy reviewing your PRs 😄. This looks perfect!

vizro-core/src/vizro/figures/_kpi_cards.py
```diff
@@ -129,6 +135,11 @@ def kpi_card_reference(  # noqa: PLR0913
         >>> vm.Page(title="Page", components=[vm.Figure(figure=kpi_card_reference(...))])
 
     """
+
+    def _get_footer_class(delta: float, reverse_color: bool) -> str:
```
antonymilne, line 139:
```suggestion
    def _get_footer_class(delta: float) -> str:
```

`reverse_color` will still be defined inside the closure without it being an explicit function argument.

antonymilne: Given this is meant to be a relatively simple copy and pastable function I would even go one step further and not use a closure here at all and instead just do:
```
pos_color, neg_color = ("color-neg", "color-pos") if reverse_color else ("color-pos", "color-neg")
footer_class = pos_color if delta > 0 else neg_color if delta < 0 else ""
```

antonymilne (approved):
Perfect! 💯 

Should @l0uden update his tests to include this? I think he may have tests for KPI cards as components in a pure Dash app and within vizro.

huong-li-nguyen:
> Perfect! 💯
> 
> Should @l0uden update his tests to include this? I think he may have tests for KPI cards as components in a pure Dash app and within vizro.

I just extend his tests 👍 Give me a sec, will update the screenshots as well.

vizro-core/tests/e2e/component_library/test_component_library.py
```diff
@@ -65,18 +62,31 @@
         title="KPI ref. with icon",
         icon="shopping_cart",
     ),
+    kpi_card_reference(
```
huong-li-nguyen, line 65: @l0uden - I've updated the configuration and screenshots already :) I was just wondering if we could take a screenshot of the whole page instead of the current view? It's currently being cut-off. If it's easy to do, would you mind taking a look?

If this is complicated to do, then don't worry. The most important thing is the change of colors, which I can also already see from the screenshot, but of course it would be nicer if I have the entire view there.
main_kpi_card_component_library

huong-li-nguyen: Discussed with @l0uden - this will not be implemented in this PR, but eventually tackled later depending how easy it is to implement. Might be interesting to research how to do so in general, but for this specific use-case we don't need it anymore, as I simplified the configuration to what actually needs to be tested.

PR #999, opened 2025-02-06
[Bug] Fix discriminator error when adding duplicate type model

maxschulz-COL:
## Description
There was a discrepancy between pydantic V1 and V2:

V1: you could have a discriminated union where the discriminator would have multiple options for choosing which model to coerce into: e.g. you could have `type = "card"` twice, and it would coerce into the last type added.

V2: In V2 the same situation would lead to an error that would say: `TypeError: Value 'card' for discriminator 'type' mapped to multiple choices`.

This situation does not arise that often, mostly in re-runs, or tests. We test for this now, and it is fixed.

Note: We are thus defining a DEFAULT behaviour, which is that the last added model takes precedence in discriminated unions.

## Notice

### DIFF
--- a/vizro-core/changelog.d/20250206_203003_maximilian_schulz_add_type_bug.md
+++ b/vizro-core/changelog.d/20250206_203003_maximilian_schulz_add_type_bug.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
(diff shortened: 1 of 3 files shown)


vizro-core/changelog.d/20250206_203003_maximilian_schulz_add_type_bug.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Fixed a bug where adding a model of the same discriminated type as an existing type would raise an error. Now added types overwrite previous type. ([#999](https://github.com/mckinsey/vizro/pull/999))
```
antonymilne, line 40:
```suggestion
- Fix a bug where `add_type` would raise an error if the `type` had already been added. ([#999](https://github.com/mckinsey/vizro/pull/999))
```

vizro-core/src/vizro/models/_base.py
```diff
@@ -136,7 +136,11 @@ def _extract_captured_callable_data_info() -> set[str]:
 
 def _add_type_to_union(union: type[Any], new_type: type[Any]):  # TODO[mypy]: not sure how to type the return type
     args = get_args(union)
-    return Union[args + (new_type,)]  # noqa: RUF005 #as long as we support Python 3.9, we can't use the new syntax
+    all_types = args + (new_type,)  # noqa: RUF005 #as long as we support Python 3.9, we can't use the new syntax
+    # The below removes duplicates by type, which would trigger a pydantic error otherwise
+    # Last added type will be the one that is kept - this is replicating V1 behavior that would other raise an error in V2
+    unique_types = tuple({t.model_fields["type"].default: t for t in all_types}.values())
```
antonymilne, line 142: This took me a which to understand... Maybe worth another comment explaining a bit how it works, like what `t.model_fields["type"].default` contains and how `values()` selects the last value if a key is duplicated.

antonymilne (approved):
Great detective work and fix! 🕵️ 🐛 🦸

huong-li-nguyen (approved):
LGTM - thank you for fixing so quickly!! 🚀 

Agree with @antonymilne on adding another comment to understand that code line 👍

PR #1001, opened 2025-02-08
Enable datasets_from_catalog to return factory-based datasets

### LINKED ISSUE #988: Kedro integration cannot load datasets created using dataset factories
### Which package?

vizro

### Package version

0.1.30

### Description

When using the kedro integration for data management, it is currently not possible to load datasets that have been generated by [dataset factories datasets](https://docs.kedro.org/en/stable/data/kedro_dataset_factories.html). This is a known problem that is being addressed by the new `KedroDataCatalog` in the most recent versions of kedro (>=0.19.9). The issue is discussed [here](https://github.com/kedro-org/kedro/issues/3312).

### How to Reproduce

1. Create a new kedro project: `kedro new --starter=spaceflights-pandas`
2. Modify `conf/base/catalog.yml` to use the dataset factories syntax. Remove 
```
companies:
  type: pandas.CSVDataset
  filepath: data/01_raw/companies.csv
  metadata:
    description: "A CSV file containing raw companies."

reviews:
  type: pandas.CSVDataset
  filepath: data/01_raw/reviews.csv
  metadata:
    description: "A CSV file containing raw reviews."
```
and replace it with:
```
"{dataset_name}#csv":
  type: pandas.CSVDataset
  filepath: data/01_raw/{dataset_name}.csv
```
3. Add vizro to the requirements and open a jupyter notebook with kedro magic: `%load_ext kedro.ipython`. The kedro catalog is now defined.
4. Write in a cell:
```
from vizro.integrations import kedro as kedro_integration

kedro_integration.datasets_from_catalog(catalog).keys()
```

### Output

The steps above should output a list of the kedro dataset names without "companies" and "reviews". However, as kedro pipelines' inputs/outputs refer to "companies" and "reviews" and they match the dataset factory defined above, they should be available in the catalog.

### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

gtauzin:
## Description
This PR adds the capabilty to the kedro integration to exploit datasets created using dataset factories.

Fixes #988.

## Notice


vizro-core/src/vizro/integrations/kedro/_data_manager.py
```diff
@@ -3,25 +3,51 @@
...
-    return datasets
+def pipelines_from_project(project_path: Union[str, Path]) -> Pipeline:
+    bootstrap_project(project_path)
+    from kedro.framework.project import pipelines
+
+    return pipelines
+
+
+def datasets_from_catalog(
+    catalog: CatalogProtocol | KedroDataCatalog, *, pipeline: Pipeline = None
```
gtauzin, line 30: @antonymilne I have updated the input type of catalog as per my comment in the issue as `DataCatalog` does not inherit from `CatalogProtocol`.

vizro-core/docs/pages/user-guides/kedro-data-catalog.md
```diff
@@ -51,7 +72,9 @@ The full code for these different cases is given below.
         from vizro.managers import data_manager
 
 
-        for dataset_name, dataset in kedro_integration.datasets_from_catalog(catalog).items():
+        pipeline = pipelines.get("my_pipeline")
+
+        for dataset_name, dataset in kedro_integration.datasets_from_catalog(catalog, pipeline=pipeline).items():
```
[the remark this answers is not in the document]
gtauzin: Additionally, it seems to me the for loop over all datasets in the catalog does not represent a typical use case. In gerneral the catalog may contains hundreds of datasets and we may want to pass a couple to vizro's data manager.

gtauzin (commented):
@antonymilne Just a few comments on the implementation you suggested in #988.

Also, when I ran `hatch run test-unit, I did not see the kedro_integration tests in the list of tests collected by pytest. Not sure why, but happy to investifate if needed.

gtauzin:
The CI is failing because I added a polars dataset to the catalog and polars is not included in the dependencies.

This is an interesting edge case because in the case where I have a large number of datasets in my catalog and I want to load a few of them (all pandas-based) for a vizro dashboards, I should not be forced to include all dependencies related to all datasets in my catalog. Does it make sense? I feel that `datasets_from_catalog` should take a list of dataset names as a parameter so we can control which ones are instantiated.

vizro-core/docs/pages/user-guides/kedro-data-catalog.md
```diff
@@ -12,7 +12,7 @@ pip install vizro[kedro]
 
 ## Use datasets from the Kedro Data Catalog
 
-`vizro.integrations.kedro` provides functions to help generate and process a [Kedro Data Catalog](https://docs.kedro.org/en/stable/data/index.html). Given a Kedro Data Catalog `catalog`, the general pattern to add datasets into the Vizro data manager is:
+`vizro.integrations.kedro` provides functions to help generate and process a [Kedro Data Catalog](https://docs.kedro.org/en/stable/data/index.html). It supports both the original [DataCatalog](https://docs.kedro.org/en/stable/data/data_catalog.html) and the more recently introduced [KedroDataCatalog](https://docs.kedro.org/en/stable/data/index.html#kedrodatacatalog-experimental-feature). Given a Kedro Data Catalog `catalog`, the general pattern to add datasets into the Vizro data manager is:
```
stichbury, line 15:
```suggestion
`vizro.integrations.kedro` provides functions to help generate and process a [Kedro Data Catalog](https://docs.kedro.org/en/stable/data/index.html). It supports both the original [`DataCatalog`](https://docs.kedro.org/en/stable/data/data_catalog.html) and the more recently introduced [`KedroDataCatalog`](https://docs.kedro.org/en/stable/data/index.html#kedrodatacatalog-experimental-feature). Given a Kedro Data Catalog `catalog`, the general pattern to add datasets into the Vizro data manager is:
```

stichbury (commented):
LGTM from docs perspective, with just two minor tweaks. Thank you so much @gtauzin for this contribution 🌟🌟🌟🌟🌟

vizro-core/docs/pages/user-guides/kedro-data-catalog.md
```diff
@@ -23,6 +23,19 @@ for dataset_name, dataset in kedro_integration.datasets_from_catalog(catalog).it
     data_manager[dataset_name] = dataset
 ```
 
+To add datasets that are defined using the [Kedro dataset factory](https://docs.kedro.org/en/stable/data/kedro_dataset_factories.html), `datasets_from_catalog` needs to access the pipelines that use them.
```
antonymilne, line 26: Please could we move all the stuff that includes dataset factories and pipelines to a whole new subsection at the bottom of the page? So it goes something like this. I just copied and pasted your changes with a few modifications - see what you think. I didn't try building it so probably went wrong somewhere with the markdown syntax so let's check the RTD preview too.

Main reason for this is that dataset factories is a relatively advanced feature which only some people would use, and I'd like someone to be able to just understand and copy & paste the existing `catalog`-only versions without seeing anything about `pipeline`. So only users who actually want to use dataset factories will look at the section on it and beginner users hopefully won't be confused by whether or not they need to supply the `pipeline` argument.

````
## Use datasets from the Kedro Data Catalog

<content as it is now including your change that mentions Kedro Data Catalog. So code snippets remain as they are now without mentioning pipelines>

### Use dataset factories

To add datasets that are defined using a [Kedro dataset factory](https://docs.kedro.org/en/stable/data/kedro_dataset_factories.html), `datasets_from_catalog` needs to resolve dataset patterns against explicit datasets. Given a Kedro `pipelines` dictionary, you should specify a `pipeline` argument as follows:

```python
kedro_integration.datasets_from_catalog(catalog, pipeline=pipelines["__default__"]) # (1)!
```
1. You can specify the name of your pipeline, for example `pipelines["my_pipeline"]`, or even combine multiple pipelines with `pipelines["a"] + pipelines["b"]`. The Kedro `__default__` pipeline is what runs by default with the `kedro run` command.

The `pipelines` variable may have been created the following ways:

1. Kedro project path. Vizro exposes a helper function `vizro.integrations.kedro.pipelines_from_project` to generate a `pipelines` given the path to a Kedro project.
1. [Kedro Jupyter session](https://docs.kedro.org/en/stable/notebooks_and_ipython/kedro_and_notebooks.html). This automatically exposes `pipelines`.

The full code for these different cases is given below.

!!! example "Import a Kedro Data Catalog with dataset factories into the Vizro data manager"
    === "app.py (Kedro project path)"
        ```python
        from vizro.integrations import kedro as kedro_integration


        project_path = "/path/to/kedro/project"
        catalog = kedro_integration.catalog_from_project(project_path)
        pipelines = kedro_integration.catalog_from_project(project_path)

        for dataset_name, dataset in kedro_integration.datasets_from_catalog(catalog, pipeline=pipelines["__default__"]).items():
            data_manager[dataset_name] = dataset
        ```

    === "app.ipynb (Kedro Jupyter session)"
        ```python


        for dataset_name, dataset in kedro_integration.datasets_from_catalog(catalog, pipeline=pipelines["__default__"]).items():
            data_manager[dataset_name] = dataset
        ```
````

File: vizro-core/docs/pages/user-guides/kedro-data-catalog.md (hunk 2 above)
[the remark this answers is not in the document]
Comment (antonymilne) [reply vizro-core/docs/pages/user-guides/kedro-data-catalog.md:77]: Very good point about renaming it `dataset_loader`, please do make that change throughout this file 👍 

Agree that this is not necessarily a representative of a typical user who only wants to use a couple of datasets, where you could instead just do:
```
datasets = kedro_integration.datasets_from_catalog(catalog).items():
data_manager["a"] = datasets["a"]
data_manager["b"] = datasets["b"]

# or maybe this, but it's a bit cryptic/weird looking for docs...
data_manager |= {dataset_name: dataset_loader for dataset_name, dataset_loader in kedro_integration.datasets_from_catalog(catalog).items() if dataset_name in {"a", "b"}}
```

However, I actually think it's easier to just add all the datasets to the data manager and then the user just specifies whichever they want to use when they develop their dashboard. Since this is all dynamic data, the dataset loading function doesn't get executed until they're needed, and any datasets that are in the Vizro data manager and not used in the dashboard won't be loaded at all so they don't do any harm in there. I thought overall this would be easier since the same code snippet just works for everyone without needing to think about what dataset you need in advance. wdyt?

This point is not at all explained in the current docs though, so I will add it later this week in a different PR if you think it make sense. If not then I'm happy to add an example that just adds a few datasets rather than looping through them all.

FYI I also considered making our Kedro functionality add to the data manager in-place so you wouldn't need to do anything other than e.g.  `kedro_integration.add_datasets_from(catalog)` to add them all to the data manager. Still not sure whether that's a good idea or not but I think probably not since it relies on `data_manager` being a global variable which it might not be in future.

**Review by antonymilne — COMMENTED** [review]
Thank you very much for your PR @gtauzin! This looks great 💯 I just looked through the docs so far (thanks very much for doing those as well!) and left a few comments but it generally looks great. I'll review the code and tests when I get a chance later today.

File: vizro-core/src/vizro/integrations/kedro/_data_manager.py
Comment (antonymilne) [inline vizro-core/src/vizro/integrations/kedro/_data_manager.py:1]: These new functions should also go in `integrations/kedro/__init__.py`.

File: vizro-core/docs/pages/user-guides/kedro-data-catalog.md (hunk 4 above)
Comment (gtauzin) [reply vizro-core/docs/pages/user-guides/kedro-data-catalog.md:26]: Done! :+1:

File: vizro-core/src/vizro/integrations/kedro/_data_manager.py
Comment (gtauzin) [reply vizro-core/src/vizro/integrations/kedro/_data_manager.py:1]: Done! :+1:

**Comment by antonymilne** [comment]
> @antonymilne Just a few comments on the implementation you suggested in #988.
> 
> Also, when I ran `hatch run test-unit, I did not see the kedro_integration tests in the list of tests collected by pytest. Not sure why, but happy to investifate if needed.

They are definitely running in CI (you can see easily because they're failing due to the polars thing):
```
tests/unit/vizro/integrations/kedro/test_kedro_data_manager.py FF        [ 13%]
```

You'll note that actually they pass on Python 3.13:
image

The reason for this is that the kedro tests are skipped on that version because kedro doesn't fully support Python 3.13 yet (at least not time I checked). So I guess what's happened for you is that `hatch run test-unit` created an environment with Python 3.13 which isn't running the kedro tests. FYI the hatch environment configuration is done like this in hatch.toml:
```
[envs.all.overrides]
# Kedro is currently not compatible with Python 3.13 and returns exceptions when trying to run the unit tests on
# Python 3.13. These exceptions turned out to be difficult to ignore: https://github.com/mckinsey/vizro/pull/216
matrix.python.features = [
  {value = "kedro", if = ["3.9", "3.10", "3.11", "3.12"]}
]
```

You can check what Python version hatch is using with `hatch run python -V`.

You should be able to fix this by doing `hatch run all.py3.12:unit-test` which will run the unit tests in a Python 3.12 environment and hopefully include the kedro tests.

File: vizro-core/docs/pages/user-guides/kedro-data-catalog.md (hunk 2 above)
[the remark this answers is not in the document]
Comment (stichbury) [reply vizro-core/docs/pages/user-guides/kedro-data-catalog.md]: Sorry to jump in here. Just a note for @antonymilne really that if this doc is going to have details of inner workings, that is really _explanatory_ content rather than _how-to guide_ content. Meaning, it should probably be split out elsewhere according to the diataxis model of our docs. 

If you make a new PR for it, I'm happy to help with the relocation of data manager docs to make sure we (approximately) follow the general guidelines:

image

**Comment by gtauzin** [comment]
Thanks for the review @antonymilne. I believe I included all your suggestions.

> The reason for this is that the kedro tests are skipped on that version because kedro doesn't fully support Python 3.13 yet (at least not time I checked). So I guess what's happened for you is that `hatch run test-unit` created an environment with Python 3.13 which isn't running the kedro tests. FYI the hatch environment configuration is done like this in hatch.toml:
> 
> ```
> [envs.all.overrides]
> # Kedro is currently not compatible with Python 3.13 and returns exceptions when trying to run the unit tests on
> # Python 3.13. These exceptions turned out to be difficult to ignore: https://github.com/mckinsey/vizro/pull/216
> matrix.python.features = [
>   {value = "kedro", if = ["3.9", "3.10", "3.11", "3.12"]}
> ]
> ```
> 
> You can check what Python version hatch is using with `hatch run python -V`.
> 
> You should be able to fix this by doing `hatch run all.py3.12:unit-test` which will run the unit tests in a Python 3.12 environment and hopefully include the kedro tests.

Thanks for the explanation, it was my first time using `hatch` (at least not as a `uv` build backend). 

Seems like it was `test-unit` instead of `unit-test`, but it failed as well:

```
(vizro) ➜  vizro-core git:(dataset_factory) ✗ hatch run all.py3.12:test-unit
ImportError while loading conftest '/home/guillaume/Workspace/vizro/vizro-core/tests/conftest.py'.
tests/conftest.py:4: in <module>
    from vizro import Vizro
E   ModuleNotFoundError: No module named 'vizro'
```

However, `hatch run all.py3.11:test-unit` ran fine!  Out of curioisity, is it possible to run `pytest` directly? That would have been easier to iterate faster.

FYI, kedro now support python 3.13 (at least according to the badges on the main README.md)

**Comment by gtauzin** [comment]
BTW, I think it might be nice to have those functions all documented in the API reference. Happy to take care of this if you agree!

File: vizro-core/docs/pages/user-guides/kedro-data-catalog.md (hunk 2 above)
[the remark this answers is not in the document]
Comment (antonymilne) [reply vizro-core/docs/pages/user-guides/kedro-data-catalog.md]: That's an interesting perspective @gtauzin - I would definitely iike to discuss more how you use vizro in relation to kedro, but let's follow up on that in #1008.

@stichbury very good point, though I don't think this will go sufficiently into the inner workings to warrant a new page. I would imagine it more as a Note callout. I have a few other modifications I'd like to make to the Kedro data catalog docs anyway that are unrelated to this PR, so let's try to get this current PR merged soon and then I'll make the other modifications to docs in a separate PR for you to review if that's ok?

**Review by antonymilne — APPROVED** [review]
Thank you for all your work on this! CI is doing something a bit strange, not sure what's happening exactly, but I may push a few commits on to this branch. Leave it with me to get a second approval on this and get it merged.

@gtauzin 
> However, hatch run all.py3.11:test-unit ran fine! Out of curioisity, is it possible to run pytest directly? That would have been easier to iterate faster.

This is strange - not sure what's happened there but basically the way to think of these hatch environments is that they are ephemeral and (thanks to uv) quick and easy to remove and recreate. So if something unexpected happens then your best bet is `hatch env remove` or `hatch env prune` and then to just do the `hatch run` command again, which will recreate your environment in a few seconds.

It is indeed possible to run `pytest` directly - if you have a look at hatch.toml then you'll see that the `test-unit` script does exactly this and is really just a very thin wrapper:
```
test-unit = "pytest tests/unit {args}"
```

The beauty of hatch is that it will (in theory) seamlessly handle all the dependencies you need to run this in the right environment, so running pytest directly won't really be any easier.

In reality I think you have been unlucky here because 99% of the time hatch has made our dev process hugely easier, faster and more reproducible. There is very rarely inconsistency between CI and what happens on your local machine, I promise! Once you get used to hatch then I can almost guarantee you will love it 😀 

> FYI, kedro now support python 3.13 (at least according to the badges on the main README.md)

Interesting! Let me take a look and see if we can do this. I also realised we need to bump our kedro dependency - I'll do that in this PR too.

> BTW, I think it might be nice to have those functions all documented in the API reference. Happy to take care of this if you agree!

Yes, I also thought this! Please do go ahead with it, but in a separate PR so I can try and fix CI in this one and get it merged.

**Comment by antonymilne** [comment]
@gtauzin I don't seem to be able to push to your fork -  is there something you can do from your side to give me permissions to do so? 🙏  e.g. following instructions on https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

**Comment by gtauzin** [comment]
> @gtauzin I don't seem to be able to push to your fork - is there something you can do from your side to give me permissions to do so? 🙏 e.g. following instructions on https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

Strange, it is checked from what I can see. 

image

Let me give you access to my fork directly.

**Comment by gtauzin** [comment]
> > However, hatch run all.py3.11:test-unit ran fine! Out of curioisity, is it possible to run pytest directly? That would have been easier to iterate faster.
> 
> This is strange - not sure what's happened there but basically the way to think of these hatch environments is that they are ephemeral and (thanks to uv) quick and easy to remove and recreate. So if something unexpected happens then your best bet is `hatch env remove` or `hatch env prune` and then to just do the `hatch run` command again, which will recreate your environment in a few seconds.
> 
> It is indeed possible to run `pytest` directly - if you have a look at hatch.toml then you'll see that the `test-unit` script does exactly this and is really just a very thin wrapper:
> 
> ```
> test-unit = "pytest tests/unit {args}"
> ```
> 
> The beauty of hatch is that it will (in theory) seamlessly handle all the dependencies you need to run this in the right environment, so running pytest directly won't really be any easier.
> 
> In reality I think you have been unlucky here because 99% of the time hatch has made our dev process hugely easier, faster and more reproducible. There is very rarely inconsistency between CI and what happens on your local machine, I promise! Once you get used to hatch then I can almost guarantee you will love it 😀

I see. I guess this is very similar to [tox](https://github.com/tox-dev/tox). I'll look into it, thank you! :)

> > BTW, I think it might be nice to have those functions all documented in the API reference. Happy to take care of this if you agree!
> 
> Yes, I also thought this! Please do go ahead with it, but in a separate PR so I can try and fix CI in this one and get it merged.

Sure, I'll get to it soon!


Thanks for the support throughout this PR!

**Comment by gtauzin** [comment]
> @gtauzin: I've seen this sort of problem before with github, where allowing maintainers access to your fork in the PR only seems to have very limited effect. I'm not sure what causes it but think it might be a bug on the github side. Have you ever had maintainers push successfully to your forks elsewhere before without needing to grant them access to the whole repo?

Never happened to me before, but I probably have less experience than you.
 
> Hatch does indeed operate a bit like tox (plus many other tools, all combined into one beautiful whole!). It makes lots of things really simple and quick and reliable for us. e.g. just now I tested that our kedro tesets pass across all Python versions in less than a minute using `hatch run all:test-unit -k kedro`. And I can pretty much guarantee that it will do exactly the same in CI and on anyone's computer.

Nice! :)

**Review by antonymilne — CHANGES_REQUESTED** [review]
I decided that we should still keep this compatible with older version of kedro (while enabling the new functionality in versions of kedro that support it). I'll push some more code to do this but let's not merge until I've done it.

**Review by petar-qb — APPROVED** [review]
Wow, @gtauzin!! 🚀🏅 Huge thanks for such an amazing PR!

The discussion you had with @antonymilne, both in [this issue](https://github.com/mckinsey/vizro/issues/988) and the PR comments was a great reminder of just how powerful Kedro is. After going through everything, it was super clear to understand the PR code and the reasoning behind it.

I have just two very minor (and totally ignorable) comments, and @antonymilne after you provide the support the older Kedro versions too, feel free to merge this PR! 🎉

File: vizro-core/tests/unit/vizro/integrations/kedro/test_kedro_data_manager.py [hunk 5]
```diff
@@ -3,23 +3,47 @@
...
+    pipeline = kp.pipeline(
+        [
+            kp.node(
+                func=lambda *args: None,
+                inputs=[
+                    "pandas_excel",
+                    "something#csv",
+                    "not_dataframe",
+                    "not_in_catalog",
+                    "pandas_parquet",
```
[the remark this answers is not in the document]
gtauzin: Yes, it will still appear. The pipeline is just used to add the factory-based dataset it uses as inputs or outputs.

antonymilne:
Thanks for the quick review @petar-qb!

@gtauzin I just pushed [0514fe3](https://github.com/mckinsey/vizro/pull/1001/commits/0514fe3cc6f97d8a3f585a1fdd7fb485c6c3af91) which hopefully is the last thing we need to do here. I'll check it over one last time with fresh eyes tomorrow morning and then merge. Thanks again for your work on this!

vizro-core/tests/unit/vizro/integrations/kedro/test_kedro_data_manager.py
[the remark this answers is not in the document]
antonymilne: Great spot @petar-qb! 🦅 This should indeed include the case that something is defined in the catalog but not used in the pipeline, which is currently doesn't 👍 I'll fix it.

vizro-core/changelog.d/20250208_114146_4648633+gtauzin.md
```diff
@@ -0,0 +1,39 @@
...
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+### Added
+
+- Kedro integration function `datasets_from_catalog` can now handle dataset factories for `kedro>=0.19.9`. ([#1001](https://github.com/mckinsey/vizro/pull/1001))
```
antonymilne, line 21:
```suggestion
- Kedro integration function `datasets_from_catalog` can handle dataset factories for `kedro>=0.19.9`. ([#1001](https://github.com/mckinsey/vizro/pull/1001))
```

vizro-core/src/vizro/__init__.py
```diff
@@ -23,7 +24,9 @@
 # This would only be the case where you need to test something with serve_locally=False and have changed
 # assets compared to main. In this case you need to push your assets changes to remote for the CDN to update,
 # and it might also be necessary to clear the CDN cache: https://www.jsdelivr.com/tools/purge.
-_git_branch = __version__ if "dev" not in __version__ else "main"
+
+
+_git_branch = __version__ if not parse(__version__).is_devrelease else "main"
```
antonymilne, line 29:
```suggestion
_git_branch = __version__ if not parse(__version__).is_devrelease else "main"
```

vizro-core/src/vizro/_vizro.py
```diff
@@ -209,7 +210,8 @@ def _make_resource_spec(path: Path) -> _ResourceSpec:
     # This would only be the case where you need to test something with serve_locally=False and have changed
     # assets compared to main. In this case you need to push your assets changes to remote for the CDN to update,
     # and it might also be necessary to clear the CDN cache: https://www.jsdelivr.com/tools/purge.
-    _git_branch = vizro.__version__ if "dev" not in vizro.__version__ else "main"
+
+    _git_branch = vizro.__version__ if not parse(vizro.__version__).is_devrelease else "main"
```
antonymilne, line 214:
```suggestion
    _git_branch = vizro.__version__ if not parse(vizro.__version__).is_devrelease else "main"
```

vizro-core/tests/unit/vizro/integrations/kedro/test_kedro_data_manager.py
```diff
@@ -1,25 +1,66 @@
...
```
antonymilne, line 49:
```suggestion
                    "pandas_excel",
                    "something#csv",
                    "something_else#csv",
                    "not_dataframe",
                    "not_in_catalog",
```

vizro-core/tests/unit/vizro/integrations/kedro/test_kedro_data_manager.py
```diff
@@ -1,25 +1,66 @@
...
+                ],
+                outputs=None,
+            ),
+        ]
+    )
+
+    datasets = datasets_from_catalog(catalog, pipeline=pipeline)
+    # Dataset factories only work for kedro>=0.19.9.
+    expected_datasets = (
+        {"pandas_excel", "pandas_parquet", "something#csv"}
```
antonymilne, line 61:
```suggestion
        {"pandas_excel", "pandas_parquet", "something#csv", "something_else#csv"}
```

antonymilne (approved):

astrojuanlu:
On behalf of the Kedro team, thanks for your contribution @gtauzin and thanks Vizro team for such a swift review!

PR #1002, opened 2025-02-10
[Feat] Enable variant argument inside `Container`

huong-li-nguyen:
## Description

- Replace `html.Div` with `dbc.Container` in the underlying model. We enable `fluid=True` to ensure containers expand to 100% of the available width, following Bootstrap's container behavior: [Bootstrap Containers](https://getbootstrap.com/docs/5.0/layout/containers/).
- Add support for the `variant` argument in `vm.Container(..., variant="filled")`. Available options: "filled", "outlined" and "plain" with "plain" being the default
- Ensure background color consistency by matching it with the Card background color.
- Update docs and tests

**Note on "elevated":** We need more testing on both the design and development side. Box shadows are not visible in dark mode, so to make "elevated" work, we'd have to use both shadows and a background color. But in a dark theme, that would end up looking like "filled." Plus, we'd need dynamic margin adjustments to keep shadows from getting cut off. After discussing with J., we’ve decided to put "elevated" on hold for now since it wasn’t part of the original plan.

## Screenshot
Screenshot 2025-03-01 at 22 02 41
Screenshot 2025-03-01 at 22 02 19

## Notice


vizro-core/src/vizro/_themes/_common_template.py
```diff
@@ -88,5 +88,7 @@ def create_template_common() -> go.layout.Template:
         yaxis_zeroline=False,
         annotationdefaults_showarrow=False,
         annotationdefaults_font_size=14,
+        modebar_bgcolor="rgba(0,0,0,0)",
```
huong-li-nguyen, line 91: I think we should generally implement this regardless of this PR, as the modebar takes too much space on the top of the chart. It looks cleaner to move it to the right.

Screenshot 2025-02-10 at 10 48 18

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -34,6 +35,14 @@ class Container(VizroBaseModel):
...
     title: str = Field(description="Title to be displayed.")
     layout: Annotated[Optional[Layout], AfterValidator(set_layout), Field(default=None, validate_default=True)]
+    # TODO: Refine API later, just for testing now
+    # TODO: Check what bootstrap adds to .container and .container-fluid - currently overflows, adds too much padding...
+    # We could be very strict and just allow bools, but my preference would be being semi-strict and allowing
+    # classNames. This way we can allow for more flexibility in the future. Default styling should then come from
+    # Bootstrap utility classes but they also have the flexibility to override them.
+    # border: bool = False
+    # background: bool = False
+    classname: str = ""
```
huong-li-nguyen, line 45: @antonymilne—I think exposing `classname` as an argument makes most sense to me. We still provide default styling, but we do it via the Bootstrap utilities (see example app). This also gives users the flexibility to deviate from it. 

It makes me generally question whether we should enable `classname` as an argument to all models where it makes sense. This would enable usage of bootstrap utilities and doesn't require the user to know the specifics of CSS.

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
```diff
@@ -102,3 +102,16 @@ label[for="theme-selector"] {
 .card-text > *:last-child {
   margin-bottom: 0;
 }
+
+.container-fluid {
```
huong-li-nguyen, line 106: This is taken from the old className attached to our previous containers `page-component-container`

stichbury (commented):
Looks perfect from a docs perspective 🌟

vizro-core/docs/pages/user-guides/container.md
antonymilne, line 1: Please can you read through this page and update other bits where appropriate? 🙏 Here's two cases I saw just skimming through:

> While there is currently no clear difference in rendering, extra functionality will be added to the `Container` soon (including controls specific to that container), enhancing the ability to manage related components.

Maybe worth mentioning possibility of styling the background here.

```
Here are a few cases where you might want to use a `Container` instead of `Page.layout`
```

Definitely add to this list.

Maybe also worth skimming through the page on layouts to see if we should mention it there when we link people towards this page on `Container`.

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -157,4 +157,75 @@ To create nested containers, add a `Container` to the `components` argument of a
 vm.Container(title="Parent Container", components=[vm.Container(title="Child Container", components=[vm.Button()])])
```
antonymilne, line 157:
```suggestion
vm.Container(
    title="Parent Container",
    components=[
        vm.Container(
            title="Child Container",
            components=[vm.Button()],
        )
    ]
)
```

This in one line at the moment makes it way too hard to understand the structure of the nested contains.

antonymilne (commented):
Looks good! I would even question whether we want the default to have `background=True` rather than `False`? No strong feelings for me, and just leaving it as you have it is definitely easiest and more consistent with our existing dashboards. I'm curious whether you would set `background=True` more often than not though?

What does this look like inside tabs? Is that something someone might want to do or does it look weird?

antonymilne (approved):

huong-li-nguyen:
> Looks good! I would even question whether we want the default to have `background=True` rather than `False`? No 

Joe asked me the same, but I wouldn't recommend enabling it by default. It looks off within the Tabs 😬. Additionally, we can't be sure how people have been using the Containers. In my experience, I've used vm.Container to manage different scroll behaviors within a chart, where the background color would also look out of place. I've also applied it in a FlexLayout, where the background color would similarly be unsuitable.

For me it's a very specific use-case where you want to "visually distinguish sections", so my gut feeling would be having it turned off by default to be less disruptive to how people could have used it until now.

antonymilne:
> Joe asked me the same, but I wouldn't recommend enabling it by default. It looks off within the Tabs 😬. Additionally, we can't be sure how people have been using the Containers. In my experience, I've used vm.Container to manage different scroll behaviors within a chart, where the background color would also look out of place. I've also applied it in a FlexLayout, where the background color would similarly be unsuitable.
> 
> For me it's a very specific use-case where you want to "visually distinguish sections", so my gut feeling would be having it turned off by default to be less disruptive to how people could have used it until now.

Cool, sounds good! Let's definitely keep as `background=False` by default then.

vizro-core/docs/pages/user-guides/container.md
huong-li-nguyen: Thanks! Updated ✅

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
```diff
@@ -102,3 +102,16 @@ label[for="theme-selector"] {
...
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  padding: 0;
+}
+
+/* Introduce vizro utility classes here */
+.bg-container {
+    background: var(--surfaces-bg-card);
+    padding: 1rem;
```
petar-qb, line 116: Why `padding` is added?

This way, for the following configuration (two same tabs with the `background=True` set in the first tab only):
```py
page = vm.Page(
    title="Page with subsections",
    components=[
        vm.Tabs(
            tabs=[
                vm.Container(
                    title="Tab 1 Title",
                    background=True,
                    components=[
                        vm.Card(text="""Hello, this is a card with a [link](https://www.google.com)"""),
                    ],
                ),
                vm.Container(
                    title="Tab 2 Title",
                    components=[
                        vm.Card(text="""Hello, this is a card with a [link](https://www.google.com)"""),
                    ],
                )
            ]
        ),
    ...
```

The first tab looks like this:
image

and the second one like this:
image

petar-qb (commented):
Hey @huong-li-nguyen @antonymilne,

I’m curious what was the reasoning behind exposing an attribute for setting the container background colour? It seems like more users have requested setting `background=False` for `vm.Card` rather than `background=True` for `vm.Container`.

Would it make sense to also expose `background: bool = True` (defaulting to True) for `vm.Card`, allowing users to set it to `False` when they for example need to display plain text over another component or whatever?

huong-li-nguyen:
> Hey @huong-li-nguyen @antonymilne,
> 
> I’m curious what was the reasoning behind exposing an attribute for setting the container background colour? It seems like more users have requested setting `background=False` for `vm.Card` rather than `background=True` for `vm.Container`.
> 
> Would it make sense to also expose `background: bool = True` (defaulting to True) for `vm.Card`, allowing users to set it to `False` when they for example need to display plain text over another component or whatever?

That's a great question, @petar-qb!

Introducing a `background` argument inside the `vm.Card` wouldn't be sufficient because the Card includes additional styles like box-shadows and effects for hover, focus, and selection. If users want to disable the styling on a Card, they are likely just looking for a simple `Text` component.

@antonymilne and I previously discussed the possibility of introducing a pure `Text` component. For example, I created a `Markdown` component as a custom component in VV: https://github.com/mckinsey/vizro/blob/main/vizro-core/examples/visual-vocabulary/custom_components.py#L44 , because I also already faced the issue, that I wanted just a simple Text component without any styling.

However, it seems more intuitive to introduce a dedicated Text/Markdown component rather than trying to strip all the styling from the Card. The styling is what defines the `Card` as a `Card`. Without it, users would likely just need a pure Text component.

Regarding the `background` argument, we're still discussing it 😅. I'll provide you with some additional context in our chat. Initially, we decided to include a `background `argument because Containers often have a background color in dashboards. If our default styling was just to enable the background color, then a background argument would have been sufficient. However, we're still iterating on the design.

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
huong-li-nguyen: Padding is added only if you enable the background color. Without it, charts and titles inside the Container would extend to the edge, which looks off (see below). The `Container` itself without a background color comes without padding.

I discussed with J. whether users might insert `Cards` into the `Container`, and we concluded that most users likely wouldn't do so. What users probably look for is a simple `Text` component, without any styling.

Screenshot 2025-02-18 at 10 46 40

petar-qb:
> However, it seems more intuitive to introduce a dedicated Text/Markdown component rather than trying to strip all the styling from the Card. The styling is what defines the Card as a Card

This solves all my questions about the vm.Card.background attribute. Thanks for the explanation. 👍

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
antonymilne: I think this is a great point @petar-qb. If `background=True` (or whatever it becomes) isn't a sensible layout when `vm.Container` is inside `vm.Tabs` then should we ignore it/override it? We already have some slightly different styling logic controlling the display of the title different when `vm.Container` is used inside and outside tabs, and I'd be happy to add more customisation like this if the container styling arguments only make sense in the context of `vm.Container` used outside tabs.

Would be curious what @huong-li-nguyen and @Joseph-Perkins think are desirable style options for containers that are nested inside `vm.Tabs`. Does it seem correct that currently it has a background by default?

antonymilne: Ahhh yes sorry, I misread something above so what I said didn't make any sense 😅 Ignore me.

antonymilne: Though I guess the point still stands that if we have a more general `styling: enum` argument then we should check that all possible options would ideally look sort of reasonable inside `vm.Tabs`. For `background=True/False` it's ok already, but just something to check if we add others. Not a biggy though.

huong-li-nguyen: They should all look fine :) Because all the additional styling options will look good as stand-alone (meaning there will be sufficient space to the borders), so it doesn't matter whether they will be added inside or outside the Tabs. It will all look fine 👍 

The only important thing is that if no styling is applied (meaning no borders are visible), we strip all margins/paddings because the spacing will then come from the page-component-container instead of the component itself. It's only when the borders are visible (let it be via elevated, filled, or outlined) that the spacing needs to come from the component.

nadijagraca (approved):
Looks good to me. 
I agree with adding padding when `background=True` as it looks much better as stand-alone. ✅

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -23,6 +24,8 @@ class Container(VizroBaseModel):
             has to be provided.
         title (str): Title to be displayed.
         layout (Optional[Layout]): Layout to place components in. Defaults to `None`.
+        theme (Optional[Literal["filled", "outlined"]]): Predefined styles to choose from.
```
huong-li-nguyen, line 27: @antonymilne @maxschulz-COL @Joseph-Perkins 

I don’t want to overcomplicate things—if everyone prefers `theme`, I’m happy to go with it. But after implementing it, I felt it wasn’t the best fit 🤔 

I see the logic in keeping it consistent with Dashboard, where we also choose from predefined styles. But "theme" feels misleading because it’s typically used for global styles like light/dark mode, while here, we’re just defining a styling variant for a single component.

Just to mention a few alternatives used in UI frameworks:

- variant → Standard term for predefined styles (e.g., Bootstrap, Material UI).
- style_variant
- appearance
- preset
- scheme

where variant is probably my preferred one. I don't strongly disagree, though, so if you guys want to keep `theme`, then I'll leave it as is.

huong-li-nguyen:
@antonymilne and @petar-qb or @nadijagraca - could I ask you guys for a new review? We decided to not introduce the `background` argument but have a `theme` argument instead that allows the user to choose from a pre-selection of styles.

vizro-core/src/vizro/models/_components/container.py
maxschulz-COL: Hm I see where you are coming from - it just feels indeed very global. I have thought about it for a while and I think we could do the following:

1. Go with our actual preferred name (I think I prefer `style_variant` or `appearance`, only issue i have with `variant` is that it's a bit general)
2. Add a validation alias for `theme` either now or later (https://docs.pydantic.dev/latest/concepts/fields/#field-aliases) - I would probably not add it for now

This would allow us to have the name we like and that is intuitive, but if we want to later allow passing a theme, then people could pass it under the more intuitive `theme` name without us needing to add an extra field. While parsing we could sort out if people passed a preset style or an entire theme.

Wdyt?

antonymilne: Yeah, I think this is a valid anti-`theme` point and I'm happy to go for something different.

Let me give your suggestions my personal rating, in order from most preferred to least:
- variant - I like this, doesn't feel too general if it's common in other frameworks
- appearance - ok
- style_variant - maybe could be convinced but too wordy for my liking
- preset - no
- scheme - no 

Another suggestion: `format`? I quite like that but maybe it sounds wrong to others. Overall I'd go for format, variant or appearance.

Agree with @maxschulz-COL that we can alias this later if we need to, but for now let's just for with whatever our preferred name is, so probably not `theme`.

vizro-core/changelog.d/20250217_105619_huong_li_nguyen_styled_containers.md
```diff
@@ -0,0 +1,46 @@
...
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+### Added
+
+- Introduce the `theme` argument inside `Container`, allowing the selection of predefined stylings for the container. ([#1002](https://github.com/mckinsey/vizro/pull/1002))
```
antonymilne, line 21:
```suggestion
- Enable styling of `vm.Container` with a new argument `theme="outlined"/"filled"`. ([#1002](https://github.com/mckinsey/vizro/pull/1002))
```

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -2,9 +2,9 @@
...
 
-A [`Container`][vizro.models.Container] complements the idea of a [`Page`][vizro.models.Page], and the two models have almost identical arguments. [`Page.layout`](layouts.md) offers a way to structure the overall layout of the page, and a `Container` enables more granular control within a specific section of that page.
+A [Container][vizro.models.Container] complements a [Page][vizro.models.Page], and both models share nearly identical arguments. While `Page.layout` provides a method for structuring the overall page layout, a `Container` offers more detailed control within a particular section of the page.
 
-While there is currently no clear difference in rendering, extra functionality will be added to the `Container` soon (including controls specific to that container), enhancing the ability to manage related components.
+Unlike the `Page.layout`, the `Container` includes a `theme` argument, allowing you to choose from predefined styles to visually distinguish it from the rest of the page content. Additional functionality will soon be added to the Container, including controls specific to it, which will further enhance the management of related components.
```
antonymilne, line 7:
```suggestion
Unlike `Page`, the `Container` includes a `theme` argument. This enables you to choose a style for your container to visually distinguish it from the rest of the page content. Additional functionality will soon be added to the Container, including controls specific to it, which will further enhance the management of related components.
```

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -154,7 +155,88 @@ Containers can be nested, providing a hierarchical structure for organizing comp
...
+    === "app.py"
+        ```{.python pycafe-link}
+        import vizro.models as vm
+        import vizro.plotly.express as px
+        from vizro import Vizro
+
+        iris = px.data.iris()
+
+        page = vm.Page(
+            title="Container with background color",
```
antonymilne, line 183:
```suggestion
            title="Containers with different styles",
```

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -154,7 +155,88 @@ Containers can be nested, providing a hierarchical structure for organizing comp
...
+        dashboard = vm.Dashboard(pages=[page])
+        Vizro().build(dashboard).run()
+        ```
+
+    === "app.yaml"
+        ```yaml
+        # Still requires a .py to add data to the data manager and parse YAML configuration
+        # See yaml_version example
+        pages:
+          - title: Container with visual distinction
```
antonymilne, line 208:
```suggestion
          - title: Containers with different styles
```

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -154,7 +155,88 @@ Containers can be nested, providing a hierarchical structure for organizing comp
...
+        from vizro import Vizro
+
+        iris = px.data.iris()
+
+        page = vm.Page(
+            title="Container with background color",
+            layout=vm.Layout(grid=[[0, 1]]),
+            components=[
+                vm.Container(
+                    title="Container I",
```
antonymilne, line 187:
```suggestion
                    title="Container with background color",
```

vizro-core/tests/unit/vizro/models/_components/test_container.py
antonymilne, line 1: Please can you add some new test cases to cover the new `theme` options?

petar-qb (commented):
First of all, I'm so happy we abandoned the `background=True/False` idea! 😅 

As `theme` feels misleading, I'll use `style_variant` in the rest of the comment as it feels  much more natural to me 😄 

Before merging this PR, I think we should establish a rough plan for `style_variant`. In future I see that we will upgrade this solution to another level and that we will have a predefined set like `"transparent"`, `"filled"`, `"outlined"`, `"elevated"`, `"rounded"`... In that case each component could have a default like for example, `"transparent"` for `Container`, `"filled"` for `Card` and `Alert`, while `Button` might include additional options like `"link"` and `"disabled"`.

Open questions:
1. Should we default Container to `"transparent"` instead of `None`?
2. Should we align our `style_variant` values with existing standards like Bootstrap, or define our own? Caution: users could require something like `"rounded-elevated"` 🤔 

While we don’t need a full plan before merging, having a clear direction will help ensure consistency. Looking forward to your thoughts! 🚀

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -34,9 +37,31 @@ class Container(VizroBaseModel):
...
+        description="Predefined styles to choose from. Options are `filled` or `outlined`. Defaults to `None`.",
+    )
 
     @_log_call
     def build(self):
+        # TODO: TBD on how to encode 'elevated', as box-shadows are not visible on a dark theme
+        # It needs to be properly designed and tested out (margins have to be added etc.).
+        # Below corresponds to bootstrap utility classnames, while 'bg-container' is introduced by us.
+        # See: https://getbootstrap.com/docs/4.0/utilities
+        variants = {"outlined": "border p-3", "filled": "bg-container p-3"}
```
antonymilne, line 51:
```suggestion
        variants = {"outlined": "border p-3", "filled": "bg-container p-3", None: ""}
```

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -34,9 +37,31 @@ class Container(VizroBaseModel):
...
+
+        return dbc.Container(
+            id=self.id,
+            children=[
+                html.H3(children=self.title, className="container-title", id=f"{self.id}_title"),
+                self._build_inner_layout(),
+            ],
+            fluid=True,
+            className=variants.get(self.theme, "") if self.theme else "",
```
antonymilne, line 60:
```suggestion
            className=variants[self.theme],
```

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -34,9 +37,31 @@ class Container(VizroBaseModel):
     )
     title: str = Field(description="Title to be displayed.")
     layout: Annotated[Optional[Layout], AfterValidator(set_layout), Field(default=None, validate_default=True)]
+    theme: Optional[Literal["filled", "outlined"]] = Field(
+        default=None,
+        description="Predefined styles to choose from. Options are `filled` or `outlined`. Defaults to `None`.",
+    )
```
antonymilne, line 43: I'm not absolutely 100% sure about this but like 90% sure that it would make more sense if we give the container default theme a name even if it's just "none". 

e.g. if we think of applying these styles to `vm.Card`, let's say we have the same 3 styles available but the default is "filled". Then what would options look like for `vm.Card` if they wanted to be able to select the same styling as is the default for a container: `theme=filled/outlined/???". We'd need a name for "???". And it's maybe a bit confusing if `theme=None` means something different for `vm.Card` vs. `vm.Container`. It would make more sense for the options to be the same but the default to be different.

```suggestion
    theme: Optional[Literal["none", "filled", "outlined"]] = Field(
        default="none",
        description="Predefined styles to choose from. Options are `none`, `filled` or `outlined`. Defaults to `none`.",
    )
```

Or do we even have a good description of what the default is other than `"none"`? Like "transparent" or something?

antonymilne (commented):
This is amazing, thank you for all your work on it! I know it's taken a lot of discussion 😅 It looks great and happy to approve now subject to a few things being resolved:

* finalise argument name
* decide what should be the possible values for `theme` (or whatever name it has) - something other than `None` might be better
* a couple of new tests

There's two other issues worth considering but I don't think we need to do anything about them before merging.

1. Implications this has for `vm.Card` and proposed new `Text` component - let's discuss in our 1-1 tomorrow
2.  Check out this [Pycafe example](https://py.cafe/snippet/vizro/v1?#c=H4sIAFeIxWcEA-2c3XIbR3aAX2UCa72URQ7n_4cxU5Epy9Sasr22l6uVoKIHwJCEBQLYwVAU5dWmKleJXyN3m9zkAVK5iC5SqVQu8wB5gbxCvtN_GPDHXNcmazkrcDDT53Sf3z7dfWaInm96w9mo7m31-urvy3rRetV8rqH-tD8dn8xnTes9H79sZv4JbScLr1p4z08uVM0ns3Zy7tcv5k29UE3mL_rTw2Z2oht4pvW-AIpxM15427TyR1Vb-QKu3ZaKUX3oHa7d3upPPT5N3Z42U-_J8xN_ZzZtq_G0btZ0lXzacTupt_u9T2Zee1yf1P3e-rJyOEPmtJ62i-0nS6x84PZRU82P1w7HR6dNvY0Wi2HVtvAWRda9F_Bc1PNqcnA2HrXHsPXOHWpST480bjibzBrBz-vhuF70e7dvd-Q_NWWLu8GGT0_bCVWjN9IG-SgP03J2UVFLcIOB98eTSYdKPm-ieYeX1bz9VCKzP51XRzVBi26fUTImWvME5Z2N22NvcTpY1MN2PJsiz7CZVOf4bRvSPVVaO2rGo-0nT4J1L1z3oqdPrVpXuwS6pWutxE7Ud8kYPZaZfK4k7QSbM7zTr9-Xnetax8x58XpOOFWNePHqQXs2u9azTh7sdE2H6XVOWqLlc4mV9-DBA8fOfq5maz-w_yOEpHy6YWnrBNdx13FTXx-KSyu_aM8ny1nxavu-n9scM_u5mqn9_DBOk4-NxE5YLxt0ib-f_T8-B7ihuKy2pCaoRtXieDCrmpGOqHsWXJNYwyq5rHs68M5mtiQh-FSFZH-q1vW12_7gdDwZrTl-t_3mdCrrem-919S_Ph03qISjyDiO23a-2NrcnJ_7w-qw3jw63qyadnxYDdvNk-Gz8XRRn2-q3GETlMylm_BabIZxFpRBlCZJvDkfz3WTjcAP_Tj1R_XzYGN-Hm9M6ZCNanrunx1PkN2ezyXLUW0B5-ft8WwKYn4-G41H9cbzwI8yP6RKz9O9rW96z-tmgdjeVoTus1n7-exM0IZVA7TeGx5jblPT6ImraStm_5ZK1Z29rTAN1nsn4-kvNRhraLceHx0jR8DxCDJ6qf4Arou6cdF2jQRpujHQbWkyr4Rvr_fq6at112aphRGUJH5UBlmaJ3GZJFmcXc0buiVLUkF_fr7KWBtuuYZp6udBGSdRmgVpGmfJtWy1Ojc7whh2kxOs_SvKOTHHVr88_X1kMhpBV5ObhNp2IlX-Xq2rziCeaTOtTqTNZrVAh8XmcMH3dNHOTnyKMDxtJlTbuG-qM_-IbOF0cIoZQwQzMHzmkovRn-ZFMYzTKAiqOgnTvB5ko2JQR0VJf46y_DCqgqLOCzMShrOm3qxfVCdz9NpcDJuqHR4fMDCu1WvWjI-s8eiAju_IAN84rqsRM943MlXMq9FoPD3amNSH7ZZXzF_8eX_6imEtvr9g92H1fIw5Pqc3x-hVpS5b_N___E-fffJRv2EuOw2CsNLXIFg5Nw92732-gpFzuN4BHGqgClkHac7nr5B12sWYczvf_dViBSNnxSeMO5jL-FCdA_hyKHyhMZfO4eLzjz7QcKf10BRWmupzcnT34d0O3CELBhQqVUp1XeecUvfg3t0vXygoBPrPv-UkUDj8UF0SVSUM--0e549sffCC0r2PVXU0plzy1cdf8T26d5_zVf3z9vydZ7wmsecOwYaDv2yfOIw-nn_MSVdqhD5e6svi411d4Pj3f-SkQz3MEsrb-106KQfZe5S-fufRZ_0-hUcnCqn7ftCfmlahEHN8dffnnIVcYcP1NQNmBqUC60u-HBIG-vi3rh6a38-vUCXaooQqAl-hDOcwMm2XCn3OuaOQGC_jZUWhli-HKOTx_Y-uOrcoX6GL4a_KuS67QzW-iNx2JX3ErqSPZ3wVN1FIH9LbF1BuqKd3NWLhEOosI5XjmQK6jtFuNUL7rcat39oXyLRY7_cUaCu9buXeu926vRXCvVXCPU0o5UDNrWEh8wPHO48kiB89lQD9SjegpI-v-AoKhcVcAfXxtWoohnNoi3RZxYHAyvQwV2fjAIU39vve_sKqStWK6hSl1miuzkPdYAXl7T_rUFzm8Mw2BzLDS8HxusyFoqgnJ1P-s06533s0N0Ud2JDZ_n6gLwplek48x5Fr4dZRn66CD_lyKFywbkvC4X0p2djleE_VmFDd0BeZx49FdKMqI8dgxbUUxAvvGdp-o6-f8j2W0NS9tip9y5FyvP6dwovHOKQdh8heE9ma_HrZn7_-naHsH-qrSF4TyS8Vwarkv3CEHEayaN7vdYzeFcGa-nrBX77-nSZUFn_Ad9eMxE5AqIL0hiqE62rJDuM7v_2pdPFP9n9zS6639n_zobQZiHAOufQHgnpxTxV3pCiTUn8gMiSUzLAQcpmXB7flJNwDcYPWQJbbwaYCzcTWlxEzkNmTQ03JQsyhnKepKo0Rw_W8zmGvc6H_hRJjffjtHRvy__I36pr-Ul1i57ngzn3TE_5P1YDpPer3f2ZrC3G2Ph6Jj7_6tVJW22K6VmRp5WxfS1lh39UImT4G3UqFkJIaQwPpKs1BgaqkcD_hxFGLYV_r8sevv9VduLRxbRXcXQWPHSjOFiDUM1FoXQPa1fy8g9QYKVzAxXdcyfkxvhOpa7D--ls9o4lI6WSCUDWg-LWZYc2aqvCRIZQ1keP1t6ZC0asazUTRS8RxfKIvqja4o6cH8akUguz1twqjGBR8TYWS8N6VGmgO12tguCgGV6qg5oylCtHrb33N4Y4ob9uJCT-7UgFNf1mBOwpveCjyK8WvSFeh15_-1iFNU4MxU4XCyElwwtni1CS0guH0rgxlAaROQkJNEQ7x_ruPZOgb4KePpKAWeY2hIOYqwKzwd9Q0saHWKFXSi7Mqevuy1nz18F1zvWWuNNEFGqh5TbGqjENl3hBJHcWCRGUYXSfI6FVVOuEI0qGadUytc4jVXS3iCiOsf6IMs0u49nohMzprt8iQae-WKxmcajUyaA2ozENAvs4DfLt4WZ-_kmxMEuJHd-36oQHhYlINhZBReKpK73PmkBZ2fbagjWa7SFtQ8HqlVhTKRXqpXsKyVkvOuSQjaP_170SmwNNmrHADUYTjN48VeSZ5gJA952uJ_v4XAAJPG907ag7RhyZLTPrgDkOqAVGHoz_9-rbCW5kvf6aIlUx9uBCxpCajh1T3XUeyJv4Oye84W18-UFgr97f6Xr0jVzRcEt5x9qp2K_dKb8_ffdaeO9AXc2tyzLfbpnt-8OEnMj9-IAPoqodG4xOeOC025YHj4vnRm_Pc6JJelx8dvU8F__fj-SLP2kMk93uefvwInCnw-bg--2AmD_oDL_CkjacreHg4AStPq4FenEymC0CxGqPPzs78s9ifNUebUASbyOn3_qI_fb_hn4s3SDSMB5Nq-Kzf2xQqeabrjUA-TMrQT0IvCTM_z_aShIffQe5Fee5HYb6bcOVB8V4axjxST7y4iPycW7A0KfwiiFy7tMj8NAj3Up67M_Uadrua-eOlDmfH4xbrLuiQRalfhIkX8QA5TZOdLCz9KCkt7GVIL0LA2E-LFHeVflJmHk38LMl2eNjsx-icyOPngmp0FRuwJQfv4DgBluYGjgpguBvyCCGpNDfcI_6PENDcCMeiMKPa6hYmflgizequ4Z0sjuBWWNjL4tzPoxjY0CchYpbcE0FDbqXj2DhCnFXOwlZ5C1vjLL013rC3rrHSreusdtazVnsNP36YBTkdW5gO3Lf9myXihHLf9KtpdXPHZqWflpZZXKBMEe3lontWeHSCH8TlbpYHfhInTlheYEMS78cBtpXpHv9T8FXsZX4ZF1THfpQ6VZSIGzUpQmQEsVXFiiqSyE_ydK_AfUUaohKxUUQ7RRr4ZYgfywR_lR5dAxnxjglRGnlFltEOP0rbPGXM4MOcMZCFXlHgy5AIyfEZQ6socz-O4aXh_TjEvCjaLYRXkENfUu8ZtFfkEf-cyYFjH64iKSRa4zBF0xLNSjTDeSEjk2tBzwVJ5MVR5IexwAw5ejgmgMKY9gRMyEiNCaBQtCFgEnqc_9L4SZEvYSIkzAvrVuOvG_1a4siyxK8Sz8RnmSWwpcs17JWpWCPxRF8jviQ-0TaBfZFGO2VMKEQ0DxI_y2KvFCvK1ItLrAZPHxB3WINPo3SnJBIy0HGOMYQzU5OfMkriDK5QWxCbyqKkuYGJ-jTJqS-Rhi8Yw3gf7pkflLADHwTFjkhnxqPXMUr4xSGzGZ1MBxEMonuuYFwl2mBbSgzQ76WYiul5TrXp9zIv_Ch3_b5TMhmm5bI-JMeT0WHpQ-4J8JVjT6-ljG0nPgzi2F8qB4gVS-XDAKujzBkHzESGK43xDjbOcbD2HaAmN6517LXnjXDbL1Y3229Wd9uvyjZsdN2O7TGgjYrVqHnswogxxXpDTzA2YnxpYOgZUzAQbRKiHP7SVWhbMFmL8XgBOdBAF4k-EV7AOPRhqnGwM97AWBWqrtDkBIB4yjAn-HJC2soOCj8jmkS3JMNxrBsMTNvNGmQAMHEH0i26FgUS5nVHLBMxC6TljT658FKC6VZmFWi1Vg40SjvYGOWIjdGOt3GKk22cJqpldJpz6qrTbxzrMgHLyhgyYZVBYqeKkDjKkr0LtR1u70RRGt6_b_hJ6sL1qsRvMjuaHYyq5tmblfddVOumtI_8DONdEpYq8ELaRxtPVxgnfc-0b9krCcEaS7Sm6V6csizJDBWxKIbRblxKOhfvJYRXQBITRyxnRb6XECmlTD2mHQsBERWS_jFFsczBa9fw7fTj1VGRkrmQ7JDgBKzm4Q4RKEPdgB7sGVmIYgAwvaRcWMtZLyWP3Em5FjLnBSGrj1RHfpknXsg8kwddmBAPStobmJAuw4h6TR-SMqRL9ixkfpjG6KKFh0xqGSPK6BayxBaJZzXX4E6KM0KGj61lEJaR1BpaUqmMZdzxFl8JtRXN6ExEtlXNwlZ1BxvTDL0z3fB3rjHyreOMds6vq25__DCVbEoWtzTdt_2aMocnAZkX_Wnqb-7PTFY1YgU-kkSw6u1l_CBGwkXCK95NM-Yh0jUjhPkfBxT7KofMuQco6QPJWCQDKaQeg0rSOVFCM79RiYw1L8LpXWNwih-k8V4esSozDUpOy9qyk7MUkZ-S8yYom3o5qzarcowWLDU5LkhllcadRRjv5AmzdyHpPabhrhybVGaHiQndkGeECCuDgfcjJtEsDnfzlG4odnLSK3rQM2gvR2ghiz6raB7lShxTP7D03k6OVFCArJth7JHrqPuoiKxVpuAcKcyf5ADcgwQptsjMjy0FeRFZm9iakg9FTNEEwhImCSBZVD41vrrRpwWqicUxlgSkakVETiuZooa9gum8IM-AJyt-7BUqPae9zO9FvpMjM5FcDJNDbkFy5GI-qxWeYgLKi4REhOUJm-jzHXiDl8QWD4vt2Jix-rAwRqANGJUkiGlGaw2yeuAcauk-_BnRHQWjUZhH0k0oy-K9I8JLsuRI0lv6HN0y6QZuLWJyLNE9IouR26aC9KWQPIQlU2IpIpILel9Gke31Ajp-HmRhsmzyUAeShDPtEFqOnCkilQFh2XNbTEK9FM-8y0NUq10hGUSxVL5kyJO_O-NYLhkpRIUx3sHGNxa2rrP01rOWv_W8FW97xqpne86qb3tWzMsx0_U8Qc5dyzIyViPn8TKU8AdZh7rJgNjC6qYkk8wChQrubVhbSHEFDghfbpnIRJhXMUAMo54MhlstDGSq4K7WwtYBDsZBBAr8NH0kd45MgpZ_xFQT5uhj5Efqto169CNS-aGgstv1toZ3CmaMNJfmpp5lMuQuz5ETPSWzkWMvt2oyqq14puiCm0unnoWt-gZ25hl6Z77h79xj5Dv3oZ_cVDn3rrr_xpFP8umFiZ4tuJ82YIfs-2VoE0lsfBq9YSlaV68_1Rzt6od8F5KFtzmaVc3CVnUHvyk52jX9qdOoP1aOdrUSb3O0PyRHu9qnbmbXK-3bHI3us73uVm0Nv83R3uZoP8oc7eqR_7-Qo6nNs-fViWxy-aHzsq4ul3Oxd7yd49lsUXuDuj2r66kXbKRBHz3a2aLa8tL-VG24UductkjEgMcNPMZDavlhFJuMLrLY8kibvHbmMQH2p2zMUGxOqhdbXn51-yiIgo0g5BAyDaUbUSlboNr6QLEAm-g2FisMNVa3XekBMfjA7BVS_ydmw8wP3xdXa3W5V_gJGH8fahbik8Xx7Mxz-7fYdTZV-9Yq2YHlqb1nh-N6JBu8K-9Xdx_uSajWvmYjPldbviXEJ-OB3fT9GaDUGXBeyYYztUd85JDKoGu2kZsmYtN37ClfVvgn1ZR_2zcL20Iik25UyNWGeke7aeb2vYmyXZon_Z7s1ev3nrJBjg18Uue2rF9oeFTNCSM2zqy2dujvIjmg4_Nr6fxfn9bN-Vq_d15Xjbe9TUxK68v82vH8gq6CEcHKsMMDNv8Mny0OJrPpkTQa-Sf1pDVbHy2JbsNWvrZuxyfsAvyyOa3tvsHx6OB51cgNkdS7rYXPq8lprWtQ46NPP_1Idi72e3fvfsZvvFXp4eNPdOn-B_r6yf29R7r08Iv7X6K249UcyBhDhlaFRrZGpJg6Baiqy35YtVR75AJO--PZfKz9cA8G9xs4r32DrsP2tJr0e1veE-4L2aQtpzwInoqyn9eHNZvDhsi29bGcuHlU9Tv45WjWnKtqeGkTjc076PLqSn3RxOkpZdHPRiVYGQL-gu2SaF-N1mRgEQ-uhZp9iQi_YevWQVu_aNdQkS2O0yNcddoebhTUKsFLni7q1957z-Fpw7g79A6Unw8OJNz6vQOUHU8PDjBKd4Xe9qlvyA8OZxPiFEG-b-7RRZVrd4ReO4m6tnYxeaMm00vaXZ5U3_G-aFnAeXmF2vEqUyXzr8yuTX00XrCJUM1IMrlO-SnPrPHGU7oBzEMdCx4zJNNkw9IlQlZnYVnXvmDzd1crz-itd4cvVO9sQMa8pjYmm-4SpN6G7BDyOWir5qhuD7Y8Qu5giG3dWhWkhzIotlxUqgr10YORVfv0hLXTDJlOvdpBveV9_NkDHsSQ-Ojx6hrIvsoto5PF_uA6VkdHdJRxtmqhPqAPDk-nwy3vpB6Nq07dG2nF4aw5YZ85Y7_bSnPRdVverW8UvOVHh-zmVA34vJHmyPbNbr3AW5KuzOeYKMLbZfUfYMFB42b2Trs_zBbH0rVZLh-dZh2THYW3Np8tmC5VAz4_TsuuHzrXWT2tj370Vl9p2e89Nvu9zuD8r7_-B8mvVTP1cQy77Ulm22rrjiHwni983jvDK51Uwrj2jaOxLG93ef5_8_Gf7ITR1JIV1Cz_3Ep3BpGwVTXCVG6yWzL6ZfVlF-CwtplNutmDVcelt5fIJ6Q3GqtfpeGI5YVH5ML2jUfkyyTKTxJS5nWPzaK5QMW6t0Gl_gpiCSkMrwwSVh27pQcWkk0-5d0d5PVyO9rNxI5J17a8LB3wkpG8TIfhKDxMsiG_QosOi6ocDMpweJj1p9V0OmsxarQhhiy2t3mO7PMQojptZ4eT6lm9vc3DOp9nAgN5lgPEj-GkwYCXyzwj7d3mUZfAw2p4XHPzKxzk55xganmvyRgSniH4PHoCdSzJXctrShi7k_FLoefhvaqajIU9D9J8HmJItimSecZOnUAb1dGGenfUNk-fFH-FHfCGkkXLi2bIY23WJzrlorJqIfntSiX_e3Lkx-3J5NpK7lCYqlaJMY5H46YBr93gNTnb_PpNaHDXAgvQTRooaEO8wqgTvsqJ49G0kiaYqG--8djGSd1WMm7Eep6_U9UuRtWUhHh2qlRCKbaIjKdfV5EQY7887mmenc7lnkha8MAQ1Pn8fIObn3oquTGUdJRQTqvm7LiaKASP6ATDC-82qsX5dDieCVZ1-fT0ZH6u2vCYVjJqMdKJl2cb8hxEuQCBYCZVK5PwaMx97zaPpHziST_J2N6Go6I6x5B2DBWPdLsInMpLgURzelzQKtjgQzDodvoNNXL3IyK1RPX-mg18VZ9ypyHkmOOTKbSKpH0JSoKNtnITgpXKACKGet6p15yrzpAAwkIi8XTezhjp6Mu_ZYRsMX4h9TzJhoIhQfsVl8JbMWtf6h6z4rhXoytjkaZUtXdt__ev-elPz-rm2cv6FMMk9uiEM8ZDuzTj5XiOD-X_cwItFpD0Xv0PQ5mZNwNRAAA) (screenshot below). Do we care that there's no way to turn the background color back to the original unfilled colour in the case of a filled container inside a filled container? i.e. the top right is the same as the bottom right and there's not an easy way to get the inverse background color (how would you do it with `extras={"className": ...}`?). Not a big deal for me and I don't want to add lots of extra complexity or any crazy dynamic changing of class names depending on parent containers. But maybe there should be another available theme that somehow does `unfilled`. But then what if you want to do `unfilled + border`, it would need to be another option... So probably best just to ignore this I think unless you have a bright idea 😁 

image

antonymilne (approved):

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -34,9 +37,31 @@ class Container(VizroBaseModel):
...
+            children=[
+                html.H3(children=self.title, className="container-title", id=f"{self.id}_title"),
+                self._build_inner_layout(),
+            ],
+            fluid=True,
+            className=variants.get(self.theme, "") if self.theme else "",
+        )
+
+    def _build_inner_layout(self):
+        """Builds inner layout and assigns components to grid position."""
```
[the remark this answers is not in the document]
nadijagraca: Moving the layout creation into a helper function makes it much easier to build custom components based on `Container`. For example, when I created the CollapsibleContainer, I had to duplicate the entire build method just to retain the layout. I assume this is the reason why @huong-li-nguyen moved it.

vizro-core/src/vizro/models/_components/container.py
petar-qb: Copied text from https://github.com/mckinsey/vizro/pull/1002#pullrequestreview-2653509111 to continue discussion here in this thread

---

First of all, I'm so happy we abandoned the `background=True/False` idea! 😅 

As `theme` feels misleading, I'll use `style_variant` in the rest of the comment as it feels  much more natural to me 😄 

Before merging this PR, I think we should establish a rough plan for `style_variant`. In future I see that we will upgrade this solution to another level and that we will have a predefined set like `"transparent"`, `"filled"`, `"outlined"`, `"elevated"`, `"rounded"`... In that case each component could have a default like for example, `"transparent"` for `Container`, `"filled"` for `Card` and `Alert`, while `Button` might include additional options like `"link"` and `"disabled"`.

Open questions:
1. Should we default Container to `"transparent"` instead of `None`?
2. Should we align our `style_variant` values with existing standards like Bootstrap, or define our own? Caution: users could require something like `"rounded-elevated"` 🤔 

While we don’t need a full plan before merging, having a clear direction will help ensure consistency. Looking forward to your thoughts! 🚀

vizro-core/src/vizro/models/_components/container.py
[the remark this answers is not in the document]
antonymilne: I think this is ok - the amount of computation here is absolutely minimal. It's just the heavy lifting that we shouldn't be repeating.

vizro-core/src/vizro/models/_components/container.py
antonymilne: FYI @petar-qb there's plans for "elevated" but that's about it. We don't intend to enable lots of different styles like rounded etc. here since they're outside our designs. I think there should always be less than 5 or so options. In future users might be able to fully customise their theme at a global level (e.g. if they want to set `rounded=True` throughout their app). It's already possible at a component level to do arbitrary customisation using CSS or the new `extra` but we don't intend to make it easy to get any style you want through this `variant` argument. It's deliberately quite restricted to things that we think will look good.

As for aligning with bootstrap, we don't want to fully match bootstrap. Users shouldn't need to read bootstrap docs to understand how to change this, and descriptive terms like `filled` are more declarative and implementation-independent. Think of our `variant` argument as a kind of higher-level translation layer into bootstrap (which is already a shortcut for CSS). And advanced users can access the full range of bootstrap configuration still through `extra` if they want to do something that's outside our very limited translation layer.

maxschulz-COL: Just saw this discussion and thought I'd chip in. I am actually against `None` meaning different things, I think it is confusing. In general working with the schema recently I think that `None` causes often more trouble than not.

If we are able to give a name to every default of components that have a `appearance/style` argument, then I think we should not rely on None. I think it is very ok to have the card be `filled` by default, and the container to be `transparent`. To me that is clear and concise.

So in summary:
- no `None`
- definitely no `"none"`
- ideally a default that is descriptive, like `transparent` or `filled` or whatever for the respective component

antonymilne: @maxschulz-COL generally agree with you that we should avoid `None` (this was my original point in this thread). Curious why do you think `"none"` is even worse? It's just a more confusing version of `None`?

Overall @petar-qb, @maxschulz-COL and I are all agreed that if we can find a name that accurately describes the style, like "transparent" or "plain" then that is best though. @huong-li-nguyen wdyt? The only argument I have against "transparent" is that "outlined" is also transparent. But maybe "plain" just sounds silly.

maxschulz-COL: Yes, I think it is almost a weaker version of `"none"` and somewhat reminiscent of the `ALL` issue. Because it is also not really true that there is no styling (see `Card` or even `Container` (still has the dashboard background and potentially padding)) - so in order to avoid potential misunderstandings I think a dedicated description is cleanest.

At first I thought `default` might be good, but it somewhat fails for the same reasons...

vizro-core/src/vizro/models/_components/container.py
huong-li-nguyen: I'll choose `variant` because it is widely used in other UI component frameworks and doesn't seem too generic to me. Seems like no one is strongly against it either.

`format` reminds me of the `value_format` arguments in the KPI cards or the general `format` function in Python, which is why I prefer `variant` over `format` @antonymilne , but happy to be convinced otherwise 😄

vizro-core/src/vizro/models/_components/container.py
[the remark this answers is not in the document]
huong-li-nguyen: Yes, exactly 👍. It will be easier to customize or override the build if we separate these out into functions.

I also believe we should break down the steps, as the layout function will likely expand in the future when we implement the flex layouts.

vizro-core/src/vizro/models/_components/container.py
huong-li-nguyen: I'm going with plain. I don't think your idea is silly at all, @antonymilne. It's also what MUI does on a global level for variants: https://mui.com/joy-ui/main-features/global-variants/

- solid (default), soft, outlined, and plain.

Just to mention some other alternatives I've seen in UI frameworks: plain, text, standard 😄 

I'm not a fan of `default` because the actual default varies per component. For instance, our default for the Card is "elevated," for the Container it's "plain," and for the Button it's "primary." Using "default" would be quite confusing as it would not be the default for every component.

vizro-core/tests/unit/vizro/models/_components/test_container.py
huong-li-nguyen: Ups! I thought I added it already 😅 Will add it in a sec!

huong-li-nguyen: https://github.com/mckinsey/vizro/pull/1002/commits/8bab8828cda08d6e7ca02203178c5c0e907015e0

huong-li-nguyen:
> This is amazing, thank you for all your work on it! I know it's taken a lot of discussion 😅 It looks great and happy to approve now subject to a few things being resolved:
> 
> * finalise argument name
> * decide what should be the possible values for `theme` (or whatever name it has) - something other than `None` might be better
> * a couple of new tests

- I went with `variant`
- Possible options are `filled`,`outlined` or `plain` - In the future we may or may not add `elevated`.
- Added unit tests here: https://github.com/mckinsey/vizro/pull/1002/commits/8bab8828cda08d6e7ca02203178c5c0e907015e0


1. Yes, let's discuss in our 1:1.
2. Yes, I did briefly discuss this case with J. and we decided to ignore it as it's such an edge case. We can in practice always iterate the filled color though. So on every second iteration, we go back to the darker color. The CSS would look like this:

```
.bg-container .bg-container {
  background-color: var(--right-side-bg);
}

.bg-container .bg-container .bg-container {
  background-color: var(--surfaces-bg-card);
}

.bg-container .bg-container .bg-container .bg-container {
  background-color: var(--right-side-bg);
}
```


Screenshot 2025-03-04 at 14 46 46

If we think we want to support that, I can just add this into our static CSS, but I am also fine just not encouraging people to do that haha 😄

vizro-core/changelog.d/20250217_105619_huong_li_nguyen_styled_containers.md
```diff
@@ -0,0 +1,46 @@
...
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+### Added
+
+- Enable styling of `vm.Container` with a new argument `theme="outlined"/"filled"`. See the user guide on [styled containers](https://vizro.readthedocs.io/en/stable/pages/user-guides/container/#styled-containers) for more details. ([#1002](https://github.com/mckinsey/vizro/pull/1002))
```
petar-qb, line 21:
```suggestion
- Enable styling of `vm.Container` with a new argument `variant="plain"/"filled"/"outlined"`. See the user guide on [styled containers](https://vizro.readthedocs.io/en/stable/pages/user-guides/container/#styled-containers) for more details. ([#1002](https://github.com/mckinsey/vizro/pull/1002))
```

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -34,9 +37,32 @@ class Container(VizroBaseModel):
...
+        "Defaults to `plain`.",
+    )
 
     @_log_call
     def build(self):
+        # TODO: TBD on how to encode 'elevated', as box-shadows are not visible on a dark theme
+        # It needs to be properly designed and tested out (margins have to be added etc.).
+        # Below corresponds to bootstrap utility classnames, while 'bg-container' is introduced by us.
+        # See: https://getbootstrap.com/docs/4.0/utilities
+        variants = {"outlined": "border p-3", "filled": "bg-container p-3", "plain": ""}
```
petar-qb, line 52: It's a nitpicking but let's align the order of variants here, in the changelog, tests and so on. I prefer that the default is in the first place and then followed by other variants ordered by importance or alphabetically.

vizro-core/tests/unit/vizro/models/_components/test_container.py
```diff
@@ -35,17 +42,30 @@ def test_mandatory_components_missing(self):
         with pytest.raises(ValidationError, match="Field required"):
             vm.Container(title="Title")
 
+    def test_invalid_variant(self):
+        with pytest.raises(ValidationError, match="Input should be 'filled', 'outlined' or 'plain'"):
+            vm.Container(title="Title", components=[vm.Button()], variant="test")
```
petar-qb, line 47:
```suggestion
            vm.Container(title="Title", components=[vm.Button()], variant="invalid_variant")
```

vizro-core/src/vizro/models/_components/container.py
maxschulz-COL: I always understood it more like `default` means just what we decide, so basically `vizro`. But yes, I am also against it because it clearly is confusing!

petar-qb (approved):

vizro-core/src/vizro/models/_components/container.py
petar-qb: Thanks for making these adjustments! 💜

huong-li-nguyen:
> If we think we want to support that, I can just add this into our static CSS, but I am also fine just not encouraging people to do that haha 😄

Just talked with J. and he wanted the first level of nesting to work, so I'll add that CSS, in which case your example will also work out of the box @antonymilne

PR #1003, opened 2025-02-10
[FOR DISCUSSION] Dev/loading spinner dynamic filters

huong-li-nguyen:
## Description

Context: 
- https://github.com/mckinsey/vizro/pull/915
- https://github.com/McK-Internal/vizro-internal/issues/1422

Quick POC to lead discussion

## Notice


antonymilne (commented):
Great idea! I like the general approach 👍 

I'm pretty confused by how it works though. I expected it to use [`data-dash-is-loading`](https://dash.plotly.com/loading-states#check-loading-states-from-components) but it doesn't?

Also to be discussed:
* should we have this new spinner and `dcc.Loading` around the components without the spinner? Either for clean code or to stop user interaction.
* should we change graphs etc. to work like this
* how it should work with radioitems and checklist on first load.
* possibility of a "flash" on finishing loading.

petar-qb:
I really like this idea! @huong-li-nguyen's sharp eye always finds something beautiful that I can't see 😄 

I think that this loading icon (but spinning one) in combination with some "Vizro universal placeholder component" would be the win win for us. (here's another reason to start thinking again about the Universal placeholder component -> https://github.com/McK-Internal/vizro-internal/issues/1494).

There's a problem if we return a visible control (e.g. vm.Checklist) as a placeholder because in the page_build phase we don't know what will be the new options that will be returned from the on_page_load. So, we always return default options and default value initially from the page_build there are calculated when the app is started. In that way it can happen that default options are ["A", "B", "C"] and that new options (that will be returned from the on_page_load phase) are ["K", "L", "M", "N", "O"]. It would be weird to look at the checklist with the ["A", "B", "C"] and the loading icon (that you suggested) that will be replaced with the Checklist that has ["K", "L", "M", "N", "O"] options. So, that's the reason why we decided to "hide" the placeholder control with the spinning loading icon (that's currently in our main branch).

PR #1005, opened 2025-02-10
[Feat] Make charts transparent when used inside dashboard

huong-li-nguyen:
## Description

**Context:** We are planning to introduce styled containers soon (see https://github.com/mckinsey/vizro/pull/1002), allowing for easy changes to the container background color and borders. To maintain a cohesive design, charts need to seamlessly adopt the background color of their parent container. This can be achieved by setting the chart background to transparent when used within a Vizro dashboard (it remains colored when used outside the dashboard).

- Set the chart background to transparent when used within a dashboard.
- Make the modebar background transparent and apply colors to icons.

While this approach may cause exported PNGs to look incorrect, we have decided to disable PNG export functionality. Instead, charts will inherit the background color of their parent container.

## Notice


vizro-core/docs/pages/explanation/faq.md
```diff
@@ -95,7 +95,7 @@ Any attempt at a high-level explanation must rely on an oversimplification that
...
 
-For a more detailed comparison, it may help to visit the introductory articles of [Dash](https://medium.com/plotly/introducing-dash-5ecf7191b503), [Streamlit](https://towardsdatascience.com/coding-ml-tools-like-you-code-ml-models-ddba3357eace) and [Vizro](https://quantumblack.medium.com/introducing-vizro-a-toolkit-for-creating-modular-data-visualization-applications-3a42f2bec4db), to see how each tool serves a distinct purpose, and could be the best tool of choice.
+For a more detailed comparison, it may help to visit the introductory articles of [Dash](https://medium.com/plotly/introducing-dash-5ecf7191b503), [Streamlit](https://blog.streamlit.io/streamlit-101-python-data-app/) and [Vizro](https://quantumblack.medium.com/introducing-vizro-a-toolkit-for-creating-modular-data-visualization-applications-3a42f2bec4db), to see how each tool serves a distinct purpose, and could be the best tool of choice.
```
stichbury, line 98:
```suggestion
For a more detailed comparison, it may help to read introductory articles about [Dash](https://medium.com/plotly/introducing-dash-5ecf7191b503), [Streamlit](https://blog.streamlit.io/streamlit-101-python-data-app/) and [Vizro](https://quantumblack.medium.com/introducing-vizro-a-toolkit-for-creating-modular-data-visualization-applications-3a42f2bec4db), to see how each tool serves a distinct purpose.
```

vizro-core/docs/pages/user-guides/custom-css.md
```diff
@@ -355,7 +355,7 @@ By default, the logo appears in the top left corner of the dashboard. You can mo
 
 If you want to make the subsections of your dashboard stand out more, you can do this by placing your components inside a [Container](container.md) and changing the container's styling, for example, background color, borders, padding, etc.
 
-To do this, you need to change the container's CSS class. Using the DevTool, as explained in the section on [identifying the correct CSS selector](#identify-the-correct-css-selector), you'll find that the CSS class for the `Container` is `page-component-container`. You can then use this class to set a new `background-color` and `padding`.
+To do this, you need to change the container's CSS class. Using the DevTool, as explained in the section on [identifying the correct CSS selector](#identify-the-correct-css-selector), you'll find that the CSS class for the `Container` is `page-component-container`. You can then use this class to set a new `background-color` and `padding`. The charts will automatically align with the background colors of the container as they are transparent.
```
stichbury, line 358:
```suggestion
To do this, you need to change the container's CSS class. Using the DevTool, as explained in the section on [identifying the correct CSS selector](#identify-the-correct-css-selector), you'll find that the CSS class for the `Container` is `page-component-container`. You can then use this class to set a new `background-color` and `padding`. The charts automatically align with the background colors of the container because they are transparent.
```

stichbury (commented):
LGTM from a docs perspective 🏆 🌟

vizro-core/docs/pages/user-guides/custom-css.md
```diff
@@ -355,7 +355,7 @@ By default, the logo appears in the top left corner of the dashboard. You can mo
 
 If you want to make the subsections of your dashboard stand out more, you can do this by placing your components inside a [Container](container.md) and changing the container's styling, for example, background color, borders, padding, etc.
 
-To do this, you need to change the container's CSS class. Using the DevTool, as explained in the section on [identifying the correct CSS selector](#identify-the-correct-css-selector), you'll find that the CSS class for the `Container` is `page-component-container`. You can then use this class to set a new `background-color` and `padding`.
+To do this, you need to change the container's CSS class. Using the DevTool, as explained in the section on [identifying the correct CSS selector](#identify-the-correct-css-selector), you'll find that the CSS class for the `Container` is `page-component-container`. You can then use this class to set a new `background-color` and `padding`. The charts automatically align with the background colors of the container because they are transparent.
```
antonymilne, line 358:
```suggestion
To do this, you need to change the container's CSS class. Using the DevTool, as explained in the section on [identifying the correct CSS selector](#identify-the-correct-css-selector), you'll find that the CSS class for the `Container` is `page-component-container`. You can then use this class to set a new `background-color` and `padding`. Chart backgrounds are transparent so they automatically match the background color of the container.
```

vizro-core/src/vizro/_themes/_colors.py
```diff
@@ -1,5 +1,7 @@
 """Color sequences by Vizro."""
 
+TRANSPARENT = "rgba(0, 0, 0, 0)"
```
antonymilne, line 3: Any reason this isn't inside `get_color`? I know it's not really part of our theme but it just looks a bit lonely here 😅 Up to you though.

vizro-core/src/vizro/_themes/_common_template.py
```diff
@@ -13,80 +13,87 @@ def create_template_common() -> go.layout.Template:
     COLORS = get_colors()
     template_common = go.layout.Template()
     template_common.layout = go.Layout(
+        annotationdefaults_font_size=14,
```
antonymilne, line 16: What has actually changed here outside all the reordering?

antonymilne (approved):
I didn't actually test it out but just skimming through the code it looks good to me! So if one more reviewer can do a more careful review then that would be great.

vizro-core/src/vizro/_themes/_common_template.py
huong-li-nguyen: map_style="carto-darkmatter" -> moved to _common_template#
modebar_activecolor="darkgrey",
modebar_bgcolor=TRANSPARENT,
modebar_color="dimgrey",

vizro-core/src/vizro/_themes/_colors.py
huong-li-nguyen: I had it there at first, but then I had to call the `get_colors()` function every time the dashboard ran because it needed to be used in `template_dashboard_overrides.py`. It seemed less efficient to keep calling the dictionary just to get the constant, but I don't have a big preference.

antonymilne: Ah ok, that makes sense and I think made the right decision to not call `get_colors` at runtime - let's reserve it for the template generation. But in this case probably I would put transparent inside `get_colors` for use in template generation and then re-define `TRANSPARENT = "rgba(0, 0, 0, 0)"` in `template_dashboard_overrides.py` just for use in that one file. We'll have duplicate definitions of `TRANSPARENT` but that's ok.

antonymilne: Either that or leave as you have it but with a comment explaining why you did it this way, up to you!

l0uden (approved):

vizro-core/changelog.d/20250210_184928_huong_li_nguyen_make_charts_transparent.md
```diff
@@ -0,0 +1,48 @@
...
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
```
nadijagraca, line 14: Should we mention that we removed toImage option?

nadijagraca (approved):
I tested and everything works. I really like this, it adds depth to our dashboards, and we are one step closer to having tiles. 🚀  👏 

The only potential issue I can think of is that the axes are not responsive to the background color set by the user. However, it's a really an edge case, so I don't think it's something we need to worry about.

vizro-core/changelog.d/20250210_184928_huong_li_nguyen_make_charts_transparent.md
huong-li-nguyen: Yes, you are right!

Issue #1010, opened 2025-02-11
Dropdown with multi=True working incorrectly for parameters


l0uden:
### Which package?

vizro

### Package version

0.1.33

### Description

Clearing dropdown with multi=True for parameters results in 500 in browser console

### How to Reproduce

1. Configure dropdown parameter with option `multi=True`
2. On the UI select clear all values from the dropdown.
3. Check browser console

### Output

```Failed to load resource: the server responded with a status of 500 (INTERNAL SERVER ERROR)```

### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

PR #1026, opened 2025-02-18
[Bug] Fix sending an empty list through parameter

petar-qb:
## Description
Fixes https://github.com/mckinsey/vizro/issues/1010

Currently on the main branch, if an empty value of the `vm.Dropdown(multi=True)` Parameter.selector is chosen, `[None]` value will be sent into its target. This behaviour is introduced in this PR https://github.com/mckinsey/vizro/pull/95.

P.S.
Mentioned PR (https://github.com/mckinsey/vizro/pull/95) actually handles a really great use-case. It enables `None` to be sent into Parameter target which is not possible in the plain dash app by using a `dcc.Dropdown` component. So, creating the following component in the plain dash app will raise and exception only because of the `{"label": "None", "value": None}` option. Options like `True`, `False` and others work correctly. :

```py
dcc.Dropdown(
    id="dropdown",
    options=[
        {"label": "None", "value": None},
        {"label": "True", "value": True},
        {"label": "False", "value": False},
        {"label": "Option A", "value": "A"},
    ],
    value="A"
),
```

## Notice

### DIFF
--- a/vizro-core/changelog.d/20250218_142411_petar_pejovic_fix_sending_empty_list_through_parameter.md
+++ b/vizro-core/changelog.d/20250218_142411_petar_pejovic_fix_sending_empty_list_through_parameter.md
@@ -1,0 +1,46 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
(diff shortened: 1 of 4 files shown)


antonymilne (commented):
Nice work! ⭐

huong-li-nguyen (approved):
LGTM 👍

PR #1042, opened 2025-02-28
[Poc] Implement collapsible container as a custom component

nadijagraca:
## Description
This PR is meant to create collapsible container as a custom component.
`CollapsibleContainer` component (for now) needs to be used within `FlexContainer` component (component that mimics flex layout). 

Discussion points:
- should we use background color in the collapsible container
- is the "open/close" arrow prominent enough in the layout

## Notice


vizro-core/examples/scratch_dev/components.py
```diff
@@ -0,0 +1,90 @@
...
+                    className="collapsible-container-header",
+                ),
+                dbc.Collapse(
+                    id=self.id,
+                    children=components_container,
+                    is_open=True,
+                ),
+            ],
+            className="bg-container",
+        )
```
huong-li-nguyen, line 90: Might be good to base this PR on this PR first: https://github.com/mckinsey/vizro/pull/1002

I refactored this part so it uses dbc.Container instead of html.Div. Hopefully it will work the same then 🤞

vizro-core/examples/scratch_dev/assets/css/custom.css
```diff
@@ -1,8 +1,29 @@
...
+.flex-container {
+  display: flex;
+  flex-direction: column;
+  gap: 24px;
 }
 
-#old-line-height p {
-    line-height: 1;
-    margin-bottom: 0.5rem;
+.bg-container {
```
huong-li-nguyen, line 7: I've changed some of the CSS here in https://github.com/mckinsey/vizro/pull/1002

vizro-core/examples/scratch_dev/assets/css/custom.css
```diff
@@ -1,8 +1,29 @@
...
+  border: 1px solid var(--border-subtleAlpha01);
+  padding: 1rem;
+}
+
+.collapsible-container-header {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  padding-right: 12px;
```
huong-li-nguyen, line 16:
```suggestion
```
Would remove so it's equal space on the left and right

vizro-core/examples/scratch_dev/assets/css/custom.css
```diff
@@ -1,8 +1,29 @@
...
+#collapsible-container_icon.material-symbols-outlined {
+  border-radius: 50%;
+  color: var(--text-disabled);
+  cursor: pointer;
+  height: 24px;
+  width: 24px;
+}
+
+#collapsible-container_icon.material-symbols-outlined:hover {
+  color: var(--text-primary);
```
huong-li-nguyen, line 28:
```suggestion
  color: var(--text-primaryHover);
```

vizro-core/examples/scratch_dev/assets/css/custom.css
```diff
@@ -1,8 +1,29 @@
...
+.collapsible-container-header {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  padding-right: 12px;
+}
+
+#collapsible-container_icon.material-symbols-outlined {
+  border-radius: 50%;
+  color: var(--text-disabled);
```
huong-li-nguyen, line 21:
```suggestion
  color: var(--text-secondary);
```

vizro-core/examples/scratch_dev/assets/css/custom.css
```diff
@@ -1,8 +1,29 @@
...
 
-#old-line-height p {
-    line-height: 1;
-    margin-bottom: 0.5rem;
+.bg-container {
+  border: 1px solid var(--border-subtleAlpha01);
+  padding: 1rem;
+}
+
+.collapsible-container-header {
```
huong-li-nguyen, line 12: Screenshot 2025-03-02 at 17 01 20

It's a bit unequally spaced at the moment. I think you need to remove the bottom padding from the h3 and then use align-items: center here, so they are properly placed in the center. Otherwise there is unequal space on the top / bottom of the heading

vizro-core/examples/scratch_dev/app.py
```diff
@@ -4,43 +4,49 @@
...
+                    title="Collapsible container 2",
+                    components=[
+                        vm.Graph(figure=px.scatter(iris, x="sepal_length", y="petal_width", color="species")),
+                        vm.Graph(figure=px.histogram(iris, x="sepal_width", color="species")),
+                    ],
+                    layout=vm.Layout(grid=[[0, 0, 0, 0], [1, 1, 1, 1]]),
+                ),
+            ]
         )
     ],
```
huong-li-nguyen, line 44: Could you add the example of how this looks inside a grid? Is it completely unusable? 😄

huong-li-nguyen (commented):
I just took a quick look—this works really well inside the FlexLayout! 🚀

I made some refactors in this PR: [#1002](https://github.com/mckinsey/vizro/pull/1002), which also impacts your PR. It might be worth rebasing your branch on that one instead.

Regarding your question about coloring, I believe this should be controlled by setting theme="XX". However, we could check with J. if we want a different default for collapsible containers compared to regular containers (currently, the default is no styling). However, I agree with you, that collapsible containers should probably come pre-styled (I don't have a preference here whether it should be outlined or filled) as otherwise it's a bit difficult to visually distinguish where the collapsible container starts / ends.

vizro-core/examples/scratch_dev/app.py
```diff
@@ -4,43 +4,49 @@
...
     components=[
-        vm.AgGrid(
-            id="my_custom_ag_grid",
-            figure=my_custom_ag_grid(
-                data_frame=df,
-                chosen_columns=df.columns.to_list(),
-            ),
+        FlexContainer(
+            components=[
+                CollapsibleContainer(
```
huong-li-nguyen, line 24: Did you try to control the scroll bar? e.g. is it possible that it expands to the available space only such that no scroll bar appears? 🤔 I think this can eventually get a bit tricky, but there should be a way how users could eventually control it

vizro-core/examples/scratch_dev/assets/css/custom.css
huong-li-nguyen:
```
.collapsible-container-header h3 {
  margin: 0;
}

.collapsible-container-header {
  display: flex;
  flex-direction: row;
  align-items: center;
}
```

vizro-core/examples/scratch_dev/assets/css/custom.css
```diff
@@ -1,8 +1,29 @@
...
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  padding-right: 12px;
+}
+
+#collapsible-container_icon.material-symbols-outlined {
```
huong-li-nguyen, line 19:
```suggestion
.collapsible-container-header .material-symbols-outlined {
```
Otherwise the collapse icons look different in the first collapsible container compared to the second

vizro-core/examples/scratch_dev/assets/css/custom.css
```diff
@@ -1,8 +1,29 @@
...
+  border-radius: 50%;
```
huong-li-nguyen, line 20:
```suggestion
```

vizro-core/examples/scratch_dev/assets/css/custom.css
```diff
@@ -1,8 +1,29 @@
...
+  color: var(--text-disabled);
+  cursor: pointer;
+  height: 24px;
+  width: 24px;
```
huong-li-nguyen, line 24: Should be by default
```suggestion
```

vizro-core/examples/scratch_dev/assets/css/custom.css
```diff
@@ -1,8 +1,29 @@
...
+}
+
+#collapsible-container_icon.material-symbols-outlined:hover {
```
huong-li-nguyen, line 27:
```suggestion
.collapsible-container-header .material-symbols-outlined:hover {
```

antonymilne:
This is looking great, nice work @nadijagraca!

Just two questions for now, for @huong-li-nguyen and/or @nadijagraca:

1. What is the plan for this component compared to `vm.Container`? Will every single `vm.Container` now be inside a `dbc.Collapse`? Or is there a new argument `collapsible=True/False` to turn on/off this behaviour? Do we want to make it possible for users to specify whether the container should be open or closed by default?
2. How is this approach better/worse/different from `dbc.Accordion`? At a glance I guess:
  * `dbc.Collapse` is better than `dbc.Accordion` because an accordion really expects multiple items whereas we just want to apply this to a single container
  * `dbc.Accordion` comes with the rotating arrow built in, while here you needed to provide the clientside callback to do that

huong-li-nguyen:
> 1. What is the plan for this component compared to `vm.Container`? Will every single `vm.Container` now be inside a `dbc.Collapse`? Or is there a new argument `collapsible=True/False` to turn on/off this behaviour? Do we want to make it possible for users to specify whether the container should be open or closed by default?

We can go into more detail later, but for now, my understanding is that this should remain a custom component. It doesn't work well within a grid and only functions properly inside a FlexContainer. To incorporate this into vizro-core, we would need to:

- Properly enable FlexLayout first.
- Either set the default to FlexLayout (which you and I previously discussed and decided against) or find a way for it to programmatically switch to FlexLayout if a collapsible container is detected. This part is still unclear.

However, I don't think we need to wait for publishing this custom component as it's ready to use as long as you put it inside a FlexLayout or you have the grid and no other component on screen 😄  And in the meantime we can work on the missing points above.

> 2. How is this approach better/worse/different from `dbc.Accordion`? At a glance I guess:
> 
> * `dbc.Collapse` is better than `dbc.Accordion` because an accordion really expects multiple items whereas we just want to apply this to a single container
> * `dbc.Accordion` comes with the rotating arrow built in, while here you needed to provide the clientside callback to do that

Everything you mentioned above and:
- With dbc.Collapse, we have more flexibility in terms of collapse behavior. dbc.Accordion only collapses vertically, while dbc.Collapse can collapse both vertically and horizontally: https://dash-bootstrap-components.opensource.faculty.ai/docs/components/collapse/
- dbc.Accordion comes pre-styled, so you'd need to disable styles for hover, focus, and normal modes. On the other hand, dbc.Collapse comes unstyled, which is easier to work with if you just want to wrap components inside.

> 3. Should our nav/controls side panel collapse in the same way as we do here? This would mean we don't need extra custom code for rotating the arrow like I think we have now.

The side panel collapse is actually similarly implemented (@nadijagraca did that one as well, and there is custom JS code that also rotates the icons - correct me here if I am wrong @nadijagraca ). Our side panel however collapses side ways, so the rotation direction is different.

vizro-core/examples/scratch_dev/app.py
huong-li-nguyen: @nadijagraca – I'm not sure if this issue has been fixed, but I discussed it with @antonymilne, and the component is essentially ready to be published (even without `Flex`), as long as it looks decent within the grid (meaning we can get rid of all these overflows)

If that works, that would be ideal! If that's not feasible, eventually ask @pruthvip15 for guidance. If it still doesn't work after that, let @antonymilne know 👍 And we can either document that `Flex` is required and add some warnings eventually 🤔

nadijagraca: I implemented all the requested changes and suggestions in another branch as I had too many rebase conflicts. I added an example in grid but it does not look good (as you said it overflows). I'll take a look tomorrow morning at fixing the overflow, and then I will push.

nadijagraca:
New PR opened: [1060](https://github.com/mckinsey/vizro/pull/1060)

PR #1048, opened 2025-03-04
[Tidy] Change light theme and apply several visual changes

huong-li-nguyen:
## Description

For context: An update to the light theme was long overdue, but it became particularly necessary after the introduction of filled containers. This PR introduces several changes aimed at streamlining our design by improving consistency, reducing the number of varying shades, and enhancing contrast between overlapping elements.

- **Light Theme Redesign:** The light theme has been updated to adjust the background color when a filled container is used. Take a look at the scratch-dev. There are basically two modes: Default and Page with filled container.
- **Collapse Button Design Update**: The collapse button now has a matching background color with the left-side panel and includes a border for improved visual consistency.
- **Accordion Item Design Update**: The hover and active states of accordion items now share the same background color for a more cohesive design.

**Default:**

Screenshot 2025-04-11 at 13 09 47

**Page with filled container**
Screenshot 2025-04-11 at 13 09 55

## Notice


vizro-core/src/vizro/static/css/bootstrap_overwrites.css
```diff
@@ -18,8 +18,8 @@ All the HEX values starting with --text-code are taken from the Github code high
 
 [data-bs-theme="light"] {
   --dropdown-label-bg: var(--primary-300);
-  --right-side-bg: var(--surfaces-bg01);
-  --left-side-bg: var(--surfaces-bg02);
+  --right-side-bg: var(--surfaces-bg02);
+  --left-side-bg: var(--surfaces-bg01);
```
huong-li-nguyen, line 22: @Joseph-Perkins - this would be the only change required in terms of CSS now to switch out the bg colors from the outer and inner containers.

I'll comment on the issues in the PR so we can come back to it if we want

vizro-core/changelog.d/20250304_201419_huong_li_nguyen_add_minor_visual_changes.md
```diff
@@ -0,0 +1,48 @@
...
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Turn `AgGrid` and `DataTable` background transparent. ([#1047](https://github.com/mckinsey/vizro/pull/1047))
+
```
huong-li-nguyen, line 29: I'll update this properly with the light theme later

maxschulz-COL (commented):
First of all wow 😍 - what really impressive work!! I am actually grateful you are doing this, I never could in the slightest... Really fun to always check what you come up with.

I guess I will mostly review the actual outcome, not so much the case, as I trust you a lot more haha:

First impression:
- urgh, bit weird grey boxes on a very white canvas

After sticking a bit with it:
- oh I like it, it seems modern for some reason, what ever this means

So some clarification questions:
- in"Default", why is the canvas where the components sit the same color as the control panel and the title panel in light mode, but not in dark mode?
- same for "Card with Graph" - why is there the asymmetry between light and dark?

Possibly stupid questions, because this may have been the whole work to work out if we need this asymmetry?

huong-li-nguyen:
> So some clarification questions:
> 
> * in"Default", why is the canvas where the components sit the same color as the control panel and the title panel in light mode, but not in dark mode?
> * same for "Card with Graph" - why is there the asymmetry between light and dark?
> 
> Possibly stupid questions, because this may have been the whole work to work out if we need this asymmetry?


Yeah, we actually already had some asymmetry between the dark and light themes before. What we’ve done now is just shift that asymmetry to a different part of the design. Let me explain :)

The main goal was to improve the design for when a filled container is added—basically, the "Page with a filled container" use case. In the dark theme, this worked fine because the contrast between the filled container’s background and the charts inside it was strong enough. But in the light theme, it didn’t work as well. The filled container’s background was grey, and according to best practices for data visualization, charts in light themes should ideally be on a white background for better clarity.

To fix this, we added an automatic switch: when a filled container is detected, its background changes to white, and the background of the right-side panel automatically switches to grey. But this caused another issue—if the side panel was also grey, it didn’t look great. So, we decided to make the side panel permanently white. Even though this means the whole background might end up being white in some cases, it’s still better than having everything look grey. Plus, having an all-white background isn’t that unusual.

Before this change, the asymmetry between the themes was different: in the dark theme, filled containers had lighter backgrounds while the right-side panel was darker. In the light theme, it was the opposite—the filled container was darker, and the right-side panel was lighter. Now, we’ve made this consistent across both themes.

At the end of the day, there will always be some asymmetry between dark and light themes because colors behave differently in each. But we made these changes on purpose now :)

vizro-core/src/vizro/static/css/container.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+/* Change filled container bg inside filled container */
+.bg-container .bg-container {
+  background-color: var(--right-side-bg);
+}
+
+/* If a filled container is detected, then the right-bg color will change to grey
+and all containers need to have a white background on that page */
+[data-bs-theme="light"] #right-side:has(.bg-container) {
```
maxschulz-COL, line 24: I have tried to [play with some nested configurations](https://py.cafe/snippet/vizro/v1?pycafe-app-view#c=H4sIAEy7_GcEA-1b3XIbx3J-lQko2xQFLPb_hzFOhaYsUTZly7YOw2OtCloAC2BFYAHvLkhCtpzbRK-RO-cqD5DKRXSRSqVymdfIK-TrnpnFgj_S8UViVR1Dg93pnunf6elZLFs_toaLUdrab8X872laViJZLiUU53GezZeLohLn2atiYSxni2q2NtLLZZGWpUhKsbyM83GxmMsJQs0-IeAK7RxyZkxzPmfGRVaKHhgYo6RKDAJ379IA_Vsmk7RfXSww4XxuPAG0G-cCnyqrZmkvbn1XrWfpSAwXeZVkeVqUcastZyyTaooJxKEDDjV-R8yS9WJV9cDwYZGNdie49J49M9vCet4Wz-y2cJ4_v6tmDxfQPU_zquw9kxj67JA2h0kxUtrQZ4evVXpZQar0GmPw2RE7OzviYFhli7wksxiLz474rkryERgJeHKUjmHCSCRyokiKVMyTUSqS8ySbJYNZKrJ8OFuNsnwi9vbgffIqeW1vT4ANcONsVqUFpuGq2OztGU2BUjdlHX2k3tMiHUPvLrsLBl-Zot1Bn79o48l0HWsN--mjQ7IeFx0xX2AfIdjSYosnfZqRBa4Pi2Q53R1nk1WR9rAZymFSgWyX9kNbXIJvmS6TWX-W5pNqCm5iDdwyrYC7yEYSNVzMFgVNXabDLMVeuNvUXX_eYYP-3GRLluN-zQz9-e3M0Z8_wyz63GQaL9O77aPP_5ONz9-hAjZfUmRJTvsM8Y7cd6vGd9u3MfqzWFwnf4_s5trIHKE93FgNlShuWAPskhm62zzl9N_A71Lwxt5btNM2a1I6h4rFTOr5gPPRLiSt5vmWqDa55-qwVFArzdqoc3CUlNPBghIlH4P3NbhLCQuy6EanJI4v7dl6jnhKGPCTrPhE3r1rDFbZbLRbM75rFKucjt1Wu1WkP6yyIp2Tv_FAMK2qZbnf7S7XxjAZp93JtJsUVTZGhu3Oh2dZXqbrLh_tXZV0u-BVdi3XtaPAiiIr6i6zpZzSMQ3LcEJjlJ6bneXa6eRY106Sr42L6Qyyq_WSHkJ4LsDlupouciCW68UoG6Wdc9OwfcPCkDzEW_v5ajZrUzimUPbZj608mROHblKWaVV2hyW-q7JazA10QbcqZhjWRhXJhTHJqulqsCrTglYPVhuIt6umua458j0ncC1nnI48y0oS00rHA98aeontee7YHZu-mSgzh4si7aaXyXwJvbrlsEiq4bQPq2_Va1FkkyxPZrQloAN03OHjaJriECzEjxRfy2REx19nlo6rfREuL_86zl_Tqnb3xNOFWOXZeFHMZ2t6cqOrKLNXqagWIpnNxHiWXnayKp2XYq8b55I7Iy-wnzrF4kIY9RwpjzfLvrB9cyPqBrrOhDZkeRu5YzbIoekBK8eDWrcvnzwS9ERxTTUsAS079QXCrxh1zpbZFnfbq7m3Xrevrf84Oc-wrAYuH87ibyt1feX_51__5clXD-MCDluZppXIu2luXYtHR_e_3cLQ1Wo3gBo14I7fQKrr-jVkrZoYda2WR38qtzB0ZT6W08Bcx1t8NcEXjfGhxFy7WuW3Dz-TcGP2UHW2psqrOzl4fNCAG2TmAJ2Ee54ca1w9jD26f_D0kiEL0H__Ay4EWcPP-ebyEDGMq2NcH-px8xK9-1_ysJ2hH-Er29_hO7n_ANeb1uf36zuv8BrFXt0Iaw3-pnpWY2Q7_xIXOSgRsr2St_LLI9lB-89_xkWGuuW76PdOmnTUN_099F7unD6JY3RO54yUaz-IczXLImK0Fwff4ErkjLXauwr0FYoD6ym-aBQGsv1HUw_J75sbVLH30YMqBN-gDK6WreZuFPoW14ZCZDztly2FKnzRSCGB73811bmD_g26KP7cD2S_bjz5KrJX92Rz6p5sZ_gyN1JINlrtK6h6q3sHElHWCL7STkU7Y6DpGOlWJTSuJK5954QgNaMdtxjUg6I5ePxxc-x4i_B4m_BYElLf5NxqhZQf0HZOKYhPn1OAvpAT0JPtBb6EgsJkLoGyveSJZDiatEj2OQ4IZtOtgK_KAYxX9hvipNSqYmhLdXRpVGnO16GcsIUSJ2cNiusczvR0QGp7Mey0KReSooIuqv9XjX7cOl2qrgxskOn1fiRvjFIrR55DC6Rw7aivt8HH-KIxzmzrHnH4lHo6dtH2eESFakfeKI9PSXTBg3bNYMu16JAX9hRtXMj71_hOKTTlqm1L369J0d7-wnjyGBrNQyPZuyRbkt8u-9u3vyjKeCzvJHmXJL9igm3Jf6gJ0ZRk0jxuNYw-IsGS-nbBT9_-IgnZ4s_wPVI7sREQ3KHV4I7V5iPbcu79_Akt8UcnP92h-52Tnz6nOQMSjka3eECoy_vcPaQuJaV4QDIolNS2IHLKy4O7dCHuJrlBakDH7aDLoEpsMe2YAWVPNE7JRIzGzpNUicSQ4TKvo-n7kuj_yGK0D9_c0yH_b3_Pd-9v-ebUnjPvPVArYXzCG6Z1Gsdf6NGQnC3bKfn4xQ-srLRFLS3JksrptaY-Yz-WCEofg-YgI6jHe2hASyU5MMg9xn2EC1pKhr2U_S_fvpFLuLFxdxs82ganNUjOJsCSmcjSrgG6HvmmgZQY6lzBOffqXu1H557Nd7P99o3MaCSSFhlByBPQfakyrDpTGW8rQjoT0d6-UQNMzyOSCdNTxKF9JW88at6T6YF8Sh3Tf_uGMcwgxFcNsIS9GzWQHG7XQHFhBjeqwDljo4L99o0hOdwj5fU8MuGLGxWQ9NcVuMd4xYPJbxS_JZ1DL85_rpFqqsKoVMEYuhCOOGscJ6EtDC4f01YmgMYoJDhF1IhPPz6lra-AT06pw4e8xKBD5jKgTvh7nCY6fEZxTx7O3BUndNa8ePyxut9Rd0yRHUzgvMasEuVQyhskqaGY6fITRtMJtHt5SD5wmN6Qs44arR2idedDnDHE-iM2TB_h0ushZXSc3SSD0t6duqdwPGuk0BLgJw8C8a09gG8TT-fzC3oaowfi0wN9fkiAuKhHDUbQLlxx71Nc0WiGPp81qKNZH9IaJLw8qZmCXSSP6g1MZzU9c27IELT__o8kk-C8yBg3IEXQfvqeyX16DiCyc3w10T_9EQDBeSFXh3OIbJLMVY8PdVOkEiB10OL85V3Ga5mvvmBililbHSKaVD3Rg1SuXUOyJH6H5J3a1lePGKvl_ix_qzfkkoYbwnu1vTxv67fS79d3X6Xn-vKmfppM8W3OaV4fff4V5cfPaAPd9NIom-MFVNnFWzSjPJ98OO-Nrul1_dXRpxiQr8bwEtiyYXFLTNNsMqWX2D6D51l68dmCXpObwhQ0R8gBeqcPLL2SBXQ5n-UlQLIaRl9cXBgXjrEoJl1QmF3IiVt_iPNPi3RYvUeiYjyYJcMz_JmNqOivs2IE5GM3sgzXEq7lG4F_DJcZthkIOwgM2wqOXNwj0z_2LMdwPFc4oW0E-AnmuaERmnY9zwt9wzOtY8_xDKRexe5IMv9-o8PFFO8nr-ng254RWi5eJnoGluXQtyLDdiMNCx_SQwugY3ihB3dFhhv5AlMM3_UPfdMzHOjsekZghhiGrmQDbAmAr2HHBUzTFWyHgMFdkdsQ4tF0xd3Gy3IT05VwWGT5GNa6Wa5hRZCmdZfwoe_Y4BZqWPhOYAS2A1jRuxbEbLi7hAa5lg7HOjbEaeU0rJXXsDZO02vjFXvtGi1du05rpz2rtZfw9499M8DChmoBT_T6-i45ITpR66pmvX9h_cjwIs3MCaFMaB8HpLsf4s10YJhOdOQHpuE6bi0sCGGD65w4JmyLvOMAK8Kx5xuRE2LYMWyvVoVFvFeT0IIM09GqaFGhaxtu4B2HcF_oWVAJsRHah6FnGpEFP0Yu_BUJLA3IEO8wwfZsEfo-5sGPNDfwsGfgwwB7wLdEGMKXFiIkgM-wtcIoMBwHvCR84lgwz7aPQuJlBqCPMC4UWoSBbeAvN4AdA1xJkoVodSwPmkbQLIJmcJ6FnYl7iJUzXVs4tm1YDsHYclhhBwFkOZiPgLGwUx0EkEXaIGBcrLhDlofBBkaEWEGo3ar89V6_RnBkFMGvFM-Iz8h3wRZLLmEReWQNxRPWGuIjxCe0dcE-9OzDyEEo2JhuuobvOyIiKyJPOBGsBh5rgLiDNfCp7R1GiAQfaOR0w0I4IzUZHnaJ44MrqDUIm6IwwnQFI-o9N8B4BGnwBfYwvA_uvmFGYAe8aYaHJB0ZD6sOo4ifYyGbYZGxQAgG0j1gGK4ibWCbhxjAukdkKkwPAgyrdY-C0LCDet0P8ac4ROpm3MIzHu0OTW_hNwF8VbPHqnnY27V4y3QcY6McQFixUd4yYbXt18YBRiKDK5XxNaycU8PSdwAluXJtzV56XgnX66J10-umddfryrbBxnrZYbsDUEfFdtR8X4cR9hTOG6wE9oYDXyoY9NhTYEDauIhy8KelgrYhkjUZDy9ADmhAZ5M-NrwA46APUk0N18YrGFZZvBSSHAFAnlLMEXwBQlrLNkPDRzSRbq4Px-HcwMbUyyxBbAAkbpOWRY5CARd5vSamRIwDUvOGPgHxYsFYVmQV0EqtalApXcPKqJpYGV3zVk6pZSunkWo-Fq126rbT37vXKQHTyWghYUWmq1OFhTjy3eMrow1uO7btWQ8eKH706IL7Ox78ZovJ4oN88mso9p5HP_zRFfbXz2ES3H7yA46-Gzf9yge_zbrYHjaic4LbkescS-hXuZ8r_9bJnEoAfmuPN3W57uQdcThdLMpUDNLqIk1z_AHatN2OaaHRX7Ml5HXsiMo2qrQ_z-o5DgY0NrncYC3zijtIev8cxYyopuCn_eX6A3DMzVpddxGV2MWtzyUL8kk5RW1BXWpCNTJcu5NQsYjgophxRpV_KPoTfzp4fExxkxqSTYyaAa4rpWibZQNdWfoEII0pcEmFhKBHIeqoRrJBt9Sqqilk0zsKVzcDxjzJsQWLUs-gMMEyMnJ7oqxvVdPqOhxStknzLG5RvVLcen5D8euViZNkiTBCGcj27Br9LpI-Nm1wK53xwyot1rtxa50mhej1EJM0-zq_Klte0ZUwJJgNG_dRyjI8K_uzBcozMWlkzNNZpQq_NImcg6qjKq2yOSqXnharVJdQZaM-yq0o8dB4XWd1nsxWqRyBGg-__vohVW_FrYODJ_hLHfcef_-V7D34TN6_enB8KnuPv3vwFGrXvIo-7THIkKpgkh4hKWqMAR667odtS6VHruCkP6hOhf1wHwweFOC8-yN0HVarZBa39sUzHPJtcndbBKaJyq249W06Tos0H0K2Hkd-bgtPjR_CL5NFseZh8JImKpsPocvrG_WFJrWe1Cf9dFQCS1vAKFHZBe2T0S5tLMRDPYNTISLCKFCI1KcK312oiGqsfAJXrapxJ8QoC97wrKN-d2-vxmMO9t1Y9NnP_T6FW9zqQ9ks7_dhlFwKWaEmz73-eDFDnEKQYaijkFS5tXjt1iRaz9WZ_YNKpte0u55UqVgap6lQxXmUKpF_KbsW6SQrUUjIGYmSa44XMgsqg8YyAPNYxgLXSC-xjVJe8e0sjJor8V0qR7RWQukti_BLXp0OyJDXuBRTLRchZSlmjaBPv0qKSVr195EVF6icbI7h6Bsl6yZmvY_CxWUTw-WZ1-ZxcI9pMzEByskZjQ8VDu4LLkTTyN_1-jV6yXrKWic5TnV3GjNC2HF9575AzZ_G4izZx48wOlIZFlQUuC8osUsMF6TuiwdUwdcRyeg8QYIbbYKr9fo5ijmRPenQb8b7FJsCRY4jZ-B7nhmEwzRIIzsdDiIvGPrBcJxiF-KPu0meLypkxlGHdC57PdPAj2rgV9ViPEvO0l4Pr5wMK84H9P4REH440oQB3HyG5NKzDPyAjfNhMpymeMQgDvTqA5iUCl0zkOBJzcB7BKCmtIUq1K0W82SG8krQ40cRD80yYo-3HUYg0yFJtvDjTEKdZIJKSfxfkx5-1bNCjB0sFlVZwWvIFnpvkU5BPYOyyNagbZikHg9Oq_ns1kGcAxWqlrfGYRxeN6oJFf3Hil4PvxSJBu4qYQFspwkMdcgrlOvBl3XORnlCU2CifMSBxzpzlC5TrJH1-H2OoaocJTnSzmLFKkEp_Dkly18mNhHD_jifJ8XZakknD83Aqwmg1st1B0dMirJPFBKTG1ixPCkupgnKqoGg10jA4D8odZJynQ-zBWF5yfPVfLkmCDXCLuUtMrIWT0-Q9LTJLoBAYGZJRRWzCG1Mw4s2sko-L_Z64Mhz1jCkykCFH8jMRiHgVJSfk-ZYcUJzsIGPCgddskxnDImUErmguQNfpSvkcyJH7BlL_P8DZl29AoqCDXMp1cNKNoDeAxGmKta8GBRAsBCRuFpWCyo479FrSCIrs0sapzdLcY4tgflbLgVvZla9kiumxeFExFI6JI1V1Wfj_33dd5xfpMXZq3QFwyj2fCCwH6qNGa-yJXwIt7LmZQmS1uv_Ba8RCAGzNgAA).

Is it our intention that this applies only to deeper nested containers? In the example above there are two things I think we might want to clarify:
- the setting "as is" causes a chart on grey background although in a container (I think that is intentional, as the outlined chart in the example shows?)
- the most inner container cannot be visually distinguished anymore, be it with our without `filled`, ie there is no "alternating" functionality or so. Is that the intentional?

The above makes me wonder why we didn't take the following alternative:
- if I want a visual distinction through container, but follow best practices, why do I not just choose `"outlined"`
- then if I want a white background on grey canvas, why don't I choose nested containers with "filled", where `filled` means different from higher level

Not sure this is better, but I feel the configuration consistency would make sense. But you may have good reasons not to do it like that :)

maxschulz-COL (approved):
Went though it in  detail, and have one bigger question, but in any case it is good to merge from my side! 👍 🚀  Very cool rework, any I like it a lot more than the previous theme, especially when flicking between old and new side by side!

nadijagraca (approved):
I like the changes, especially how the layout looks with cards and styled containers (much better than before 🙌 ).  🚀  🎨

antonymilne (approved):
I'm going to comment on this in order of things I love to things I do not love 😅 

> * **Collapse Button Design Update**: The collapse button now has a matching background color with the left-side panel and includes a border for improved visual consistency.
> * **Accordion Item Design Update**: The hover and active states of accordion items now share the same background color for a more cohesive design.

Love these - looks much nicer and more consistent now I think 👍 

> * **Clickable Card Design Update**: The clickable Card component has been redesigned. The previous "jumping" effect has been removed, and a border is now added to indicate interactivity. The Card bg changes if a filled container is on the same page.

Love the idea here (I never liked the jumping) but I think it still needs a little work. But the current implementation causes a different sort of jumping which causes other problems e.g. it can shift other elements and even alter the wrapping of text inside the card on mouseover, which isn't right. 

We should by default set a transparent 1px border or similar to avoid this: https://stackoverflow.com/questions/9612758/add-a-css-border-on-hover-without-moving-the-element
https://css-tricks.com/almanac/properties/o/outline/

> * **Light Theme Redesign:** The light theme has been updated to adjust the background color when a filled container is used. Take a look at the scratch-dev. There are basically two modes: Default and Page with filled container.

I know how much discussion this has taken so I tried really, really hard to like it but I still don't, sorry 😅 
* the asymmetry between dark and light theme feels wrong. I still feel there should be a rule where `color of area 1 in light theme == color of area 2 in light theme` if and only if `color of area 1 in dark theme ==  color of area 2 in dark theme`
* I don't like how the background flips when you put a filled container in. It's surprisingly and unexpected behaviour to me as a user and doesn't match the rest of the dashboard
* breaking the symmetry leads to more special rules/edge cases that need to be considered, e..g what happens if there's a card inside a container? We could no doubt resolve these one by one, and after a certain point of nesting we shouldn't worry about too much, but this change just seems to introduce new potential inconsistencies for (IMO) little added benefit

Just on the question of card inside a `vm.Container(variant="filled")`, I'm not sure what's expected here but background colors looks a bit off in both light and dark themes to me:

Light theme is grey then white then white:
image

Dark theme is black(ish) then grey 1 then grey 2:
image

> Before this change, the asymmetry between the themes was different: in the dark theme, filled containers had lighter backgrounds while the right-side panel was darker. In the light theme, it was the opposite—the filled container was darker, and the right-side panel was lighter. Now, we’ve made this consistent across both themes.
>
> At the end of the day, there will always be some asymmetry between dark and light themes because colors behave differently in each. But we made these changes on purpose now :)

This I half agree with. My proposed rule that `color of area 1 in light theme ==  color of area 2 in light theme` if and only if `color of area 1 in dark theme ==  color of area 2 in dark theme` doesn't say anything about how colors compare to each other in each theme. e.g. we could have `color 1 < color 2 in light theme` and `color 1 > color 2 in dark theme` or `color 1 < color 2 in light theme` and `color 1 < color 2 in dark theme`, where < and > means "lighter than/darker than". Either of these is valid and we can do whichever works best.

Of course, maybe this rule doesn't work in practice (it's just something I made up, although it does strongly feel correct to me) in which case we can break it. But the current breaking just seems too flagrant and not an overall improvement to me.

I am approving the PR anyway because:
* in reality I don't think many (any?) users will notice the asymmetry
* everyone else seems to like the changes, and I don't want my personal design preference to overrule that
* I know it's been a lot of effort to get to this point and I don't think there's much value in debating it further now that we've reached some kind of conclusion

(Don't take the criticism of the light theme changes as a criticism of the work itself... I think you've done a great job here finding a solution to an impossible problem. I just would personally have preferred a different solution.)

vizro-core/src/vizro/static/css/container.css
maxschulz-COL: Ah amazing, I guess we don't need infinite layers of nesting, but a few might be good :) 

As for the last two points:

> Choosing "outlined": You can simply choose the "outlined" style, and it will look good! There will be no color switching. That said, some users might prefer the "filled" style over "outlined," so we needed to optimize for both preferences. However, there is nothing stopping you from just using "outlined" as the visual distinction for all containers.

Indeed! But my point was more around the fact that IF we might want to prevent the canvas changing (currently part of this PR), there is still the option to visually distinguish. Which you would not have without `"Outlined"`. Visual distinction is at least one reason to use containers. Then you could use a container with `"filled"` for grey background (and if a chart is in it, it's ugly yes), but with another container with `"filled"` it is white background again.

This would make it more consistent as to what is happening precisely. Am I making sense 😄 ? Or quick call?

> Clarification on "filled": Currently, "filled" means "distinct from higher-level container", as you describe (might be clearer now that I pushed the fix?). The key distinction is that the first "filled" level must always have a white background, such that charts are displayed on a white bg.

Yes, with the bug fixed, this is 100% clear now 🚀 . Then my only point is: do we need the canvas change if we have this?

huong-li-nguyen: Yep, I think the answer to both is: Yes, we’d still need it. 😄  

In your setup, it would be a white right side, then a grey container, and then another white container for the chart inside the grey container, right? That would add an extra nested container visually, which we don’t really want. Plus, the default for the first filled container would make the chart sit on a grey background, and we’re trying to avoid that too.  

In the end, this is the kind of look and feel J. is trying to get us to, which can ultimately only be achieved via the switch 😬 

Screenshot 2025-04-14 at 14 16 18

huong-li-nguyen:
@antonymilne - If you dislike it that much, I wouldn’t rule out the idea of rediscussing it again. 😄  

I think the real question here is: Should a design like this be something Vizro supports out of the box or not? This seems to be where most of the disagreement comes from. You’re not particularly interested in it, so you don’t see much value in adding all the extra rules just for this. For J., though, it’s one of the most important visual features we need to make possible out of the box.

I’m kind of in the middle here. Would it be great if this was easy to implement? Yes. Could I just handle it with custom CSS? Yes. Do I think others should handle it with custom CSS? Maybe, but who would actually think about doing this rather than just accepting charts on grey boxes 😄 

Screenshot 2025-04-14 at 14 16 18

vizro-core/src/vizro/static/css/container.css
huong-li-nguyen: Both – it'll add redundant nesting in the configurations and visually I wouldn't recommend **nested filled containers**. It adds a lot of unnecessary visual clutter to the chart.

maxschulz-COL: Ok, I completely trust you that this solution is better then :) Just wanted to probe a bit for consistency's sake

antonymilne:
Don't worry, I don't strongly dislike it. I just strongly know that I dislike it even though I tried really hard to like it. If that makes sense 😅 

I suppose one of the reasons I'm not too fussed about creating a design that looks exactly like that is because I don't ever use light theme myself anyway. But that's not a good argument against any of these changes, just a personal bias that means my opinion should probably count for less here anyway! I definitely agree that we should try and enable best practice design out of the box without custom CSS. It's just a question for me of the balance between best practice design vs. simplicity (for users and us).

Regardless, we should absolutely merge these changes and not let the discussion hold it up!

huong-li-nguyen:
On your two remaining points / questions:

- Yes, I agree - we should add a transparent border. I added that into the [vizro-bootstrap PR](https://github.com/mckinsey/vizro-bootstrap/pull/7) directly and this has been merged together with the updates on the Card here: https://github.com/mckinsey/vizro/pull/1112. So I removed it from this PR, as it fits better into the other :)
- The `Card` isn’t really meant to be used inside a filled `Container`. In those cases, it’s better to just use `Text` instead 😄. That said, the `Card` should generally match the container’s color. This already works fine in the light theme. For the dark theme, though, the shades are a bit off right now. I’m waiting on feedback from a designer to figure out the best colors for the card and container backgrounds. I tried making the card background the same as the container background before, but it made the KPI cards look worse, so I left it as is for now as discussed with J. It’s on my list to discuss with S. though to come up with a nice gray shade that works well for both the container and card backgrounds.

PR #1052, opened 2025-03-05
[Tidy] Remove custom js and clientside callback for `AgGrid`

huong-li-nguyen:
## Description
- Follow-up on https://github.com/mckinsey/vizro/pull/1047#discussion_r1980958989
- Removal of clientside callback inside AgGrid and custom JS code to switch themes inside AgGrid
- Change pagination style to use primary and secondary font-color

It looks like we don't actually need these code snippets. I checked both dash_ag_grid and custom_ag_grid, and they both switch themes correctly without them. This is mainly because we have custom CSS for AgGrid that utilizes our CSS variables, so we don't rely on switching the default from quartz dark to light, etc.

The only initial issue was that pagination wasn't visible in the light theme, but we can resolve that with CSS.

Should we remove these snippets unless someone recalls a specific reason for keeping them? 😄 There's a slight chance there might be other issues like the pagination, but I'm not sure if we should keep the code just in case or simply address any future issues with static CSS as they come up.

## Notice


petar-qb:
I'm just looking at the code and it looks like the "sort" and the "filter" column icons are invisible on the light theme. 

| Visible | Invisible |
|--------------|----------------|
| image            | image             |

Similar happens for checkboxes in the aggrid cells if the `"cellRenderer": 'agCheckboxCellRenderer'` is defined within the `columnDefs`. 

It looks like there are countless similar examples and I'm now really sorry for raising a false alarm. 😞 
I think it's obvious that we don't want these changes then.

Perhaps, we can add a comment above the `ag-theme-quartz` and `ag-theme-quartz-dark` that explain why they exist which could be done in https://github.com/mckinsey/vizro/pull/1047.

huong-li-nguyen:
> It looks like there are countless similar examples and I'm now really sorry for raising a false alarm. 😞 I think it's obvious that we don't want these changes then.
> 
> Perhaps, we can add a comment above the `ag-theme-quartz` and `ag-theme-quartz-dark` that explain why they exist which could be done in #1047.

Ah yes, I see! Don't worry, I think it was a good point! Would have been nice to get rid of it 😄 Let's not change it then!

PR #1053, opened 2025-03-05
[Feat] Add info icon to dashboard title

nadijagraca:
## Description
POC implementation

API needs to be agreed

## Notice


huong-li-nguyen:
I just looked at the design for now as you asked :)

I think overall it looks good, I would just ensure:

- It's `center-aligned` with the title
- Add color: var(--text-primaryHover) to the icon on hover
- Change font-hierarchy inside tooltip so the text is not bigger than the actual title (it probably should use `h3`/`h4` max or just use `p`)
- Optimally, the font-size needs to be updated based on the title it is attached to (feel free to discuss this one with @Joseph-Perkins ) as well. So I feel the current font size is too small for the dashboard title.

See this article: https://hanjing.medium.com/best-practice-of-i-icon-ab6713e3aac9
- For headers: 20x20 pixels - 24x24
- For body text: 16x16 pixels - 18x18 pixels

Overall question: I am not sure if I like the position of this icon, but this is personal preference 😬 I think in general it makes sense to have them close to the title, so for any other title I would agree it should be right next to it. However, for this specific case, you have (icon on dashboard-level) it belongs next to the toggle-switch for me.

vizro-core/src/vizro/models/_components/title.py
```diff
@@ -0,0 +1,49 @@
...
+        title (str): Dashboard title to appear on every page on top left-side. Defaults to `""`.
+        icon (bool): Boolean to add information icon to dashboard title. Defaults to `""`.
+        tooltip_text (str): Markdown string to create  text that appears on icon hover.
+
+    """
+
+    type: Literal["title"] = "title"
+    title: str = Field(default="", description="Dashboard title to appear on every page on top left-side.")
+    icon: bool = False
+    tooltip_text: str = Field(description="Markdown string to create card text that appears on icon hover.")
```
antonymilne, line 25:
```suggestion
    text: str = Field(description="Dashboard title to appear on every page on top left-side.")
    icon: str = "info-mark"  # optional, can take any Material icon and defaults to info
    tooltip: str = Field(description="Markdown string to create card text that appears on icon hover.")
```

antonymilne: FYI @nadijagraca: @huong-li-nguyen, @maxschulz-COL, @Joseph-Perkins, @lingyielia and I just had a chat about this and think this would be the best API. Let me know if it makes sense.

maxschulz-COL: Yes I agree, avoiding the double `title` is good 👍

vizro-core/src/vizro/models/_page.py
```diff
@@ -64,7 +64,9 @@ class Page(VizroBaseModel):
 
     # TODO[mypy], see: https://github.com/pydantic/pydantic/issues/156 for components field
     components: conlist(Annotated[ComponentType, BeforeValidator(check_captured_callable_model)], min_length=1)  # type: ignore[valid-type]
-    title: str = Field(description="Title to be displayed.")
+    title: Union[str, Title] = Field(
+        default="", description="Dashboard title to appear on every page on top left-side."
+    )
```
maxschulz-COL, line 69:
```suggestion
    ) # random change that doesnt do anything
```

nadijagraca:
Currently, `vm.Title` is designed to return the dashboard title component, which includes both the title and an icon (with a tooltip).
As agreed, at the dashboard level, the icon should be positioned on the right, next to the theme toggle. However, if we want to reuse `vm.Title` in other components that also have a title property, some adjustments are needed to make the icon positioning more flexible.
Some of the possible approaches:
1. Via private property: Introduce a private property, ` _icon_position` , which determines the icon's position based on where `vm.Title` is being used.
2. Expose arguments: This could possibly be implemented in two ways: 

    - Add properties like `icon_size` and `icon_position` directly to the `vm.Title` model.
    - Create a separate `vm.Icon` model, which would have `size` and `position` arguments. 
3. Have one standard size and position of the icon, and advise users to use custom css to alter size/position/any other css property of the icon. 


@antonymilne @Joseph-Perkins I would like to hear your thoughts. These changes can be implemented in this PR,  or in one of the following PRs, or not implemented at all.

petar-qb (commented):
Great work @nadijagraca!! If we don't change the configuration API, this will be merged very soon 🚀

vizro-core/src/vizro/models/_components/title.py
```diff
@@ -0,0 +1,50 @@
...
+                html.H2(id="dashboard-title-text", children=self.text),
+                html.Span(self.icon, className="material-symbols-outlined", id=f"{self.id}-icon"),
+                dbc.Tooltip(
+                    id=f"{self.id}-tooltip",
+                    children=dcc.Markdown(self.tooltip, dangerously_allow_html=True, id="dashboard-title-markdown"),
+                    placement="left",
+                    target=f"{self.id}-icon",
+                ),
+            ],
+        )
```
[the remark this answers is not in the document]
nadijagraca: The `id` is one of the issues we need to address if we want to reuse  this within other components. 

I agree that `"dashboard-title"` cannot be hardcoded in that case. However, when `vm.Title` is used specifically as the dashboard title, the id of the returned html.Div must be `"dashboard-title"`, as it is referenced by ID in `_dashboard.py`. 

```
def _arrange_page_divs(self, page_divs: _PageDivsType):
    logo_title = [page_divs["logo"], page_divs["logo-dark"], page_divs["logo-light"], page_divs["dashboard-title"]]
```

One possible solution is to explicitly pass `vm.Title(id='dashboard-title' ...)` when used as a dashboard title, and while keeping `self.id` when used within other components.

vizro-core/src/vizro/models/_components/title.py
```diff
@@ -0,0 +1,50 @@
...
+    def build(self):
+        return html.Div(
+            id="dashboard-title",
+            children=[
+                html.H2(id="dashboard-title-text", children=self.text),
+                html.Span(self.icon, className="material-symbols-outlined", id=f"{self.id}-icon"),
+                dbc.Tooltip(
+                    id=f"{self.id}-tooltip",
+                    children=dcc.Markdown(self.tooltip, dangerously_allow_html=True, id="dashboard-title-markdown"),
+                    placement="left",
```
antonymilne, line 46:
```suggestion
```
auto placement (the default) seemed to work fine for me testing out your scratch_dev example. Is there a reason you fixed this as left?

If auto is ok then this would make things easier when it comes to reuse this component in other parts of the dashboard.

vizro-core/src/vizro/models/_components/title.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @_log_call
+    def build(self):
+        return html.Div(
+            id="dashboard-title",
+            children=[
+                html.H2(id="dashboard-title-text", children=self.text),
+                html.Span(self.icon, className="material-symbols-outlined", id=f"{self.id}-icon"),
+                dbc.Tooltip(
+                    id=f"{self.id}-tooltip",
```
antonymilne, line 44: Two more arguments we should consider here:
* let's set `autohide=False` so it's possible to select text/click links/etc. in the tooltip unless this causes other problems
* we have `fade=True` by default but I don't actually see any animated fade in/out, do you? 🤔 Doesn't really matter but our CSS is somehow accidentally disabling the animation then I'd be curious to see what it looks like. Not important though.

vizro-core/src/vizro/models/_components/title.py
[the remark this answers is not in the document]
antonymilne: Great points from both of you 👍 I have an idea, but let me play around with it first and will then respond here when I've tried it out.

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -228,3 +243,14 @@
...
+
+#dashboard-title-markdown h1,
+#dashboard-title-markdown h2,
+#dashboard-title-markdown h3,
+#dashboard-title-markdown h4,
+#dashboard-title-markdown h5,
+#dashboard-title-markdown h6,
+#dashboard-title-markdown p {
+  color: var(--text-primary-inverted) !important;
+}
```
antonymilne, line 256: Is there a better way to do this so that it applies to all components in the Markdown? e.g. try out this:

```
* asdf
* asdf

[link](http://www.google.com)
```

The list appears in the wrong text colour and the link is ok until you hover over it.

We used to have a similar problem with `Card` where we had to individually give styles for h1, h2, ..., p, li, a etc. and _think_ we solved it there. Is there something similar we could do here to generally solve the styling of markdown in a tooltip without having to add every possible HTML element manually?

If not then don't worry about it, just a few manual special cases like you have here is ok.

vizro-core/src/vizro/models/_components/title.py
```diff
@@ -0,0 +1,50 @@
...
+                    children=dcc.Markdown(self.tooltip, dangerously_allow_html=True, id="dashboard-title-markdown"),
```
antonymilne, line 45:
```suggestion
                    children=dcc.Markdown(self.tooltip, dangerously_allow_html=False, id="dashboard-title-markdown"),
```

Unless there was a good reason for it?

vizro-core/changelog.d/20250310_162006_nadija_ratkusic_graca_add_info_icon.md
```diff
@@ -0,0 +1,46 @@
...
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+### Added
+
+- Added `Title` model to use within `vm.Dashboard`, enabling support for info icon and tooltip in the dashboard header. ([#1053](https://github.com/mckinsey/vizro/pull/1053))
```
antonymilne, line 21:
```suggestion
- Add `Title` model to use within `vm.Dashboard`, enabling support for an info icon with tooltip in the dashboard header. ([#1053](https://github.com/mckinsey/vizro/pull/1053))
```

vizro-core/src/vizro/models/_components/title.py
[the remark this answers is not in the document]
antonymilne: Definitely let's try to `Title` generic, so it shouldn't have `dashboard` or anything specific to `dashboard` to it mentioned anywhere. This includes the `html.H2` since that is dashboard-title specific and won't apply to a title for e.g. a checklist.

Let's try this:
* `Title.build` returns `html.Div([self.text, html.Span(self.icon), dbc.Tooltip()])`. Not sure what `id` is attached to, let's figure that out later. The `html.Div` might be redundant here but is necessary to obey our rule that `build` can only return one component
* the thing that calls `title.build` is responsible for specific styling. In this case, that means that for the dashboard styling you'd do `html.H2(title.build())` so the whole thing is inside `html.H2`. For a checklist it would be like `html.Legend(title.build())`
* if the `html.Div` from `Title.build` is not wanted then you can just remove it and do `html.H2(title.build().children)` or similar instead
* we can also modify ids etc. outside the `Title.build` in the calling component. We shouldn't overuse this but it's ok to some extent. We do it with the navigation and layouts (which needs to work for both `Page` and `Container` without hardcoding `page` into the `Layout` model) and it works pretty well

vizro-core/src/vizro/static/css/layout.css
nadijagraca: Would the below work? It's more concise than the one I had before. 
```
#dashboard-title-markdown :is(h1, h2, h3, h4, h5, h6, p) {
  color: var(--text-primary-inverted);
}
```

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -110,13 +112,17 @@ def pre_build(self):
         self.pages[0].path = "/"
         meta_img = self._infer_image("app") or self._infer_image("logo") or self._infer_image("logo_dark")
 
+        from vizro.models import Title
+
+        dashboard_title = self.title.text if isinstance(self.title, Title) else self.title
```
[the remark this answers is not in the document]
maxschulz-COL: This is a very good question, thanks for pointing this out and including me in it! I have to admit that I have to remind myself everytime of what is the right solution when thinking about this.

### This may be a long one, so TLDR:

- vizro-ai plays no role here (probably?) so no concerns here
- the main problem comes from `to_python`, which struggles with after the fact conversions, because it then may emit the wrong configuration --> this is not as much of a problem as with the actions chain, because not array is involved AND `vm.Title` is (I assume?) also a valid and (equally important) public input

--> if we go for `title = Annotated[Union[str,vm.Title],AfterValidator(convert_str),Field("Foo",description="bar") then ok with me (but please add this to the `to_python` tests

### Now for the long one - what _should_ we be doing if we were 100% correct!

Ultimately I think we should follow pydantic, and they say here https://docs.pydantic.dev/latest/concepts/models/#the-long-version:
> In essence, Pydantic's primary goal is to assure that the resulting structure post-processing (termed "validation") precisely conforms to the applied type hints.

We definitely do not always obey this, but I think eventually we should. In the ideal case, our type hint in the above example would be `vm.Title` and this alone, and the user would need to configure this. This raises 2 important points:

1. If the user is allowed to configure additional types?

In our context this would be the `str` - so what if we want to allow `Union[str,vm.Title]` (but ultimately convert to `vm.Title`) as input. I think the answer is given here: https://docs.pydantic.dev/latest/concepts/validators/#json-schema-and-field-validators

> While the type hint for value is str, the cast_ints validator also allows integers. To specify the correct input type, the json_schema_input_type argument can be provided:

```python
  from typing import Any, Union
  
  from pydantic import BaseModel, field_validator
  
  
  class Model(BaseModel):
      value: str
  
      @field_validator(
          'value', mode='before', json_schema_input_type=Union[int, str]
      )
      @classmethod
      def cast_ints(cls, value: Any) -> Any:
          if isinstance(value, int):
              return str(value)
          else:
              return value
  
  
  print(Model.model_json_schema()['properties']['value'])
  #> {'anyOf': [{'type': 'integer'}, {'type': 'string'}], 'title': 'Value'}
```

So in our example, what we really should be doing is keeping the type hint at `vm.Title`, but having `json_schema_input_type=Union[str,vm.Title]`.

Beware this is for `before` validators, `after` validators are a different story again. While pydantic recommends the above, I am not sure this is so good in our case. We are (atm) not working so much with json and json schema, and our models are the main configuration way, so I am not sure what the `vm.Title` type hint would do with a different json schema input thing? Would IDE's know about this and the fact that `str` is also allowed? I don't think so... Vizro-AI would would be happy though, and so would a schema based GUI.

For `after` validators, there are some strong opinions flying about from the maestro himself - it "breaks type hints": https://github.com/pydantic/pydantic/discussions/3997#discussioncomment-3099169. I think we are doing this precise thing a fair bit lol. I agree with him actually, although of course our models are not really consumed much by anyone but us (and maybe vizro-ai) so it's not so problematic - except for the actions chain (see below)

2. What if the final type is not a subset of the input type (think for above example of just `json_schema_input_type=str`) or even worse if the coerced type is not even public (think of the actions chain in the case of `after` validators).

The is problematic for the reasons described above, but it also concretely affects some of our code: `to_python`. We fixed it for the actions chain, but in essence the model you would serialize contains things that you can't re-instantiate the model with. Not fun.

In the `str` example, this could be hacked in the `before` validator passing through any received `vm.Title`, but it's very ugly. In the case of the actions chain this would fail unless we hacked serialization (which we did).

### Summary
My preferred options are:
- `title = Annotated[Union[str,vm.Title],AfterValidator(convert_str),Field("Foo",description="bar")` (with downside that the Union type is plain wrong - but maybe not so important for us)
- `title = Annotated[vm.Title, BeforeValidator(convert_str,json_schema_input_type=Union[str,vm.Title]),Field("Foo",description="bar")` (with downside of being less clear to developers)

#### What we should definitely avoid:
- `AfterValidators` that convert into things that are not public
- `BeforeValidators` that convert into things that are not in JSON schema

This was a long write-up, hope I didn't confuse anything... so very curious about your opinions!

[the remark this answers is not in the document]
antonymilne: Wow, thanks for the great write up @maxschulz-COL! Let's make sure this is all recorded somewhere (maybe https://github.com/mckinsey/vizro/wiki) so we can find it again.

I basically agree with everything you say. I had forgotten the problem was with `to_python` itself, that's what I was thinking of 👍 I just have a few additional points:
* It has always been our intention to, wherever practical, follow the pydantic guidance that the type hint guarantees the coerced type. Just to understand, the problem you point with doing `title: Title` is that IDEs/mypy don't like it if you then specified `title` as a `str`? This is indeed quite annoying and I don't think there's much we can do about it, but let's double check with a search because surely (hopefully) someone must have wondered about this before (?).
* The other problem I foresee with this is that if and when we generate API docs automatically, it's slightly confusing because the the API docs here would give `vm.Title` without mentioning `str`, even though `str` is actually the most common user input. That's not a huge problem though so long as we explain it, especially since most of our narrative docs examples will make it clear that `title` can be a string because that's what we'll do 99% of the time.
* "For after validators, there are some strong opinions flying about from the maestro himself - it "breaks type hints": https://github.com/pydantic/pydantic/discussions/3997#discussioncomment-3099169. I think we are doing this precise thing a fair bit lol." Agree that in principle this is wrong but actually I don't think we do it that much: the only examples I can think of are `Actions/ActionsChain` and possibly in navigation (though glancing at it I'm not 100% sure it's wrong). Are there others? If so I think we should make a new issue to see if we can fix them. `Actions/ActionsChain` is definitely the most problematic and will change as part of the actions v2 work.

I'm actually quite tempted now to go for the correct `vm.Title` (no union) with a validator doing the `str` to `vm.Title` conversion. It really just depends on how annoying that is for users. Let's check if there's any way to trick IDEs/mypy into thinking that `title="string"` is a valid input for such an annotation _without a user having to `mypy: ignore`_.

* if yes and it's not a ridiculous solution on our end: use `vm.Title` as annotation
* If no or it's just crazy ugly to fix on our end: consider how annoying that IDE warning is.
  * too annoying: keep the `Union` type
  * not too annoying: go for the correct non-`Union` type and the user does `mypy: ignore`

For the case of title, which is used multiple times all over a user's dashboard configuration, I think the answer here would be "too annoying". For other things that aren't so central to the dashboard maybe it's "not too annoying" because it's just less common.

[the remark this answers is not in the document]
maxschulz-COL: > Just to understand, the problem you point with doing `title: Title` is that IDEs/mypy don't like it if you then specified `title` as a `str`?

Yes exaclty, and not just with IDE warnings, also with things like hovering over the class and check arguments, API docs (as you mentioned below) etc. I have not checked what works and doesn't work, so we should definitely do that

> Agree that in principle this is wrong but actually I don't think we do it that much: the only examples I can think of are `Actions/ActionsChain` and possibly in navigation (though glancing at it I'm not 100% sure it's wrong). 

That would be great, I also didn't have many concrete ones besides ActionsChain in my head, but just the vague thinking that we LOVE `AfterValidators` :) - but if not, all the better! I am all keen for an issue that fixes the remaining stuff

>
> I'm actually quite tempted now to go for the correct `vm.Title` (no union) with a validator doing the `str` to `vm.Title` conversion. 

We should use `json_schema_input_type=str` on top though as well right?

> 
> * if yes and it's not a ridiculous solution on our end: use `vm.Title` as annotation
> * If no or it's just crazy ugly to fix on our end: consider how annoying that IDE warning is.
>   
>   * too annoying: keep the `Union` type, store `_title_text: str` in a private property
>   * not too annoying: go for the correct non-`Union` type and the user does `mypy: ignore`
> 
> For the case of title, which is used multiple times all over a user's dashboard configuration, I think the answer here would be "too annoying". For other things that aren't so central to the dashboard maybe it's "not too annoying" because it's just less common.

Good plan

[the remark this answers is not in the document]
antonymilne: Oh, I just realised that I don't quite agree with you @maxschulz-COL in this case
> * `title = Annotated[Union[str,vm.Title],AfterValidator(convert_str),Field("Foo",description="bar")` (with downside that the Union type is plain wrong - but maybe not so important for us)

If we have `Union[str, vm.Title]` then I wouldn't do a validator to coerce string to `vm.Title`. The field would be a genuine union, and all subsequent code needs to work with both `str` and `vm.Title`.

In reality, what this means in this case is having the new `_title_text: str` private property that is populated by `title.text` in the case that `isinstance(title, Title)` or `title` when it's a `str`.

So for me the options here are:
* `Title` with coercion to `Title`. Annotation is correct. Potential problem with IDEs/mypy but we should check if there's a nice solution
* `Union[str, Title]` with no coercion. Annotation is correct. Definitely no problem with IDEs/mypy

vizro-core/src/vizro/static/css/layout.css
antonymilne: It's more concise but suffers the same problem that it only includes the HTML elements we've explicitly mentioned. Is there any way around that?

Is the styling for markdown inside a tooltip with theme=dark basically just the same as markdown _outside a tooltip_ with theme=light?

nadijagraca: The below would apply to all child elements inside `#dashboard-title-markdown`

```
#dashboard-title-markdown * {
  color: var(--text-primary-inverted);
}
```

maxschulz-COL:
> > > 3. Have one standard size and position of the icon, and advise users to use custom css to alter size/position/any other css property of the icon.

> Maybe it's unpopular but actually I don't really like the info icon appearing next to the theme switch. It feels like it gets lost over there when personally I would expect it to appear next to the dashboard title, just like it will do for other components. Then we don't need to worry about this question at all and we always put the icon directly to the right of the title.

> 3. My preferred solution since I don't actually like the icon where it appears now 😅

I agree with @antonymilne here. I would even say somewhat strongly. It looks weird to have the two things (icon and toggle) next to one another, and the icon definitely is completely disconnected from the title, especially on big screen. Currently on a large monitor, and it looks silly!

Hope I am not opening a can of works here....

antonymilne:
Thanks for all your work on @nadijagraca. It's now continued in #1053.

PR #1050, opened 2025-03-05
[Feat] Enable `Flex` layout model

huong-li-nguyen:
## Description
- Enable `vm.Flex()` as an argument for `layout` inside `Page` and `Container`
- Overall, our default is still the grid `Layout`. So if no `layout` is specified, it still takes the grid `Layout`.
- Users can now easily switch to a `Flex` layout by specifying: `layout=vm.Flex()`. No arguments are required, but optional ones are `direction`, `gap` and `wrap`. 
- `layout=vm.Flex()` returns a Flexbox with `direction=column` (to align with our current default stacking), `gap=24px` and `nowrap`.
- Introduce `LayoutType` as a discriminating Union with callable discriminator
- Height/width of flex-items can be changed directly via the relevant `style` arguments or via custom CSS. Depending on what they want to change, we might recommend either way.
    - Change dimension of individual or multiple graphs or tables -> Prefer `style` argument
    - Change dimension of all graphs or all tables -> Depends on user skill set. `style` argument requires duplication, but might be easier to implement if CSS is unfamiliar to them. Otherwise CSS is probably easier.
    - Change dimension of all flex-items of different types -> Prefer custom CSS.
  

**TO DO:**

- [x] Create Flex model with arguments: `direction`, `gap` and `wrap`
- [x] Enable `Flex` as input for `layout` inside `Page` and `Container` and `Form`
- [x] Allow non-pixel units inside `Layout` and `Flex`
- [x] Check whether flex items need to have `height` unset - currently all our components are set to height: 100% / width: 100% to fill out their cell inside the grid
- [x] Improve scroll bars visually -> can't change the ones for the horizontal scroll (deprioritised for now)
- [x] Add unit tests for `Flex`
- [x] Add docs for `Flex` -> https://github.com/mckinsey/vizro/pull/1090
- [ ]  Rename `Layout` to `Grid` -> **will be done in another PR**

**This needs to be properly tested as I am sure there will be more layout hick-ups eventually:**

- [x] Testing all kinds of combinations of components and check scroll - done by @nadijagraca 
- [x] Testing all kinds of combinations of containers / no containers / tabs / no tabs and scheck scroll - done by @nadijagraca 
- [x] Testing all kinds of nested structures of grids and flex etc. - done by @nadijagraca 
- [x] Testing combinations of `Flex` arguments visually
- [x] Setting heights for the flex-items, and see if it behaves as expected
- [x] Add unit test for non-pixel units for `Layout` and `Flex`
[...]

PR #1054, opened 2025-03-05
[Tidy] Convert built-in actions to pydantic models

antonymilne:
# Description

> [!NOTE]  
> `AbstractAction` is not yet public so all user-defined actions for now must be done as functions.

tl;dr: 
* actions are now more powerful and easier to use
* forthcoming changes after this PR will improve it even more
* this is the big first step that needs to change lots of things simultaneously, future PRs will be smaller I promise 😅 

### Built in actions

**User API:**
Before: `actions=[vm.Action(export_data(...))]`
After:  `actions=[export_data(...)]`

The old version still works for backwards compatibility but will go in 0.2.0. `export_data` will stay in 0.2.0 but `filter_interaction` will not (it will be replaced by something much better, contingent on the Dash bug fix).

What's changed?
* these actions are now _pydantic models_ and hence have a schema, validation, etc.
* `_callback_mapping` is completely gone. Instead these are done next to the function definition in the `export_data` etc. class
* no more `action.function._function.__name__ != "_filter"` comparisons or `_parameter.__wrapped__` things; instead we can do `isinstance(action, _filter)`

### Custom actions

**User API:**
Before: `actions=[vm.Action(function=my_custom_action(), inputs=["dropdown_id.value"], outputs=["card.children"])]`
After:  `actions=[vm.Action(function=my_custom_action("dropdown_id.value"), outputs=["card.children"])]`

Before: define action with `@capture("action")` and a function definition
After: same still works but there's also a new possibility to do `class my_custom_action(AbstractAction)` for advanced users. These are slightly more powerful than the function variety and can be used as if they were built in actions, like `actions=[my_custom_action(...)]` without needing the `vm.Action` wrapper.

The old version still works for backwards compatibility but will go in 0.2.0. In 0.2.0 passing static arguments to a custom action is only possible through the class definition approach.

What's changed?
* users can use special arguments like `_controls` directly in their actions to use built in Dash states
* inputs are specified directly as function arguments rather than through `inputs` which is way clearer

## Other smaller changes

- `ModelID` is now public and an alias instead of `NewType` so no need to do the annoying `cast(ModelID, ...)` everywhere. We should use `ModelID` instead of `str` where appropriate
- if `targets` not specified in `filter_interaction` then it targets everything on the page (previously it did nothing). Very unlikely to be of practical use, and I don't care much about `filter_interaction` any more since we're going to remove it, but now it matches everything else that way.

## How to review the PR

- Watch the two team learning recordings if you didn’t already see them
- Play around with the scratch app - look at the comments marked `FOR REVIEWERS`. Try writing your own actions to make sure it works as expected and you understand the new syntax and what is now possible
- Ask questions about anything you don’t understand
- Don't worry about looking through the code. Unless you're @petar-qb, who needs to go through and read every line 😀 
- Don't worry about the stuff marked `TODO NOW CHECK` - these are just things we need to manually test still
- Don't worry about the stuff marked `TODO AV2` - these are the basis for new tickets
- There aren't many new tests here, but don't worry about it
- There aren't any narrative docs in this PR, just the API ones. Read through those if you like but it's not terrible if you don't
- Nothing has deprecation warnings yet but let's all try to use the new syntax whenever we can (which should basically be always)

## TODOs

To do in this PR:
- [x] @maxschulz-COL fix generation of schema and check it looks correct
- [x] @antonymilne and @petar-qb manual testing including the `TODO NOW CHECK` points (FYI I didn't test the new validation stuff I added in [e21038c](https://github.com/mckinsey/vizro/pull/1054/commits/e21038c96af41f910440ad1bdc84bef473cdbf51) at all yet)
- [x] @petar-qb fix existing tests - let me know if there's any problems
- [x] @petar-qb to review all code
- [x] @petar-qb to check #363 and see if there's any missing TODOs or things we forgot, then close it 😮 All TODOs are in `vizro-core/examples/scratch_dev/scratch-todo.py` for now
- [ ] @petar-qb and/or @antonymilne write new tests - let's not spend too long on this though
- [ ] ~@petar-qb to read through the built API docs and see if they make sense~ No longer needed since `AbsractAction` is private for now
- [x] @antonymilne to make new tickets based on the TODOs. Look at Petar's slack message with links also

## Rough scheme of code structure 

image

## Notice


vizro-core/src/vizro/actions/export_data_action.py
```diff
@@ -1,59 +1,152 @@
...
+            for target in targets
+        }
+
+    @property
+    # TODO NOW: put these thoughts somewhere
+    # For multiple files could use single dcc.Download but zip file.
+    # Will need some way to add new components on the fly for other actions though.
+    # e.g. for key-value pairs on screen
+    # This could be built into e.g. KeyValuePairs model.
+    # Petar qn: what other actions would require new components on page?
```
petar-qb, line 116: Other actions requiring the insertion of dash_components could be:
- `drill_through_action` – if we choose to implement it using dcc.Store instead of relying on Dash routing callback inputs.
- `navigate_to_page("page-2")` – if we opt to use [dcc.Location](https://dash.plotly.com/dash-core-components/location) for navigation.

Also, since dash_components are "magically" hidden, we should clarify in the documentation how this argument is used, where these components live in the HTML DOM, and why they are hidden.

antonymilne: My idea here is that we have some components that are available for all actions to use as output with some reserved names, e.g. maybe prefixed with `vizro_`:
* definitely useful: `vizro_download`, a `dcc.Download` component
* definitely useful: `vizro_location`, a `dcc.Location` component (same as the Dash pages one but public)
* possibly usfeul: `vizro_store`: a `dcc.Store`

Then you'd be able to write an action with `output=["vizro_download"]` for example to use this. So individual actions wouldn't need to ever create new components, just use existing built in ones.

However, there are some assumptions here we should discuss since I'm not sure whether they hold...
* all these components are hidden and globally available on all pages
* it is ok for there to only be one of each component. e.g. we would need to change the current implementation of `export_data` to zip the files into one file and then send that to the single `vizro_download` object rather than dynamically creating n `dcc.Download` components like it does now. Would this work ok with `dcc.Store`?

If we're not 100% sure but think this scheme will _probably_ work then I'd like to try and do it because it would help us remove the whole `dash_components` thing which is only necessary for a couple of actions.

wdyt?

vizro-ai/src/vizro_ai/dashboard/_response_models/controls.py
```diff
@@ -1,4 +1,4 @@
-"""Controls plan model."""
+"""_Controls plan model."""
```
antonymilne, line 1:
```suggestion
"""Controls plan model."""
```

vizro-core/src/vizro/actions/export_data_action.py
petar-qb: As far as I remember, the approach of creating a separate `dcc.Download` component for each `export_data` action was originally implemented back when duplicate callback outputs weren’t allowed. Since that limitation no longer exists, this might not be necessary anymore. Another reason for this approach was to enable exporting multiple files without zipping them.

Personally, I prefer downloading multiple individual files over a single zip, though I don’t have a strong opinion on that. However, I do feel strongly that we should remove the `dash_components` property. At least from the configuration API, since it doesn’t really belong in the action definition.

I also like the general idea of having `vizro_download`, `vizro_location`, and `vizro_store` components. That said, we don’t currently use `dcc.Location` anywhere, so I don’t think we need to introduce a global `vizro_location` just yet (but in future definitely yes). The same applies to `vizro_store`. Also, we can break `vizro_store` to `vizro_session_store`/`vizro_local_store` or something like that.  

If we want to avoid exporting a zip file, we should explore other ways to handle multiple file exports. One option is to continue creating separate `dcc.Download` components per target but without relying on the `dash_components` property. Alternatively, we could make `_dash_components` a private property used only for internal actions.

If we do remove or privatise `dash_components`, we should provide an alternative way to add global helper Dash components via a more user-friendly configuration API. Right now, users can achieve this through:
- custom Dashboard components
- `app.dash.layout.children.append(dcc.Store(...))`

However, we might want to simplify this process to make it more accessible.

petar-qb: This is a great recap that I completely agree with and I don't think there's anything missing.

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -17,8 +18,8 @@ class Button(VizroBaseModel):
     Args:
         type (Literal["button"]): Defaults to `"button"`.
         text (str): Text to be displayed on button. Defaults to `"Click me!"`.
+        actions (list[ActionsType]): See [`ActionsType`][vizro.models.types.ActionsType]. Defaults to `[]`.
         href (str): URL (relative or absolute) to navigate to. Defaults to `""`.
```
antonymilne, line 22:
```suggestion
        href (str): URL (relative or absolute) to navigate to. Defaults to `""`.
        actions (list[ActionsType]): See [`ActionsType`][vizro.models.types.ActionsType]. Defaults to `[]`.
```

vizro-core/src/vizro/models/_components/form/checklist.py
```diff
@@ -1,36 +1,35 @@
...
+from vizro.models.types import ActionsType, MultiValueType, OptionsType
 
 
 class Checklist(VizroBaseModel):
     """Categorical multi-option selector `Checklist`.
 
     Can be provided to [`Filter`][vizro.models.Filter] or
-    [`Parameter`][vizro.models.Parameter].
+    [`Parameter`][vizro.models.Parameter]. Based on the underlying
+    [`dcc.Checklist`](https://dash.plotly.com/dash-core-components/checklist).
```
antonymilne, line 20:
```suggestion
    [`Parameter`][vizro.models.Parameter].
```

vizro-core/src/vizro/models/_components/form/checklist.py
```diff
@@ -40,24 +39,11 @@ class Checklist(VizroBaseModel):
     ]
     title: str = Field(default="", description="Title to be displayed")
     actions: Annotated[
-        list[Action],
+        list[ActionsType],
         AfterValidator(_action_validator_factory("value")),
         PlainSerializer(lambda x: x[0].actions),
         Field(default=[]),
     ]
```
antonymilne, line 46:
```suggestion
        Field(default=[]),
    ]
        extra: SkipJsonSchema[
        Annotated[
            dict[str, Any],
            Field(
                default={},
                description="""Extra keyword arguments that are passed to `dbc.Checklist` and overwrite any
            defaults chosen by the Vizro team. This may have unexpected behavior.
            Visit the [dbc documentation](https://dash-bootstrap-components.opensource.faculty.ai/docs/components/input/)
            to see all available arguments. [Not part of the official Vizro schema](../explanation/schema.md) and the
            underlying component may change in the future. Defaults to `{}`.""",
            ),
        ]
    ]
```

vizro-core/src/vizro/models/_components/form/date_picker.py
```diff
@@ -8,16 +8,19 @@
...
 from vizro.models._models_utils import _log_call
+from vizro.models.types import ActionsType
 
 
 class DatePicker(VizroBaseModel):
     """Temporal single/range option selector `DatePicker`.
 
     Can be provided to [`Filter`][vizro.models.Filter] or [`Parameter`][vizro.models.Parameter].
+    Based on the underlying [`dmc.DatePicker`](https://www.dash-mantine-components.com/components/datepicker) or
+    [`dmc.DateRangePicker`](https://www.dash-mantine-components.com/components/datepicker#daterangepicker).
```
antonymilne, line 23:
```suggestion
```

vizro-core/src/vizro/models/_components/form/radio_items.py
```diff
@@ -6,32 +6,32 @@
...
+from vizro.models.types import ActionsType, OptionsType, SingleValueType
 
 
 class RadioItems(VizroBaseModel):
     """Categorical single-option selector `RadioItems`.
 
     Can be provided to [`Filter`][vizro.models.Filter] or
-    [`Parameter`][vizro.models.Parameter].
+    [`Parameter`][vizro.models.Parameter]. Based on the underlying
+    [`dcc.RadioItems`](https://dash.plotly.com/dash-core-components/radioitems).
```
antonymilne, line 21:
```suggestion
    [`Parameter`][vizro.models.Parameter].
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -48,6 +47,11 @@
     "categorical": SELECTORS["numerical"] + SELECTORS["temporal"],
 }
 
+# TODO: Remove DYNAMIC_SELECTORS along with its validation check when support dynamic mode for the DatePicker selector.
+# Tuple of filter selectors that support dynamic mode
+DYNAMIC_SELECTORS = (Dropdown, Checklist, RadioItems, Slider, RangeSlider)
+DynamicNonCategoricalSelectorType = Union[Slider, RangeSlider]
+
```
antonymilne, line 54:
```suggestion
```

antonymilne:
FYI @petar-qb this is ready for you to review now!! 🎉 I'll update the description to help any other reviewers in due course.

To do in this PR:
- [ ] @maxschulz-COL fix generation of schema and check it looks correct
- [ ] @antonymilne and @petar-qb manual testing (FYI I didn't test the new validation stuff I added in [e21038c](https://github.com/mckinsey/vizro/pull/1054/commits/e21038c96af41f910440ad1bdc84bef473cdbf51) at all yet)
- [ ] @petar-qb fix existing tests - let me know if there's any problems
- [ ] @petar-qb to review all code
- [ ] @petar-qb to check #363 and see if there's any missing TODOs or things we forgot. All TODOs are in `vizro-core/examples/scratch_dev/scratch-todo.py` for now
- [ ] @petar-qb and/or @antonymilne write new tests - let's not spend too long on this though
- [ ] @petar-qb to read through the built API docs and see if they make sense
- [ ] @antonymilne to make new tickets based on the TODOs

There won't be any narrative docs in this PR, just the API ones.

@petar-qb feel free to push anything to this branch now, even outside `tests`. I don't intend to make any more big changes to the `src` code.

vizro-core/docs/pages/API-reference/action-callables.md
```diff
@@ -4,5 +4,9 @@
 API reference for all pre-defined action functions.
 
 ::: vizro.actions
+    options:
+      filters: ["!^_","!function", "!outputs", "!AbstractAction"] # Don't show underscore methods and function method
```
antonymilne, line 8:
```suggestion
     # Don't show underscore method, function or output. AbstractAction is done separately below.
      filters: ["!^_","!function", "!outputs", "!AbstractAction"]
```

vizro-core/src/vizro/_constants.py
```diff
@@ -6,8 +6,8 @@
 NONE_OPTION = "NONE"
 MODULE_PAGE_404 = "not_found_404"
 EMPTY_SPACE_CONST = -1
-ON_PAGE_LOAD_ACTION_PREFIX = "on_page_load_action"
-FILTER_ACTION_PREFIX = "filter_action"
-PARAMETER_ACTION_PREFIX = "parameter_action"
+ON_PAGE_LOAD_ACTION_PREFIX = "__on_page_load_action"
```
antonymilne, line 9: Just renamed these to make them more private.

vizro-core/src/vizro/actions/__init__.py
```diff
@@ -1,7 +1,5 @@
-from vizro.actions._filter_action import _filter
-from vizro.actions._on_page_load_action import _on_page_load
-from vizro.actions._parameter_action import _parameter
-from vizro.actions.export_data_action import export_data
-from vizro.actions.filter_interaction_action import filter_interaction
+from vizro.actions._abstract_action import AbstractAction
```
antonymilne, line 1: Note these are deliberately in `vizro.actions`, not `vizro.models`, even though they're pydantic models. So you do `from vizro.actions import export_data`, just like now.

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,42 +1,159 @@
...
+#         species = clicked_point["customdata"][0]
+#         card_1_text = f"Clicked point has sepal length {x}, petal width {y}."
+#         card_2_text = f"_Controls are `{controls}`"
+#         return card_1_text, card_2_text
+#
+#     @property
+#     def outputs(self):
+#         return (
+#             ["my_card_2.children", "my_card_1.children"] if self.swap else ["my_card_1.children", "my_card_2.children"]
+#         )
```
petar-qb, line 110: It works fine when `_transformed_outputs` is overwritten and used to return the dictionary output format from the function 👍

petar-qb (commented):
Manual testing has been done and everything is working properly

petar-qb:
I'm testing docs examples and it looks like yaml examples are broken. I'm getting the following error for the [yaml docs example](https://vizro.readthedocs.io/en/stable/pages/user-guides/actions/#__tabbed_1_2)

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Dashboard
pages.0.components.2.button.actions.0
  Unable to extract tag using discriminator 'type' [type=union_tag_not_found, input_value={'function': {'_target_': 'export_data'}}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/union_tag_not_found
```

antonymilne:
Thanks for all the testing @petar-qb! Great point on the YAML docs examples breaking. @huong-li-nguyen realised the same thing happened in #1050 - basically changing a union to a discriminated union is a breaking change because the yaml configuration needs `type` added 🤦 

The solution will be:
* keep it as a non-discriminated union for now
* possibly use functional discriminator - I'll try this out later and see how it works

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,42 +1,159 @@
...
-            title="Title",
-            footer="Footer",
-            header="Header",
             figure=px.scatter(df, x="sepal_length", y="petal_width", color="species", custom_data=["species"]),
             actions=[
                 vm.Action(
-                    function=custom_action_obtiene_seleccion(),
-                    inputs=["scatter_chart.clickData"],
-                    outputs=["my_card.children", "scatter_chart_header.children"],
+                    function=my_custom_action("scatter_chart.clickData"),
```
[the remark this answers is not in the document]
petar-qb: I see, it doesn't make sense to conduct such a validation due to:
```py
        if "inputs" in self.model_fields_set:
            legacy = True
        else:
            # If all supplied arguments look like states `<component_id>.<property>` then assume it's a new type of
            # action. For the case that there's no arguments and no inputs, this gives legacy=False.
            try:
                legacy = not all(re.fullmatch("[^.]+[.][^.]+", arg_val) for arg_val in self._runtime_args.values())
            except TypeError:
                # arg_val isn't a string so it must be treated as a legacy action.
                legacy = True
```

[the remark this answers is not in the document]
antonymilne: Yes, so if I understand your question right then it's answer by this comment in `_action.py` I think?
```
        # Validate that the runtime arguments are in the same form as the legacy Action.inputs field, so a string
        # of the form component_id.component_property. Currently this code only runs for subclasses of
        # AbstractAction but not vm.Action instances because a vm.Action that does not pass this check will
        # have already been classified as legacy in Action._legacy. In future when vm.Action.inputs is deprecated
        # then this will be used for vm.Action instances also.
        self._validate_dash_dependency(self._runtime_args.values(), type="input")
```

vizro-core/src/vizro/models/_action/_actions_chain.py
```diff
@@ -13,15 +15,30 @@ class Trigger(NamedTuple):
...
+
+    converted_actions = []
+
+    # Convert any built in actions written in the old style vm.Action(function=filter_interaction(...)) or
+    # vm.Action(function=export_data(...)) to the new style filter_interaction(...) or export_data(...).
+    # We need to delete the old action models from the model manager so they don't get built. After that,
+    # built in actions are always handled in the new way.
+    for action in actions:
+        if isinstance(action.function, (export_data, filter_interaction)):
+            del model_manager[action.id]
```
petar-qb, line 33: It makes sense that the `vm.Action` knows nothing about the `export_data` and `filter_interaction` implementations. However, it's worth questioning: should we raise a warning if the ID is assigned to the vm.Action that uses the built-in action?

vizro-core/src/vizro/models/_page.py
```diff
@@ -119,15 +127,17 @@ def pre_build(self):
         targets = figure_targets + filter_targets
 
         if targets:
+            # TODO-AV2 A 3: can we simplify this to not use ActionsChain, just like we do for filters and parameters?
```
petar-qb, line 130: Reposting from the https://github.com/mckinsey/vizro/pull/363#discussion_r2021020062:

This was added because the `on_page_load` trigger ID isn't `Page.id` (from `values["id"]`) but rather `dcc.Store.id`. Store ID is generated using `component_id_prefix` and Page ID. 

This was added into the old av2 PR so that the `Page.actions` could be customised as I can remember.

vizro-core/src/vizro/models/types.py
```diff
@@ -527,7 +546,35 @@ class OptionsDictType(TypedDict):
...
+    Union[
+        "Action",
+        "export_data",
+        "filter_interaction",
+        SkipJsonSchema["_filter"],
+        SkipJsonSchema["_parameter"],
+        SkipJsonSchema["_on_page_load"],
+    ],
+    Field(discriminator="type", description=""),
+]
```
petar-qb, line 565: Is the json schema the only reason why the `ActionsType` is not annotated as `Union[Action, AbstractAction]`?

petar-qb (commented):
I am speechless. You know what I mean about this PR 🔝 🔝 🔝 Can't wait to see this merged and released 🚀 

There are a few comments below, but they are mostly minor things.

Half of my brain is constantly trying to find a better solution for class-based and inputs/outputs management, but to no avail 😄

vizro-core/src/vizro/actions/_export_data.py
```diff
@@ -0,0 +1,109 @@
...
+
+from vizro.actions import AbstractAction
+from vizro.actions._actions_utils import _apply_filters, _get_unfiltered_data
+from vizro.managers import model_manager
+from vizro.managers._model_manager import FIGURE_MODELS
+from vizro.models._models_utils import _log_call
+from vizro.models.types import FigureType, ModelID, _Controls
+
+
+# TODO NOW: fix generation of schema and check it.
```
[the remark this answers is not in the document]
antonymilne: Yes, it is! I'll remove the TODO.

vizro-core/src/vizro/actions/_on_page_load_action.py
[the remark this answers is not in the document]
antonymilne: No, I decided those are special cases because otherwise we have two files called `_filter.py` and `_parameter.py` (for the `vizro.models` and `vizro.actions`) and that's annoying.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -1,114 +1,203 @@
...
+
+    @property
+    def _runtime_args(self) -> dict[str, _IdProperty]:
+        raise NotImplementedError
+
+    @property
+    def _action_name(self) -> str:
+        raise NotImplementedError
+
+    def _validate_dash_dependency(self, /, dependencies, *, type: Literal["output", "input"]):
```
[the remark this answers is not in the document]
antonymilne: Why? It's called with `type="output"` or `"inputs"`.

vizro-core/src/vizro/models/_action/_actions_chain.py
antonymilne: Could you explain this a bit more please? I don't understand 😅

vizro-core/src/vizro/models/_action/_action.py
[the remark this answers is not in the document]
petar-qb: I thought if we should change this in all places 😄, because it's dependencies (plural) and then validated below as a list of string.
```py
            TypeAdapter(list[Annotated[str, StringConstraints(pattern="^[^.]+[.][^.]+$")]]).validate_python(
                dependencies
            )
```

vizro-core/src/vizro/models/_action/_actions_chain.py
petar-qb: So, if users write `vm.Action(id="my_export_data_action_id", function=export_data(targets=['x']))`, they would expect that  the "my_export_data_action_id" exists somewhere (on the html or within the model_manager). This is really and an edge case and we probably don't have to care about it.

vizro-core/src/vizro/actions/_abstract_action.py
```diff
@@ -0,0 +1,99 @@
...
+        # Note order of parameters doesn't matter since we always handle things with keyword arguments.
+        return set(inspect.signature(self.function).parameters)
+
+    @property
+    def _runtime_args(self) -> dict[str, _IdProperty]:
+        # Since function is not a CapturedCallable, input arguments have not yet been bound. They correspond to the
+        # model fields that are present in the function signature. This is just the user-specified runtime arguments, as
+        # static arguments are not in the function signature (they're in self) and built in runtime arguments are not
+        # model fields. These will be of the form {"argument_name": "dropdown.value"}.
+        return {arg_name: getattr(self, arg_name) for arg_name in self.model_fields if arg_name in self._parameters}
```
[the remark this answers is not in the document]
antonymilne: Not yet but there should be in future. The only things that can be in `self._parameters` but not `self.model_fields` would be `builtin_args` (defined in `_transformed_inputs`), and currently that's just `_controls`.

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -204,12 +210,15 @@ def _get_parametrized_config(
...
+            #  Potentially this function would move to the filter_interaction action. That will be deprecated so
+            #  no need to worry too much if it doesn't work well, but we'll need to do something similar for the
+            #  new interaction functionality anyway.
+            if not isinstance(action, _parameter):
                 continue
 
-            for dot_separated_string in _get_target_dot_separated_strings(
-                action.function["targets"], target, data_frame
-            ):
+            for dot_separated_string in _get_target_dot_separated_strings(action.targets, target, data_frame):
```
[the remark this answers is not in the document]
antonymilne: Actually I _think_ (?) it might be ok because what that comment about wasn't syncing `action.function["targets"]` with `action.targets` but syncing `Parameter.targets` with `_parameter.targets`, and I don't think we actually fetch `targets` from `Parameter` anywhere.

vizro-core/src/vizro/models/_action/_actions_chain.py
antonymilne: Ah I see, you are right. Potentially I could even fix this by modifying the action id to the user specified one but I'm happy to ignore this because I don't see any reason why anyone would set an id for an action currently and then want to use that id to refer to it elsewhere, do you? (in future I can see some uses for this, but I'm sure no one does it now unless you can think of a reason for it)

vizro-core/src/vizro/models/_page.py
antonymilne: Thanks, I've added it to the comment.

vizro-core/src/vizro/models/types.py
antonymilne: That's the main reason, yes. But also I think a type hint of `AbstractAction` doesn't quite make sense because really what it needs to take is a subclass of `AbstractAction`. Just like the type hint of `_BaseAction` is correct in terms of `isinstance` checks but not a very useful type hint for the actual model field, because no one is going to actually try and provide a `_BaseAction` instance to the class.

vizro-core/src/vizro/actions/_abstract_action.py
```diff
@@ -0,0 +1,99 @@
...
+
+from vizro.models._action._action import _BaseAction
+from vizro.models.types import _IdProperty
+
+
+class AbstractAction(_BaseAction, abc.ABC):
+    """`AbstractAction` to be used in an `actions` field.
+
+    To use this class, you must subclass it and define `function` and `outputs` to make a concrete action class. All
+    built in actions follow this pattern, and it's also possible for user-defined actions. This class is not relevant
```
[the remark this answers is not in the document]
antonymilne: Good thinking but I'm not sure about the best terminology - let me have a think about it when I have a fresh head and update then!

antonymilne:
FYI @petar-qb I did a little bit of tidying and fixed the problem with YAML configuration in the last few commits. I don't have any more code changes to make for now apart from these:
https://github.com/mckinsey/vizro/pull/1054#discussion_r2024682969
https://github.com/mckinsey/vizro/pull/1054#discussion_r2024788875

maxschulz-COL:
> Play around with the scratch app - look at the comments marked FOR REVIEWERS. Try writing your own actions to make sure it works as expected and you understand the new syntax and what is now possible

@antonymilne There are not such comments?

vizro-core/src/vizro/actions/_abstract_action.py
```diff
@@ -0,0 +1,100 @@
...
+    for user-defined actions using `@capture("action")`.
+
+    When subclassing, you can optionally define model fields. These can be either static or runtime arguments and
+    define the configuration that a user specifies to use the action:
+
+      * static arguments, e.g. `file_format = "csv"`, are fixed upfront and do not change depending on the state of the
+      dashboard.
+      * runtime arguments are Dash State are determined at runtime depending on the state of the dashboard. These
```
maxschulz-COL, line 25: Here is a typo still - not sure what is the intention though?

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,42 +1,212 @@
...
+data_manager["gapminder"] = load_gapminder
+
+page_3 = vm.Page(
+    title="Experiment with data loading",
+    components=[
+        vm.AgGrid(
+            id="grid",
+            figure=dash_ag_grid(data_frame=df_gapminder),
+            actions=[
+                vm.Action(function=click_counts_heatmap("grid.cellClicked"), outputs=["card_cross_filter.children"])
```
maxschulz-COL, line 188: This action does not display the cell clicked in the Card, what am I doing wrong? @petar-qb @antonymilne

petar-qb: The problem occurs because the `id="grid"` is assigned to the outer `html.Div` that wraps the actual `AgGrid` object. 

It works if it's configured in the following way:
```py
      vm.AgGrid(
            id="grid",
            figure=dash_ag_grid(id="underlying_grid", data_frame=df_gapminder),
            actions=[
                vm.Action(function=click_counts_heatmap("underlying_grid.cellClicked"), outputs=["card_cross_filter.children"])
            ],
        ),
```

This is the known problem I discussed with Antony last week. We will try to address this when we find some time and probably as the part of this ticket -> https://github.com/McK-Internal/vizro-internal/issues/1610.

antonymilne: FYI just to be clear, this is the same before and after this PR and has been an annoyance for a while. We now have a potential route to solve it but it will remain like this until we do that.

petar-qb:
Hey @antonymilne, the following exception started occurring after [this commit](https://github.com/mckinsey/vizro/pull/1054/commits/261d50df3812a1c02250ac4bfbbc6211a5ba167b):


```
pydantic.errors.PydanticUserError: `Tag` not provided for choice {'type': 'definition-ref', 'schema_ref': '__main__.f:5210560576'} used with `Discriminator`
```

Any idea what might be causing this?

vizro-core/src/vizro/actions/_abstract_action.py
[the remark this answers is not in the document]
antonymilne: I think "built in" and "custom" is the best terminology to use but we can also say "user-defined" since it makes sense. Let's avoid using "predefined" though. Let me update this and put it in the docs ticket too.

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,35 +1,490 @@
...
+                x="gdpPercap",
+                y="lifeExp",
+                color="continent",
+            ),
+            footer="This is the result of the slow calculation/filtering operation - filtered by the continent of the selected row.",
+        ),
+        vm.Button(
+            id="button_delayed_via_state",
+            text="Make it happen",
+            # actions=[
```
maxschulz-COL, line 435: This does not cause the User input to trigger, but that is known

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,35 +1,491 @@
...
+            #         outputs=["card_via_store_model.children", "user_input_via_store_model.value"],
+            #     )
+            # ],
+        ),
+        VizroState(
+            id="store_model",
+            data={"dropdown_value": ["Asia"]},
+            # This does not work, but the callback does!
+            actions=[
+                vm.Action(
```
maxschulz-COL, line 447: This does not work, and I am not sure if it should.

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,35 +1,491 @@
...
+    title="[MS Experiment] Advanced filter interaction and caching calculations",
+    components=[
+        vm.AgGrid(
+            id="grid",
+            figure=dash_ag_grid(id="underlying_grid", data_frame=df_gapminder),
+            footer="Clicking here will trigger a slow calculation/filtering operation, but if you click the same row multiple times, it will use the cached result.",
+            # TODO: This doesn't work, but it somewhat should in the future!
+            # actions=[
+            #     vm.Action(
+            #         function=parameter_interaction("underlying_grid.cellClicked"),
```
maxschulz-COL, line 238: This does not work, but it is known

maxschulz-COL (approved):
🚀 Insane work you two 🚀 

It really is impressive. I spoke to @antonymilne for a while, so here just the short run-down.

Focussed on building actions that are different from the standard, because I trust you guys tested those enough.

Found some that work, some don't. A should be aware of most.

petar-qb (commented):
@maxschulz-COL, @antonymilne, I’m nearly there with fixing the actions unit tests. I'd appreciate your input on finalising the Pydantic-related fixes for the actions unit tests, and here's the current situation:

The plan is to first fix the existing unit tests, and then, after this PR is merged, add new unit tests for `_BaseAction`, `Action`,` _AbstractAction`, and other related action models. It's decided to merge new unit tests separately because [they are still WIP](https://github.com/mckinsey/vizro/pull/1083) and so that we don't overwhelm this PR with too many changes. 

So, I pushed   https://github.com/mckinsey/vizro/pull/1054/commits/638b943aa13a409da7b2912677b1d84a64bbc2c7 to this PR and it resolves most of the current test failures. However, there are eight unit tests that still fail (for two reasons). 


###  [SOLVED] ~1. Six failing tests in `test_text_area.py` and `test_user_input.py`:~

~These all fail for the same reason:~

```
>       text_area = TextArea()
E       pydantic.errors.PydanticUserError: `TextArea` is not fully defined; you should define `Action`, then call `TextArea.model_rebuild()`.
E       
E       For further information visit https://errors.pydantic.dev/2.10/u/class-not-fully-defined
```

### 2. Two failing tests in `test_base.py`:
- `test_to_python_builtin_actions`
- `test_to_python_complete_dashboard`

These fail because the built-in actions are incorrectly referenced as `vm.export_data()` instead of `va.export_data()`.

vizro-core/src/vizro/actions/_abstract_action.py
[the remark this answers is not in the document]
petar-qb: Is it possible for users to overwrite `_controls` for the custom abstract action? It's currently disallowed by Pydantic as I can see. If _controls: str is added as a class model field I get:
```
 E       pydantic_core._pydantic_core.ValidationError: 1 validation error for class_action_with_one_arg_and_controls
 E       _controls
 E         Extra inputs are not permitted [type=extra_forbidden, input_value='component.property', input_type=str]
```

petar-qb (approved):
I manually verified that everything is working correctly, and after reviewing the code once more, I can confidently say that this PR is ready to be merged as soon as we address the couple of test cases I flagged here: https://github.com/mckinsey/vizro/pull/1054#pullrequestreview-2760240571.

Huge thanks for this masterpiece! It's truly the cornerstone for all the upcoming backend improvements, and I'm really excited to see this unblocking so many long-planned initiatives. 🪨 🚀

maxschulz-COL:
> ### 1. Six failing tests in `test_text_area.py` and `test_user_input.py`:
> These all fail for the same reason:
> 
> ```
> >       text_area = TextArea()
> E       pydantic.errors.PydanticUserError: `TextArea` is not fully defined; you should define `Action`, then call `TextArea.model_rebuild()`.
> E       
> E       For further information visit https://errors.pydantic.dev/2.10/u/class-not-fully-defined
> ```

@antonymilne and I decided that we use `model_rebuild` on all models, so this should fix itself

> ### 2. Two failing tests in `test_base.py`:
> * `test_to_python_builtin_actions`
> * `test_to_python_complete_dashboard`
> 
> These fail because the built-in actions are incorrectly referenced as `vm.export_data()` instead of `va.export_data()`.

This can probably fixed in one place in the `to_python` bit - do you need help here?

vizro-core/src/vizro/actions/_abstract_action.py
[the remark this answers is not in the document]
petar-qb: Explanation: This is possible in general, it just doesn’t work here because field names with initial `_` are private in pydantic so aren’t actually model fields. When it becomes public `controls` then it will work exactly as expected.

vizro-core/src/vizro/models/_action/_actions_chain.py
petar-qb: I can't remember any good reason, so let's leave it as it is for now and then we can get back on it if needed.

vizro-core/changelog.d/20250415_102924_petar_pejovic_rewrite_actions.md
petar-qb, line 1: @antonymilne it's time for the final chef's touch 🧑‍🍳

Maybe we should mention Action → ActionType conversion under the "Changed" section.

vizro-core/examples/scratch_dev/app.py
maxschulz-COL: @antonymilne This is the one I meant

vizro-core/changelog.d/20250415_102924_petar_pejovic_rewrite_actions.md
```diff
@@ -0,0 +1,48 @@
...
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
```
antonymilne, line 30:
```suggestion
### Changed

- `actions` now have type `list[ActionType]` rather than `list[Action]`. ([#1054](https://github.com/mckinsey/vizro/pull/1054))
```

vizro-core/changelog.d/20250415_102924_petar_pejovic_rewrite_actions.md
antonymilne: Done in https://github.com/mckinsey/vizro/pull/1054/commits/d62890d502b855acf4116cc64ab5660698fda61f

vizro-core/examples/scratch_dev/app.py
petar-qb: This intentionally doesn't work because Vizro actions break its executions if called outside the actions_chain. This ensures that fully deterministically you know what action will be trigged when trigger component changes -> Only actions that are defined in the actions chain (in the order they are defined). It won't cause triggering other actions/callbacks defined in other places in your app even though their inputs are affected. That's the biggest difference between vizro actions and dash callbacks (that can trigger other callbacks indirectly).

vizro-core/changelog.d/20250415_102924_petar_pejovic_rewrite_actions.md
petar-qb: Any other announcements like calling built-in actions directly within the actions (like: `actions=[export_data()]`). Or defining custom action runtime inputs directly inside the custom action function?

.github/workflows/test-vizro-ai-ui.yml
```diff
@@ -45,6 +45,9 @@ jobs:
       - name: Install Hatch
         run: pip install hatch
 
+      - name: Install local Vizro
+        run: hatch run pip install ../vizro-core
+
```
petar-qb, line 50: FYI: @l0uden, @lingyielia

In terms to enable vizro-ai github actions to pass when something in the vizro-core is "broken" @l0uden added `hatch run pip install ../vizro-core` within the `vizro-ai/hatch.toml` (see [commit](https://github.com/mckinsey/vizro/commit/7d7d4486d44c97ecad42b86580415ee418d67794)), and I moved it here in the github action job. 

This step has failed because in the vizro-core `actions` type is changed from `list[Action]` -> `list[ActionType]` (see changes in this PR in the `vizro-ai/examples/dashboard-ui/components.py` file). So, github action wasn't able to import the `ActionType` from the `vizro.models.types` as the `ActionType` is newly added type (in this PR), and vizro-ai has been installing the latest released vizro-core version (instead of the local one).

vizro-core/changelog.d/20250415_102924_petar_pejovic_rewrite_actions.md
antonymilne: No, I don't want to add those to the changelog until it's properly documented, which won't be until we've added a few more features. So all this is undocumented features for now and hence not in the changelog. The only thing that potentially affects users is the type annotation.

petar-qb: Then, the PR is ready to be merged!

PR #1059, opened 2025-03-10
[Feat] Apply Visual Vocabulary to vizro-ai

lingyielia:
### How does vivivo become available in vizro-ai?
Assuming with current approach, whenever we make vizro-ai releases, run
```bash
# under vizro-core
hatch run examples:gen-vivivo
```
to generate and copy the latest `visual_vocabulary.json` to vizro-ai directory.

### How is vizro-ai leveraging vivivo now?
1. A list of example chart types is populated from vivivo and added into the field `chart_type` of `BaseChartPlan`
2. A dictionary of {chart_type: example_code} is populated from vivivo to provide code suggestion/augment to vizro-ai

For example,
augment=False:
Screenshot 2025-03-13 at 00 46 51

augment=True: 
Screenshot 2025-03-13 at 00 47 08

--------

TODOs:
- [ ] Decide whether rules like color preference should be enforced even when `augment=False`

LIMITATIONS of vizro-ai:
- Because the code generated by `.plot` needs to be compatible with `.dashboard`, we enforced that the code should always in this format
```py
def custom_chart(data_frame):
    fig = < do something >
    return fig
```
This means the vizro-ai generated code couldn't be as flexible as Visual Vocabulary. e.g., https://github.com/mckinsey/vizro/blob/77c898b620d10092c9e5bb88e1e08e42e818426b/vizro-core/examples/visual-vocabulary/pages/examples/waterfall.py#L7-L12

## Notice


vizro-core/hatch.toml
```diff
@@ -38,6 +38,7 @@ PYTHONPATH = "tests/e2e/vizro/dashboards/default:tests/tests_utils:tests/e2e/viz
 [envs.default.scripts]
 download-static-files = "python tools/download_static_files.py"
 example = "hatch run examples:example {args:scratch_dev}"  # shortcut script to underlying example environment script.
+gen-vivivo = "hatch run examples:python examples/visual-vocabulary/tools/generate_vivivo_json.py"
```
antonymilne, line 41: Can we just define this script in the `examples` env only? There's no need for the top level shortcut `hatch run gen-vivivo` here; it's fine to just do `hatch run examples:gen-vivivo`.

vizro-ai/hatch.toml
```diff
@@ -37,6 +37,7 @@ installer = "uv"
 VIZRO_AI_LOG_LEVEL = "DEBUG"
 
 [envs.default.scripts]
+copy-vivivo = "python tools/copy_visual_vocabulary.py"
```
antonymilne, line 40: I would actually just remove this script altogether in favour of:
* `copy-vivivo = "cp visual_vocabulary.json path/to/destination`
* even better: `hatch run examples:gen-vivivo` should do the file copying too, so you'd just define the script there as `["python generate_vivivo_json.py", "cp ..."]`. This way there's no need to run two hatch commands to do the whole generate + copy process.

Maybe also change `cp` to `mv` and then there's no need to `.gitignore` it also?

vizro-core/examples/visual-vocabulary/tools/generate_vivivo_json.py
```diff
@@ -0,0 +1,114 @@
...
+    possible_files = [
+        f"{group_name}_{var_name}.py",  # e.g., time_column.py
+        f"{var_name}.py",  # e.g., column.py
+    ]
+
+    for filename in possible_files:
+        example_file = examples_dir / filename
+        if example_file.exists():
+            with open(example_file) as f:
+                return f.read()
```
antonymilne, line 45:
```suggestion
            return example_file.read_text()
```

vizro-core/examples/visual-vocabulary/tools/generate_vivivo_json.py
```diff
@@ -0,0 +1,114 @@
...
+                chart_groups[group_name]["charts"].append(chart_info)
+        except ImportError as e:
+            raise ImportError(f"Error importing module for {group_name}: {e}")
+
+    visual_vocabulary = {"chart_groups": chart_groups}
+
+    output_path = Path(__file__).resolve().parent.parent / "visual_vocabulary.json"
+
+    with open(output_path, "w") as f:
+        json.dump(visual_vocabulary, f, indent=2)
```
antonymilne, line 108: Let's use `path.write_text()` here.

vizro-core/examples/visual-vocabulary/tools/generate_vivivo_json.py
```diff
@@ -0,0 +1,114 @@
+"""Generate visual vocabulary JSON from vivivo directory."""
+
+import importlib
+import json
+import re
+import sys
+from pathlib import Path
+
+# Add parent directory to path so we can import chart_groups
+sys.path.append(str(Path(__file__).resolve().parent.parent))
```
antonymilne, line 10: If we just put this script one level up then can we avoid doing this?

vizro-core/examples/visual-vocabulary/tools/generate_vivivo_json.py
```diff
@@ -0,0 +1,114 @@
...
+                var_name = page_info.title.lower().replace(" ", "_").replace("-", "_")
+                page_var = f"{var_name}_page"
+                if not hasattr(module, page_var):
+                    continue
+
+                page = getattr(module, page_var)
+
+                card_text = ""
+                for component in page.components:
+                    if component.type == "card":
```
antonymilne, line 83: Can we do this with `isinstance(vm.Card)` instead?

vizro-core/examples/visual-vocabulary/tools/generate_vivivo_json.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def get_chart_name_from_example(var_name, group_name=None):
+    """Get the chart name from the example file."""
+    if group_name and var_name.startswith(f"{group_name}_"):
+        return var_name[len(group_name) + 1 :]
+    return var_name
+
+
+def generate_visual_vocabulary():
```
antonymilne, line 57: I think this function could do with a few more comments to make it clear what it's doing and what its assumptions are. e.g. if we change the layout of the ViViVo in future, which bit here is likely to break? We assume a certain structure and that Card is used to hold the content, etc.

antonymilne (commented):
I just skimmed through at a high-level and generally looks good! 🙂 I just left a few suggestions.

vizro-core/examples/visual-vocabulary/tools/generate_vivivo_json.py
lingyielia: Yes, if one level up we could do direct import.

vizro-ai/src/vizro_ai/plot/_response_models.py
```diff
@@ -222,20 +224,41 @@ class ChartPlan(BaseChartPlan):
...
+            validate_code: Whether to validate the code.
+            chart_type_examples: List of chart type examples.
 
         Returns:
-            Chart plan model with additional validation
+            Chart plan model with additional enhancements.
         """
+        examples = chart_type_examples or []
+        description = f"""
+        {chart_plan.model_fields["chart_type"].description}
```
maxschulz-COL, line 247: Are we modifying the models description here?

maxschulz-COL (commented):
Tried to skim over the vizro-ai part (not the json creation) and I think it's a bit late for me as I am struggling to make full sense of it :( Will have to wait for Monday if you would like a review from me...

In general I would suggest making types a little clearer, for example what data am I providing where. That might make it a bit easier for me to digest.

General questions:
- I take it you have opted for the route we discussed, where we do not package vivivo, but we provide it as an occasionally updated file?
- you have taken the route of rerunning the request again if the `augment` is true?

As for the second point, I may have an alternative: have you tested how fast in terms of latency a single short LLM request is that just asks for the chart type, and gives as context the avaialble charts from vivivo (maybe with when to use info), but of course allows model to go beyond if nothing fits.

If that is fast, I'd argue why not do that first, and then send a single enhanced request that sends the example code for that chart type - and not have the model produce code twice.

If that works well, maybe we can even scrap augment? If it's hardly slower, then why not. We even have the minimal argument, which still makes it super quick

vizro-ai/src/vizro_ai/plot/_response_models.py
lingyielia: Yes, to dynamically add examples to the  `chart_type` field

vizro-ai/src/vizro_ai/_vizro_ai.py
```diff
@@ -47,9 +47,9 @@ def __init__(self, model: Optional[Union[BaseChatModel, str]] = None):
 
         """
         self.model = _get_llm_model(model=model)
-        self.components_instances = {}
+        self.chart_type_enum = _create_chart_type_enum()
+        self.chart_type_examples = [member.value for member in self.chart_type_enum]
```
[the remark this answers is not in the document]
lingyielia: chart types (chart titles). this is added into field `chart_type` eventually as examples.

```
class BaseChartPlan(BaseModel):
    """Base chart plan used to generate chart code based on user visualization requirements."""

    chart_type: str = Field(
        description="""
        Describes the chart type that:
        1. Best reflects the user request.
        2. Is feasible to implement given the data schema.
        """,
        examples=["bar", "line", ...] # all chart types now defined in vivivo 
    )

File: vizro-ai/src/vizro_ai/utils/helper.py [hunk 10]
```diff
@@ -14,3 +33,108 @@ class DebugFailure(Exception):
...
+            templates[chart_type] = example_code
+
+    return templates
+
+
+def _create_chart_type_enum() -> Enum:
+    """Create an Enum of all available chart types.
+
+    Returns:
+        Enum with all chart types available in the visual vocabulary.
```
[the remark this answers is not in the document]
lingyielia: No this enum only includes chart names (chart types)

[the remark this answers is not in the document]
lingyielia: I simplified this to directly generate a list of examples instead of "Enum then list"

vizro-ai/src/vizro_ai/plot/_response_models.py
```diff
@@ -222,20 +224,41 @@ class ChartPlan(BaseChartPlan):
...
 
 class ChartPlanFactory:
-    def __new__(cls, data_frame: pd.DataFrame, chart_plan: type[BaseChartPlan] = ChartPlan) -> type[BaseChartPlan]:
-        """Creates a chart plan model with additional validation.
+    def __new__(
+        cls,
+        data_frame: pd.DataFrame,
+        chart_plan: type[BaseChartPlan] = ChartPlan,
+        validate_code: bool = True,
+        chart_type_examples: Optional[list[str]] = None,
```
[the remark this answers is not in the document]
lingyielia: a list of possible chart types

vizro-ai/src/vizro_ai/_vizro_ai.py
```diff
@@ -78,13 +79,19 @@ def plot(
...
             _minimal_output: Internal flag to exclude chart insights and code explanations and
                 skip validation. Defaults to `False`.
+            augment: Flag to augment the BaseChartPlan.chart_code with Vizro best practices. Defaults to `False`.
 
         Returns:
             go.Figure or ChartPlan pydantic model
 
         """
         chart_plan = BaseChartPlan if _minimal_output else ChartPlan
-        response_model = ChartPlanFactory(data_frame=df, chart_plan=chart_plan) if validate_code else chart_plan
```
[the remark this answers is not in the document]
lingyielia: the `if validate_code` is moved into the `ChartPlanFactory`

```
        if validate_code:
            validators = {"validator1": field_validator("chart_code")(_test_execute_chart_code(data_frame))}
```

lingyielia:
> * I take it you have opted for the route we discussed, where we do not package vivivo, but we provide it as an occasionally updated file?
> * you have taken the route of rerunning the request again if the `augment` is true?
> 
> As for the second point, I may have an alternative: have you tested how fast in terms of latency a single short LLM request is that just asks for the chart type, and gives as context the avaialble charts from vivivo (maybe with when to use info), but of course allows model to go beyond if nothing fits.
> 
> If that is fast, I'd argue why not do that first, and then send a single enhanced request that sends the example code for that chart type - and not have the model produce code twice.
> 
> If that works well, maybe we can even scrap augment? If it's hardly slower, then why not. We even have the minimal argument, which still makes it super quick

Yes! you and @antonymilne suggested similar approach and I really like the byproduct (the JSON representation of vivivo), which I believe could be useful in other genai settings. In terms of when to update the file, we could make `hatch run examples:gen-vivivo` as an additional step in vizro-ai release process.

I thought about the alternative and I decided not to pursue it somehow. Let me give it another try. Either I will remember why I quit last time or prove it's a good route. You can take another look on Monday.

vizro-core/hatch.toml
```diff
@@ -125,14 +125,20 @@ dependencies = [
   "plotly==6.0.0"  # to leverage new MapLibre features in visual-vocabulary,
 ]
 installer = "uv"
-scripts = {example = "cd examples/{args:scratch_dev}; python app.py"}
```
antonymilne, line 128: Sorry I wasn't clearer before, but this line should stay in the default environment; otherwise we won't be able to do `hatch run example` and would need to do `hatch run examples:example` instead, which is annoying (or I'm just lazy).

The idea of the jobs that are in the default environment that refer to other environments is that they're shortcuts like `hatch run lint` that we run often so we don't have to explicitly specify the environment every time. For things like `gen-vivivo` that aren't run all the time we don't need the top-level wrapper, but for `example` we do still want it.

vizro-core/examples/visual-vocabulary/chart_groups.py
```diff
@@ -143,7 +143,7 @@ class ChartGroup:
 
 
 part_to_whole_intro_text = """
-#### Part-to-whole helps you show how one whole item breaks down into its component parts. If you consider the size of\
+#### Part-to-whole helps you show how one whole item breaks down into its component parts. If you consider the size of \
```
antonymilne, line 146: Is this intentional? Does it still render ok or has is squashed two words together?

maxschulz-COL:
> I thought about the alternative and I decided not to pursue it somehow. Let me give it another try. Either I will remember why I quit last time or prove it's a good route. You can take another look on Monday.

Ok interesting. Yes thinking about it a little more, I think that could save some tokens and LLM confusion. Let me know once you have tried, or if you would like to catch-up to discuss this a bit more. I saw the tomorrow is not possible anymore!

vizro-core/examples/visual-vocabulary/chart_groups.py
lingyielia: Yes, it was squashing two words together and cause linting error in the output json file

vizro-core/hatch.toml
lingyielia: You can still run
```
hatch run example
hatch run example visual-vocabulary
```
because in default env there is a shortcut command
```
example = "hatch run examples:example {args:scratch_dev}"  # shortcut script to underlying example environment script.
```

lingyielia: would you like me to replace this command in default env:
```
example = "hatch run examples:example {args:scratch_dev}"  # shortcut script to underlying example environment script.
```

with this command:
```
example = "cd examples/{args:scratch_dev}; python app.py"
```

so we can remove this line in example env:
```
example = "cd examples/{args:scratch_dev}; python app.py"
```

PR #1060, opened 2025-03-11
[POC] Add collapsible container as custom component

nadijagraca:
## Description
Closed PR [1042](https://github.com/mckinsey/vizro/pull/1042) and implemented all action point from that PR. 

To-do:

- [x] fix overflow when used in grid

## Notice


antonymilne:
This is great, thanks for all your work on it @nadijagraca 🙌 

I've played around with it for a while now and I think it works well enough. I've noticed it can be a little bit glitchy at times on pycafe - I guess something to do with limited screen width. Probably there's nothing we can do about it but please do take a look. The expand/collapse for the bottom container doesn't open until you've made the screen wider and then it works ok after that.

https://github.com/user-attachments/assets/7267cfe2-a05f-4c04-97d2-8c811d3984eb

Here's my current open questions:
1. Is there anything we can do about this glitch? Don't worry about it much though since I guess probably not.
2. Can we put border just around the title in the collapsed state like we discussed before? Let's ask @huong-li-nguyen when she comes back. If we can't figure it out then doesn't matter though since it ony affects the grid layout and is a minor thing.
3. Can we add a tooltip like we do when you hover over the side panel collapse button that says "Show content" and "Hide content"?
4. Should we modify the side panel collapse button to be the same single arrow rather than double to match this? See what @Joseph-Perkins and @huong-li-nguyen think. I have a weak preference for them all being single arrows so we have a single UI for this behaviour. Let's save for another PR either way.
5. Can you think of a way to get this working nicely with `dimension="width"`?  This would be nice but definitely not essential if we can't have it.
  - this is relevant for a flex layout with `flex-direction: row`
  - we would have left/right arrow instead of up/down
  - collapsed state of container should still just contain the title, but I don't know if it should then go vertical or horizontally?

All of these are minor things apart from the last one is because it affects how we structure the models. So let's talk again once we've figured that out, but I'll note down my thoughts here so I don't forget them...

_If_ we can have a version that does horizontal collapse then let's do this:

```
# in Container, a new field
collapse: Union[bool, Collapse] = False
# if user sets collapse=True then validator sets collapse=vm.Collapse()
# Then in build method we do `if not collapse`: behaviour is same as now, no collapsing
# else: use collapse.build to surround the Container.build
# Actually implementing this will be tricky so let's talk about the best way to do it.

class Collapse(VizroBaseModel):
    direction: Literal["vertical", "horizontal"] = "vertical"
    open: bool = False
```

If we can't then probably there's no point having `Collapse` model because it would only have a single field (`open`). Instead let's do as something like `collapse = True/False` and then add another option for "collapsible but open" if we really think we need it. I've gone off the idea for a more general `display` field unless we can really think of a good use for it, which currently I can't. Implementation will be simpler without a separate `Collapse` model also which is a definite advantage.

vizro-core/examples/scratch_dev/components.py
```diff
@@ -0,0 +1,96 @@
...
+
+    type: Literal["collapse_container"] = "collapse_container"
+    is_open: bool = True
+
+    @_log_call
+    def build(self):
+        """Returns custom collapsible container."""
+        clientside_callback(
+            """
+            (n_clicks, is_open) => {
```
antonymilne, line 56: This callback all looks good to me 👍 Let's put it in its own js file when this goes beyond a POC, just like we do for other clientside callbacks.

I think it's correct that it doesn't share code with `collapse_nav_panel` but it would be nice to add some comments explaining their similarities/differences. This took me a while to figure out just reading through code now.

antonymilne:
After some experimenting and discussions with the team we decided that there's no nice to way implement the horizontal collapse and no one was super keen to have it as a feature anyway. This is great news because (a) it means we can finalise the API and (b) it's the simplest possible solution.

So here's what we should:
* `vm.Container` takes a new argument `collapse: Optional[bool] = None`
* `if collapse is None`: container is like a normal one that we have now, no collapsing
* Otherwise, container is inside `dbc.Collapse` with `is_open = not collapse` i.e. if you set `collapse = True` then it's collapsed and if you set `collapse = False` then it's collapsed

@nadijagraca please do feel free to go ahead and implement this fully now, either in this PR or another one 🎉 And then I'll review properly.

PR #1079, opened 2025-03-27
[Feat] Enable collapsible containers

nadijagraca:
## Description
Enable collapsible containers by adding `collapse` argument to `vm.Container`. 

To-do:

- [ ] unit tests
- [ ] Docs

## Notice


vizro-core/src/vizro/models/_components/container.py
```diff
@@ -103,3 +117,38 @@ def _build_inner_layout(self):
...
 
         return components_container
+
+    def _build_container(self):
+        """Returns collapsible container."""
+        return (
+            dbc.Collapse(
+                id=f"{self.id}_main_container",
+                children=self._build_inner_layout(),
+                is_open=self.collapse,
```
antonymilne, line 127: This is the wrong way round!
```suggestion
                is_open=not self.collapse,
```

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -103,3 +117,38 @@ def _build_inner_layout(self):
...
+
+        return (
+            html.Div(
+                id=f"{self.id}_title",
+                children=[
+                    title,
+                    html.Span("keyboard_arrow_up", className="material-symbols-outlined", id=f"{self.id}_icon"),
+                    dbc.Tooltip(
+                        id=f"{self.id}_tooltip",
+                        children="Hide Content" if self.collapse is True else "Show Content",
```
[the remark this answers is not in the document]
antonymilne: In fact, do we actually need this line at all? Maybe the clientside callback just does it for us automatically on first load?

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -103,3 +117,38 @@ def _build_inner_layout(self):
...
+        """Returns collapsible container."""
+        return (
+            dbc.Collapse(
+                id=f"{self.id}_main_container",
+                children=self._build_inner_layout(),
+                is_open=self.collapse,
+                className="collapsible-container",
+            )
+            if self.collapse is not None
+            else self._build_inner_layout()
```
antonymilne, line 131: Very minor, but can we do in the order `if self.collapse is None` instead please?

antonymilne (commented):
Just did a first look through and it looks good! 🙌 

Let me put here the remaining open questions:
1. Is there anything we can do about the two graph glitch? Don't worry about it much though since I guess probably not. @huong-li-nguyen said she'll take a look at it. If we can't solve it then no worries though.
2. Can we put border just around the title in the collapsed state like we discussed before? @huong-li-nguyen said she'll take a look at it. If we can't solve it then no worries though.
3. Separate PR: should we modify the side panel collapse button to be the same single arrow rather than double to match this? @huong-li-nguyen thought it might look weird though and we deliberately did it double arrow. It should be quick to check though.

Possibly it would make sense to leave writing docs on this until we've done flex layout actually, because the point of the component will make much more sense then. Assuming the flexlayout can be merged within the next ~2 weeks, that is. wdyt @huong-li-nguyen?

vizro-core/src/vizro/static/js/models/container.js
```diff
@@ -0,0 +1,28 @@
+function toggle_container(n_clicks, is_open) {
```
antonymilne, line 1: Please can you add a short comment referencing the `collapse_nav_panel` callback just to explain their similarities/differences?

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,21 +1,30 @@
...
+            layout=vm.Layout(grid=[[0, 1]]),
+            variant="outlined",
+        ),
+        vm.Container(
+            title="Container title 2",
+            components=[
+                vm.Graph(figure=px.scatter(iris, x="sepal_length", y="petal_width", color="species")),
+            ],
+            collapse=True,
+            variant="outlined",
```
huong-li-nguyen, line 77: @antonymilne @nadijagraca, I think the default variant should be `outlined` when the collapse is activated. Wdyt?

Otherwise, the collapse button looks a bit lost, and it's unclear where the container starts / ends and, therefore, also where it expands/collapses.

vizro-core/src/vizro/models/_components/container.py
nadijagraca: So the container will be closed by default?

huong-li-nguyen (commented):
This looks great already! 🚀 Have a few comments and questions.

huong-li-nguyen:
> 1. Is there anything we can do about the two graph glitch? Don't worry about it much though since I guess probably not. @huong-li-nguyen said she'll take a look at it. If we can't solve it then no worries though.

@antonymilne - I couldn't reproduce the glitch anymore. I checked with small screen widths and on PyCafe and based on the current branch I couldn't reproduce it at least 🤔 I did see it once, but it quickly updated. So I couldn't reliably reproduce the glitch such that I can fix it 😅 Can you double-check if it still happens on your side with the current branch? 

Screenshot 2025-03-28 at 12 54 16
Screenshot 2025-03-28 at 12 54 28


> 2. Can we put border just around the title in the collapsed state like we discussed before? @huong-li-nguyen said she'll take a look at it. If we can't solve it then no worries though.

I've fixed it 👍 Pushed as https://github.com/mckinsey/vizro/pull/1079/commits/1e9807f5d3f9bcf396605f9ba8a7094363dac897

> 3. Separate PR: should we modify the side panel collapse button to be the same single arrow rather than double to match this? @huong-li-nguyen thought it might look weird though and we deliberately did it double arrow. It should be quick to check though.

I've also pushed that already - in general it looks ok, but we needed to reposition the icon in the collapsed state. Otherwise, the arrow wasn't visible. Pushed in https://github.com/mckinsey/vizro/pull/1079/commits/8bc004b7e97879d2dad5eda9aecae83354847ea1

> Possibly it would make sense to leave writing docs on this until we've done flex layout actually, because the point of the component will make much more sense then. Assuming the flexlayout can be merged within the next ~2 weeks, that is. wdyt @huong-li-nguyen?

I think we can already write docs, because it already works inside the grid layout and still makes sense to be added there.

vizro-core/examples/scratch_dev/app.py
nadijagraca: I agree! 👍

antonymilne: Yep, good idea!

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -49,6 +50,7 @@ class Container(VizroBaseModel):
         description="Predefined styles to choose from. Options are `plain`, `filled` or `outlined`."
         "Defaults to `plain`.",
     )
+    collapse: Optional[bool] = None
```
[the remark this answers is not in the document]
nadijagraca: I agree with @antonymilne, I actually like `collapsed` over `collapse` - it's more clearly describes the initial state of the container, but its not a hard preference.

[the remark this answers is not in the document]
huong-li-nguyen: Let's go with that then 👍

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -69,16 +71,28 @@ def build(self):
...
             "id": self.id,
             "children": [
-                html.H3(children=self.title, className="container-title", id=f"{self.id}_title"),
-                self._build_inner_layout(),
+                self._build_container_title(),
+                self._build_container(),
             ],
             "fluid": True,
-            "class_name": variants[self.variant],
+            "class_name": variants[self.variant] if self.collapse is None else variants["outlined"],
```
huong-li-nguyen, line 95: I think you have the cover the case when they configure the variant. In they want the collapse, they should still be able to configure variant. Only if not configured, it should take outlined 👍 Implementing this via a validator might be cleaner

antonymilne:
One minor thought: should we have `cursor: "pointer"` for the container title itself here to match what happens when you mouseover the collapse/expand arrow? Or even swap between "s-resize/n-resize" if that's easy to do and doesn't look weird (it might do)? wdyt @huong-li-nguyen?

Don't worry about it much though, it's just a very minor thing.

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -233,3 +233,35 @@
...
+.container-title {
+  align-items: center;
+  display: flex;
+  justify-content: space-between;
+}
+
+/* Change cursor for the whole container title when it's a collapsible container.
+Doing has(.material-symbols-outlined) isn't a very nice way to identify when the title
+is for a collapsible container but due to the order of elements, sibling selectors
+with collapsible-container don't work here. */
```
antonymilne, line 246: @huong-li-nguyen can you see any nicer way to do this?

Also should we move all this stuff to a new `container.css` file?

huong-li-nguyen (commented):
This looks amazing @nadijagraca 🚀  Thank you so much for finishing this up! Have a few comments. After updating the docs, could you also ask @stichbury to review?

vizro-core/src/vizro/static/css/layout.css
antonymilne: Cool, I was thinking of doing it that way too but didn't know if we wanted the two different `container-title-collapse` vs. `container-title` classes. It's definitely more robust than the current solution though so let's do it 👍

huong-li-nguyen:
Could you also add the example to the `dev` demo app? :)

vizro-core/examples/dev/app.py
```diff
@@ -321,6 +322,7 @@
                     )
                 )
             ],
+            collapsed=True,
```
huong-li-nguyen, line 325: I think it might be better if we create a new page - you can just use your flex example from the docs :) And remember to also add the yaml example into the `dev` example as well 👍

vizro-core/src/vizro/static/css/container.css
```diff
@@ -0,0 +1,30 @@
+.container-title,
+.container-title-collapse {
+  align-items: center;
+  display: flex;
+  justify-content: space-between;
+  margin-bottom: 0;
+}
```
huong-li-nguyen, line 7: We only want to remove the margin-bottom for the heading when it's collapsible :) 

```suggestion
.container-title,
.container-title-collapse {
  align-items: center;
  display: flex;
  justify-content: space-between;
}

.container-title-collapse {
  margin-bottom: 0;
}
```

vizro-core/src/vizro/static/css/container.css
```diff
@@ -0,0 +1,30 @@
...
+
+/* Change cursor for the whole container title when it's a collapsible container. */
+.container-title-collapse  {
+  cursor: pointer;
+}
+
+.container-title .material-symbols-outlined {
```
huong-li-nguyen, line 14:
```suggestion
.container-title-collapse .material-symbols-outlined {
```

vizro-core/src/vizro/static/css/container.css
```diff
@@ -0,0 +1,30 @@
...
+  color: var(--text-secondary);
+}
+
+.container-title .material-symbols-outlined:hover {
```
huong-li-nguyen, line 18:
```suggestion
.container-title-collapse .material-symbols-outlined:hover {
```

huong-li-nguyen (approved):
Approving already, but would be great if you could incorporate the last PR comments 👍 Thank you so much - this looks great! 🚀

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -60,6 +69,12 @@ class Container(VizroBaseModel):
         ]
     ]
 
+    @model_validator(mode="before")
```
antonymilne, line 72: This sort of works but is a bit wrong:
* we now normally write validators with `Before/AfterValidator` rather than `@model_validator`
* this one could be an after validator rather than before
* actually it doesn't need to be a validator at all but could be done with a [default factory](https://docs.pydantic.dev/latest/concepts/fields/#default-values)

I suggest you move `variant` underneath `collapsed` in the field list (so it can use `data["collapsed"]` in the default factory and do as:
```
 variant: Literal["plain", "filled", "outlined"] = Field(
        description="Predefined styles to choose from. Options are `plain`, `filled` or `outlined`."
        "Defaults to `plain` (or `outlined` for collapsible container).",
        default_factory=lambda data: "plain" if data["collapsed"] is None else "outlined"
    )
```

Do test this out to make sure it works as expected 🙏

antonymilne: Remember to update the docstring too to match - I added `` (or `outlined` for collapsible container)``.

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -16,17 +21,21 @@ Unlike `Page`, the `Container` includes a `variant` argument. This enables you t
 
 ## When to use containers
 
-In general, any arbitrarily granular layout can already be achieved by [using `Page.layout`](layouts.md) alone and is our recommended approach if you want to arrange components on a page with consistent row and/or column spacing.
+In general, any arbitrarily granular layout can already be achieved by [using `Page.layout`](layouts.md) alone, which is our recommended approach if you want to arrange components on a page with consistent row and/or column spacing.
 
-`Page.layout` has a `grid` argument that sets the overall layout of the page. `Container.layout` also has a `grid` argument. This enables you to insert a further `grid` into a component's space on the page, enabling more granular control by breaking the overall page grid into subgrids.
+Both `Page.layout` and `Container.layout` now accept either a [Grid][vizro.models.Layout] (default) or a [Flex][vizro.models.Flex] layout model, giving you full flexibility when arranging components.
```
antonymilne, line 26:
```suggestion
Both `Page.layout` and `Container.layout` now accept either a [`Grid`][vizro.models.Grid] (default) or a [`Flex`][vizro.models.Flex] layout model, giving you full flexibility when arranging components.
```

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -16,17 +21,21 @@ Unlike `Page`, the `Container` includes a `variant` argument. This enables you t
...
+
+- `Page.layout` defines the overall layout of the page using either Grid or Flex.
```
antonymilne, line 28:
```suggestion
- `Page.layout` defines the overall layout of the page using either `Grid` or `Flex`.
```

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -249,6 +258,161 @@ To make the `Container` stand out as a distinct section in your dashboard, you c
...
+To make a Container collapsible, use the `collapsed` argument:
+
+- Set `collapsed=True` to have the container start in a collapsed (hidden) state.
+- Set `collapsed=False` to have it expanded (visible) by default.
+
+Once defined, dashboard users can toggle the container’s visibility interactively.
+
+Collapsible containers are supported in both `Grid` and `Flex` layouts. However, we recommend using them within a `Flex` layout for optimal behavior, as `Flex` is better suited to dynamic sizing and more efficient use of space when content is shown or hidden.
+
+!!! example "Collapsible container inside Flex"
```
antonymilne, line 272:
```suggestion
!!! example "Collapsible container inside `Flex`"
```

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -249,6 +258,161 @@ To make the `Container` stand out as a distinct section in your dashboard, you c
...
+        ```{.python pycafe-link}
+        import vizro.models as vm
+        import vizro.plotly.express as px
+        from vizro import Vizro
+
+        iris = px.data.iris()
+
+        page = vm.Page(
+            title="Collapsible containers",
+            layout=vm.Flex(direction="column"),
```
antonymilne, line 285:
```suggestion
            layout=vm.Flex(),
```

`direction="column"` is the default.

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -249,6 +258,161 @@ To make the `Container` stand out as a distinct section in your dashboard, you c
...
+    === "app.yaml"
+
+        ```yaml
+        # Still requires a .py to add data to the data manager and parse YAML configuration
+        # See yaml_version example
+        pages:
+          - title: Collapsible containers
+            layout:
+              type: flex
+              direction: column
```
antonymilne, line 313:
```suggestion
```

huong-li-nguyen:
@nadijagraca - I fixed your remaining docs issue. It wasn't your hatch env, but the actual formatting of your docstring :)

antonymilne (approved):
This is great, thank you for all your work on it! Just a few small comments but feel free to merge when it's ready.

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -249,6 +258,161 @@ To make the `Container` stand out as a distinct section in your dashboard, you c
...
+                collapsed: false
+        ```
+
+    === "Result"
+
+        [![CollapsibleContainerFlex]][collapsiblecontainerflex]
+
+Collapsible containers can be used in `Grid` layout as well.
+
+!!! example "Collapsible container inside Grid"
```
antonymilne, line 345:
```suggestion
!!! example "Collapsible container inside `Grid`"
```

antonymilne (approved):

vizro-core/src/vizro/models/_components/container.py
antonymilne: FYI @huong-li-nguyen this is worth knowing in general if you didn't see it before.

huong-li-nguyen: Yes, @maxschulz-COL just told me about it in one of my previous PRs, but I didn't know it before. Did you guys document that somewhere - that we don't use `model_validator` etc.? Is there anything we shouldn't use / use?

maxschulz-COL: I mentioned it in the Team Learning on the V2 migration, but we didn't explicitly write it down as it didn't seem super important as it's just a convenience most of the time. The most important things can be found in the Wiki:
- https://github.com/mckinsey/vizro/wiki/Developer-docs#how-should-types-of-vizro-pydantic-models-and-the-vizro-schema-work-andor-be-interpreted 
- (mentioned in Wiki, but only in passing):`AfterValidator > BeforeValidator` whenever possible

maxschulz-COL:
> @maxschulz-COL Looks like we need to bump to `pydantic>=2.10.0` (from October 2024) to use `default_factory=lambda data: ...` like I suggested here. Do you think bumping will cause problems (e.g. on pycafe)? I can rewrite this bit of code differently if it would be problematic, but if it's fine then I would bump.

This will not be possible for PyCafe atm, for the exact same reason that `json_schema_input_type` is not possible...There is issues with the internal consistency of pyodide that is the blocker here, so we cannot upgrade to this pydantic version! @maartenbreddels showed me how to contribute to pyodide to fix it, but I just don't find the time...

vizro-core/src/vizro/models/_components/container.py
antonymilne: @huong-li-nguyen Max's team learning and notes in #vizx-dev are also worth looking at for a summary of how we do things in pydantic.

@maxschulz-COL Nothing I'm afraid... Default factories don't give anything for `default` in the schema at all.

PR #1083, opened 2025-03-28
[QA] - Rewrite actions unit tests

petar-qb:
## Description
Decided to open a separate PR as many changes are made. It's easy to merge it soon if we agree it's a better approach. 

## Notice


vizro-core/src/vizro/actions/_export_data.py
```diff
@@ -35,9 +35,6 @@ class export_data(AbstractAction):
         default="csv", description="Format of downloaded files. Defaults to `csv`."
     )
 
-    # TODO NOW: CHECK: continue testing this and eventually remove.
-    runtime_arg: str
-
```
petar-qb, line 40: This is one of the reasons why I haven't pushed tests changes into the feature PR. I needed to remove this field to make most of the tests to pass and this field is great for testing (so we should keep it still in the feat branch).

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -91,6 +91,10 @@ def __getitem__(self, arg_name: str):
     @property
     def _filter_interaction_input(self):
         """Required properties when using pre-defined `filter_interaction`."""
+        # We need to call pre_build here to ensure that the self._input_component_id is set.
+        # This is required for the filter_interaction.pre_build validation.
+        self.pre_build()
+
```
petar-qb, line 97: There were some tests that failed non-deterministically. The reason is that the `filter_interaction.pre_build` could be called before the `AgGrid.pre_build` which causes raising the AttributeError because of the unknown `self._input_component_id`.

"Fun" fact: 
Initially, I faced this problem by seeing the following exception: `Model ag-grid-id does not have required attribute '_filter_interaction_input'.`

That was really weird to me until I found that [hasattr method doesn't work as we would expect with Python properties](https://stackoverflow.com/questions/30143957/why-does-hasattr-execute-the-property-decorator-code-block).

Under the hood, `hasattr` calls `getattr(obj, name)` and returns False if `AttributeError` is raised. Guess what: The AttributeError was raised by unknown `self._input_component_id` which causes that the `hasattr` has been returning `False` sometimes (non-deterministically).

vizro-core/tests/unit/vizro/actions/test_export_data_action.py
```diff
@@ -202,19 +216,18 @@ def test_graphs_no_targets(self, ctx_export_data, gapminder_2007):
 
     @pytest.mark.usefixtures("managers_one_page_two_graphs_one_button")
     @pytest.mark.parametrize(
-        "ctx_export_data, targets",
-        [
-            ([["scatter_chart", "box_chart"], None, None, None], None),
```
petar-qb, line 207: Before, it was possible to assign the `None` to the `actions.targets` (e.g.` export_data(targets=None)`), which is currently not. And, I think there's any reason to support that, hence this test-case is removed.

vizro-core/tests/unit/vizro/actions/test_export_data_action.py
```diff
@@ -167,10 +179,10 @@ class TestExportData:
     @pytest.mark.parametrize("ctx_export_data", [([[], None, None, None])], indirect=True)
     def test_no_graphs_no_targets(self, ctx_export_data):
         # Add action to relevant component
-        model_manager["button"].actions = [vm.Action(id="test_action", function=export_data())]
+        model_manager["button"].actions = [vm.Action(function=export_data(id="test_action"))]
 
         # Run action by picking the above added action function and executing it with ()
-        result = model_manager["test_action"].function()
+        result = model_manager["test_action"].function(_controls=None)
```
petar-qb, line 185: When we get rid of `ctx`, all controls inputs will be sent through the `_controls` input action function argument. Our tests will then look much cleaner without all those `ctx_* pytest fixtures`😃

vizro-core/src/vizro/models/_components/ag_grid.py
antonymilne: Wow, great catch! 🙌 

Is it possible to solve this without needing to fix the order of `pre_build` though? Can we change the `hasattr` check to do `if attribute in dir(triggered_model)` instead? With an explanation in a comment about why we don't do `hasattr` that links to that stackoverflow post.

vizro-core/tests/unit/vizro/actions/test_export_data_action.py
antonymilne: Yes, that will be great 👍 Turning these into more "pure" functions that are more easily unit-testable is definitely a signal that it's a good change to make.

vizro-core/tests/unit/vizro/actions/test_export_data_action.py
antonymilne: Agreed.

vizro-core/tests/unit/vizro/actions/test_export_data_action.py
antonymilne, line 1: For the two public actions `export_data` and `filter_interaction` ideally we would still test the old syntax still works until we fully remove it. Looking at this file I think it will be too annoying to do by parametrising every test to handle both the new and old syntax, so instead I would just keep the old file and rename it `test_legacy_export_data_action.py`.

If you think it's too much effort to be worth doing then no worries though, it's not super important.

antonymilne (commented):
This looks great, thank you very much for doing it! I felt a few comments but generally I think you should carry on with this approach 🙏

vizro-core/tests/unit/vizro/actions/_callback_mapping/test_get_action_callback_mapping.py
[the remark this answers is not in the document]
antonymilne: These tests are less valuable now that the previously mapped data is directly part of the relevant classes so there's not much that could go wrong. But I think they're worth keeping if it's easy to do so:

* `test_action_callback_mapping_inputs` and `outputs` ➡️ `test_action.py` file. Feel free to just dump it there for now and we can work out how to rewrite this file afterwards because we should probably write some new tests in there anyway for `_transformed_inputs/outputs` or similar (would be better to test the actual registered callback rather than private method but not sure how feasible that is)
* `test_export_data_*` ➡️ `test_export_data.py`. Possibly some of these cases aren't needed any more though or it could be simplified a bit? Not sure.
* `test_known_action_unknown_argument` and `test_custom_action_mapping` no longer needed I think?

vizro-core/src/vizro/models/_components/ag_grid.py
petar-qb: Unfortunately it's not possible because the `filter_interaction.pre_build` also makes the following check:
`if "modelID" not in triggered_model._filter_interaction_input: raise ValueError(...)`

It accesses the `triggered_model._filter_interaction_input`, and the `AttributeError` will be raised anyway.

vizro-core/tests/unit/vizro/actions/test_export_data_action.py
petar-qb: Did it here -> https://github.com/mckinsey/vizro/pull/1083/commits/9af00f718a70f2ecf22365bf2e2e19589f2dceaa

We can delete the `test_legacy_export_data_action.py` and the `test_legacy_filter_interaction_action.py` when we deprecate the old way of configuration.

vizro-core/src/vizro/models/_components/ag_grid.py
petar-qb: > Or maybe comment it out with a TODO explaining why we had to remove it, and maybe we'll figure out a better way to put it back in when we rework filter interaction: 

I agree, let's do that.

petar-qb: Done that in the -> https://github.com/mckinsey/vizro/pull/1083/commits/f4f35d82868c76710be5d2d05dca2159261cad6a

vizro-core/tests/unit/vizro/actions/_callback_mapping/test_get_action_callback_mapping.py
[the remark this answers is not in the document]
petar-qb: I partially addressed what you suggested and will continue working on it after reviewing the feature PR.

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -48,47 +102,30 @@ def test_create_action_mandatory_only(self, identity_action_function):
...
-            pass
-
-        function = identity_action_function()
-        my_action = MyAction(function=function)
-
-        assert hasattr(my_action, "id")
-        assert my_action.function is function
-        assert my_action.inputs == []
-        assert my_action.outputs == []
+        # TODO: Should we test private attributes?
```
antonymilne, line 105: In this case yes, we should. In fact we should also test `_parameters` and `_runtime_args` unless there's a good reason not to?

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -99,98 +136,173 @@ def test_inputs_outputs_valid(self, inputs, outputs, identity_action_function):
             ["component.property.property"],
         ],
     )
-    def test_inputs_invalid(self, inputs, identity_action_function):
+    def test_model_filed_inputs_invalid(self, inputs, identity_action_function):
```
antonymilne, line 139:
```suggestion
    def test_model_field_inputs_invalid(self, inputs, identity_action_function):
```

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -48,47 +102,30 @@ def test_create_action_mandatory_only(self, identity_action_function):
...
+                [State("component", "property"), State("component", "property")],
             ),
         ],
     )
-    def test_inputs_outputs_valid(self, inputs, outputs, identity_action_function):
-        action = Action(function=identity_action_function(), inputs=inputs, outputs=outputs)
+    def test_model_field_inputs_valid(self, identity_action_function, inputs, _transformed_inputs):
+        action = Action(function=identity_action_function(), inputs=inputs)
 
+        assert action._legacy is True
```
antonymilne, line 126:
```suggestion
        assert action._legacy
```

antonymilne: And `assert not action._legacy` instead of `is False` also. I know this is a weaker check but it just looks weird to me currently and we know that `legacy` is just a bool.

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -99,98 +136,173 @@ def test_inputs_outputs_valid(self, inputs, outputs, identity_action_function):
...
-        assert_component_equal(predefined_filter_action, predefined_action_build_expected)
+        assert action._legacy is False
+        assert action.inputs == []
+        assert action._transformed_inputs == _transformed_inputs
 
+    # TODO: Add more tests here and reorder the tests to be more logical.
+    # TODO: Should the "_legacy" testing be removed?
+    # TODO: Should we test both runtime arguments as list and dict?
+    @pytest.mark.parametrize(
+        "custom_action_with_parameters, runtime_inputs, model_field_inputs, expected_legacy, _transformed_inputs",
```
antonymilne, line 184: I was expecting these tests to handle only the `@capture("action")` case and not the `AbstractAction` case so not sure why `model_field_inputs` are relevant?

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -1,45 +1,99 @@
...
+            pass
+    """)
+
+    local_vars = {}
+    exec(func_code, {}, local_vars)
+
+    return local_vars["custom_action"]
 
 
+# TODO: This is used in the single test only. Should it be moved there?
```
antonymilne, line 49: I think it's good here!

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -1,45 +1,99 @@
...
+
+        @capture("action")
+        def custom_action({parameters}):
+            pass
+    """)
+
+    local_vars = {}
+    exec(func_code, {}, local_vars)
+
+    return local_vars["custom_action"]
```
antonymilne, line 46: This is very clever but I think we should avoid doing it if at all possible 😅 Can we instead just do just by defining 3 separate functions for 0, 1 and 2 arguments? Those can still be passed in as `pytest.mark.parametrize`.

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -99,98 +136,173 @@ def test_inputs_outputs_valid(self, inputs, outputs, identity_action_function):
...
-    def test_predefined_export_data_action_build(self, predefined_action_build_expected):
-        predefined_filter_action = model_manager["test_page"].controls[0].selector.actions[0].actions[0].build()
-        assert_component_equal(predefined_filter_action, predefined_action_build_expected)
+        assert action._legacy is False
+        assert action.inputs == []
+        assert action._transformed_inputs == _transformed_inputs
 
+    # TODO: Add more tests here and reorder the tests to be more logical.
+    # TODO: Should the "_legacy" testing be removed?
+    # TODO: Should we test both runtime arguments as list and dict?
```
antonymilne, line 182: So generally I've established a pattern in #1098 that I think we should try to follow:
> Where feasible and worth the effort to do so, copy and paste tests that use old API and label as `legacy` (e.g. `legacy_layout`). Write new tests to check `FutureWarning` is emitted where expected and otherwise filter out `FutureWarning` to ensure the old tests still pass. Depending on how important/widespread the change is, this might require copy and pasting a single file or several tests spread throughout different files across the test codebase. Don't `pytest.mark.parameterize` tests to check old and new API, just copy and paste instead. This will make it easier to remove the legacy features in future without needing to rewrite tests.

We haven't put the deprecation warnings in yet (that will be a whole different PR after we've merged #1054) but we should try to arrange our tests already so that tests of legacy behaviour and new behaviour are clearly separated. e.g. with the `test_legacy_export_data` file, this is great because when it comes to Vizro 0.2.0 all we need to do is delete that file to update tests.

So here I would also try to split off `legacy` tests. This could be done various ways and I don't mind which, it depends on which tests are duplicated really:
* new `test_legacy_action` file
* new `TestLegacyXYZ` class(es)
* new `test_legacy_xyz` function(s)

Importantly it **shouldn't** be done with `pytest.mark.parameters` running the same test in both legacy and non-legacy mode. Take a look at the test diff in #1098 to see how this is done.

In practice what this means is:
* keep old tests that are relevant but make sure they're labelled as `legacy`
* when legacy tests would need to be updated, it depends on the situation whether it's worth the effort to do so or not, e.g. whether we should test `_transformed_inputs` for legacy actions. Probably in this case it is worth the effort
* write the new tests independently of those legacy tests (but obviously many cases will be inspired by the old tests)
* don't worry if there's some redundancy/many tests/things which would be neater if written with parametrized - it's ok in this context

Sorry this wasn't clear before - I only really thought of it this morning.

vizro-core/tests/unit/vizro/models/_action/test_action.py
petar-qb: Sorry for bad wording.. I used the `"model_fileld_inputs"` for the inputs that are assigned to the `Action.inputs` and the `"runtime_inputs"` for the inputs that are sent as the `Action.function` arguments like here:
`action = Action(function=custom_action_with_parameters(**runtime_inputs), inputs=model_field_inputs)`

I'll rename "model_fileld_inputs" to something like "action_inputs" or you have another suggestion.

antonymilne: Ah I see. `action_inputs` is good I think 👍 

This will all be much clearer once it's split into legacy tests separately from non-legacy tests because then it's obvious that `action_inputs` only exists in the legacy case.

I think we should have a parameterised test for "is this action detected as legacy" that just tests `assert (not) action._legacy` and doesn't need to check `_transformed_inputs`. Then the `_transformed_inputs` would be separate tests elsewhere.

antonymilne (approved):

PR #1089, opened 2025-04-02
[Tidy] Replace __init__ with model_post_init

maxschulz-COL:
## Description
Essentially just a small tidy exercise that:
- makes the API docs nicer
- should improve the IDE behaviour of pydantic models

## Notice


antonymilne (approved):
Thanks for doing this!

Actually the effect of this is more than just a small tidy:
1. our API docs have changed:
  * this bit is no longer there, which is good and part of the point of doing this
image
  * `Page.model_post_init` now appears in the API docs and shouldn't, so please add it to the `filters` list to ignore in `models.md`
2. most importantly, and really the main reason for making this change, this _should_ hopefully improve our IDE completion if you have some pydantic plugin installed🤞 If you type `vm.Card(` in VS Code and PyCharm then does it now suggest the field names. If yes then that's great and we should put this in the changelog and maybe also a tip in the docs linking to the suitable pydantic extension in VS Code and PyCharm.

Just to double check: are there any other models besides `Page` where we had defined `__init__` before? I don't think so.

maxschulz-COL:
> 1. our API docs have changed:

Ah good shout, I had not thought about this

>    * `Page.model_post_init` now appears in the API docs and shouldn't, so please add it to the `filters` list to ignore in `models.md`
👍 

> 2. most importantly, and really the main reason for making this change, this _should_ hopefully improve our IDE completion if you have some pydantic plugin installed🤞 If you type `vm.Card(` in VS Code and PyCharm then does it now suggest the field names. If yes then that's great and we should put this in the changelog and maybe also a tip in the docs linking to the suitable pydantic extension in VS Code and PyCharm.

Will try out, but I think if I remember correctly, the VS Code one is less good than the PyCharm one, let's see

> Just to double check: are there any other models besides `Page` where we had defined `__init__` before? I don't think so.

Will also double check, Page was the only one caught by tests.

maxschulz-COL:
As for the IDE, is that what you mean @antonymilne :

image

maxschulz-COL:
> As for the IDE, is that what you mean @antonymilne :
> 
> image

Someone should probably check for PyCharm. For me autocomplete works, but the strict check for required args doesn't..

vizro-core/changelog.d/20250402_101816_maximilian_schulz_model_post_init.md
```diff
@@ -0,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+
+### Highlights ✨
+
+- Vizro models now have autocomplete functionality for configuration when using the relevant [configuration/plugin](https://docs.pydantic.dev/latest/integrations/visual_studio_code/) ([#1089](https://github.com/mckinsey/vizro/pull/1089))
```
maxschulz-COL, line 10:
```suggestion
- Vizro models now have autocomplete functionality in IDE configuration when using the relevant [configuration/plugin](https://docs.pydantic.dev/latest/integrations/visual_studio_code/) ([#1089](https://github.com/mckinsey/vizro/pull/1089))
```

antonymilne (changes requested):
De-approving just temporarily because the pydantic plugin in PyCharm doesn't seem to work properly for me at all so it's hard to check. Let me check with @petar-qb later to see how it works for him, and then I'll add a note to the docs about it if it works well in that. If not we'll still merge anyway, but I'd like to check this first.

maxschulz-COL:
> LGTM 👍 Not sure if relevant, but might be worth double-checking vizro-ai? I found this:
> 
> https://github.com/mckinsey/vizro/blob/ba5a8234cd326e49e4110c2813c87d8962c492f9/vizro-ai/src/vizro_ai/dashboard/_response_models/page.py#L86

I think it is fine to keep here as people never instantiate the model directly. One would need to test if there is any benefit of the post init, but I would deprioritize atm, unless you disagree @lingyielia

petar-qb:
Here's what I have of the PyCharm plugins installed:
image


When I type `vm.Button()`, this is all I get (same from the `main` and from the `feat/model_post_init` branch):
image

When I type `vm.Graph(ti)`, this is all I get (same from the `main` and from the `feat/model_post_init` branch):
image

vizro-core/changelog.d/20250402_101816_maximilian_schulz_model_post_init.md
```diff
@@ -0,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+
+### Highlights ✨
+
+- Vizro models now have autocomplete functionality in IDE configuration when using the relevant [configuration/plugin](https://docs.pydantic.dev/latest/integrations/visual_studio_code/) ([#1089](https://github.com/mckinsey/vizro/pull/1089))
```
antonymilne, line 10:
```suggestion
- Vizro models now have autocomplete functionality in IDEs when using the pydantic [plugin for VS Code](https://docs.pydantic.dev/latest/integrations/visual_studio_code/) or [for PyCharm](https://docs.pydantic.dev/latest/integrations/pycharm/). ([#1089](https://github.com/mckinsey/vizro/pull/1089))
```

antonymilne (approved):
Thanks for checking this @petar-qb! It confirms that there's something wrong with my PyCharm... 

@maxschulz-COL given this does seem to work well, please can you put a `!!! tip` or `!!! note` in some place in our docs pointing to the plugins? (Not sure where tbh...)

PR #1092, opened 2025-04-02
[PoC Info icon for dashboard, page and dropdown

antonymilne:
## Description

/*TODO NOW: fix mouseover links,. --bs-link-hover-color-rgb-inverted*/
# TODO NOW: note in docs it only shows when there's a title
feels a bit low
Note could deuplidcate Tooltip but not for now

## Screenshot

image

## Notice

### DIFF
--- a/vizro-core/changelog.d/20250401_113146_antony.milne_info_icon.md
+++ b/vizro-core/changelog.d/20250401_113146_antony.milne_info_icon.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Security
+
+- A bullet item for the Security category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
--- a/vizro-core/src/vizro/models/_components/form/dropdown.py
+++ b/vizro-core/src/vizro/models/_components/form/dropdown.py
@@ -4,14 +4,15 @@ from typing import Annotated, Any, Literal, Optional, Union, cast
 
 import dash_bootstrap_components as dbc
 from dash import dcc, html
-from pydantic import AfterValidator, Field, PrivateAttr, StrictBool, ValidationInfo, model_validator
+from pydantic import AfterValidator, Field, PrivateAttr, StrictBool, ValidationInfo, model_validator, BeforeValidator
 from pydantic.functional_serializers import PlainSerializer
 from pydantic.json_schema import SkipJsonSchema
 
-from vizro.models import Action, VizroBaseModel
+from vizro.models import Action, Tooltip, VizroBaseModel
 from vizro.models._action._actions_chain import _action_validator_factory
 from vizro.models._components.form._form_utils import get_options_and_default, validate_options_dict, validate_value
 from vizro.models._models_utils import _log_call
+from vizro.models._tooltip import coerce_str_to_tooltip
 from vizro.models.types import MultiValueType, OptionsDictType, OptionsType, SingleValueType
 
 
@@ -71,6 +72,7 @@ class Dropdown(VizroBaseModel):
             [`MultiValueType`][vizro.models.types.MultiValueType]. Defaults to `None`.
         multi (bool): Whether to allow selection of multiple values. Defaults to `True`.
         title (str): Title to be displayed. Defaults to `""`.
+        description (Optional[Tooltip]): Additional information about the selector shown in a tooltip.
         actions (list[Action]): See [`Action`][vizro.models.Action]. Defaults to `[]`.
         extra (Optional[dict[str, Any]]): Extra keyword arguments that are passed to `dcc.Dropdown` and overwrite any
             defaults chosen by the Vizro team. This may have unexpected behavior.
@@ -93,6 +95,11 @@ class Dropdown(VizroBaseModel):
         Field(default=True, description="Whether to allow selection of multiple values", validate_default=True),
     ]
     title: str = Field(default="", description="Title to be displayed")
+    description: Annotated[
+        Optional[Tooltip],
+        BeforeValidator(coerce_str_to_tooltip, json_schema_input_type=Union[str, Tooltip]),
+        Field(default=None, description="Additional information about the selector shown in a tooltip."),
+    ]
     actions: Annotated[
         list[Action],
         AfterValidator(_action_validator_factory("value")),
@@ -127,6 +134,7 @@ class Dropdown(VizroBaseModel):
         full_options, default_value = get_options_and_default(options=options, multi=self.multi)
         option_height = _calculate_option_height(full_options)
         altered_options = _add_select_all_option(full_options=full_options) if self.multi else full_options
+        description = self.description.build().children if self.description is not None else [None]
 
         defaults = {
             "id": self.id,
@@ -141,7 +149,7 @@ class Dropdown(VizroBaseModel):
 
         return html.Div(
             children=[
-                dbc.Label(self.title, html_for=self.id) if self.title else None,
+                dbc.Label([self.title, *description], html_for=self.id) if self.title else None,
                 dcc.Dropdown(**(defaults | self.extra)),
             ]
         )
--- a/vizro-core/src/vizro/models/_dashboard.py
+++ b/vizro-core/src/vizro/models/_dashboard.py
@@ -4,7 +4,7 @@ import base64
 import logging
 from functools import partial
 from pathlib import Path
-from typing import TYPE_CHECKING, Annotated, Literal, Optional, TypedDict, cast
+from typing import TYPE_CHECKING, Annotated, Literal, Optional, TypedDict, cast, Union
 
 import dash
 import dash_bootstrap_components as dbc
@@ -22,15 +22,16 @@ from dash import (
     html,
 )
 from dash.development.base_component import Component
-from pydantic import AfterValidator, Field, ValidationInfo
+from pydantic import AfterValidator, Field, ValidationInfo, BeforeValidator
 
 import vizro
 from vizro._constants import MODULE_PAGE_404, VIZRO_ASSETS_PATH
 from vizro._themes.template_dashboard_overrides import dashboard_overrides
 from vizro.actions._action_loop._action_loop import ActionLoop
-from vizro.models import Navigation, VizroBaseModel
+from vizro.models import Navigation, VizroBaseModel, Tooltip
 from vizro.models._models_utils import _log_call
 from vizro.models._navigation._navigation_utils import _NavBuildType
+from vizro.models._tooltip import coerce_str_to_tooltip
 
 if TYPE_CHECKING:
     from vizro.models import Page
@@ -88,7 +89,8 @@ class Dashboard(VizroBaseModel):
             Defaults to `vizro_dark`.
         navigation (Navigation): See [`Navigation`][vizro.models.Navigation]. Defaults to `None`.
         title (str): Dashboard title to appear on every page on top left-side. Defaults to `""`.
-
+        description (Optional[Tooltip]): Additional information about the dashboard shown in a tooltip and used for
+            description meta tag.
     """
 
     pages: list[Page]
@@ -99,6 +101,15 @@ class Dashboard(VizroBaseModel):
         Optional[Navigation], AfterValidator(set_navigation_pages), Field(default=None, validate_default=True)
     ]
     title: str = Field(default="", description="Dashboard title to appear on every page on top left-side.")
+    description: Annotated[
+        Optional[Tooltip],
+        BeforeValidator(coerce_str_to_tooltip, json_schema_input_type=Union[str, Tooltip]),
+        Field(
+            default=None,
+            description="Additional information about the dashboard shown in a tooltip and used for description "
+            "meta tag.",
+        ),
+    ]
 
     @_log_call
     def pre_build(self):
@@ -111,10 +122,13 @@ class Dashboard(VizroBaseModel):
         meta_img = self._infer_image("app") or self._infer_image("logo") or self._infer_image("logo_dark")
 
         for order, page in enumerate(self.pages):
+            # Dash also uses the dashboard-level description passed into Dash() as the default for page-level
+            # descriptions, but this would involve extracting dashboard.description and inserting it into the Dash app
+            # config in Vizro.build. What we do here is simpler but has the same effect.
             dash.register_page(
                 module=page.id,
                 name=page.title,
-                description=page.description,
+                description=page.description.text if page.description is not None else self.description,
                 image=meta_img,
                 title=f"{self.title}: {page.title}" if self.title else page.title,
                 path=page.path,
@@ -190,8 +204,9 @@ class Dashboard(VizroBaseModel):
 
     def _get_page_divs(self, page: Page) -> _PageDivsType:
         # Identical across pages
+        dashboard_description = self.description.build().children if self.description is not None else [None]
         dashboard_title = (
-            html.H2(id="dashboard-title", children=self.title)
+            html.H2(id="dashboard-title", children=[self.title, *dashboard_description])
             if self.title
             else html.H2(id="dashboard-title", hidden=True)
         )
@@ -219,7 +234,8 @@ class Dashboard(VizroBaseModel):
 
         # Shared across pages but slightly differ in content. These could possibly be done by a clientside
         # callback instead.
-        page_title = html.H2(id="page-title", children=page.title)
+        page_description = page.description.build().children if page.description is not None else [None]
+        page_title = html.H2(id="page-title", children=[page.title, *page_description])
         # cannot actually be None if you check pages and layout field together
         navigation: _NavBuildType = cast(Navigation, self.navigation).build(active_page_id=page.id)
         nav_bar = navigation["nav-bar"]
--- a/vizro-core/src/vizro/models/_models_utils.py
+++ b/vizro-core/src/vizro/models/_models_utils.py
@@ -17,6 +17,11 @@ def _log_call(method):
     return _wrapper
 
 
+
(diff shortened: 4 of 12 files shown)


vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -99,6 +101,15 @@ class Dashboard(VizroBaseModel):
         Optional[Navigation], AfterValidator(set_navigation_pages), Field(default=None, validate_default=True)
     ]
     title: str = Field(default="", description="Dashboard title to appear on every page on top left-side.")
+    description: Annotated[
+        Optional[Tooltip],
+        BeforeValidator(coerce_str_to_tooltip, json_schema_input_type=Union[str, Tooltip]),
```
antonymilne, line 106: FYI @maxschulz-COL maybe in future we should define some more of these `Annotated[...]` types instead of copy and pasting long types between different models. it will make documentation a bit of a hassle though, so I think it's ok to just leave like this for now.

vizro-core/src/vizro/static/css/tooltip.css
```diff
@@ -0,0 +1,11 @@
+.tooltip-icon
+{
+    cursor: pointer;
+    font-size: max(20px, 1em);
```
antonymilne, line 4: @huong-li-nguyen I played around for a while with the size of this. Since I put the icon inside the `h2/legend/whatever` I wanted to set `font-size` as a fixed % of the the parent element. Unfortunately this didn't work well because for selector titles with `html.Legend` the font-size is very small and for headings it's very big, and there's no fixed % that looks sensible for both of these. Any size less than 20px for selector titles I found just looked too small, but then if you apply that same % (20px / 14px = 1.42) to h2 then the icon is huge. Hence why I came up with `max(20px, 100%)` so it can't shrink too small.

Actually this as a `max` is kind of pointless because currently the icon is only in two possible places: `h2`, which has font-size of 24px (the default icon size anyway) and `legend`, where the icon is 20px. So alternatively we could just set `.tooltip-icon` to 20px when it's in legend.

wdyt?

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
```diff
@@ -113,6 +113,19 @@ label[for="theme-selector"] {
   padding: 0;
 }
 
+.tooltip-inner {
+  max-width: 500px !important;
+}
+
+/* TODO NOW: fix light theme readability */
```
antonymilne, line 120: Is it just me or is this very hard to read in the light theme? I feel like we need more contrast or a bigger font weight.

image

@maxschulz-COL @huong-li-nguyen @nadijagraca please tell me if this is just my eyes or whether you also think the tooltips are ok to read in dark theme but not light.

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
```diff
@@ -113,6 +113,19 @@ label[for="theme-selector"] {
...
+.tooltip-inner * {
+  color: var(--text-primary-inverted);
+}
+
+.tooltip-inner a:hover {
+  color: var(--text-primaryHover-inverted);
+}
```
antonymilne, line 127: @huong-li-nguyen something like this is needed to ensure that the link doesn't disappear when you hover over it. I wasn't sure the right variable to go for but this sounded like it might be the right choice?

vizro-core/src/vizro/static/css/tooltip.css
```diff
@@ -0,0 +1,11 @@
...
+    padding-left: 8px;
+    vertical-align: middle;
```
antonymilne, line 6: Somehow the icon feels very slightly too low to me compared to the title in all the cases in the example app, even though I _think_ it's correctly aligning it (?). wdyt @huong-li-nguyen?

gtauzin:
Hi @antonymilne and thank you so much for this, it would be super useful!

> Maybe in future, but don't do for now: implement `description` on Button (this is something @gtauzin mentioned but not sure exactly what it would look like given there's no `title` on this component)

This is true, I did not think of it :/ I guess what would make sense for my actual use case is a [Popover](https://dash-bootstrap-components.opensource.faculty.ai/docs/components/popover/).

antonymilne:
@gtauzin, I think a tooltip would still work rather than a popover (they're very similar really), it's just a question of what triggers opening it. I guess it could work when you hover over the button, like this:
image

I'm not sure how much I like it tbh, curious what others think. If we did want to enable it then we would add a new argument `target` to `vm.Tooltip` which would enable it to operate in "standalone mode". In this case you wouldn't set an icon and `build` would instead just return the `vm.Tooltip`, which could be attached to an arbitrary `target` (in the case of a `vm.Button`, the target would just be the button).

maxschulz-COL:
> Also any idea why pycafe didn't run here? It was originally a draft PR, like the one we saw before that didn't work, although other draft PRs seem to work ok 🤔

I think the reason is that the PyCafe fails at the moment! So it never got around to posting any link.

```bash
File "/lib/python3.12/site-packages/vizro/models/_components/form/dropdown.py", line 61, in <module>
    class Dropdown(VizroBaseModel):

  File "/lib/python3.12/site-packages/vizro/models/_components/form/dropdown.py", line 100, in Dropdown
    BeforeValidator(coerce_str_to_tooltip, json_schema_input_type=Union[str, Tooltip]),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

TypeError: BeforeValidator.__init__() got an unexpected keyword argument 'json_schema_input_type'
```

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -93,6 +95,11 @@ class Dropdown(VizroBaseModel):
         Field(default=True, description="Whether to allow selection of multiple values", validate_default=True),
     ]
     title: str = Field(default="", description="Title to be displayed")
+    description: Annotated[
+        Optional[Tooltip],
+        BeforeValidator(coerce_str_to_tooltip, json_schema_input_type=Union[str, Tooltip]),
```
maxschulz-COL, line 100: After some checking, it turns out this feature `json_schema_input_type` is only available from `pydantic>=2.9.0`: https://github.com/pydantic/pydantic/releases/tag/v2.9.0

This should probably show up in our lower bounds tests as well - but it seems to also just be a straight out problem for PyCafe. I will reach out to them.

Before we decide on what to do, will not review further

maxschulz-COL: Here is the PR in question, which is oart of the 2.9 release: https://github.com/pydantic/pydantic/pull/10094

antonymilne: Ahh, this makes sense, thank you! Agreed that including `json_schema_input_type` is not so important here, so I'll remove it and put a TODO in to add it in future.

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -141,7 +151,7 @@ def __call__(self, options):
 
         return html.Div(
             children=[
-                dbc.Label(self.title, html_for=self.id) if self.title else None,
+                dbc.Label([self.title, *description], html_for=self.id) if self.title else None,
```
petar-qb, line 154: If we make the following change, would it be easier for users to move the icon with the tooltip to float right? 
```suggestion
                dbc.Label([self.title], html_for=self.id) if self.title else None,
                dbc.Label([*description], html_for=self.id) if self.title else None,
```

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
maxschulz-COL: Age is just a number

maxschulz-COL: Aside: when I click this on PyCafe it fails! Is that also the case locally?

antonymilne: Yes, that's just because I didn't put the `http://` in the link.

So the tooltip looks fine to your young eyes then?! I am wearing my spare glasses this week actually since I sat on top of my normal ones at the weekend, so maybe that's why...

vizro-core/src/vizro/static/css/tooltip.css
antonymilne: Oh, I thought 14px for the legend looked ridiculously small... Maybe this is just a thing with my eyes again 😅 Curious to know what other people think.

@nadijagraca @petar-qb @maxschulz-COL What do you think of Li's two screenshots above?

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
huong-li-nguyen: I think the font size is too small—we don't go below 14px, but it seems a lot smaller. I think something is off because even the font size for the tooltip for the collapse icon is now smaller. 🤔  In general, the tooltip text should be 14 px, like our body text. I think you might have changed that accidentally?

Screenshot 2025-04-03 at 15 51 38

But yes, contrast and font-color seem fine to me 👍

vizro-core/src/vizro/static/css/tooltip.css
petar-qb: The current icon size for controls/page/dashboard looks a bit large to me.

The sizes in the second screenshot (Dashboard and Page) seem perfect. The filter icons might be slightly small, but just maybe (disclaimer: I wear glasses too). I’m concerned that increasing them from 14px to 16px might make them look too big compared to the filter titles, but I haven’t tested it yet.

antonymilne: @petar-qb said he liked your version better so I'm happy to go for that with its comically small icons 😆 Then we can just do `font-size: inherit` which is simple anyway.

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
huong-li-nguyen: Never mind - the font-size for the tooltip is 12px according to QB design 😬 So just leave it as is.

In general, you would add your changes inside the bootstrap_overwrites.css file. I check every now and then when to bring changes over to vizro-bootstrap.

vizro-core/src/vizro/static/css/tooltip.css
huong-li-nguyen: I'll push some changes and summarise in a separate comment. Closing this thread.

vizro-core/src/vizro/static/css/tooltip.css
huong-li-nguyen: I am so proud of your eyes 💯  Yes, they were actually not centered correctly, because they had different line heights. It's easier if you turn the titles into flex-boxes and use `align-items: center` there. Let me push it 👍

huong-li-nguyen:
> * @huong-li-nguyen: does the styling look ok to you? I have a couple of issues with it still, see what you think

I've just pushed some changes [here](https://github.com/mckinsey/vizro/pull/1092/commits/6c712cc8d31e975ad67800788c49f873c005b7a7): 

- I moved some of the code from `bootstrap_overwrites.css` to `tooltip.css`— placing it inside `bootstrap_overwrites.css` was definitely correct. I occasionally check if changes are generic enough to go into `vizro-bootstrap`. By "generic," I mean they should work as part of our overall Bootstrap theme. In this case, though, I think it’s more specific to Vizro, so it makes sense to keep it separate rather than applying it to all tooltips, hence should live inside `tooltip.css`.

- I talked with Steph, and turns out the QB design library has fixed icon sizes: 16px (small), 24px (regular), and 32px (large). So, we don’t actually need to set a font size for icons since 24px is already the default from the Material Icon library. We only need to define sizes if we’re using small or large icons. So I applied 16px to all form labels, so it should now cover all control titles.

- You were spot on that the icon was not center-aligned! Great catch—most people wouldn’t have even noticed that tiny detail! I turned the titles into flex-containers, where the items align on the center. This now fixes it.

vizro-core/src/vizro/models/_components/form/dropdown.py
huong-li-nguyen: I just pushed some changes where I turned the page-title, dashboard-title, and form-label into flexboxes. Not only does this fix the centering issue @antonymilne pointed out, but it also makes it super easy to position the icon on the right using some custom CSS, without needing any further changes here 👍 

You could just do:

```
.form-label {
}
```

And then you get this:

Screenshot 2025-04-04 at 14 53 38

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -88,7 +89,8 @@ class Dashboard(VizroBaseModel):
             Defaults to `vizro_dark`.
         navigation (Navigation): See [`Navigation`][vizro.models.Navigation]. Defaults to `None`.
         title (str): Dashboard title to appear on every page on top left-side. Defaults to `""`.
-
+        description (Optional[Tooltip]): Additional information about the dashboard shown in a tooltip and used for
```
huong-li-nguyen, line 92: Similar comment about explaining when and where the tooltip/info-icon appears

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -111,10 +124,13 @@ def pre_build(self):
...
 
         for order, page in enumerate(self.pages):
+            # Dash also uses the dashboard-level description passed into Dash() as the default for page-level
+            # descriptions, but this would involve extracting dashboard.description and inserting it into the Dash app
+            # config in Vizro.build. What we do here is simpler but has the same effect.
             dash.register_page(
                 module=page.id,
                 name=page.title,
-                description=page.description,
+                description=page.description.text if page.description is not None else self.description,
```
huong-li-nguyen, line 133: I'm still not sure how I feel about the `description` argument for pages automatically turning into an info-icon with a tooltip now. I guess it's just something I'll have to get used to 😆  For me, the description was always more about adding extra details for link previews when sharing the dashboard URL, not something that needed to be visible. But, since the text is hidden in the tooltip, it might not be so bad after all.

vizro-core/changelog.d/20250401_113146_antony.milne_info_icon.md
```diff
@@ -0,0 +1,48 @@
...
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
```
huong-li-nguyen, line 11: Were you also planning to add some additional docs? Given that the argument is called `description` I think we need lots of mentioning of the `tooltip` inside the docs, otherwise I am not sure if people would find this?

vizro-core/src/vizro/models/_tooltip.py
```diff
@@ -0,0 +1,73 @@
...
+        AfterValidator(validate_icon),
+        Field(description="Icon name from Google Material icons library."),
+    ]
+    extra: SkipJsonSchema[
+        Annotated[
+            dict[str, Any],
+            Field(
+                default={},
+                description="""Extra keyword arguments that are passed to `dbc.Tooltip` and overwrite any
+                defaults chosen by the Vizro team. This may have unexpected behavior. Visit the
```
huong-li-nguyen, line 51: Why may this have unexpected behavior? 🤔

huong-li-nguyen (approved):
This looks great! 🚀 I think I have to get used to the `description` argument, but I don't oppose it strongly.  

I think some docs might be a nice add-on, but can be done here or in a separate PR  👍

vizro-core/changelog.d/20250401_113146_antony.milne_info_icon.md
maxschulz-COL: Agreed, tbh it should have a tiny section on every how-to guide that describes a model usage that now has description

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -71,6 +72,7 @@ class Dropdown(VizroBaseModel):
             [`MultiValueType`][vizro.models.types.MultiValueType]. Defaults to `None`.
         multi (bool): Whether to allow selection of multiple values. Defaults to `True`.
         title (str): Title to be displayed. Defaults to `""`.
+        description (Optional[Tooltip]): Additional information about the selector shown in a tooltip.
```
maxschulz-COL, line 75:
```suggestion
        description (Optional[Tooltip]): Additional information about the selector shown in a tooltip. Defaults to `None`.
```

maxschulz-COL: And similar on the other models

vizro-core/src/vizro/models/_tooltip.py
```diff
@@ -0,0 +1,73 @@
...
+    """A tooltip that displays text when hovering over an icon.
+
+    Args:
+        text (str): Markdown string for text shown when hovering over the icon. Should adhere to the CommonMark Spec.
+        icon (str): Icon name from [Google Material icons library](https://fonts.google.com/icons).
+        extra (Optional[dict[str, Any]]): Extra keyword arguments that are passed to `dbc.Tooltip` and overwrite any
+            chosen by the Vizro team. This may have unexpected behavior. Visit the
+           [dbc documentation](https://dash-bootstrap-components.opensource.faculty.ai/docs/components/tooltip/)
+           to see all available arguments. [Not part of the official Vizro schema](../explanation/schema.md) and the
+           underlying component may change in the future. Defaults to `{}`.
```
maxschulz-COL, line 33:
```suggestion
        extra (Optional[dict[str, Any]]): Extra keyword arguments that are passed to `dbc.Tooltip` and overwrite any
            defaults chosen by the Vizro team. This may have unexpected behavior. Visit the
           [dbc documentation](https://dash-bootstrap-components.opensource.faculty.ai/docs/components/tooltip/)
           to see all available arguments. [Not part of the official Vizro schema](../explanation/schema.md) and the
           underlying component may change in the future. Defaults to `{}`.
```

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -20,6 +20,10 @@ def _wrapper(self, *args, **kwargs):
     return _wrapper
 
 
+def validate_icon(icon) -> str:
```
maxschulz-COL, line 23:
```suggestion
def validate_icon(icon: str) -> str:
```

vizro-core/src/vizro/models/_dashboard.py
maxschulz-COL: So that means you cannot separate the links preview description from the tooltip? Hm, I am not so sure I like this either... Annoying that the names clash here. Would it be crazy to use a different field name here? Probably yes right?

maxschulz-COL (approved):
LGTM!

vizro-core/src/vizro/models/_dashboard.py
maxschulz-COL: Yes I think that could be interesting. But you are right, if we slightly want to change the description, it starts being annoying..

vizro-core/src/vizro/static/css/tooltip.css
antonymilne: Ah, perfect! I was going to flex box them exactly like you say but wasn't sure if it was just my eyes or not again 😅

vizro-core/src/vizro/static/css/tooltip.css
```diff
@@ -1,11 +1,31 @@
...
+  color: var(--text-primaryHover-inverted);
+}
+
+/* We differentiate between 3 different icon sizes: 16px, 24px, 32px. 24px is our default.
+Use 16px for all text. Use 24px for all headings. */
+.form-label .tooltip-icon {
+    font-size: 16px;
+}
+
+#dashboard-title, #page-title, .form-label {
```
antonymilne, line 31: @huong-li-nguyen does this definitely belong in `tooltip.css`? It's not exactly to do with tooltips, although that's why it matters I guess. I'm happy to put it wherever you think is best but just wanted to check it wasn't an accident.

vizro-core/src/vizro/models/_components/form/dropdown.py
antonymilne: Nice, this is a perfect solution I think 👍

vizro-core/src/vizro/models/_tooltip.py
antonymilne: This is the standard warning message we put on all `extra` fields so I would just leave it as is 🙂

vizro-core/src/vizro/static/css/tooltip.css
huong-li-nguyen: I don't really care where we put it :) So feel free to move it 👍

PR #1094, opened 2025-04-03
[Feat] Enable controls inside container

nadijagraca:
## Description
Enable `controls` argument inside `vm.Container`. 

`Checklist` and `RadioItems` are by default assigned `inline=True` via `extra` argument. 

Users are still able to have vertically aligned selector values by explicitly providing `extra={"inline": False}`

To do:

- [x] Fix selectors alignment 

## Screenshot
Screenshot 2025-05-21 at 14 15 27

## Notice


antonymilne:
Oooh I'm super excited for this to become a reality, it looks really cool already!!

Just in terms of design, I have a couple of ideas that are probably not as good as what you have done here but just want to check.

This is what you have done:
**Inline controls at top of container**
image

Here's the only two sensible alternatives that spring to mind - did you try them out and decide they didn't work?

**Normal (not inline) controls at top of container** - I guess this uses up too much vertical space?
image

**Normal (not inline) controls at side of container** - i guess this uses up too much horizontal space?
image

I think that for controls in containers we can enable all selector types but realistically dropdown will always be much more practical for many options instead of radioitems/checklist, and that's ok.

Whatever the design, possibly we need some way to visually separate the controls from the figures a bit more? e.g. a different background colour and/or a keyline. Probably the keyline is easier to avoid needing to consider 100 different permutations of container filled on different backgrounds etc. 😅 

Possibly we should check with Steph to see if she has any ideas too, wdyt @huong-li-nguyen? Let's not let lack of a perfect design significantly delay though, because what @nadijagraca has done here + maybe some slight visual separation is good enough for me anyway.

nadijagraca:
> Oooh I'm super excited for this to become a reality, it looks really cool already!!
> 
> Just in terms of design, I have a couple of ideas that are probably not as good as what you have done here but just want to check. The container is shown in grey background in all the below just to show its boundaries. It doesn't matter if it's filled or not, the layout would be the same.
> 
> **Inline controls at top of container** - this is what you have done: image
> 
> Here's the only two sensible alternatives that spring to mind - did you try them out and decide they didn't work?
> 
> **Normal (not inline) controls at top of container** - I guess this uses up too much vertical space? image
> 
Yes, this approach takes up too much vertical space in the container, especially if selectors like checklist or radioitems are used that have a lot of options. 
> **Normal (not inline) controls at side of container** - i guess this uses up too much horizontal space? image
> 
Controls at the side of the container take up too much horizontal space, and might visually clash with the control panel filters .
> I think that for controls in containers we can enable all selector types but realistically dropdown will always be much more practical for many options instead of radioitems/checklist, and that's ok.
> 
> Whatever the design, possibly we need some way to visually separate the controls from the figures a bit more? e.g. a different background colour and/or a keyline. Probably the keyline is easier to avoid needing to consider 100 different permutations of container filled on different backgrounds etc. 😅
> 
> Possibly we should check with Steph to see if she has any ideas too, wdyt @huong-li-nguyen? Let's not let lack of a perfect design significantly delay though, because what @nadijagraca has done here + maybe some slight visual separation is good enough for me anyway.

huong-li-nguyen (commented):
I'll discuss this with Steph next week, but for now, here's what I suggest implementing:

- **Keyline for Visual Distinction**: The keyline you currently have works well as a visual separator, so let's keep it as is for now.  
- **Default Placement**: Place the filter section at the top by default. However, we should start considering how it could be configured if placed on the side. It really depends on the use case if there is enough space, but I wouldn't disregard this option.
- **Title and Selector Layout**: For now, keep the title and selector on separate lines.  
- **Dropdowns and Radio Items on One Line**: Test how it looks when dropdowns and radio items are placed on the same line. Do we need to adjust the height or make other layout changes?  
- **Support for inline=False**: We should ensure the layout supports the case where inline=False, even when the filter section is placed at the top. While this doesn't need to be the default configuration, users should have the option to enable it. I also don't think it takes too much space (e.g. radio-items with just two options - could either be inline or not and both would look nice)

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -7,9 +7,7 @@ A [Container][vizro.models.Container] complements a [Page][vizro.models.Page], a
 Unlike `Page`, the `Container` model offers additional visual customization options:
 
 - **`variant`**: Allows you to select a visual style for the container, making it stand out from the rest of the page content.
```
stichbury, line 9:
```suggestion
- **`variant`**: Enables you to select a visual style for the container, making it stand out from the rest of the page content.
```

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -7,9 +7,7 @@ A [Container][vizro.models.Container] complements a [Page][vizro.models.Page], a
...
-- **`collapsed`**: Enables collapsible behavior, letting you define whether the container's contents are initially shown or hidden.
-
-More functionality will be introduced to `Container` soon, including container-specific controls to support better grouping and interaction of related components.
+- **`collapsed`**: Enables collapsible behavior, letting you define whether the container's contents are initially shown or hidden
```
stichbury, line 10:
```suggestion
- **`collapsed`**: Enables collapsible behavior. You can define whether the container's contents are initially shown or hidden.
```

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -497,6 +495,91 @@ You can provide markdown text as a string to use the default info icon or a [`To
 
         [![ContainerInfoIcon]][containerinfoicon]
 
+## Add controls to container
+
+The `Container` now accept `controls` argument, where you can define container-specific controls for better grouping and interaction of related components.
```
stichbury, line 500:
```suggestion
The `Container` now accepts a `controls` argument, where you can define container-specific controls to group related components for better interaction.
```

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -497,6 +495,91 @@ You can provide markdown text as a string to use the default info icon or a [`To
...
+
+By default, a control affects only components within its container when no targets are specified. To affect the components outside the container, use the `targets` argument of [Filter][vizro.models.Filter] or [Parameter][vizro.models.Parameter] to explicitly define them.
```
stichbury, line 502:
```suggestion
By default, a control affects only components within its container when no targets are specified. To affect components outside the container, use the `targets` argument of [Filter][vizro.models.Filter] or [Parameter][vizro.models.Parameter] to explicitly define them.
```

stichbury (approved):
LGTM for docs!

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -216,6 +213,13 @@ def pre_build(self):
             ),
         ]
 
+        # set default inline=True for container selectors
+        _is_page_control = self._is_page_control(page=page)
+
+        if not _is_page_control and isinstance(self.selector, (Checklist, RadioItems)):
+            self.selector.extra = self.selector.extra or {}
+            self.selector.extra.setdefault("inline", True)
```
huong-li-nguyen, line 221: Can you add any logic we add to the PR description as well? This basically means we default to inline inside the Container right for these Selectors?

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -216,3 +216,18 @@ Current logic:
...
+}
+
+.container-controls-panel {
+  border-bottom: 1px solid var(--border-subtleAlpha01);
+  display: flex;
+  flex-wrap: wrap;
+  gap: 12px;
+  justify-content: start;
+  margin-bottom: 0.5rem;
+  padding-bottom: 0.5rem;
```
huong-li-nguyen, line 232: Do we need both margin-bottom and padding-bottom?

huong-li-nguyen (changes requested):
Hi @nadijagraca,

This is looking great — really excited to see it coming together! 🚀 

I had a few thoughts and questions. If we're aiming to make inline the default layout for controls inside Container, we might need to refine the visuals a bit more. Also, for the docs, I think we could add clearer examples — ideally something that shows when using controls inside Container is necessary (i.e., when Page.controls alone isn't sufficient). That might help users better understand the use case.

A few visual notes I noticed:

**Bottom border** – I’d suggest we keep it by default, but hide it when Tabs come directly after. That should hopefully be doable in CSS, but let me know if it’s tricky!

Screenshot 2025-05-12 at 12 03 00

**Dropdown sizing** – The dropdowns feel a bit too narrow right now. Could we add a minimum width to improve readability?
Screenshot 2025-05-12 at 12 03 33

**Alignment across control types** – This one’s probably the trickiest. Ideally, all control types (dropdowns, sliders, etc.) should align nicely. Currently, slider titles seem slightly misaligned, and the options aren’t centered either. Might require some experimentation to get it feeling good.
Screenshot 2025-05-12 at 12 03 42

huong-li-nguyen:
> ## Description
> POC for controls inside containers
> 

Could you also update your PR description? :)

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -69,7 +69,7 @@ def __iter__(self) -> Generator[ModelID, None, None]:
     def _get_models(
         self,
         model_type: Optional[Union[type[Model], tuple[type[Model], ...], type[FIGURE_MODELS]]] = None,
-        page: Optional[Page] = None,
+        page: Optional[Union[Page, VizroBaseModel]] = None,
```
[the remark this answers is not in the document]
petar-qb: I completely agree with @huong-li-nguyen. The current `page` parameter is actually any Vizro model, but it has been used specifically for Vizro pages until now, which explains the original naming and type. Its type should be updated to `Optional[VizroBaseModel]`.

Regarding the new parameter name, I suggest `root_model`, as it clearly describes its purpose.

@nadijagraca Once you update the name, please ensure it is consistently changed across all occurrences (there are over 30, unfortunately 😞).

vizro-core/src/vizro/models/_controls/parameter.py
```diff
@@ -82,6 +82,7 @@ def pre_build(self):
         self._check_categorical_selectors_options()
         self._set_selector_title()
         self._set_actions()
+        self._set_container_control_default()
```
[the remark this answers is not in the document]
petar-qb: Also, let's write a simple unit test for that inside the `_test_parameter.py::TestPreBuildMethod`.

vizro-core/src/vizro/models/_controls/filter.py
petar-qb, line 1: Let's add some unit tests within `test_filter.py::TestPreBuildMethod`.

vizro-core/examples/scratch_dev/app.py
```diff
@@ -5,42 +5,360 @@
...
-    layout=vm.Flex(gap="40px"),
+    title="Controls in the tabs",
+    components=[
+        vm.Container(
+            title="Container with tabs and controls",
+            controls=[
+                vm.Filter(
+                    column="species", selector=vm.Checklist(title="Species", extra={"inline": True}), targets=["graph6"]
+                )
+            ],
```
petar-qb, line 110: Is it intentional that this control is not applied to the `graph5` within the `tab-3`?

vizro-core/examples/scratch_dev/app.py
petar-qb, line 1: Have you checked that the dynamic filters works well too? If not, let’s add a new page within the `scratch/app.py` and optionally a new test for that.

vizro-core/examples/scratch_dev/app.py
```diff
@@ -5,42 +5,360 @@
...
+                vm.Graph(
+                    title="Scatter chart",
+                    figure=px.scatter(
+                        iris, x="sepal_length", y="petal_width", color="species", custom_data=["species"]
+                    ),
+                ),
+            ],
+            layout=vm.Layout(grid=[[0, 1]]),
+            controls=[
+                vm.Filter(column="species", selector=vm.Dropdown()),
```
petar-qb, line 248: Maybe we should pin the dropdown width. That would ensure that other container's controls don’t move every time the dropdown filer value is selected. Should we then fix its height to maximum 3 rows or 1 row?? IDK, maybe one row.

petar-qb (changes requested):
Awesome job @nadijagraca!!! 🚀 🌟  I'm so glad to see this huge feature around the corner. In my opinion it's a good candidate for the "⭐ Highlight" announcing in the changelog section.

There are several comments below, but there's nothing major.

P.S. I'll get back to the `Filter.pre_build` again when you address Li's comments there. Currently, Li's suggestion looks good, but I just want to double check the final version of it.

vizro-core/examples/scratch_dev/app.py
nadijagraca: Yes, it is intentional. I was trying to test if different combinations of container/tabs nesting work well with setting targets. :)

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -182,3 +183,16 @@ def _build_container_title(self):
...
+        return (
+            html.Div(
+                id=f"{self.id}-control-panel",
+                children=controls_content,
+                hidden=not controls_content,
+                className="container-controls-panel",
+            )
+            if self.controls
+            else html.Div(hidden=True)
+        )
```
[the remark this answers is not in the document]
nadijagraca: I don't have a strong preference. I decided on the `html.Div(hidden=True)` just to be more flexible and preserve the layout structure, but I'm okay with using `None` as well.

vizro-core/examples/scratch_dev/app.py
petar-qb: Does it mean that container.controls will automatically be applied only to the figures that are direct child of the container? And, that for the nested children, these controls won't be applied?

petar-qb: Ahh sorry my bad 😆 Can I delete this conversation from the internet?? 😆

vizro-core/src/vizro/models/_components/container.py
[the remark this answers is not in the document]
huong-li-nguyen: We have some guidelines on this in the wiki: https://github.com/mckinsey/vizro/wiki/Developer-docs#htmldivhiddentrue-vs-none :)

In this case, I’d also return `None`, even though we technically provide an `id`. However, the `id` isn’t present when no controls are passed and we don't seem to try to access the `id` elsewhere, so to me it still fits under the `None` case 👍

I’d suggest moving the `if/else` logic up one level though —so instead of handling it inside the method, you call it like this:

```python
self._build_control_panel() if self.controls else None
```

This makes it clearer that the control panel is only added to `Container.children` when controls exist, rather than hiding that logic inside the method :)

huong-li-nguyen (commented):
**Just looking at the visuals:**

1. The Tabs are currently to close to the controls I think: 
Screenshot 2025-05-19 at 18 38 51

2. Further refine alignment if possible. I don't know how difficult it is to do these so if it's too hard, then leave it out for now, but optimally this should be fixed:
- RangeSlider and Slider should have the same width. It looks a bit off if you use both, and one is shorter than the other. Maybe use the same width as you do on the Dropdown?
- I think the RadioItem and Checklist items are currently not perfectly centered. Could we center them to the middle of the Dropdown for example? We might have to set a height on these, such that we can align them :) 
Screenshot 2025-05-19 at 18 44 28

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -497,6 +496,91 @@ You can provide markdown text as a string to use the default info icon or a [`To
...
+
+The `Container` now accepts a `controls` argument, where you can define container-specific controls to group related components for better interaction.
+
+By default, controls only affect components inside their own container.
+
+!!! example "Container with controls"
+
+    === "app.py"
+
+        ```{.python pycafe-link hl_lines="14-18"}
```
huong-li-nguyen, line 509: This example makes a lot more sense 👍 🚀  Wondering if we should explicitly rename them "Container with gapminder data" and "Container with iris data". WDYT?

vizro-core/src/vizro/models/_controls/_controls_utils.py
```diff
@@ -19,3 +19,11 @@ def check_targets_present_on_page(control: ControlType) -> None:
         target_id = target.split(".")[0] if "." in target else target
         if target_id not in page_figure_ids:
             raise ValueError(f"Target {target_id} not found within the page {control_page.id}.")
+
+
+def set_container_control_default(control, control_id, selector):
```
huong-li-nguyen, line 24: Add typing?

vizro-core/src/vizro/models/_controls/_controls_utils.py
```diff
@@ -19,3 +19,11 @@ def check_targets_present_on_page(control: ControlType) -> None:
...
+    page = model_manager._get_model_page(control)
+    is_page_control = any(control.id == control_id for control in page.controls)
+
+    if not is_page_control and isinstance(selector, (Checklist, RadioItems)):
+        return selector.extra.setdefault("inline", True)
```
huong-li-nguyen, line 29:
```suggestion
    control_is_on_page = any(control.id == control_id for control in page.controls)

    if not control_is_on_page and isinstance(selector, (Checklist, RadioItems)):
        return selector.extra.setdefault("inline", True)
    return None
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -335,3 +335,28 @@ def _get_options(
         # changes. See https://pandas.pydata.org/docs/whatsnew/v2.1.0.html#whatsnew-210-enhancements-new-stack.
         targeted_data = pd.concat([targeted_data, pd.Series(current_value)]).stack().dropna()  # noqa: PD013
         return sorted(set(targeted_data) - {ALL_OPTION})
+
+    def _get_proposed_targets(self, page):
```
huong-li-nguyen, line 339: Add typing

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -335,3 +335,28 @@ def _get_options(
...
+            return self.targets
+
+        parent_container = self._find_parent_container(page)
+        target_scope = parent_container if parent_container else page
+
+        return [
+            model.id for model in cast(Iterable[VizroBaseModel], model_manager._get_models(FIGURE_MODELS, target_scope))
+        ]
+
+    def _find_parent_container(self, page):
```
huong-li-nguyen, line 351: Add typing

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -335,3 +335,28 @@ def _get_options(
...
+        """Find the container that contains this control, if any."""
+        page_containers = cast(Iterable[Container], model_manager._get_models(Container, page))
```
huong-li-nguyen, line 353: Do we actually need to cast?
```suggestion
        page_containers =model_manager._get_models(Container, page)
```

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
```diff
@@ -86,3 +86,7 @@ label[for="theme-selector"] {
   margin-right: 0;
   padding-left: 0;
 }
+
+.dropdown {
```
huong-li-nguyen, line 90: I am sure it will look fine, but just double-check how it looks like on mobile 👍

vizro-core/tests/unit/vizro/models/_controls/conftest.py
```diff
@@ -17,3 +17,22 @@ def managers_one_page_two_graphs(gapminder):
...
+    """Instantiates a simple model_manager and data_manager with a page, and two graph models and gapminder data."""
+    vm.Page(
+        id="test_container",
+        title="My first dashboard",
+        components=[
+            vm.Container(
+                title="",
+                components=[
+                    vm.Graph(id="scatter_chart", figure=px.scatter(gapminder, x="lifeExp", y="gdpPercap")),
+                ],
```
petar-qb, line 33: What do you think about this approach:
```suggestion
                components=[
                    vm.Graph(id="scatter_chart", figure=px.scatter(gapminder, x="lifeExp", y="gdpPercap")),
                ],
                controls=[
                    vm.Filter(id="container_filter", column="continent", selector=vm.Checklist(value=["Europe"])),
                    vm.Parameter(
                        id="container_parameter",
                        targets=["scatter_chart.x"],
                        selector=vm.Checklist(options=["lifeExp", "gdpPercap", "pop"], value=["lifeExp"]),
                    )
                ]
```

In tests you can just `filter=model_manager["container_filter"]` and `assert`. Similar for Parameter.

Also, you can remove `Vizro._pre_build()` from here and then explicitly call it from the tests.

petar-qb (commented):
Amazing job @nadijagraca!! This looks much better than the last time I reviewed it and I see this PR merged by the end of the week (and then release the next week) 🎉 🎉

vizro-core/examples/scratch_dev/app.py
```diff
@@ -22,29 +24,345 @@
...
+                    ]
+                )
+            ],
+        )
+    ],
+)
+page3 = vm.Page(
+    title="Controls in the tabs",
+    components=[
+        vm.Container(
```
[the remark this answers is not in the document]
nadijagraca: Some of the pages, this one included, do not look good. They are there to represent how the page would look like with tab/container nesting without custom styling. But I agree, removing the title does look much cleaner and better.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -335,3 +335,28 @@ def _get_options(
...
+
+        return next(
+            (
+                container
+                for container in page_containers
+                if any(control.id == self.id for control in container.controls)
+            ),
+            None,
+        )
```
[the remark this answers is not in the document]
nadijagraca: I like it, lets do it this way.

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
nadijagraca: It look fine on mobile as well. 

Screenshot 2025-05-20 at 14 29 32

vizro-core/src/vizro/actions/_filter_interaction.py
```diff
@@ -51,7 +51,8 @@ def pre_build(self):
         figure_ids_on_page = [
             model.id
             for model in cast(
-                Iterable[FigureType], model_manager._get_models(FIGURE_MODELS, page=model_manager._get_model_page(self))
+                Iterable[FigureType],
```
antonymilne, line 54: Rough comment to myself: look at action TODOs and figure out how to make this work inside container.

vizro-core/src/vizro/models/_controls/parameter.py
```diff
@@ -78,6 +78,7 @@ class Parameter(VizroBaseModel):
     @_log_call
     def pre_build(self):
         check_targets_present_on_page(control=self)
```
antonymilne, line 80: Just to understand: we don't check anywhere the `targets` are inside the container? Is this something we want to do? Similarly for `Filter`.

if we do want to do this (would be a separate PR) then possibly the logic for it should go in `Page/Container` rather than `controls`. In fact maybe eventually `_get_proposed_targets` should live there too.

Let me think about it...

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -135,13 +135,10 @@ def pre_build(self):
...
-        # Possibly in future this will change (which would be breaking change).
-        proposed_targets = self.targets or [
-            model.id
-            for model in cast(
-                Iterable[FigureType], model_manager._get_models(FIGURE_MODELS, model_manager._get_model_page(self))
-            )
-        ]
+        # If filter used within container and if targets aren't explicitly provided it will target all figures within
+        # that container. Possibly in future this will change (which would be breaking change).
+        proposed_targets = self._get_proposed_targets()
```
antonymilne, line 140:
```suggestion
        proposed_targets = self.targets or self._get_proposed_targets()
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -335,3 +335,25 @@ def _get_options(
         # changes. See https://pandas.pydata.org/docs/whatsnew/v2.1.0.html#whatsnew-210-enhancements-new-stack.
         targeted_data = pd.concat([targeted_data, pd.Series(current_value)]).stack().dropna()  # noqa: PD013
         return sorted(set(targeted_data) - {ALL_OPTION})
+
+    def _get_proposed_targets(self):
+        """Get all valid figure model targets for this control based on its location in the page hierarchy."""
+        if self.targets:
+            return self.targets
+
```
antonymilne, line 343:
```suggestion
```

vizro-core/src/vizro/models/_controls/filter.py
antonymilne: Together with below suggestion or removing the early `return` from `_get_proposed_targets`.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -335,3 +335,25 @@ def _get_options(
...
+        page = model_manager._get_model_page(self)
+        page_containers = model_manager._get_models(model_type=Container, root_model=page)
+
+        # Find the control's parent model. Set it as the control's parent container it exists.
```
antonymilne, line 347: I think we might be able to simplify this logic a bit if we move it into `Page`/`Container`, but that also might make a horrible mess. Let me try it out and see.

vizro-core/changelog.d/20250509_121204_nadija_ratkusic_graca_controls_inside_container.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Enable `controls` inside [Container][vizro.models.Container]. See the [user guide on container](https://vizro.readthedocs.io/en/stable/pages/user-guides/container/#add-controls-to-container) for more details.([#1094](https://github.com/mckinsey/vizro/pull/1094))
```
antonymilne, line 22:
```suggestion
- Enable controls inside containers. See the [user guide on container](https://vizro.readthedocs.io/en/stable/pages/user-guides/container/#add-controls-to-container) for more details.([#1094](https://github.com/mckinsey/vizro/pull/1094))
```

Definitely I'd make this a highlight.

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -497,6 +496,95 @@ You can provide markdown text as a string to use the default info icon or a [`To
 
         [![ContainerInfoIcon]][containerinfoicon]
 
+## Add controls to container
+
+The `Container` now accepts a `controls` argument, where you can define container-specific controls to group related components for better interaction. By default, controls only affect components inside their own container.
```
antonymilne, line 501:
```suggestion
The `Container` has a `controls` argument, where you can define container-specific controls to group related components for better interaction. By default, controls only affect components inside their own container.
```

vizro-core/docs/pages/user-guides/container.md
```diff
@@ -497,6 +496,95 @@ You can provide markdown text as a string to use the default info icon or a [`To
...
+
+Here are a few cases where you might want to use `controls` within `Container`:
```
antonymilne, line 503:
```suggestion
Here are a few cases where you might want to use `controls` within a `Container`:
```

vizro-core/src/vizro/models/_controls/_controls_utils.py
```diff
@@ -19,3 +19,11 @@ def check_targets_present_on_page(control: ControlType) -> None:
         target_id = target.split(".")[0] if "." in target else target
         if target_id not in page_figure_ids:
             raise ValueError(f"Target {target_id} not found within the page {control_page.id}.")
+
+
+def set_container_control_default(control: ControlType) -> None:
```
antonymilne, line 24: Potentially this should move to `Checklist/RadioItems` or to `Page/Container` 🤔

antonymilne (commented):
This is looking awesome, thank you for all your work on it @nadijagraca and the excellent reviewing @huong-li-nguyen @petar-qb 🙌 

Just a few comments for now - most of them you don't have to do anything about, they're just notes to myself. Here's my plan:
* there's a couple of small changes I think you could make now
* when you've pushed final changes, I'll do a final review and ✅ 
* I'll have a think about a few possible refactorings but save them for a different PR

vizro-core/src/vizro/models/_controls/parameter.py
nadijagraca: No, we don’t currently validate whether the provided targets are within the container—for either filters or parameters. If no targets are specified for a filter, we default to using all components within the container. However, users are allowed to specify targets that reside outside the container. Should we restrict users to only specify targets that are within the container? What do you think @petar-qb 

I discussed this briefly with Joe, and he mentioned that in future this can be resolved with synchronised filters.

maxschulz-COL: Just saw this while skimming - I feel our "opinionated" approach should forbid using targets outside containers?

vizro-core/src/vizro/models/_controls/_controls_utils.py
petar-qb: From top of my head, moving this logic into the `Checklist/RadioItems` doesn't make much sense for me. Components like Checklist or RadioItems shouldn't care about the Filter specific logic. They are able to render in the `inline` layout, but they shouldn't care about the internal logic when this happens or not. That kind of logic should live within the Filter or Page/Container in my opinion.

vizro-core/src/vizro/models/_controls/parameter.py
petar-qb: I'm not convinced we should introduce this restriction just because we don't currently see any strong use cases for that behaviour. By adding this limitation, we're effectively introducing a new rule that comes with long-term maintenance and communication overhead. It also opens up several questions we’d need to address, such as:
- Should the same restriction apply to the upcoming filter_interaction? If so, how would drill-through scenarios work?
- Should we also prevent Button.action.inputs and outputs from referencing components outside the Button’s container?
- What if there are valid use cases we haven’t yet considered?

Ultimately, if a user explicitly targets a component outside the container, it likely means they do have a reason to do so. Why to make an effort to prevent users from implementing the dashboard behaviour they want (especially if it’s not causing any issues)?

Unless we have a concrete problem this change solves, I would rather lean towards flexibility (with no effort) than restrictions (that requires described effort).

vizro-core/tests/unit/vizro/models/_controls/conftest.py
```diff
@@ -17,3 +17,30 @@ def managers_one_page_two_graphs(gapminder):
...
+                        id="container_parameter",
+                        targets=["scatter_chart.x"],
+                        selector=vm.Checklist(options=["lifeExp", "gdpPercap", "pop"], value=["lifeExp"]),
+                    ),
+                ],
+            ),
+            vm.Graph(id="bar_chart", figure=px.bar(gapminder, x="country", y="gdpPercap")),
+        ],
+    )
+    # Vizro._pre_build()
```
petar-qb, line 46:
```suggestion

```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -135,13 +135,10 @@ def pre_build(self):
...
         # This is the case when bool(self.targets) is False.
-        # Possibly in future this will change (which would be breaking change).
-        proposed_targets = self.targets or [
-            model.id
-            for model in cast(
-                Iterable[FigureType], model_manager._get_models(FIGURE_MODELS, model_manager._get_model_page(self))
-            )
-        ]
+        # If filter used within container and if targets aren't explicitly provided it will target all figures within
+        # that container. Possibly in future this will change (which would be breaking change).
```
petar-qb, line 139: What could lead here to the breaking change? 
```suggestion
        # If filter used within container and if targets aren't explicitly provided it will target all figures within
        # that container.
```

petar-qb (approved):
In my opinion, after addressing the remaining comments, this looks good to go 🎉

vizro-core/src/vizro/models/_controls/parameter.py
antonymilne: Hmmm, I'm not sure about this. My instinct is that we should forbid it, and if we're not sure whether allowing it is a good idea or not then it's actually safest to forbid it from the start (since relaxing the restriction in future is non-breaking, while introducing a new restriction could be interpreted as breaking). So if we ban it to begin with and then discover a valid use case for it we can always loosen the restriction.

If we introduce the restriction then it would be purely for `controls` affecting `components`, not `components` affecting a `control` (otherwise as you say drill-through wouldn't make sense). So it wouldn't affect `Button` either unless that was somehow in `controls` (which isn't possible currently without `add_type` anyway).

A separate question we should think about is whether other/all actions should be container-aware, e.g. if you do `export_data` inside a `vm.Container` then by default should it just download data from that container? I think probably yes, it should. Then should we forbid it targeting data outside the container? Probably not in this case? 🤔 

So I think _probably_ we should forbid controls affecting components outside the container it but let's all discuss after standup 🙂

vizro-core/examples/scratch_dev/app.py
antonymilne: Agree with this 👍 Even without extra controls next to a dropdown, it's a bit weird that as you add option it just expands indefinitely.

Can we make it behave the same way it does in the side control panel? 

If this is hard to do then no worries though, it would just be a nice tweak to have.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -335,3 +335,22 @@ def _get_options(
...
+        root_model = next(
+            (
+                container
+                for container in page_containers
+                if any(control.id == self.id for control in container.controls)
+            ),
+            page,
+        )
+        return [
+            model.id for model in cast(Iterable[VizroBaseModel], model_manager._get_models(FIGURE_MODELS, root_model))
```
antonymilne, line 355:
```suggestion
            model.id for model in cast(Iterable[FigureType], model_manager._get_models(FIGURE_MODELS, root_model))
```

Assuming mypy doesn't object.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -335,3 +335,22 @@ def _get_options(
...
+        page = model_manager._get_model_page(self)
+        page_containers = model_manager._get_models(model_type=Container, root_model=page)
+
+        # Find the control's parent model. Set it as the control's parent container it exists.
+        # Otherwise set it as the control's page.
+        root_model = next(
+            (
+                container
+                for container in page_containers
+                if any(control.id == self.id for control in container.controls)
```
antonymilne, line 350:
```suggestion
                if self in container.controls
```

Probably works?

vizro-core/tests/unit/vizro/models/_components/test_container.py
```diff
@@ -203,3 +205,24 @@ def test_container_build_with_description(self):
                 id="container_title",
             ),
         )
+
+    def test_container_build_with_controls(self):
+        result = vm.Container(
+            id="container", components=[vm.Button()], controls=[vm.Filter(column="species", selector=vm.Dropdown())]
+        ).build()
+        assert_component_equal(
+            result, dbc.Container(id="container", class_name="", fluid=True), keys_to_strip={"children"}
```
antonymilne, line 228:
```suggestion
            result, dbc.Container(id="container", fluid=True), keys_to_strip={"children"}
```
Maybe?

vizro-core/tests/unit/vizro/models/_components/test_container.py
```diff
@@ -217,3 +219,24 @@ def test_container_build_with_description(self):
...
+            result.children[1],
+            html.Div(
+                id="container-control-panel",
+                children=[],
+                className="container-controls-panel",
+            ),
+            keys_to_strip={"children"},
+        )
+        # Test if correct selector is added
+        assert isinstance(result.children[1].children[0].children[1], dcc.Dropdown)
```
antonymilne, line 242: These tests are nicer if we can minimise reference to `children[1]` etc. and instead use `id` to refer to components. See the changes I made here for an example: https://github.com/mckinsey/vizro/pull/1169/commits/f10e91bd5152b7f67773fa96af7ffca4c0d3c661

So this would become something like this:
```suggestion
        assert_component_equal(
            result["container-control-panel"],
            html.Div(
                id="container-control-panel",
                className="container-controls-panel",
            ),
            keys_to_strip={"children"},
        )
        assert_component_equal(result["dropdown-id"], dcc.Dropdown(), keys_to_strip=STRIP_ALL)
```

where you'll need to specify `vm.Dropdown(id="dropdown_id")`

antonymilne:
Writing it here so we can look back on it in future:
* filter behaviour of `vm.Container.controls` mirrors that of `vm.Page.controls`:
  * by default, target everything inside that container (or nested subcontainer etc.)
  * if a figure doesn't have the relevant column, just ignore it
* maybe in future this scheme changes to one of the following:
  * more restrictive: all figures in the (sub)container must contain the relevant column
  * less restrictive: you can target figures in different containers (but probably not cross-page)

vizro-core/src/vizro/models/_controls/parameter.py
antonymilne: Conclusion is here: https://github.com/mckinsey/vizro/pull/1094#issuecomment-2901694647.

vizro-core/src/vizro/models/_controls/_controls_utils.py
petar-qb, line 1: I played a bit with the new code you pushed in purpose to reduce the code duplication and here's what I have. 

I renamed `check_targets_present_on_page` to `check_control_targets` (please if you can remember any better name that would be awesome) and extracted most of it's content to the util `_get_control_parent`. There's now more code than it is in the original solution you implemented. However, with this proposed solution it's possible to reuse the newly added `_get_control_parent` from the `Filter._get_proposed_targets`. You can maybe even delete the `Filter._get_proposed_targets` as it can be replaced with the single line list comprehension. 

I've also added a TODO below. Something that might spark someone's thoughts when we find time to further improve thhe model_manager.

```py
# TODO: Consider rewriting the model_manager._get_model_page to model_manager._get_model_parent()
#  This would make the following renaming logical: model_manager._get_models -> model_manager._get_model_children.
#  These two new methods could have the same signature.
#  Consider adding the parent_model_id to the VizroBaseModel and use that to find the parent model more easily.
def _get_control_parent(control: ControlType) -> Optional[VizroBaseModel]:
    """Get the parent model of a control."""

    # Return None if the control is not part of any page.
    if (page := model_manager._get_model_page(model=control)) is None:
        return None

    # Return the Page if the control is its direct child.
    if control in page.controls:
        return page

    # Otherwise, return the Container that contains the control.
    page_containers = model_manager._get_models(model_type=Container, root_model=page)
    return next((container for container in page_containers if control in container.controls))


def check_control_targets(control: ControlType) -> None:
    if (root_model := _get_control_parent(control=control)) is None:
        raise ValueError(f"Control {control.id} should be defined within a Page object.")

    _validate_targets(targets=control.targets, root_model=root_model)
```

huong-li-nguyen (approved):
Ah, it's so satisfying to see all the selectors nicely aligned and equally sized now! 🚀 Amazing job @nadijagraca !!

Sorry to be a bit nitpicky here, but would it be possible to add the same spacing between the last item and the horizontal line as there is between the selector title and the first item? 😬 

Screenshot 2025-05-30 at 13 19 39

PR #1098, opened 2025-04-07
Deprecate vm.Layout in favour of vm.Grid

antonymilne:
## Notice

### DIFF
--- a/vizro-core/src/vizro/models/_grid.py
+++ b/vizro-core/src/vizro/models/_grid.py
@@ -5,6 +5,7 @@ import numpy as np
 from dash import html
 from numpy import ma
 from pydantic import AfterValidator, Field, PrivateAttr, ValidationInfo
+from typing_extensions import deprecated
 
 from vizro._constants import EMPTY_SPACE_CONST, GAP_DEFAULT
 from vizro.models import VizroBaseModel
@@ -30,10 +31,10 @@ def _get_unique_grid_component_ids(grid: list[list[int]]):
 
 # Validators for reuse
 def set_layout(layout, info: ValidationInfo):
-    from vizro.models import Flex, Layout
+    from vizro.models import Flex, Grid
 
     # No validation for Flex layout
-    # Possibly some of this validator should be attached directly to the grid Layout model type rather than
+    # Possibly some of this validator should be attached directly to the Grid model type rather than
     # the LayoutType Union to avoid needing this check.
     if isinstance(layout, Flex):
         return layout
@@ -46,7 +47,7 @@ def set_layout(layout, info: ValidationInfo):
 
     if layout is None:
         grid = [[i] for i in range(len(info.data["components"]))]
-        return Layout(grid=grid)
+        return Grid(grid=grid)
 
     unique_grid_idx = _get_unique_grid_component_ids(layout.grid)
     if len(unique_grid_idx) != len(info.data["components"]):
@@ -143,7 +144,7 @@ def _do_rectangles_overlap(r1: ColRowGridLines, r2: ColRowGridLines) -> bool:
 
 
 def _validate_grid_areas(grid_areas: list[ColRowGridLines]) -> None:
-    """Validates `grid_areas` spanned by screen components in `Layout`."""
+    """Validates `grid_areas` spanned by screen components in `Grid`."""
     for i, r1 in enumerate(grid_areas):
         for r2 in grid_areas[i + 1 :]:
             if _do_rectangles_overlap(r1, r2):
@@ -164,8 +165,7 @@ def _get_grid_lines(grid: list[list[int]]) -> tuple[list[ColRowGridLines], list[
     return component_grid_lines, space_grid_lines
 
 
-# TODO: Deprecate and rename to Grid
-class Layout(VizroBaseModel):
+class Grid(VizroBaseModel):
     """Grid specification to place chart/components on the [`Page`][vizro.models.Page].
 
     Args:
@@ -217,7 +217,7 @@ class Layout(VizroBaseModel):
     # We could have _LayoutBuildType as a return type annotation, but it would need to be generated
     # dynamically which is tricky and not amenable to type checking anyway.
     # Possibly in future we would have a public method to generate this string or maybe even
-    # a new method Layout.inject or similar that handles the injection of components into the grid for us.
+    # a new method Grid.inject or similar that handles the injection of components into the grid for us.
     # Another alternative is to take [component.build() for component in components] as an argument
     # in the build method here.
     @_log_call
@@ -248,3 +248,8 @@ class Layout(VizroBaseModel):
             id=self.id,
         )
         return component_container
+
+
+@deprecated("Blah balh", category=FutureWarning)
+class Layout(Grid):
+    type: Literal["legacy_layout"] = "legacy_layout"  # type: ignore[assignment]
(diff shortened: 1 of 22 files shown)


huong-li-nguyen (approved):
This looks great! Very clear and concise 🚀 

I did spot a few mentions of `vm.Layou`t in `vizro-ai` and the `goodread-tutorial.md`, but I’m guessing we’re not updating the `goodread-tutorial.md` since the article is already published? Or can we update the article in theory as well? @stichbury

For `vizro-ai`, we need to release our changes first, then update the vizro version before making any tweaks—right?

huong-li-nguyen:
🥳 

Screenshot 2025-04-08 at 08 59 40

Screenshot 2025-04-08 at 09 00 42

vizro-core/docs/pages/API-reference/deprecations.md
```diff
@@ -0,0 +1,7 @@
+# Deprecations and breaking changes
+
+This page lists Vizro features that are now deprecated and forthcoming breaking changes for Vizro 0.2.0.
+
+## `Layout` model
```
maxschulz-COL, line 5: Not super important just now, but I would prefer a tabular structure going forward. Depends how much there is going to be, but I picture this page becoming a bit of a log, where the sections are the breaking versions, and the table content has a clear and same structure like: what breaks, what is the replacement, comments, or smth similar

vizro-core/src/vizro/models/types.py
```diff
@@ -24,8 +25,18 @@ def _get_layout_discriminator(layout: Any) -> Optional[str]:
     # YAML/dictionary configuration in which `type` is not specified. This function is needed to handle the legacy case.
     if isinstance(layout, dict):
         # If type is supplied then use that (like saying discriminator="type"). Otherwise, it's the legacy case where
-        # type is not specified, in which case we want to use vm.Layout, which has type="grid".
-        return layout.get("type", "grid")
+        # type is not specified, in which case we want to use vm.Grid, which has type="grid", but emit a warning.
+        # We don't instantiate vm.Layout in this case since that would require it being added to LayoutType.
```
maxschulz-COL, line 29: Is it not part of LayoutType if I read below?

vizro-core/src/vizro/models/types.py
```diff
@@ -24,8 +25,18 @@ def _get_layout_discriminator(layout: Any) -> Optional[str]:
...
+        try:
+            return layout["type"]
+        except KeyError:
+            warnings.warn(
+                "`layout` without an explicit `type` specified will no longer work in Vizro 0.2.0. To ensure "
```
maxschulz-COL, line 34: I can imagine this was a fair bit of work... Now that we have it good, but I would have not even considered it all that breaking if we had just introduced the discriminated union :) I presume the `yaml` config is quite rare. Props for doing it though!

vizro-core/src/vizro/models/_grid.py
```diff
@@ -248,3 +248,14 @@ def build(self):
             id=self.id,
         )
         return component_container
+
+
+@deprecated(
```
maxschulz-COL, line 253: Is there any way to auto color these things. I find that especially in debug mode, it completely gets lost in all messages. Even without debug, it is not super visible...

maxschulz-COL (approved):
Yay I am happy! Good job :) Tested as well and got the messages, although I overlooked them initially 😂  see comment

maxschulz-COL:
As for `vizro-ai` I think we should definitely do this now, ideally in this PR. It is just that `vizro` as to be released before `vizro-ai`

vizro-core/src/vizro/models/types.py
antonymilne: Oh yes, good catch! I had to change how this worked a bit but didn't update the comment.

vizro-core/src/vizro/models/types.py
antonymilne: It was actually not too bad since I'd already heard of the callable discriminator so had the idea in mind for a few months. Now I've worked out the general pattern it should all be good in future too 🙂

vizro-core/src/vizro/models/_grid.py
antonymilne: Unfortunately no, at least not a simple way that also works nicely across Unix and Windows. We would need to use `rich` and turn warnings into logs, but that would be a bigger project and potentially cause other problems. Python warnings and tracebacks are just ugly and hard to spot I'm afraid.

antonymilne:
Thanks for the swift reviews both!

Great point about Vizro-AI, which I must admit I had completely forgotten about here... Agree we should fix that, but not in this PR and ideally not by me also 😅 As far as I can tell this PR won't immediately cause any Vizro-AI tests to fail. It looks the only tests that rely on local dev vizro (instead of PyPI released vizro) is "Run vizro-ai e2e plot tests with local vizro" part of `test-e2e-plot-vizro-ai.yml`, but that doesn't use `Layout` anywhere I presume.

Once there's a release of Vizro then the Vizro-AI unit tests would start failing, which is correct:
```
FAILED tests/unit/vizro-ai/dashboard/_response_models/test_layout.py::TestLayoutCreate::test_layout_create_valid[layout_grid_template_areas0-component_ids0-grid0] - FutureWarning: The `Layout` model has been renamed `Grid`, and `Layout` will no longer exist in Vizro 0.2.0. To ensure future compatibility, replace your references to `vm.Layout` with `vm.Grid`.
FAILED tests/unit/vizro-ai/dashboard/_response_models/test_layout.py::TestLayoutCreate::test_layout_create_valid[layout_grid_template_areas1-component_ids1-grid1] - FutureWarning: The `Layout` model has been renamed `Grid`, and `Layout` will no longer exist in Vizro 0.2.0. To ensure future compatibility, replace your references to `vm.Layout` with `vm.Grid`.
FAILED tests/unit/vizro-ai/dashboard/_response_models/test_layout.py::TestLayoutCreate::test_layout_create_invalid[layout_grid_template_areas0-component_ids0-Build failed for `Layout] - FutureWarning: The `Layout` model has been renamed `Grid`, and `Layout` will no longer exist in Vizro 0.2.0. To ensure future compatibility, replace your references to `vm.Layout` with `vm.Grid`.
FAILED tests/unit/vizro-ai/dashboard/_response_models/test_layout.py::TestLayoutCreate::test_layout_create_invalid[layout_grid_template_areas1-component_ids1-Calculated grid which caused the error:] - FutureWarning: The `Layout` model has been renamed `Grid`, and `Layout` will no longer exist in Vizro 0.2.0. To ensure future compatibility, replace your references to `vm.Layout` with `vm.Grid`.
```

Then Vizro-Ai should be fixed. Or it can be done pre-emptively if whoever makes the changes is confident about it.

As for `goodread-tutorial.md`, I think we probably leave it alone because it's published already but please do weigh in when you're back @stichbury.

maxschulz-COL:
> Great point about Vizro-AI, which I must admit I had completely forgotten about here... Agree we should fix that, but not in this PR and ideally not by me also 😅 So I assign that task to @maxschulz-COL or @lingyielia 🙏 As far as I can tell this PR won't immediately cause any Vizro-AI tests to fail. It looks the only tests that rely on local dev vizro (instead of PyPI released vizro) is "Run vizro-ai e2e plot tests with local vizro" part of `test-e2e-plot-vizro-ai.yml`, but that doesn't use `Layout` anywhere I presume.

@lingyielia can you take over this task?

PR #1103, opened 2025-04-09
[Tidy] Make `Container.title` optional

huong-li-nguyen:
## Description

- Make `Container.title `optional by default.  
- Enforce` Container.title` as mandatory via a validator only when the `Container` is used within `Tabs`.

**Context:**
Previously, `Container.title` was required in all cases, similar to `Page.title.` However, unlike Page, `Container.title` is not necessary for the `path` generation. Instead, it should be optional, aligning with components like `Graph`, `Table`, and `AgGrid`, where a title is not displayed unless provided.  

With the introduction of the `Flex` layout, having to always provide a `Container.title` has become a bit of a hassle, especially since it often needs to be hidden with CSS when creating sections without titles. By making `Container.title` optional, we make things more flexible while still ensuring it's required when used in Tabs, where a title is actually needed.

## Notice

### DIFF
--- a/vizro-core/changelog.d/20250409_100011_huong_li_nguyen_make_container_title_optional.md
+++ b/vizro-core/changelog.d/20250409_100011_huong_li_nguyen_make_container_title_optional.md
@@ -25,7 +25,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 ### Changed
 
-- Make `Container.title` mandatory only when used within Tabs.([#1](https://github.com/mckinsey/vizro/pull/1))
+- Make `Container.title` mandatory only when used within Tabs.([#1103](https://github.com/mckinsey/vizro/pull/1103))
 
 
 <!--
--- a/vizro-core/changelog.d/20250409_100011_huong_li_nguyen_make_container_title_optional.md
+++ b/vizro-core/changelog.d/20250409_100011_huong_li_nguyen_make_container_title_optional.md
@@ -25,7 +25,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 ### Changed
 
-- Make `Container.title` mandatory only when used within Tabs.([#1103](https://github.com/mckinsey/vizro/pull/1103))
+- Make `Container.title` mandatory only when used within `Tabs`. ([#1103](https://github.com/mckinsey/vizro/pull/1103))
 
 
 <!--
--- a/vizro-core/changelog.d/20250409_100011_huong_li_nguyen_make_container_title_optional.md
+++ b/vizro-core/changelog.d/20250409_100011_huong_li_nguyen_make_container_title_optional.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Make `Container.title` mandatory only when used within Tabs.([#1](https://github.com/mckinsey/vizro/pull/1))
+
+
+<!--
+### Deprecated
+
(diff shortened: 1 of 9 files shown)


vizro-core/src/vizro/models/_components/tabs.py
```diff
@@ -25,6 +25,12 @@ class Tabs(VizroBaseModel):
     # TODO[mypy], see: https://github.com/pydantic/pydantic/issues/156 for tabs field
     tabs: conlist(Container, min_length=1)  # type: ignore[valid-type]
 
+    @field_validator("tabs")
```
maxschulz-COL, line 28: We are following the `Annotated` pattern, could you also do that in this case? https://docs.pydantic.dev/latest/concepts/validators/#field-after-validator

vizro-core/src/vizro/models/_components/tabs.py
```diff
@@ -25,6 +25,12 @@ class Tabs(VizroBaseModel):
...
+    def validate_tabs_have_titles(cls, tabs: list[Container]) -> list[Container]:
+        if any(tab.title == "" for tab in tabs):
+            raise ValueError("Each `Container` in `tabs` requires a `title` provided.")
```
maxschulz-COL, line 31:
```suggestion
            raise ValueError("Each `Container` in `tabs` requires an explicit `title`.")
```
Maybe one for the native speakers - found the above sounded weird. We should also stress that it needs to be explicitly set, because it already does have a default

maxschulz-COL (commented):
Quick comment before finish review

vizro-core/changelog.d/20250409_100011_huong_li_nguyen_make_container_title_optional.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Make `Container.title` mandatory only when used within Tabs.([#1103](https://github.com/mckinsey/vizro/pull/1103))
```
antonymilne, line 28:
```suggestion
- Make `Container.title` mandatory only when used within `Tabs`. ([#1103](https://github.com/mckinsey/vizro/pull/1103))
```

vizro-core/src/vizro/models/_components/tabs.py
huong-li-nguyen: Ah, I missed that. Any particular reasons why?

antonymilne: Note that for `list[X]` there's a different way to do this now in pydantic v2 also where the validator is attached to `X` rather than the list. Overall it should look something like this:

```
# outside model
def validate_tab_has_title(tab: Container) -> Container:
    if not tab.title:
        raise ValueError("`Container` must have a `title` explicitly set when used inside `Tabs`.")
    return tab

# model field:
tabs: conlist(Annotated[Container, AfterValidator(validate_tab_has_title)], min_length=1)  # type: ignore[valid-type]
```

vizro-core/src/vizro/models/_components/tabs.py
antonymilne: See above for my suggested wording :)

vizro-core/src/vizro/models/_components/tabs.py
maxschulz-COL: There are some pros and cons listed in the pydantic docs. It mostly is convention though, and we just happen to have decided for this one.

I personally prefer attaching the validation to the type as it is more expressive on where it is applied.

antonymilne (approved):
I'd been wondering about doing this for a while and still have mixed feelings on whether I like it given that the field is mandatory in the `Tabs` context. But overall it makes sense and is a good practical solution so I like it.

Please can you check the docs quickly too in case we say that title is mandatory for `Container` there?

maxschulz-COL:
> I'd been wondering about doing this for a while and still have mixed feelings on whether I like it given that the field is mandatory in the `Tabs` context. But overall it makes sense and is a good practical solution so I like it.

@antonymilne This is actually an interesting case for where the schema becomes not the sole source of truth (`title` is actually required in some cases, although the schema tells otherwise). Because JSON schema is limited, advanced validators come into play, hence this is the correct solution. Do you see it the same?

antonymilne:
> @antonymilne This is actually an interesting case for where the schema becomes not the sole source of truth (`title` is actually required in some cases, although the schema tells otherwise). Because JSON schema is limited, advanced validators come into play, hence this is the correct solution. Do you see it the same?

Yes exactly. It's similar to cases where you have a model and can only specify one of two fields or can't specify one without specifying the other etc. These can't be encoded in JSON schema but are legitimate constraints so have to go into validators.

vizro-core/src/vizro/models/_components/tabs.py
huong-li-nguyen: Thanks guys!

maxschulz-COL (approved):
LGTM! 🚀

PR #1107, opened 2025-04-09
Check mdformat linting

l0uden:
## Notice


.pre-commit-config.yaml
```diff
@@ -85,6 +85,10 @@ repos:
         exclude: ^vizro-core/docs/pages/API-reference|^vizro-ai/docs/pages/API-reference|vizro-core/docs/pages/user-guides/custom-components.md|^vizro-core/changelog.d|^vizro-ai/changelog.d
         additional_dependencies:
           - mdformat-mkdocs[recommended]==3.1.1
+          # fixed the mdformat_admon version to the latest working, because from v2.1.0 it doesn't support
+          # mdformat-mkdocs. Actually the docs says that it is not support mdformat-mkdocs>=4.0.0 and in our case
+          # it should be working, but it is not. You can check README here: https://github.com/KyleKing/mdformat-admon
+          - mdformat_admon==2.0.6
```
huong-li-nguyen, line 91: And what if we don't use the pinned version? Does it fix it? I don't think we use the mdformat_admon dependency actively 🤔 I don't see why we have to explicitly add it here.

```suggestion
          - mdformat-mkdocs
```

huong-li-nguyen (commented):
Why do the screenshots need to be updated here? 🤔

.pre-commit-config.yaml
antonymilne: From reading the repos more, I think the best solution is to bump `mdformat-mkdocs`:
```suggestion
          - mdformat-mkdocs[recommended]==4.1.2
```

Note the `[recommended]` here is important because it installs lots of other useful packages that work well with `mdformat-mkdocs`. It's also best with these pre-commit configs to pin a specific version rather than leave open.

l0uden:
> Why do the screenshots need to be updated here? 🤔

For some reason from the latest merge to main there were failing screenshots tests. I've fixed it here.

.pre-commit-config.yaml
l0uden: Looks like Antony's suggestion worked!

l0uden: But it change all of the md files

antonymilne (approved):

huong-li-nguyen (approved):

Issue #681, opened 2024-09-04
Setting a default columnSize value for dash_ag_grid


petar-qb:
Reference link: https://dash.plotly.com/dash-ag-grid/column-sizing#size-to-fit-and-responsive-size-to-fit

Should Vizro set `columnSize="responsiveSizeToFit"` for the `dash_ag_grid` figure function by default?

huong-li-nguyen:
@maxschulz-COL and I once discussed this several times 😄 I actually don't know what we ended up with.

So this is how I would expect it to work:
- The Grid always takes on the full width of the screen size so even if there are only 3 columns, the width of the grid will be the entire page (as in the dash_data_table)
- If there are more columns in the Grid that it fits in width, a scroll bar will appear horizontally
- All columns are always well visible and not squeezed (so should be fully readable)

Now, I don't know what configuration achieves this (if any?)

maxschulz-COL:
Mostly agree with the above. But I would note a few things:
- having columns and their content fully readable can also be annoying when there are outliers
- I remember that problems also occurred also when pre-setting things, but this issue may have been removed now with your work and providing just a Div, and not an object

huong-li-nguyen:
I quickly discussed this with Joe and from a design perspective we would prefer that it expands to its full width (which it currently does not do). However, there are pros/cons for both approaches and the preference is not significant. So we decided to not change anything for now and let users configure it themselves 👍 

This means for our demos, please set `columnSize="responsiveSizeToFit"` manually for now if the table does not take the full width :)

antonymilne:
[...]

PR #1108, opened 2025-04-09
[Fix] Fix CSS for nested grid / flex layouts

antonymilne:
## Description

Check the very minor difference in the last page to see the effect of this... Basically a card inside a grid inside a flex now behaves as expected.

@huong-li-nguyen I didn't check through every page but at a glance this doesn't seem to have broken anything. Probably we should would wait until @l0uden tests are complete before merging though?

## Notice


vizro-core/src/vizro/static/css/flex.css
```diff
@@ -1,5 +1,9 @@
-/* To reset all heights coming from our parent containers otherwise there will be sizing issues */
-.flex-item, .flex-item .figure-container, .flex-item .card {
+/* To reset all heights coming from our parent containers otherwise there will be sizing issues.
```
antonymilne, line 1: @huong-li-nguyen doing `.flex-item > .figure-container` doesn't work since it's much more nested than that. I'm sure there is a way to do it with CSS selectors but not sure how robust it would be so probably not worth doing.... In which case I'm not even sure the card one is a good idea since then it's a bit inconsistent. wdyt?

huong-li-nguyen: Take a look at my latest pushed changes—I think we should approach this the other way around. Specifically, we should introduce a `grid-item` class and define dimensions there. We only want to set dimensions when elements are inside a `grid-item`. For `flex-item`, we can simply rely on the default behavior without explicitly setting dimensions.

For `card`, `loading-container`, and `figure-container,` I believe I’ve resolved the issue using inheritance and direct ancestry. Inside a `grid-item`, we set dimensions for the direct children (either `loading-container` or `card`). Any nested `loading-container` or `figure-container` elements will inherit these dimensions from their parent. This means the nesting level no longer matters—they will work correctly inside both `grid-item` and `flex-item`, regardless of how deeply they are nested.

The table-container is more challenging, as you mentioned, because we need to explicitly set dimensions for it in both `grid-item` and `flex-item`. In this case, the **nesting does matter**. While we can solve it by adding nested selector chains like .`grid-item .flex-item .grid-item`, it’s not the most elegant solution 😅. That said, I think it’s good enough for now, unless you can think of another way. Realistically, I don’t think people will go beyond three levels of nesting, as layouts become difficult to manage at that point.

huong-li-nguyen (approved):
The example app seems to look fine and the screenshot tests also passed, so approval from my side, but feel free to stress-test further 👍

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,34 +1,460 @@
...
+        page20,
+        page21,
+        page22,
+        page23,
+        page24,
+        page25,
+        page26,
+        page27,
+        page28,
+        page29,
```
huong-li-nguyen, line 454: The relevant ones are pages 28 and 29. I checked the others as well and all looks fine, but here you can see the difference. You can also see why three levels of nesting can get really tricky to maintain with the scroll bars/dimensions, etc., so I generally wouldn't recommend doing that unless people understand CSS and the interactions of nesting flex inside grid inside flex inside grid 😄 .

vizro-core/src/vizro/static/css/flex-grid.css
```diff
@@ -0,0 +1,46 @@
...
+.flex-item .table-container, .grid-item .flex-item .table-container, .flex-item .grid-item .flex-item .table-container {
+  height: 450px;
+  width: 700px;
+}
+
+/* stylelint-disable-next-line no-descending-specificity */
+.grid-item .table-container, .flex-item .grid-item .table-container, .grid-item .flex-item .grid-item .table-container {
+  height: 100%;
+  width: 100%;
+}
```
antonymilne, line 46: Let me just check something here... There's 8 possible combinations of 3-level nesting (I agree this is a good cutoff point). What happens in each case?

What we want to happen is that the last, innermost class takes priority so it's always 100%, 100% for ones ending "grid" and 450x, 700px for ones ending "flex"?

```
grid grid grid
grid grid flex
grid flex grid - 100%, 100%
grid flex flex
flex grid grid
flex grid flex - 450px, 700px
flex flex grid
flex flex flex
```

huong-li-nguyen: Yes, exactly 👍  I checked these cases already and it works like that.

```
grid grid **grid** - 100%, 100%
grid grid **flex** - 450px, 700px
grid flex **grid** - 100%, 100%
grid flex **flex** - 450px, 700px
flex grid  **grid** - 100%, 100%
flex grid **flex** - 450px, 700px
flex flex **grid** - 100%, 100%
flex flex  **flex** - 450px, 700px
```

vizro-core/src/vizro/static/css/flex-grid.css
```diff
@@ -0,0 +1,46 @@
...
+
+.grid-item > .loading-container {
+  height: 100%;
+  width: 100%;
+}
+
+/* Apply dimensions to .table-container based on its closest ancestor (up to 3 levels) */
+/* stylelint-disable-next-line no-descending-specificity */
+.flex-item .table-container, .grid-item .flex-item .table-container, .flex-item .grid-item .flex-item .table-container {
```
nadijagraca, line 37: Did you consider using `.flex-row` and `.flex-column` to conditionally set the width of the container? I know its not the cleanest solution and it will have a lot of repetitive code, but width of 700px looks weird in `flex direction: column`.

nadijagraca (approved):
Other than one small comment/question I had all looks really good. 🙌

vizro-core/src/vizro/static/css/flex-grid.css
huong-li-nguyen: That's definitely an interesting idea, but I do think it might make things more complicated, especially since there’s no one-size-fits-all solution. For example, if we change the default width for `.flex-column`, what should it be? Should it take up the entire right side? That might look great if the` Flex` container spans the full width and the table data fills it, but what if someone only uses half the right side? Or if the AgGrid only has only two columns of data? That increased width would then look off. With a fixed width, there are always going to be cases where it doesn’t quite work. 😬  

Right now, the dimensions are based on Plotly figures, so the behavior stays consistent. I think the width is already a bit bigger than usual. And since Plotly figures don’t adjust their logic for rows or columns, I feel like keeping the same approach for the table works ok for now.  

That said, I think the best solution would be for the `table-container` to automatically adjust its width based on the table’s actual content, no matter if it’s in a `flex-row` or `flex-column`. The `ag-header-row` has the relevant number in its `style` attribute, so we’d probably need to use JavaScript to dynamically set the width and make sure it works across all the nested setups. However, I don't know how complicated it is to implement this, but might be worth exploring in the next sprint if the fixed width doesn't work for us? It is good enough for me at the moment though, so up to you guys and how much capacity you have :) 

Screenshot 2025-04-11 at 12 39 16

vizro-core/src/vizro/static/css/flex-grid.css
antonymilne: Perfect, thanks for checking!

vizro-core/src/vizro/static/css/flex-grid.css
antonymilne: Let's leave it for now I think - agree it's impossible to come up with a perfect solution that works for all possible setups here, and this feels good enough for now. So we can come back to it in future if it turns out that the `table-container` width is problematic for common layouts.

@huong-li-nguyen please do add the comment to https://github.com/mckinsey/vizro/issues/681 if you think it's relevant there.

vizro-core/changelog.d/20250410_105936_huong_li_nguyen_nested_flex_css.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
```
antonymilne, line 28: A small entry here would be nice.

antonymilne (commented):
I can't approve the PR because I raised it, but I approve it 😀 Thanks for the fix @huong-li-nguyen, it's much better than what I had done 👍

PR #1112, opened 2025-04-12
[Tidy] Update styling of `Card`

huong-li-nguyen:
## Description
- Remove `text` className from `Text`
- Move some of the CSS specifications from `bootstrap-overwrites.css` to `vizro-bootstrap.css`
- Update `card` styling on hover (no jumping effect anymore)

Changes inside vizro-bootstrap:
- Update `card` bg-color on light theme
- Remove `card-text` font styles from `vizro-bootstrap.css`, allowing it to inherit the default body text styling
- Update `card-nav` styling on `hover`: The previous "jumping" effect has been removed, and a border is now added to indicate interactivity.

## Notice


vizro-core/src/vizro/models/_components/text.py
```diff
@@ -46,7 +46,6 @@ def build(self):
             "id": self.id,
             "children": self.text,
             "dangerously_allow_html": False,
-            "className": "text",
```
maxschulz-COL, line 49: Not entirely sure if I understand correctly, but to me the styling of the `text` component looks the same after removal, so all good 👍

maxschulz-COL (approved):
Compared the versions, and the card text and text are not the same as far as I can tell. So good form my side 👍

vizro-core/src/vizro/models/_components/text.py
antonymilne: `vm.Text` should look the same but `vm.Card` has a better line-spacing now I think? @huong-li-nguyen

antonymilne (approved):
The line spacing looks good now 👍 Is there anything else this should have fixed also?

FYI if you didn't realise it, `className="card-text"` appears in the bootstrap app.py so do check that still works as you'd like.

vizro-core/src/vizro/models/_components/text.py
maxschulz-COL: Yes indeed, that is also what I observed

huong-li-nguyen:
> The line spacing looks good now 👍 Is there anything else this should have fixed also?
> 
> FYI if you didn't realise it, `className="card-text"` appears in the bootstrap app.py so do check that still works as you'd like.

Yes, yes. I changed the underlying vizro-bootstrap file first, before this PR 😄  https://github.com/mckinsey/vizro-bootstrap/pull/7

So card-text is coming from the vizro-bootstrap file and I've changed it there like I wanted

vizro-core/src/vizro/models/_components/text.py
huong-li-nguyen: They are the same haha 👀 

Both the `Markdown` inside the `Card` and the `Text` inherit their styles from the `body`. No other CSS is applied, except we remove the `margin-bottom` from the last item in the `Card` because the extra spacing isn’t needed, as the `Card` has padding.

The reason they look different is because the `Card` has padding, margin, height, and width set, which changes how it looks overall. But that doesn’t come from the actual `Markdown` inside. If you add the same styling (like the Card's padding, margin, height, and width) to the outer container of the `Text`, everything will line up and be the same line height :)

Screenshot 2025-04-14 at 15 59 51

PR #1114, opened 2025-04-14
Feat/enable variant button

huong-li-nguyen:
## Notice

### DIFF
--- a/vizro-core/changelog.d/20250412_200111_huong_li_nguyen_enable_variant_button.md
+++ b/vizro-core/changelog.d/20250412_200111_huong_li_nguyen_enable_variant_button.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Enable styling of `vm.Button` with a new argument `variant="plain"/"filled"/"outlined"`. See the user guide on styled buttons for more details. ([#1113](https://github.com/mckinsey/vizro/pull/1113))
+
+
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Fixed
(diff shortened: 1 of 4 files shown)


vizro-core/src/vizro/models/_components/button.py
```diff
@@ -52,11 +59,16 @@ class Button(VizroBaseModel):
...
+        variants = {"plain": "link", "filled": "primary", "outlined": "secondary"}
+
         defaults = {
             "id": self.id,
             "children": self.text,
             "href": get_relative_path(self.href) if self.href.startswith("/") else self.href,
             "target": "_top",
+            # dbc.Button includes `btn btn-primary` as a  class by default and appends any class names provided.
+            # To prevent unnecessary class chaining, the button's style variant should be specified using `color`.
+            "color": variants[self.variant],
```
huong-li-nguyen, line 71: Screenshot 2025-04-14 at 10 16 00

This is what happens if you provide it via `class_name` instead of `color`. So to keep this clean, we deviate a bit from the container implementation and provide via `color` instead

vizro-core/changelog.d/20250412_200111_huong_li_nguyen_enable_variant_button.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Enable styling of `vm.Button` with a new argument `variant="plain"/"filled"/"outlined"`. See the [user guide on styled buttons](https://vizro.readthedocs.io/en/stable/pages/user-guides/button/#styled-buttons) for more details. ([#1114](https://github.com/mckinsey/vizro/pull/1114))
```
petar-qb, line 22: I see the order of button variants in the changelog, docstrings, code and tests is `variant="plain"/"filled"/"outlined"`, but the default is `filled`. Should we introduce a convention and always list the default value as the first element like: `variant="filled"/"plain"/"outlined"`? Or to always stick with the alphabetical order instead (even though I like `default-value/the-rest-of-the-options` more)?

petar-qb (approved):
Very clean description, code and tests! 🥇 

I have one question, but other than that everything looks great!

vizro-core/changelog.d/20250412_200111_huong_li_nguyen_enable_variant_button.md
huong-li-nguyen: I used the same order as inside the `variant` argument of `Container`, because I preferred to have the same order everywhere 😄 I don't mind which order we take, as long as I can then just copy/paste this variant argument to various models.

nadijagraca (approved):
Very nice and clean. I like the new buttons. 🚀

vizro-core/docs/pages/user-guides/button.md
```diff
@@ -158,6 +158,62 @@ vm.Page.add_type("controls", vm.Button)
 ...
 ```
 
+## Styled buttons
+
+There are three predefined button styles that can be customized using the `variant` argument.
```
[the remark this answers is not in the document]
huong-li-nguyen: Alright, let me just remove the default button. 😄 I think unlike the Container examples, we should display all three button styles here. The reason I didn't include the default style in the Container is that it doesn't have any styling applied. However, for the Button, all three styles—including the default—have distinct styling, so it makes sense to show them all 👍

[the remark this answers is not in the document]
antonymilne: Agreed, I think this makes perfect sense.

PR #1128, opened 2025-04-18
[Tidy] Fix position of collapse-icon and remove border color

huong-li-nguyen:
## Description
- Add back left margin to collapse-icon when side-panel is collapsed, otherwise it's not visible
- Remove border-color on hover

## Notice


petar-qb:
Quick thought: Should we bring back the smooth transition for the collapsible panel and apply the same effect to the collapsible containers too? I personally think it looks a lot nicer as it gives a sense of something actually collapsing/expanding rather than just suddenly appearing or disappearing.

huong-li-nguyen:
> Quick thought: Should we bring back the smooth transition for the collapsible panel and apply the same effect to the collapsible containers too? I personally think it looks a lot nicer as it gives a sense of something actually collapsing/expanding rather than just suddenly appearing or disappearing.

From what I see, @nadijagraca has already implemented that. For both, the collapse_nav_panel, as for the collapse_container she uses `transition: "transform 0.35s ease-in-out",`. So if that doesn't work as you expect, feel free to open a ticket and she can take a look :) But she has implemented above from what I can see in pure code.

PR #1145, opened 2025-05-05
[Feat] Allow model IDs as inputs/outputs via default input/output properties

huong-li-nguyen:
## Description

* Introduced a new `_get_component_id_and_property` method that parses a string like `"my-graph.figure"` into a component ID and property. If only a component ID is provided (e.g., `"my-graph"`), it automatically uses the model’s default input or output property.
* Updated `_validate_dash_dependencies` to raise a `ValueError` when a component ID is provided without an explicit property and the model lacks a default input or output.
* Extended `_transformed_inputs` and `_transformed_outputs` to support component IDs without explicit properties by using their default input/output properties.
* Added `_input_default_property` and `_output_default_property` to each model to enable the above functionality.

**What does it mean?**
These changes make actions more user-friendly, as our users don't need to look up the input / output property anymore:

* **Before** (explicit property required):
  `outputs = ["card-id.children"]`
  `inputs = ["button-id.n_clicks"]`

* **After** (property inferred via model default):
  `outputs = ["card-id"]`
  `inputs = ["button-id"]`

**TO DO:**

- [ ] Align with @antonymilne how this works together with https://github.com/mckinsey/vizro/pull/1150
- [ ] Docs and tests will follow in a separate PR: https://github.com/mckinsey/vizro/pull/1148
----

**Quick overview of default inputs and outputs + open questions:**

* When multiple inputs apply, we list the most common one. For the Table and AgGrid, I went for selected rows. 
* Each component has an Input and Output for consistency, though some are rarely used as inputs/ouputs. Shall we remove the defaults there? 

Component | Default Input Property | Default Output Property | Open question
-- | -- | -- | --
Form | children | children | Rarely used as input -  remove?
Container | children | children | Rarely used as input - remove?
Figure | children | children | Rarely used as input and output - remove?
AgGrid | selectedRows | children | Agree with default?
Table | selected_rows | children | Agree with default?
Graph | clickData | figure | ✅ 
Button | n_clicks | children | ✅ 
Card | children | children | Rarely used as input - remove?
Text | children | children | Rarely used as input - remove?
Alert | is_open | children | ✅ 
TextArea | value | value | ✅ 
UserInput | value | value | ✅ 
Checklist | value | value | ✅ 
DatePicker | value | value | ✅ 
Dropdown | value | value | ✅ 
RadioItems | value | value | ✅ 
RangeSlider | value | value | ✅ 
Slider | value | value | ✅ 
Tabs | active_tab | children | Rarely used as output - remove?
Filter | children | children | Rarely used as input - remove?
Parameter | children | children | Rarely used as input and output -  remove?
Dashboard | children | children | Rarely used as input and output - remove?
Flex | children | className | Rarely used as input -  remove?
Grid | children | className | Rarely used as input -  remove?
Accordion | active_item | children | Rarely used as output - remove?
NavBar | children | children | Rarely used as input and output - remove?
NavLink | active | children | Rarely used as input and output - remove?
Navigation | children | children | Rarely used as input and output - remove?
Page | children | children | Rarely used as input and output -  remove?
Tooltip | is_open | is_open | Rarely used as input -  remove?

## Notice


vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -309,15 +361,17 @@ class Action(_BaseAction):
...
-    inputs: list[DotSeparatedStr] = Field(
+    inputs: list[str] = Field(
         default=[],
-        description="""List of inputs provided to the action function, each specified as `<component_id>.<property>`.
+        description="""List of inputs provided to the action function. Each input can be specified as either
+            `<component_id>.<property>` or just `<component_id>` if the model has a default input property defined.
             Defaults to `[]`.""",
     )
-    outputs: Union[list[DotSeparatedStr], dict[str, DotSeparatedStr]] = Field(  # type: ignore
+    outputs: Union[list[str], dict[str, str]] = Field(  # type: ignore
```
huong-li-nguyen, line 370: Not completely sure about this change, so sharing some context:

I originally added a custom type `ModelIDOrDotSeparatedStr` since we accept two formats: a dot-separated string like `component-id.property`, or just a model ID (as long as it has a default `_input_default_property` or `_output_default_property`). The goal was to catch errors early with Pydantic and improve type hints/IDE support.

But I realized we still need to run the same validation in `_validate_dash_dependencies` inside `BaseAction`, because the Pydantic field validator doesn't apply to `AbstractAction`.

That meant duplicating the same validation logic in two places — the custom type `ModelIDOrDotSeparatedStr` and `_validate_dash_dependencies` — which felt unnecessary. So I removed the custom type and just went with `str` as the type hint, since:

* `_validate_dash_dependencies` is called for both `Action` and `AbstractAction`
* the logic is shared anyway
* and this keeps the codebase simpler

Happy to bring the custom type back if we think the IDE/type checking benefits are worth it. I am still undecided 🤔 

```
class ModelIDOrDotSeparatedStr(str):
    """A string that is either a model ID or a dot-separated string.
    
    This type validates that the string is either:
    1. A dot-separated string in the format `<component_id>.<property>` (e.g., "graph-1.figure")
    2. A valid model ID that has a default input/output property defined (e.g., "graph-1")
    """
    @classmethod
    def __get_validators__(cls):
        yield cls.validate

    @classmethod
    def validate(cls, v, *, type: Literal["input", "output"] = "output"):
        if not isinstance(v, str):
            raise ValueError("Value must be a string")
        
        # If it's a dot-separated string, validate it
        if "." in v:
            return DotSeparatedStr.validate(v)
        
        # If it's just a model ID, validate it exists and has default property
        try:
            model = model_manager[v]
            property_name = "_input_default_property" if type == "input" else "_output_default_property"
            if not hasattr(model, property_name):
                raise ValueError(
                    f"Model '{v}' does not have a default {type} property defined. "
                    f"Please specify the property explicitly as '{v}.<property>'."
                )
            return v
        except KeyError:
            raise ValueError(
                f"Action {type}s {v} must be a string of the form "
                "<component_name>.<component_property> or a valid model ID with a default {type} property."
            )
```

petar-qb:
> Hey @petar-qb, before I get into docs and tests, could you take a quick look to see if this is on track with what you and @antonymilne were thinking? :) I’ve got a few open questions — check the comments.

Hey @huong-li-nguyen, I quickly skimmed through the PR and it's definitely on the right track. Feel free to continue working on docs/tests and other. 🚀  🚀

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: Ah, yes I see, I forgot about that 😬  I'll leave it as is for now, but maybe @antonymilne has a better idea or we just live with `str` as a type here. It will be caught anyway when running `_validate_dash_dependencies`, so also suffices for me.

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -465,3 +399,87 @@ def test_action_callback_function_outputs_mapping_return_value_keys_not_match(se
             ValueError, match="Keys of action's returned value .+ do not match the action's defined outputs {'output'}."
         ):
             action._action_callback_function(inputs={}, outputs={"output": Output("component", "property")})
+
+
+class TestBaseActionValidateDependencies:
```
huong-li-nguyen, line 404: I moved all these tests down here and now call `_validate_dash_dependencies` directly, since all the validation happens here now instead of in the Pydantic field validator. That said, these tests feel only kind of useful at this point—it's not really about the syntax anymore. The KeyErrors are just coming from the fact that the IDs aren't registered in the model manager. Still unsure if the validation should stay where it is, but depending on our decision there, I might rewrite all of these tests in a separate PR here: https://github.com/mckinsey/vizro/pull/1148

vizro-core/src/vizro/actions/_filter_interaction.py
```diff
@@ -100,7 +100,7 @@ def outputs(self):
 
         for target in self.targets:
             component_id = target
-            component_property = cast(FigureType, model_manager[target])._output_component_property
+            component_property = cast(FigureType, model_manager[target])._output_default_property
```
antonymilne, line 103: @huong-li-nguyen I _think_ one of the acceptance criteria for this should be that we can just do `return self.targets` as per this part of the TODO:

```
When this change is made the outputs property for filter, parameter and on_page_load should just
become `return self.targets` or similar.
```

Does that work with what you have here already?

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -60,32 +60,82 @@ def _runtime_args(self) -> dict[str, _IdProperty]:
     def _action_name(self) -> str:
         raise NotImplementedError
 
+    def _get_component_id_and_property(self, reference: str, *, type: Literal["input", "output"]) -> tuple[str, str]:
```
antonymilne, line 63: Generally agree with the logic here but I have a slightly reworking of it which will also make it easier to integrate @petar-qb's forthcoming changes - will share tomorrow.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -60,32 +60,82 @@ def _runtime_args(self) -> dict[str, _IdProperty]:
...
+
+        Returns:
+            Tuple of (component_id, component_property)
+        """
+        if "." in reference:
+            component_id, component_property = reference.split(".")
+            return component_id, component_property
+        component_id = reference
+        property_name = "_input_default_property" if type == "input" else "_output_default_property"
+        return component_id, getattr(model_manager[component_id], property_name)
```
maxschulz-COL, line 82: Just glanced over this to check for things like this. @antonymilne do we need to pre-think any revised MM logic, just worried that we dig a deeper and deeper whole. Might be just a case of keeping things in mind

antonymilne: This is a very legitimate concern and exactly the same thought occurred to me in the shower this morning 🚿 .

Does it concern me? A little bit, yes... But I think we need to make this change anyway:
* there's no other way to do this that I can think of without `model_manager`, and it really is an essential feature IMO
* overall it will hopefully reduce our dependence on `model_manager` rather than increase it because it will actually centralise this sort of logic to one place since access to the `model_manager` will be removed from all our actions themselves where it's currently done multiple times (not done in this PR yet but will be). So if we need to manually inject `model_manager` this will actually make it required in fewer rather than more places ultimately

antonymilne: The places we should be able to remove references to `model_manager` as a result are e.g. here: https://github.com/mckinsey/vizro/pull/1145#discussion_r2075884753

huong-li-nguyen:
Closing as duplicated in: https://github.com/mckinsey/vizro/pull/1153

PR #1143, opened 2025-04-29
[Feat] Enable dict output format for `Action.outputs`

huong-li-nguyen:
## Description

- Enable dict `output` format for `vm.Action.outputs`

**Code Example**

```
# This already worked
action_output_list = vm.Action(
    function=action_function_output_list("button-id.n_clicks"),
    outputs=["card-id-1.children", "card-id-2.children"],
)

# This didn't work, but now works
action_output_dict = vm.Action(
    function=action_function_output_dict("button-id.n_clicks"),
    outputs={
        "card-id-1": "card-id-1.children",
        "card-id-2": "card-id-2.children",
    },
)
```

## Notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

    - I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
    - I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
    - I certify that the use of this contribution as authorized by the Apache 2.0 license does not violate the intellectual property rights of anyone else.
    - I have not referenced individuals, products or companies in any commits, directly or indirectly.
    - I have not added data or restricted code in any commits, directly or indirectly.

PR #1150, opened 2025-05-06
[Tidy] Validate AbstractAction outputs in only one place

antonymilne:
## Description

I thought again about the changes in #1143 and have a few more modifications here:
* move the checking so it only occurs for `AbstractAction` and not `Action` as well by moving it to `AbstractAction._transformed_outputs` and using `super()`
* validation is now a bit more robust since we don't test `outputs.values()` for the dict case but instead test the whole `dict[str, DotSeparatedString]`. This means we'll catch cases like `{1: "component.property"}` or just wrong types like `1` which we wouldn't before
* `validate_dash_dependencies` is no longer used anywhere. The function definition remains since it still has `TODO`s in it and it might turn out with these that we actually still want it
* `DotSeparatedString` made private - good idea to introduce it, but not quite yet I think. I put in a new TODO to think about this with the remaining `TODO-AV2 D` tickets

Test app still works the same as before, although validation error messages are a little bit different now in the `AbstractAction` case.

## Notice


huong-li-nguyen:
This will probably affect this PR: https://github.com/mckinsey/vizro/pull/1145

I actually had to remove the annotated type `DotSeparatedStr` there, but I did have several open questions to where the validation should live etc. If this PR will affect how I should do https://github.com/mckinsey/vizro/pull/1145, let me know 👍

huong-li-nguyen (approved):
Nice! I’m pre-approving this now since I think we can cover the other open questions during our call later. But I’m still a bit unsure about the `DotSeparatedStr` validation in the Pydantic field, which we can deal with separately anyway 🤔

---

Let me know if you’d like to adjust it further!

PR #1153, opened 2025-05-06
[Feat] Allow model IDs as action outputs and inputs

antonymilne:
## Description

- Added `_transform` method in `_BaseAction` to validate action outputs and inputs. 
  - Validates references are in format `component-id.property` or `component-id`
  - For implicit references (just `component-id`), uses default output/input property
  - Raises appropriate errors for invalid formats (e.g. `component-id..property`) or if the relevant component id doesn't have a default output/input in case the implicit version is used
- Added private property of `_outputs` and `_inputs` to models according to table below

**Related issues**
- Closes: https://github.com/McK-Internal/vizro-internal/issues/1609
- Enables: https://github.com/McK-Internal/vizro-internal/issues/1610

**What does it mean?**
These changes make actions more user-friendly, as our users don't need to look up the input / output property anymore:

* **Before** (explicit property required):
  `outputs = ["card-id.children"]`
  `inputs = ["button-id.n_clicks"]`

* **After** (property inferred via model default):
  `outputs = ["card-id"]`
  `inputs = ["button-id.n_clicks"]`

| Component   | Default Input Property | Default Output Property      | Additional output properties to support        |
| ----------- | ---------------------- | ---------------------------- | ---------------------------------------------- |
| Form        |                        |                              |                                                |
| Container   |                        |                              |                                                |
| Figure      |                        | children                     | figure                                         |
| AgGrid      |                        | children                     | title, header, footer, figure                  |
| Table       |                        | children                     | title, header, footer, figure                  |
| Graph       |                        | figure                       | title, header, footer, figure (already works?) |
| Button      |                        | Not sure so don't do for now |                                                |
| Card        |                        | children                     | text                                           |
| Text        |                        | children                     | text                                           |
| Alert       |                        |                              |                                                |
| TextArea    | value                  | value                        | title, description                             |
| UserInput   | value                  | value                        | title, description                             |
| Checklist   | value                  | value                        | title, description                             |
| DatePicker  | value                  | value                        | title, description                             |
| Dropdown    | value                  | value                        | title, description                             |
| RadioItems  | value                  | value                        | title, description                             |
| RangeSlider | value                  | value                        | title, description                             |
| Slider      | value                  | value                        | title, description                             |
| Tabs        |                        |                              |                                                |
| Filter      | maybe in future        | maybe in future              |                                                |
| Parameter   | maybe in future        | maybe in future              |                                                |
| Dashboard   |                        |                              |                                                |
| Flex        |                        |                              |                                                |
| Grid        |                        |                              |                                                |
| Accordion   |                        |                              |                                                |
| NavBar      |                        |                              |                                                |
| NavLink     |                        |                              |                                                |
| Navigation  |                        |                              |                                                |
| Page        |                        |                              |                                                |
| Tooltip     |                        |                              |                                                |

## Notice


vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -135,14 +135,40 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
...
         # _AbstractAction instance.
-        runtime_args = {arg_name: State(*arg_value.split(".")) for arg_name, arg_value in self._runtime_args.items()}
+        runtime_args = {
+            arg_name: State(*self._transform(arg_value).split("."))
+            for arg_name, arg_value in self._runtime_args.items()
+        }
 
         return builtin_args | runtime_args
 
+    @staticmethod
```
huong-li-nguyen, line 150: Just adding a few notes ahead of our call so I don’t forget:

My main question is about the typing for `Action.outputs` and `Action.inputs` if we also want to support passing a `ModelID`. As it stands, using something like `outputs=["my-card"]` causes Pydantic validation to fail, since it expects a `DotSeparatedStr`.

Defining a type like `DotSeparatedStrOrModelID` doesn’t seem viable, though — from what I understand, Pydantic doesn’t have access to the full list of registered models from the model manager, so it can’t validate whether a given string corresponds to a valid model.

So to me it seemed like all validation for `_AbstractAction` and `Action` needs to happen here `_transform` or `_validate_dash_dependencies` instead? We’d lose the benefits of Pydantic validation, but at least the checks would be centralized in one place, which again has its pros and cons. Did I understand that correctly or is there another way we can keep both? Separate validation for `_AbstractAction` and pydantic validation for `Action`, while still allowing for valid Model IDs? 🤔

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -135,14 +135,40 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
...
+    def _transform(output):
+        # need to make this work with outputs or inputs
+        # Need to also handle case that _outputs isn't defined
+        if "." in output:
+            component_id, component_property = output.split(".")
+            if component_id in model_manager and component_property in model_manager[component_id]._outputs:
+                return model_manager[component_id]._outputs[component_property]
+
+            # not a problem if either case of and is False, just leave output unchanged
+            return output
```
huong-li-nguyen, line 160: While testing the different use cases, I wasn’t completely sure why we don’t catch this error. I noticed you're returning the output unchanged if either case of and is False.

I’m talking about cases where the syntax is valid (i.e., it's a `DotSeparatedStr`), but the referenced `ModelID` doesn’t actually exist. For example:

```python
outputs = ["wrong-id.prop"]
# or
outputs = {"anything": "wrong-id.prop"}
```

If we return this as-is and the component ID isn’t found in the model manager, a Dash callback error gets raised at runtime. What’s our general approach here — is it fine to let that happen, or do we prefer to catch these kinds of issues earlier? In this case, I figured it might be cleaner to throw a validation error upfront, especially since it doesn’t add much extra code, but not sure what our general approach is for this 🤔

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -135,14 +135,40 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
...
+        component_id, component_property = output, "__default__"
+        if component_id not in model_manager:
+            # this is a problem
+            raise
+        elif component_property not in model_manager[component_id]._outputs:
+            # this is a problem
+            raise
+
+        return model_manager[component_id]._outputs["__default__"]
+
```
huong-li-nguyen, line 171: There’s one use case I’m not sure we can even catch (and I didn’t handle it in the other PR either). It’s when the value uses dot notation, the model ID exists, but the specified property doesn’t.

Right now, the dashboard still runs without any errors, but the interaction (e.g., clicking a button) just silently fails — there’s no clear indication of what went wrong, which makes it pretty hard to debug.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -135,14 +135,40 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
...
         return builtin_args | runtime_args
 
+    @staticmethod
+    def _transform(output):
+        # need to make this work with outputs or inputs
+        # Need to also handle case that _outputs isn't defined
+        if "." in output:
+            component_id, component_property = output.split(".")
+            if component_id in model_manager and component_property in model_manager[component_id]._outputs:
+                return model_manager[component_id]._outputs[component_property]
```
huong-li-nguyen, line 157: Just a note for myself: If we decide to loosen the pydantic field validations for `Action.inputs` and `Action.outputs`  (so we can support raw `ModelID`s), then I think we need to check for a valid `DotSeparatedStr` here again, as the dot check otherwise doesn't capture invalid inputs 🤔

huong-li-nguyen:
The tables look good 👍 I’m also for keeping the pre-selection smaller. One thing I was thinking, though, is maybe we should add the title property for things like Container, Page, and Tabs too. I often find myself wanting to change the title based on filter selections, for example. If we're doing it for Graph, AGGrid, and Table, why not for Container, Page, and Tabs too? 🤔

vizro-core/src/vizro/models/_action/_action.py
petar-qb: This is the great question! I think we should not validate it `upfront` because only VizroBaseModel models live within the model_manager. So if users want to output the `"custom-component-id_title.children"` we shouldn't forbid that. The only correct place to validate that, unfortunately, is at `runtime`.

vizro-core/src/vizro/models/_action/_action.py
antonymilne: In case it wasn't clear after our conversation: this is the circle case below and we should interpret as in pure Dash as `component_id.property`, where `component_id` coincides with a model ID. Think e.g. of `vm.Graph(id="my_graph")` and then doing `my_graph.style`. So if that exists in the layout then it should work, and if it doesn't `my_graph.non_existence_property` then it wouldn't - but the checks and error messages are done entirely by Dash, nothing from us here.
image

vizro-core/src/vizro/models/_action/_action.py
antonymilne: In case it wasn't clear from the call: @petar-qb is right here 👍 Anything like this should just get passed straight to Dash as an `Output/Input`, and then it's down to Dash to work our how and when to raise the error.

antonymilne:
> The tables look good 👍 I’m also for keeping the pre-selection smaller. One thing I was thinking, though, is maybe we should add the title property for things like Container, Page, and Tabs too. I often find myself wanting to change the title based on filter selections, for example. If we're doing it for Graph, AGGrid, and Table, why not for Container, Page, and Tabs too? 🤔

I think this is very legitimate 👍 I'm happy to make `title`  and `description` available as output (but not yet input) properties for those models. FYI @petar-qb

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: Yes, that makes it clearer, but I still think the above use case is not handled properly 😄 I might raise an issue on Dash for this though! Because I think this should be raised on their side then!

huong-li-nguyen: I haven't tried it yet :)

huong-li-nguyen: Yes, it's also reproducible in Dash btw: [Example](https://py.cafe/snippet/dash/v1?pycafe-app-view#c=H4sIAK2GG2gEA41UbWvbMBD-K0L7koLjvtENAoU13Yf2y1q6wBhxMYp1mUVlSZPObU3Jf99JjvMCGZkCce7tee4en_LBKyuBT_gnVtVQvTDbIqsRXZicnkoR6txpi7rLK9ucsqX1TNqqbcCgQGVNYT6xGyMP1DoqEUs4bYTwCGbhQUrQIUE01gODd9E4DaEwqnHWI4tshVl626SfbO2usdEZe2jRtZixe5Mesqr26sqFtRjQC1dSo84aajAwEZhcUGJhhHPsOmXm3-hrVJZGNFCWGbWB4I3QZcCOuqkBMFzPqSzHGhoI-fThYfZj9nTz-HyyRsq16OKoBEhpt5a0UAb8aF4YRic6py2iNaOC32pFwjRQ8IwpeV3wphsvUjB6KqutJ6fzqhG-Sy4tQvhOzcVcHF8W_CTb4t4KLweeeAbf1Mpu1x9PFC6_u4pNUAabKdSpjV2GiiJj7CMDz3BS_SOV3xuFSmgWkxmSYHmESdP09eQ6DJwCJFvCoxMlTM_95PWchekl_ho1roTWC1G9jPqK_v1TMzvAGSOzVlp6IDWH9tOGUOKe0AWX2wzn4ZX2o1T9XGWkup75FgpD_BKWrHVSIFDAyzJyjUxZxRcZTiY9hAdsvWHLgv-yLc1CMSBtamA9J_sYKlYMVQOjcEKiDfvjWzNKk_KMe_jTKk-bRgtL1zCuKPVA3wdXmgqwc_G-xhSyXIe1NWS7zkolYfx6ll98zs8ppNOW8skHfwUf6LLyyQXxEeqTfYvuNZInK9sIySfzTQTFIgBS8E1JrPnk_Oos440yP3vzsrfuQP2uiSeaSlLZUmmYEmoAv7kc_2CIqeNFn0spTkRczlfPq2yTs-niEAAFt3VRW9fx_dpteFfqHN9xn8Yfl-H46OtRjo09TLzXwIamXqOef7n6H076_yK30MdIh7zIGj-rLMlPWzd_Xv0FSAWbEQgGAAA)

It throws a nice Dash error if the component ID doesn't exist, but if the component ID exists but we use a property that doesn't exist, it currently behaves like in the pyCafe app. It silently fails, you click on the Button, but don't get an obvious error message. There are some errors in the console though.

huong-li-nguyen: Yeah, I tried both debug=True and debug=False and in both cases I don't get that pop-up window 🤔 I only get it for the case of an invalid component ID. Strange, I can show you in our catch-up, but I guess if it's working for you, then all good?

petar-qb: > (just tried with Dash 2, not 3)

The PyCafe example (callback output non-existing property) raises the following error in the `Dash v3` as well.

image

However, in the Vizro app (with the action output non-existing property) it's silent (id both Dash v2 and v3 ). Let's keep this comment unresolved and I can investigate it later this week.

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -489,3 +420,86 @@ def test_action_callback_function_outputs_mapping_return_value_keys_not_match(se
...
+    )
+    def test_action_outputs_invalid(self, outputs):
+        # LQ: Check why some of these don't throw ValidationErrors instead of KeyError
+        with pytest.raises(
+            KeyError,
+            match="Component with ID .* not found. Please provide a valid component ID or use the explicit "
+            "format '<component-id>.<property>'.",
+        ):
+            action = Action(function=action_with_no_args(), outputs=outputs)
+            action._transform(outputs, type="output")
```
huong-li-nguyen, line 475: **TODO:** Discuss with A and P how to handle the current tests. They currently fail with `KeyError` because the component isn't found as they are not registered in the model manager. We likely need to replace these with tests that explicitly check for registered models instead, otherwise it's not really checking for the wrong syntax. Also, the `ValidationError` is no longer raised in some cases where I expected it to be—need to investigate why that’s happening.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -141,14 +129,65 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
...
+        Raises:
+            KeyError: If component not found or no default mapping exists
+            ValueError: If reference format is invalid e.g. "model-id.prop.prop"
+        """
+        property_name = "_outputs" if type == "output" else "_inputs"
+
+        if "." in reference:
+            # LQ: Check whether we need TypeAdapter here or something else. Currently it doesn't catch cases such as
+            # `component-id.prop.prop` well. However, adding below leads to a bunch of unit tests to fail. Check later.
+            # TypeAdapter(Union[list[_DotSeparatedStr], dict[str, _DotSeparatedStr]]).validate_python(reference)
```
huong-li-nguyen, line 169: QQ - Are we fine with the simple "." in reference check or shall we go for something stricter? At the moment, it doesn't capture cases such as `component-id.prop.prop` unless you uncomment the line with the `TypeAdapter`

vizro-core/src/vizro/models/_components/figure.py
```diff
@@ -35,6 +35,10 @@ class Figure(VizroBaseModel):
 
     _validate_figure = field_validator("figure", mode="before")(validate_captured_callable)
 
+    @property
+    def _outputs(self) -> dict[str, str]:
+        return {"__default__": f"{self.id}.children"}
+
```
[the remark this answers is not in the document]
petar-qb: @antonymilne wrote this, but it definitely looks like `_outputs` is here to replace the `_output_component_property `.

vizro-core/src/vizro/models/_action/_action.py
antonymilne: Ah, is it because we set `suppress_callback_exceptions=True`?

vizro-core/src/vizro/models/_components/figure.py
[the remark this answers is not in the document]
antonymilne: Yes, it's meant to replace it 👍

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -32,13 +38,19 @@ class ControlsStates(TypedDict):
     filter_interaction: list[dict[str, State]]
 
 
+# TODO-AV2 D 3: try to enable properties that aren't Dash properties but are instead model fields e.g. header,
+# title. See https://github.com/mckinsey/vizro/issues/1078.
+# Try to fix AgGrid problem with underlying input component id.
+# Note this is needed for inputs in both vm.Action and _AbstractAction but outputs only in _AbstractAction.
```
antonymilne, line 44:
```suggestion
```

This is no longer relevant

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -141,14 +129,65 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
...
+        # Qn: should this work for legacy actions too? Think about docs.
+        runtime_args = {
+            arg_name: State(*self._transform(arg_value, type="input").split("."))
+            for arg_name, arg_value in self._runtime_args.items()
+        }
 
         return builtin_args | runtime_args
 
+    @staticmethod
+    def _transform(reference: str, type: Literal["output", "input"]) -> str:
```
antonymilne, line 146:
```suggestion
    def _transform_dependency(dependency: str, type: Literal["output", "input"]) -> str:
```

Input/Output/State all subclass `DashDependency`. I don't much like the name `dependency` but I think it's a bit clearer than `reference`.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -141,14 +129,65 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
...
+
+        if "." in reference:
+            # LQ: Check whether we need TypeAdapter here or something else. Currently it doesn't catch cases such as
+            # `component-id.prop.prop` well. However, adding below leads to a bunch of unit tests to fail. Check later.
+            # TypeAdapter(Union[list[_DotSeparatedStr], dict[str, _DotSeparatedStr]]).validate_python(reference)
+            component_id, component_property = reference.split(".")
+            if component_id in model_manager and hasattr(model_manager[component_id], property_name):
+                if component_property in getattr(model_manager[component_id], property_name):
+                    return getattr(model_manager[component_id], property_name)[component_property]
+            return reference
```
antonymilne, line 174: EAFP would make this clearer I think:

```suggestion
            try:
                return getattr(model_manager[component_id], property_name)[component_property]
            except KeyError, AttributeError:
                 # Some comment here explaining what this clause is for - could happen if component_id not in MM
                 # or intputs/outputs not defined or component_property not in that dictionary
                return dependency
```

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -141,14 +129,65 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
...
+                f"the explicit format '<component-id>.<property>'."
+            )
+
+        elif not hasattr(model_manager[component_id], property_name):
+            raise KeyError(
+                f"Component with ID '{component_id}' does not have implicit {type} properties defined. "
+                f"Please specify the {type} explicitly as '{component_id}.<property>'."
+            )
+
+        return getattr(model_manager[component_id], property_name)[component_property]
```
antonymilne, line 189: This doesn't cover the case that `model_manager[component_id].outputs` exists but `__default__` isn't an entry. I don't know if we have that case anywhere but maybe worth covering just in case we do in future? Up to you. Something like:
```
try: 
    getattr(model_manager[component_id], property_name)[component_property]
except AttributeError, KeyError:
    ...
```

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -83,6 +83,10 @@ class AgGrid(VizroBaseModel):
 
     _validate_figure = field_validator("figure", mode="before")(validate_captured_callable)
 
+    @property
+    def _outputs(self) -> dict[str, str]:
```
antonymilne, line 87:
```suggestion
    def _outputs(self) -> dict[str, _IdProperty]:
```

I think?

vizro-core/src/vizro/models/types.py
```diff
@@ -531,12 +544,16 @@ def wrapped(*args, **kwargs) -> CapturedCallable:
 
 # For "component_id.component_property", e.g. "dropdown_id.value".
 _IdProperty = NewType("_IdProperty", str)
+"""A string that must be in the format 'component-id.component-property'."""
```
antonymilne, line 547: Please do check that adding this docstring doesn't make it appear in the API docs as if it's public - previously I was doing `#` comments above the private members of this file.

vizro-core/src/vizro/models/types.py
```diff
@@ -531,12 +544,16 @@ def wrapped(*args, **kwargs) -> CapturedCallable:
...
 
 # Really this should be NewType and used for models like VizroBaseModel.id, but that clutters the code with casts and
 # means that to get user code to type-check successfully they would need to cast to ModelID.
 ModelID: TypeAlias = str
 """Represents a Vizro model ID."""
 
+# LQ: mypy complains if we use NewType here, as the type needs to be subclassable and the current Union does not work
+_IdOrIdProperty: TypeAlias = Union[ModelID, _IdProperty]
```
antonymilne, line 555: Let me check this. It's not a big worry though.

vizro-core/src/vizro/models/types.py
```diff
@@ -531,12 +544,16 @@ def wrapped(*args, **kwargs) -> CapturedCallable:
...
+"""Represents either a component ID or a string in the format 'component-id.component-property'."""
```
antonymilne, line 556:
```suggestion
"""Represents either a model ID or a string in the format 'component-id.component-property'."""
```

vizro-core/src/vizro/models/types.py
antonymilne: Is `_IdProperty` still used anywhere by itself or are they all `_IdOrIdProperty` now?

vizro-core/src/vizro/models/_components/form/_text_area.py
```diff
@@ -37,6 +37,14 @@ class TextArea(VizroBaseModel):
     # See comment thread: https://github.com/mckinsey/vizro/pull/298#discussion_r1478137654
     _set_actions = _action_validator_factory("value")
 
+    @property
+    def _outputs(self) -> dict[str, str]:
+        return {"__default__": f"{self.id}.value"}
+
+    @property
+    def _inputs(self) -> dict[str, str]:
+        return {"__default__": f"{self.id}.value"}
```
antonymilne, line 46: The current names feel a bit too vague to me so I would rename like this:

```suggestion
    @property
    def _action_outputs(self) -> dict[str, str]:
        return {"__default__": f"{self.id}.value"}

    @property
    def _action_inputs(self) -> dict[str, str]:
        return {"__default__": f"{self.id}.value"}
```

vizro-core/src/vizro/actions/_abstract_action.py
antonymilne, line 1: L81 here should change to whatever the new type hint for `Action.outputs` is.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -141,14 +129,65 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
         # have already been classified as legacy in Action._legacy. In future when vm.Action.inputs is deprecated
         # then this will be used for vm.Action instances also.
         TypeAdapter(dict[str, _DotSeparatedStr]).validate_python(self._runtime_args)
```
antonymilne, line 131: This should change to just `dict[str, str]` and the comment above should be updated.

vizro-core/src/vizro/models/_action/_action.py
antonymilne: By this point we should know `reference` is a string but nothing else about it. If that's right then it would be worth adding some comments in `transformed_outputs/inputs` to explain the different routes that could be taken to get there:
* `Action.outputs`: already validated by pydantic as `list[str]` or `dict[str, str]`
* `_AbstractAction._transformed_outputs`: manually does the same validation with `TypeAdapter`
* `runtime_args` for both `Action` and `_AbstractAction`: validated as `dict[str, str]` in `_transformed_inputs`

(we don't care about `Action: inputs` because it will be removed in future)

In this case what the check here needs to do isn't the whole `Union` thing, it's just validation of `reference`, which we could do just with a single regex or maybe a couple of `if`s - no need to use `TypeAdapter`. Basically we just need to check it's one of two valid forms:
* `id.property`: `^[^.]+[.][^.]+$`
* `id`: `^[^.]+$`

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -53,37 +65,13 @@ def _parameters(self) -> set[str]:
...
     @property
-    def _runtime_args(self) -> dict[str, _IdProperty]:
+    def _runtime_args(self) -> dict[str, _IdOrIdProperty]:
         raise NotImplementedError
 
     @property
     def _action_name(self) -> str:
         raise NotImplementedError
 
-    def _validate_dash_dependencies(self, /, dependencies, *, type: Literal["output", "input"]):
```
antonymilne, line 63: Make sure you don't forget this part:
```
When this change is made the outputs property for filter, parameter and on_page_load should just become `return self.targets` or similar. 
```

vizro-core/src/vizro/models/_action/_action.py
petar-qb: Yes it is! Great catch. When the `suppress_callback_exception` is omitted or set to False I get:
image

The `suppress_callback_exceptions` is preset to `True` to prevent console errors when `filter_interaction` is attached to the `Table` (for some reason it works without any issues when attached to the `AgGrid`).

So, I suggest to leave the `suppress_callback_exception=True` for now, and than get back to it when we deprecate the `vm.Table`. If you agree, could you add a comment in the `Vizro` constructor (or in the vm.Table) to remove the preset `suppress_callback_exceptions` when the Table become deprecated?

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -141,14 +129,65 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
...
+            type: "input" or "output" to check _inputs or _outputs
+
+        Returns:
+            Mapped property value or original reference for explicit format
+
+        Raises:
+            KeyError: If component not found or no default mapping exists
+            ValueError: If reference format is invalid e.g. "model-id.prop.prop"
+        """
+        property_name = "_outputs" if type == "output" else "_inputs"
```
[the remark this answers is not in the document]
petar-qb: Also, why do we have `type` as the `_transform` parameter? Couldn't it be `property_name`? In that case we would be able to skip the conversion in this line completely.

vizro-core/src/vizro/models/_action/_action.py
petar-qb: Great catch! It actually comes from the initialised `page_one_b` object that is, at the end, not added to the final `vm.Dashboard.pages`. If you add that page to the dashboard.pages, no errors should occur.

From top of my head, this shouldn't be too hard to fix. We just have to build the actions loop on top of the `registered pages`, and not all Page objects from the model manager. I think that's the trick. May you mention this unexpected behaviour in the TODO as well?

vizro-core/src/vizro/models/_components/ag_grid.py
huong-li-nguyen: Yes 👍 It does require a lot of `cast` statements below, then. Otherwise mypy complains. Let me implement it for now and then we can decide 👍

vizro-core/src/vizro/models/_components/figure.py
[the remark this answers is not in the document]
huong-li-nguyen: I've implemented that now, but just a heads-up — as part of this, I also introduced `_action_outputs` for the `Filter`, which wasn’t originally planned (see Table). The `_Filter` previously had the `_output_component_property` though.

vizro-core/src/vizro/models/types.py
huong-li-nguyen: It doesn't appear in the public API btw, even if it has a docstring 👍 I saw your question somewhere before

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: Yes, I actually considered adding that! But I figured this situation would never occur 🤔 Why would we define `_action_outputs` or `_action_inputs` without also setting a `__default__`? As soon as there's at least one item, that item could just be the `__default__`. That’s why I left it out, but I can definitely add it as a safeguard in case we forget to define the `__default__`, if you think it’s worth covering that edge case?

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -92,12 +92,12 @@ class Filter(VizroBaseModel):
     selector: Optional[SelectorType] = None
 
     _dynamic: bool = PrivateAttr(False)
-
-    # Component properties for actions and interactions
-    _output_component_property: str = PrivateAttr("children")
-
     _column_type: Literal["numerical", "categorical", "temporal"] = PrivateAttr()
 
+    @property
```
huong-li-nguyen, line 97: See comment: https://github.com/mckinsey/vizro/pull/1153#discussion_r2081175475

vizro-core/src/vizro/actions/_filter_action.py
```diff
@@ -38,12 +38,6 @@ def function(self, _controls: _Controls) -> dict[ModelID, Any]:
...
-
-        for target in self.targets:
-            component_id = target
-            component_property = cast(FigureType, model_manager[target])._output_component_property
-            outputs[target] = f"{component_id}.{component_property}"
-
-        return outputs
+        return {
+            target: cast(FigureType, model_manager[target])._action_outputs["__default__"] for target in self.targets
+        }
```
huong-li-nguyen, line 43: I didn't fully understand this:

```
When this change is made the outputs property for filter, parameter and on_page_load should just
become `return self.targets` or similar.
```

Is this what you were expecting? Or should this really just work with `self.targets`? Is this always supposed to return a dict by the way?

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: Closing here, so we continue discussion here: https://github.com/mckinsey/vizro/pull/1153#discussion_r2081237566

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: I think this is because it should match the format used in `Action.inputs`. Since we replaced `DotSeparatedStr` with a plain `str` there, it needs to change here to keep the validations consistent. The same is true for `_AbstractAction`

huong-li-nguyen: See: https://github.com/mckinsey/vizro/pull/1153#discussion_r2079439183

vizro-core/src/vizro/models/types.py
huong-li-nguyen: 1. It's not appearing in the public api 👍 
2. _IdProperty is still being used by itself
3. _DotSeparatedStr will not be used anywhere now after the recent changes.

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: Antony's comment here was also helpful for me to understand above: https://github.com/mckinsey/vizro/pull/1153#discussion_r2079452763

vizro-core/src/vizro/actions/_abstract_action.py
```diff
@@ -9,7 +9,7 @@
 from pydantic import TypeAdapter
 
 from vizro.models._action._action import _BaseAction
-from vizro.models.types import _DotSeparatedStr, _IdProperty
+from vizro.models.types import _DotSeparatedStr, _IdOrIdProperty
```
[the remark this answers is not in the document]
huong-li-nguyen: It's not used anywhere anymore, but it also wasn't replaced with `_IdProperty` but just with `str` 👍

vizro-core/src/vizro/models/_action/_action.py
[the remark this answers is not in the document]
huong-li-nguyen: I can rename it to `property_type` if that makes things clearer?

I'm not entirely sure I understand your second question. We need `_transform_dependency` to work for both inputs and outputs, so the function needs to know whether to access `_action_outputs` or `_action_inputs` as an attribute. It’s essentially the same pattern we used before in the `_validate_dependencies` function, where we also passed in the `type` to distinguish between them. 🤔

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: I added it now 👍

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: I've kept the try/except structure for both cases (dot-notation and model id only ), but had to add more logic inside, so do take a look if it's easier to read now 👍

vizro-core/src/vizro/models/types.py
```diff
@@ -630,14 +646,6 @@ class OptionsDictType(TypedDict):
...
 FigureType = Union["Graph", "Table", "AgGrid", "Figure"]
 
-# TODO-AV2 D 3: think about how to make this public and use in our inbuilt actions. Compare to _IdProperty.
-# Consider how it works if you just specify model name and not a dot separated string. Do we need a new composite type
-# for Union[list[_DotSeparatedStr], dict[str, _DotSeparatedStr]] too? Consider parameter target form too and whether
-# it should work similarly.
-_DotSeparatedStr = Annotated[str, StringConstraints(pattern="^[^.]+[.][^.]+$")]
-"""A string that must contain exactly one dot ('.'), with at least one character on both sides.
-For example: 'model-id.children'."""
-
```
huong-li-nguyen, line 640: Not sure if this is already counting as solving for D3 😄 We don't use `_DotSeparatedStr` anywhere anymore, so I've removed it for now. If we do want to enable this publicly, maybe we can take this on in a separate PR. But for now I would just remove it.

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -356,30 +412,59 @@ def test_outputs_valid(self, outputs, expected_transformed_outputs):
...
+        vm.Page(title="Page Title", components=[button])
+        Vizro._pre_build()
+
+        with pytest.raises(
+            AttributeError,
+            match="Component with ID 'test_button' does not have implicit output properties defined. "
+            "Please specify the output explicitly as 'test_button.<property>'.",
+        ):
+            action = Action(function=action_with_no_args(), outputs=["test_button"])
+            action._transformed_outputs
```
huong-li-nguyen, line 467: The tests are largely unchanged—I've just split up the examples to account for the fact that we're now returning different types of errors (like ValidationError, KeyError, and AttributeError) that need to be checked separately. Regarding the unit tests, I'm still a bit unsure about the level of coverage. For instance, is the current depth sufficient, or should we also be testing more interactions between registered models and actions?

vizro-core/src/vizro/models/_action/_action.py
petar-qb: Resolving this as the discussion will be continued in another thread.

vizro-core/src/vizro/actions/_filter_action.py
petar-qb: > Is this always supposed to return a dict by the way?

We always work with dict-like outputs for builtin actions because then we don't have to care about the order of outputs we return when the action is triggered. 

> Is this what you were expecting?

As we now attach the `__default__` action output if the ModelID is propagated as the dictionary value, then the following should also work: 

```suggestion
        return {target: target for target in self.targets}
```

Let's apply this change to all builtin actions where it suits:
- filter_action
- parameter_action
- on_page_load
- filter_interaction

vizro-core/tests/unit/vizro/models/_components/form/test_dropdown.py
```diff
@@ -25,6 +25,8 @@ def test_create_dropdown_mandatory_only(self):
         assert dropdown.title == ""
         assert dropdown.description is None
         assert dropdown.actions == []
+        assert dropdown._action_outputs == {"__default__": f"{dropdown.id}.value"}
+        assert dropdown._action_inputs == {"__default__": f"{dropdown.id}.value"}
```
petar-qb, line 29: I'm not sure private properties have to be tested too.

vizro-core/tests/unit/vizro/actions/test_abstract_action.py
petar-qb, line 1: What the status for the tests? Do we still want to push them in a separate PR?

petar-qb: Also, what's the status/plan for https://github.com/mckinsey/vizro/pull/1148?

vizro-core/src/vizro/models/_action/_action.py
petar-qb, line 1: There is the following line within the `_get_control_states` method:
`State(component_id=control.selector.id, component_property=control.selector._input_property)`

As we currently have `_action_inputs` defined, I think we can adjust this line and also get rid of all `_input_property` in models.

petar-qb, line 1: I'm writing this just so I don't forget.

We are replacing `_input_property` and `_output_default_property` with the `_action_inputs` and `_action_outputs` dictionaries which is great! 

Question: Does it make sense to re-use the similar approach for the `trigger_property` used to trigger the action chain?

Currently, that information is written/hardcoded here:
```py
    actions: Annotated[
        list[ActionType],
        AfterValidator(_action_validator_factory("value")),  # <--- HERE
        PlainSerializer(lambda x: x[0].actions),
        Field(default=[]),
    ]
```

If we extract this (`"value"` from the example above) inside the `def _action_trigger(...) ...`, then users would be able to overwrite it more easily. 
I know this shouldn't be the part of this PR, but I added it here so I don't forget. 😄 

What do you think @huong-li-nguyen @antonymilne?

petar-qb (commented):
This PR is very very close to being merged 🎉 🎉

vizro-core/src/vizro/models/_action/_action.py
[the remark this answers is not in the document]
huong-li-nguyen: I've renamed to `attribute_type`

vizro-core/src/vizro/actions/_filter_action.py
huong-li-nguyen: Ah I see - Yes, that works! Thanks!!

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -135,20 +125,90 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
...
-        runtime_args = {arg_name: State(*arg_value.split(".")) for arg_name, arg_value in self._runtime_args.items()}
+        # Qn: should this work for legacy actions too? Think about docs.
+        runtime_args = {
+            arg_name: State(*self._transform_dependency(arg_value, type="input").split("."))
+            for arg_name, arg_value in self._runtime_args.items()
+        }
 
         return builtin_args | runtime_args
 
+    @staticmethod
```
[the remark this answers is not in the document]
huong-li-nguyen: Yes, sure :)

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -135,20 +125,90 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
...
+        # vm.Action that does not pass this check will have already been classified as legacy in Action._legacy.
+        # In future when vm.Action.inputs is deprecated then this will be used for vm.Action instances also.
+        TypeAdapter(dict[str, str]).validate_python(self._runtime_args)
+        # Do exactly same lookup but we only have "__default__" defined in inputs dictionary so will only ever use that
+        # case
         # User specified arguments runtime_args take precedence over built in reserved arguments. No static arguments
         # ar relevant here, just Dash States. Static arguments values are stored in the state of the relevant
         # _AbstractAction instance.
-        runtime_args = {arg_name: State(*arg_value.split(".")) for arg_name, arg_value in self._runtime_args.items()}
+        # Qn: should this work for legacy actions too? Think about docs.
```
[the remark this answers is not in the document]
huong-li-nguyen: I think @antonymilne added that in. At least I don't remember adding that 😄 And I also don't know what it means.

vizro-core/tests/unit/vizro/models/_components/form/test_dropdown.py
huong-li-nguyen: I think it should be in this case, since we don’t want these values to change unless we actively decide to

vizro-core/tests/unit/vizro/actions/test_abstract_action.py
huong-li-nguyen: Depends on what we decide on here: https://github.com/mckinsey/vizro/pull/1153#discussion_r2081741095 :)

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -118,7 +105,10 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
         """
         if self._legacy:
             # Must be an Action rather than _AbstractAction, so has already been validated by pydantic field annotation.
-            return [State(*input.split(".")) for input in cast(Action, self).inputs]
+            return [
+                State(*self._transform_dependency(input, type="input").split("."))
+                for input in cast(Action, self).inputs
```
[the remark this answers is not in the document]
huong-li-nguyen: You tell me haha It was like that before, but it seems like I can just replace it with `self.inputs` and remove the code comment above, then, since we don't have the validation inside the pydantic field anymore.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -32,13 +38,18 @@ class ControlsStates(TypedDict):
...
+
+
 class _BaseAction(VizroBaseModel):
     # The common interface shared between Action and _AbstractAction all raise NotImplementedError or are ClassVar.
     # This mypy type-check this class.
     # function and outputs are overridden as fields in Action and abstract methods in _AbstractAction. Using ClassVar
     # for these is the easiest way to appease mypy and have something that actually works at runtime.
     function: ClassVar[Callable[..., Any]]
-    outputs: ClassVar[Union[list[_IdProperty], dict[str, _IdProperty]]]
+    outputs: ClassVar[Union[list[_IdOrIdProperty], dict[str, _IdOrIdProperty]]]
```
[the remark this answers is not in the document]
huong-li-nguyen: I think you are right! 🤔 I'll change it to str and to make it consistent with Action.outputs

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: Ah yes, I didn't see that one! Thanks!

huong-li-nguyen: I don’t have a strong opinion on this, but if it works, that’s great 👍  From what I understood, you were planning to change the trigger logic anyway, right? And the inputs aren’t supposed to be visible anymore? So anything that fits into that plan works for me 👍

vizro-core/tests/e2e/vizro/custom_components/custom_components/dropdown.py
```diff
@@ -1,55 +1,21 @@
...
 MultiOptionType = Union[list[bool], list[float], list[str]]
 
 
-class CustomDropdown(VizroBaseModel):
-    """Categorical single/multi-selector `Dropdown` to be provided to `Filter`."""
+class CustomDropdown(vm.Dropdown):
+    """Custom Dropdown that has multi=False as default."""
 
     type: Literal["custom-dropdown"] = "custom-dropdown"
-    options: Optional[MultiOptionType] = Field(default=None, description="Possible options the user can select from")
```
huong-li-nguyen, line 20: @l0uden - I just simplified this code. From what I saw in the configuration, you just wanted to use a custom component selector. So I think in this case, we can just inherit from an existing selector and only change its default to `multi=False` for example.

vizro-core/src/vizro/models/types.py
petar-qb: To close D 3, at least two more things need to be done:
- https://github.com/McK-Internal/vizro-internal/issues/1610
- https://github.com/McK-Internal/vizro-internal/issues/1710

vizro-core/src/vizro/models/_action/_action.py
petar-qb: I like the way you changed it 👍

vizro-core/tests/unit/vizro/models/_action/test_action.py
petar-qb: I like the way you separated `invalid` inputs/outputs tests. However, I think a couple of more use-cases for `valid` inputs/outputs should be added because our `action.inputs/outputs` API is now expanded.

Do you also think we should add tests (for Action and AbstractAction) for valid inputs/outputs if only a `ModelId` is propagated?

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: As we agreed to test the content of the `_action_outputs/inputs` property, could you test it for the filter as well?

vizro-core/tests/unit/vizro/models/_action/test_action.py
huong-li-nguyen: Ah, sorry for the confusion — I should've explained better. I didn't add more valid inputs/outputs because they no longer make sense with our current setup (since there's no interaction with the model manager). So most of the previously valid inputs/outputs are now invalid, as they're not registered models.

My actual question was: do we want to extend the tests to include checks for valid model IDs (as you also suggested)? To do that, I'd need to pre-register models first. And then, how thorough should those tests be? Should I test every model that has `_action_outputs` defined, or is it enough to just test one valid model ID?

petar-qb: I don't believe every model with `_action_outputs` defined needs to be tested individually. Since `_action_outputs` are already tested across all models (which I'm actually glad we're keeping 😄), we can focus on testing just one model within the `model_manager`. Specifically, we can test how it behaves when used as a `ModelId` in Action/AbstractAction, with (inputs)/outputs as either a `list` or a `dictionary`. Something like that..

vizro-core/src/vizro/models/types.py
huong-li-nguyen: Ah, but I thought D3 will be done by you in a separate PR after this is merged?

vizro-core/tests/unit/vizro/models/_action/test_action.py
antonymilne: Fully agree with @petar-qb. Let's have new test cases for:
* `action = Action(function=action_with_one_arg("model_id"))` and then `action._transformed_inputs == ...`
* not yet, but to come with D 3: `Action(function=action_with_one_arg("model_id.argument_name"))`

+ same sort of thing for outputs
+ same thing for `AbstractAction`

I don't mind whether they're new cases in an existing parameterised test or whole new tests, whatever makes more sense.

Ideally do this by defining a new `CustomModel` with the required `_action_outputs/inputs` field and make sure it's added to `model_manager` instead of using a "real" model from `vm`, but if that's too cumbersome then fine to use a real model.

vizro-core/tests/unit/vizro/actions/test_abstract_action.py
```diff
@@ -168,19 +170,41 @@ def test_inputs_valid(self, action_class, inputs, expected_transformed_inputs):
...
+            action_with_one_runtime_arg(arg_1=input)._transformed_inputs
+
+    @pytest.mark.parametrize(
+        "input",
+        [
             "component",
+            "component_property",
+        ],
+    )
+    def test_inputs_invalid_model_id(self, input):
```
antonymilne, line 186: See comments in `test_action.py` 🙂

antonymilne (commented):
Generally this looks great! I can't approve my own PR but would do so if I could 😀 

Anything marked with "Come back to this" is for me to look at again, not for you to do anything about yet. I need to check some of the open conversations still too but basically we're nearly there 👍 

@petar-qb one thing I didn't check is that this change was implemented for all the components inputs/outputs that we put in the table, so please do make sure your review covers that 🙏

vizro-core/src/vizro/models/_action/_action.py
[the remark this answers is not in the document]
antonymilne: This was a very temporary rough note to myself, definitely should be removed.

vizro-core/src/vizro/models/_action/_action.py
antonymilne: Yeah, this stuff will very likely change indeed. You're right that depending on what it ends up it might be improved by isolating  the `"value"` part in a new `_trigger_property` or similar to make it cleaner.

Let's see where #1163 ends up since it completely changes the `set_actions` validator and so might rewrite this too in due course.

vizro-core/src/vizro/models/_components/figure.py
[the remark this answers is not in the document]
antonymilne: Cool, thanks for noting it!

@petar-qb I presume the reasons for this already existing is dynamic filters?

We'll need to think about how to handle this since ideally in future we'd be able to do target `filter_id` (e.g. in `interact`) and have it automatically update the underlying `selector_id.value`, which would mean something like `{"__default__": f"{self.selector.id}.value"}` instead.

vizro-core/changelog.d/20250507_135806_huong_li_nguyen_actions_shortcut_dash_properties.md
```diff
@@ -0,0 +1,48 @@
...
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
```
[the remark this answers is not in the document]
antonymilne: I think let's leave this out of the changelog until we've figured out what we're doing with docs. Otherwise it's announced in the changelog but not explained anywhere which isn't great.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -118,7 +169,10 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
         """
         if self._legacy:
             # Must be an Action rather than _AbstractAction, so has already been validated by pydantic field annotation.
-            return [State(*input.split(".")) for input in cast(Action, self).inputs]
+            return [
+                State(*self._transform_dependency(input, type="input").split("."))
+                for input in cast(Action, self).inputs
+            ]
```
antonymilne, line 175: Unless you have strong feelings otherwise, let's revert this so that the new behaviour is only possible for non-legacy actions.

vizro-core/changelog.d/20250507_135806_huong_li_nguyen_actions_shortcut_dash_properties.md
huong-li-nguyen: I was planning to update the docs in a separate PR, but I'll stop that then 👍

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -264,6 +292,8 @@ class TestActionInputs:
         [
             (action_with_no_args, {}, {}),
             (action_with_one_arg, {"arg_1": "component.property"}, {"arg_1": State("component", "property")}),
+            # LQ: Check why this doesn't pass?
+            # (action_with_one_arg, {"arg_1": "custom-model-id"}, {"arg_1": State("custom-model-id", "children")}),
```
huong-li-nguyen, line 296: @petar-qb - qq - I would have expected this to work, but it didn't. Can I not reuse the `action_with_one_arg` function? or do I have to do this differently? 🤔

huong-li-nguyen (approved):

vizro-core/src/vizro/models/_action/_action.py
antonymilne: ok, let's keep it in then (both the transform and using the transformed values). I was just thinking it might be clearer to make a clean split between legacy actions where no new features are possible vs. new actions where all the new features are possible. But here it's easy to keep the new feature in and you get the benefit of validation which we'd be losing otherwise for the legacy actions. So leave as you had it.

vizro-core/tests/unit/vizro/models/_action/test_action.py
petar-qb: This one looks like done.

petar-qb:
> petar-qb one thing I didn't check is that this change was implemented for all the components inputs/outputs that we put in the table, so please do make sure your review covers that 🙏

Done ✅ 

---

@huong-li-nguyen @antonymilne 
Seeing some unit-test failing (`tests/unit/vizro/models/test_models_utils.py::TestSharedValidators::test_check_captured_callable`).

Here's the exception that's raised:
```
E           pydantic_core._pydantic_core.ValidationError: 1 validation error for Page
E           components.0
E             Unable to extract tag using discriminator 'type' [type=union_tag_not_found, input_value=vizro.tables._dash_ag_gri...[1704 rows x 8 columns]), input_type=CapturedCallable]
E               For further information visit https://errors.pydantic.dev/2.11/v/union_tag_not_found
```

but this one is expected:

```
match="A callable of mode `ag_grid` has been provided. Please wrap it inside `vm.AgGrid(figure=...)`",
```

---

FYI @huong-li-nguyen: I'm investigating this one -> https://github.com/mckinsey/vizro/pull/1153#discussion_r2086732662

vizro-core/tests/unit/vizro/models/_action/test_action.py
huong-li-nguyen: I discussed with @antonymilne using an inbuilt model instead of a custom model; otherwise, the tests won't get cleaned up nicely. So I'll push these changes in a sec.

huong-li-nguyen:
> match="A callable of mode `ag_grid` has been provided. Please wrap it inside `vm.AgGrid(figure=...)`",
> ```

It's because I used a CustomModel and used the `add_type` method :) The tests don't get cleaned up and it's difficult to do so. I just discussed with @antonymilne that we don't use a CustomModel but an inbuilt one instead.

vizro-core/tests/unit/vizro/models/_action/test_action.py
huong-li-nguyen: I believe I’ve fixed it here – could you take a quick look to confirm? :) https://github.com/mckinsey/vizro/pull/1153/commits/4c4b55ffb330c5ad0e4b286ff1c6ce1d26761e9e

I had to adjust the logic for detecting legacy actions. Without the change, it was incorrectly labeling the action as "legacy" even though the test was actually running for a non-legacy action.

vizro-core/src/vizro/models/_components/figure.py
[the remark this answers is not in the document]
antonymilne: Sounds good but can we relabel this as `TODO-AV2 E`? And then we'll deal with it later when we do `interact`.

vizro-core/tests/unit/vizro/models/_action/test_action.py
```diff
@@ -65,35 +65,55 @@ class TestLegacyActionInputs:
...
+        "runtime_inputs",
+        [
+            ["unknown_model_id"],
+        ],
+    )
+    def test_action_inputs_invalid_model_id(self, runtime_inputs):
+        with pytest.raises(KeyError, match="Model with ID .* not found. Please provide a valid component ID."):
+            action = Action(function=action_with_one_arg(), inputs=runtime_inputs)
+            action._transformed_inputs
+
```
petar-qb, line 99: Should we add `test_inputs_invalid_missing_action_attribute` here,  in the `TestActionInputs`, and in the `TestAbstractActionInputs` to align with the outputs tests?

vizro-core/tests/unit/vizro/models/_action/test_action.py
petar-qb: Great spot! We should probably add two new test cases within the `TestIsActionLegacy::test_mixed_runtime_and_runtime_inputs` like: 
- (action_with_one_arg, {}, ["known-model-id"], True),
- (action_with_one_arg, {"arg_1": "known-model-id"}, [], False),

petar-qb (approved):
Excellent work, @huong-li-nguyen and @antonymilne! 🚀⚡ 

After a thorough review, I've identified just a couple of missing test cases. Once those are added, this PR will be good to go! ✅

Issue #1140, opened 2025-04-25
Multiple `AgGrid` with the same `id` makes all pages load as blank


l0uden:
### Which package?

vizro

### Package version

0.1.38

### Description

There should be validation before dashboard starts on the same `id` for `AgGrid`.
Now it allow start the dashboard, but all pages will be blank.
Also there's error in the browser console:
<img width="940" alt="Image" src="https://github.com/user-attachments/assets/54b05af0-dbdb-4cbd-b253-f2e51e52f412" />

### How to Reproduce

[...]

Issue #1139, opened 2025-04-25
Multiple `Table` can be created with same `id`


l0uden:
### Which package?

vizro

### Package version

0.1.38

### Description

If we create two `Table` with the same `id` there should be error on start of the dashboard

### How to Reproduce

[PyCafe link](https://py.cafe/snippet/vizro/v1?pycafe-app-view#c=H4sIAAaVC2gEA11S0W7bIBT9lSv6YkspTrpqD5E8KVPbvURbH7JpUx1ZxJAYBQMF7Mat-u-74HTq6hcD95x7zj3wQhrDBVmSC_gln50B6YFpMFboS2961wgIxqijDLA3DhonWJD6AJ3hvWIOOAsMBul7puQzlowGZq2STVp7WukLaFrRHMH0AdoQrF8WxUGGtt_RxnRF1xyl9mIshiQfRTrjBEi9N8B2kZWMxUZM86nZ-16JR9EXD63gpvFUmkLowge2U6JIHfG474QOyRR6qrTsrHEBnUeyVSaokYqTdcLj_B7sqdJ7Z7oJAGf02ch_XAxCqMQZuvccmuT9G5Uz39YxrDqdRwtBWg8lStF4TuM2y2PBsoPAwtDRe1xllQb8ggxKlBVZr_78-Lmp79a3v-tvq_t69f2m3qy-rm8rMpuQio2YT4n0OyVO2YFZpF3PcSSSnyEYvDUaA_HlA-I20VK2l4feifKD0yz6moHk2OTxSbgwohD4MCoxAcqXijxJDL8iS6jIYj5PUq95nqKXeJPgmMZBPuVb1E8jRpGdYY5Pc968bbM4O5qKvy0iU-JZTne9VDz7x8qp63UMi8yIE4-9dCJer8d3nML_Us7pgl5dJ0AYbXzgqYBbO4bWaDywo-GSi8thTq8-0wWWpuTIUvdKzche4v2R5cP29S-2T6e7JQMAAA) to reproduce

### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

PR #1159, opened 2025-05-08
[Fix] Fix silent duplicate ID bug of Table and AgGrid

maxschulz-COL:
## Description
Closes https://github.com/mckinsey/vizro/issues/1140 and https://github.com/mckinsey/vizro/issues/1139
## Notice


vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -139,6 +140,21 @@ def _filter_interaction(
     @_log_call
     def pre_build(self):
         self._input_component_id = self.figure._arguments.get("id", f"__input_{self.id}")
+        # Check if any other AgGrid figure function has the same input component ID
+        existing_models = [
+            model
+            for model in model_manager._get_models(vm.AgGrid)
+            if hasattr(model, "_input_component_id")
```
antonymilne, line 147: What is this check for? Won't it always be true?

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -139,6 +140,21 @@ def _filter_interaction(
     @_log_call
     def pre_build(self):
         self._input_component_id = self.figure._arguments.get("id", f"__input_{self.id}")
+        # Check if any other AgGrid figure function has the same input component ID
+        existing_models = [
+            model
+            for model in model_manager._get_models(vm.AgGrid)
```
antonymilne, line 146: Do we need to check for clashes between `vm.AgGrid` and `vm.Table`? Or cover `vm.Figure` too? e.g. by doing `_get_models(FIGURE_MODELS)`.

maxschulz-COL: Hm, of the clashes between Grid and Table I didn't think of tbh. For the Figure models, I don't think we should be doing this test. Components can be arbitrarily nested etc, so I don't think it is up to us to cover that case.

As for the tables and aggrid, it [actually works](https://py.cafe/snippet/vizro/v1?pycafe-app-view#c=H4sIANCuHGgEA41UUW_TMBD-K5b30kqZ024MUKWCYMAm8QBCEzwsVeUm18SaY3u20zWb-t85O2m7TkwlfWjOd_d9332280RzXQCd0BNyQn6LR6uJcIQrog2oU6cbmwPxWss74clSW5Jb4F6oktS6aCS3pOCek5VwDZfiEVNaEW6MFHl8dyxTATqvIL8juvGk8t64SZqWwlfNguW6Tuv8TigHbbqKAgJNrS0QoZaa8EXoitI6KK6KDu45WuxkqK3wFRQ6d0zoFFTqPF9ISCMmLjc1KB-Foa5Midpo61F9aMaBQOLsjqzqFykjtZctg7Wx4GKJWWdqaXXdFZC-ule5T7BI77b5grtqzst5aUWRdFGwbx6rgiAvjCNTRGdhnYVwMAyJTBlewjnmVjX7ia-DTBF8vPASphnNrXZuLoqMJl1C8hb9mWL1NwnrQckNVr0ZoW467EvQeqMVGuKmt91KeLDjJsgZLEXZWJi-UDkImhIiCoS7fwDrkZE430ro8tOnjD4I3IWMTkhGx6NR5NwMt6zhQY5P5RWacEDSG_M6wwF2QN6jzvA_GhWQFprborPqyzYcBP9w0GjjrPc07tdgyBaNkMVg1zpktlGd7ycn5OYHubz-evl9Qrh03ensVIeTmGJ0ZbmpPmaKJtTCfSMshFPm8FLFM_BhOmJjdv4-FvjWhNsWExia1lda4YJpdSEKOF2N2NlbNsZUt4F08kRXYB2eWDo5QwKt_S_9EJZ7KItRQvMKJ7CARbe7DO6HA4_J6BmdjC9GCa2F-tOF5110DaKskCeEosC2pZDwGVEd2EuNt0UosK8whNLTRVeLJYYHXEo3s02yq9mp-BcAJvd9-NlgpqWHvfv0c2-ZX_tDGnvchuOj96McG3s78YGAHU3Vo47fXfwPpweLy1weI93WBdbw2yTRfjxmt7PNX5JvLk7HBQAA), do you still think we should rule that out?

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -142,7 +142,7 @@ def pre_build(self):
         # Check if any other AgGrid figure function has the same input component ID
         existing_models = [
             model
-            for model in model_manager._get_models(AgGrid)
+            for model in model_manager._get_models(self.__class__)
```
petar-qb, line 146: Let's import the model_manager

petar-qb (approved):
The current implementation works for the case where AgGrid and Table share the same ID. Therefore, I believe this PR is ready for merge.

However, if we decide to forbid ID sharing between any two components in Vizro, we should refactor this logic. Specifically in that case, we should validate this restriction in the centralised way (`vm.Dashboard.pre_build`, `model_manager`, or even in the `vm.Dashboard.build`), although that may be overkill at this stage.

vizro-core/src/vizro/models/_components/ag_grid.py
maxschulz-COL: Decided we do not act upon this

antonymilne: As discussed, I think this is fine how you have it ignoring `Figure` models and not worrying about the AgGrid vs. Table clash - if it works ok then all good.

petar-qb:
What happens if the `AgGrid` and its callable have the same ID? Should we forbid that too? 

```py
AgGrid(id="my-id", figure=dash_ag_grid(df, id="my-id"))
```

PR #1160, opened 2025-05-08
Remove dash<3.0.0 upper-bound

petar-qb:
## Description
- Set `dash>3.0.0` lower bound in the `vizro-core/pyproject.toml`
- Avoid facing browser console errors (`RangeError: Invalid array length`) by introducing some changes in the `build_action_loop_callbacks.js` and other vizro js files.
- Setting (in `vizro-core/pyproject.toml` and in the `vizro-core/hatch.toml` for the `lower-bounds` env)
  -  `"dash_bootstrap_components>=2"`, See https://github.com/facultyai/dash-bootstrap-components/releases/tag/2.0.0
  -  `"dash-ag-grid>=31.3.1"`,  See https://github.com/plotly/dash-ag-grid/releases/tag/v31.3.1
  -  `"dash_mantine_components>=1"`,  See https://github.com/snehilvj/dash-mantine-components/releases/tag/1.0.0

to avoid facing browser console warnings (`Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.`). 

## Notice


vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js
```diff
@@ -25,7 +25,10 @@ function gateway(
   ) {
     // If there's no more actions to execute, stop the loop perform.
     if (remaining_actions.length == 0) {
-      return dash_clientside.PreventUpdate;
+      // Return list of dash.no_update for all outputs to circumvent the JS "RangeError" exception for the dash==3.0.4
```
antonymilne, line 28: I think we can do `throw dash_clientside.PreventUpdate` like Python's `raise PreventUpdate` rather than all these `return dash.no_update`?

vizro-core/pyproject.toml
```diff
@@ -15,13 +15,13 @@ classifiers = [
   "Programming Language :: Python :: 3.13"
 ]
 dependencies = [
-  "dash>=2.18.0,<3",  # 2.18.0 needed as the 'id' attribute is exposed for dcc.Loading
-  "dash_bootstrap_components",
-  "dash-ag-grid>=31.0.0",
+  "dash>=2.18.0",  # 2.18.0 needed as the 'id' attribute is exposed for dcc.Loading
```
antonymilne, line 18: Looking at the lower bound tests I think we're going to need to bump to `dash>=3.0.0`, which is fine by me.

We should add all these new lower bounds to the `lower-bounds` env in hatch.toml too.

vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js
petar-qb: You're right! All errors disappear when `return dash_clientside.PreventUpdate` is replaced with the `throw dash_clientside.PreventUpdate`

vizro-core/tests/e2e/vizro/test_dom_elements/test_datepicker.py
petar-qb, line 1: Hey @l0uden 👋

This test was failing at this line: `check_graph_is_loading(dash_br, cnst.BAR_POP_RANGE_ID)`. To temporarily bypass this, I replaced it with `sleep(1)`. Unfortunately, I couldn't debug it further due to the lack of `chromedriver` on my machine (as I mentioned to you privately).

Could you please assist investigate this? It would be great to replace the current `sleep(1)` workaround with a more reliable, deterministic solution. Let me know when you're available and we can solve it online together.

Thank you!

vizro-core/pyproject.toml
petar-qb: I bumped it to the `dash>=3.0.0` along with other libraries 👍.

Does it look good now?

antonymilne: Yeah, looks good! 

Just remember the point of doing this change was to get the bump to Dash 3 working in advance so that once the Dash persistence bug fix is released we can bump to the fixed version of dash. So we should either wait until that fix is released with Dash 3.0.5 and change to `dash>=3.0.5` before we merge this PR or merge this PR and then do another one when 3.0.5 is released to bump the lower bound again. Either one is good for me, up to you.

petar-qb: I think it's easier to merge this one now and then bump only `dash >= 3.0.5` in a separate PR. 
P.S.I really don't have idea when they plan to release the bugfix.

l0uden (approved):

PR #1163, opened 2025-05-08
[PoC] Simplify action loop

antonymilne:
Closes: https://github.com/McK-Internal/vizro-internal/issues/1613 

## Description

Very rough initial PoC to show an idea to @petar-qb that we discussed before.

I've tried removing the action loop completely and replacing with just logic that allows you to chain previously unlinked actions together, i.e. `actions = [action_1, action_2]` still behaves like before. We do this without an `ActionsChain` model.

* different from before: you can now trigger the initial action in an actions chain from a pure Dash callback, which will then trigger the remaining actions in that chain as per normal Dash callbacks. This will be useful for the `interact` functionality.
* same as before: actions _after_ the initial one in an actions chain still can't be triggered outside the action chain

The big advantage is this is much simpler than the old system and clearer about how actions interact with Dash callbacks.

There's still some stuff to be worked out (see the TODOs) but I'm curious to see what, if anything, breaks by making this change... So let's see the tests do.

## Notice


petar-qb:
Here's the acceptance criteria (and some open questions) for the actions loop system (that I can currently remember). This sounds like a good starting point for this development and as soon as the solution satisfies it (reds and yellows), I'm okay with the approach (the easier the better).

OQ - Open Question
🔴  - Hard requirements, 
🟡 - Soft requirements, 
🟢  - Nice to have requirements.

1. 🔴 (initial-page-build) When a page is opened, the first http request sent to the server responds with the lightweight answer (build methods that make objects that do not depend on the data)
2. 🔴 (on-page-load) After the "initial-page-build", if there's a need (if there are page components that are based on the data), the second (or more) request is sent to the server.
3. 🔴 By default: no other callbacks are triggered after the "initial-page-build" / "initial-page-build" + "on-page-build".
Example: This means that export_data should not be triggered if attached to the button component when the page is opened.
4. 🟡  Enable that users can explicitly set that some callbacks (one or more) are triggered immediately after the "initial-page-build".
This was planned to be achieved by exposing page.actions and on-page-load ("update_figure_action"). In that case users would be able to apply any custom logic before or after the on-page-load.
5. 🔴 Enable that actions could be define in a list ("actions_chain") where the order of the execution is guaranteed.
  - OQ: Should we enable that "action" within the "actions_chain" or the entire "actions_chain" can be triggered from the another callback / action? In the current system, the answer is "No".
  - OQ: Should we enable that the "action" from the "actions_chain" can trigger another callback / action? In the current system, the answer is "No".
  - OQ: Should we enable parallel "actions_chain" executions?
  - OQ: Enable or disable an "actions_chain" to be triggered again before the previous "actions_chain" finished if the UI component is clicked multiple times.
6. 🟢 Enable to trigger multiple "actions_chains" from a single component by modifying different component properties.
7. 🟢 Enable dash callback features like: Patch, background=True etc.
8. 🟢 Enable that actions can be defined once and the easily reused in multiple places.

nadijagraca:
@antonymilne  FYI: Already discussed this with Petar and he is aware of this bug. 
While working on vizro_download and vizro_url PR, I encountered a bug. 
`dcc.Store(id=f"{ON_PAGE_LOAD_ACTION_PREFIX}_trigger_{self.id}")` needs to stay inside the `components_container` as moving it to the outside container prevents on_page_load action from triggering. 

dcc.Store inside components_container
Screenshot 2025-07-02 at 15 54 45

dcc.Store outside of components_container (as you can see here on_page_load_action was not triggered)
Screenshot 2025-07-02 at 15 54 01

petar-qb:
Just to mention that the bug @nadijagraca explained happens in the main branch. We agreed to post it here as a comment just so we're aware and so we can investigate while implementing the new actions loop.

Also, it's worth to mention that in both cases when the `dcc.Store(opl-trigger)` is inside or outside the components_container, it's returned from the `page.build`.

antonymilne:
Superseded by https://github.com/mckinsey/vizro/pull/1301.

PR #1178, opened 2025-05-15
[Feat] Actions outputs mappings for model fields

petar-qb:
Closes: https://github.com/McK-Internal/vizro-internal/issues/1610
Closes: https://github.com/McK-Internal/vizro-internal/issues/1710
 
## Description

- Make model fields accessible/targetable from the action inputs/outputs. For example `outputs=["graph-id.title"]`.  
- Move `f”{self.id}_title”` for the form components (like checklist, radioitems) directly to the title text so it can be accessed/changed from the action without changing the tooltip.
- **NEW**: For the AgGrid, all All `"outer-ID.property"` are mapped to `"inner-ID.property"`. It means, that Vizro users don't have to care about the `underlying-ag-grid-id` anymore. 🎉 
- Move AgGrid/Table `_input_component_id` initialisation to the `model_post_init` method.
- Add `description` model field to `Tabs`, `UserInput`, `TextArea` models. 

P.S. Wherever it's possible, accessing/targeting `“component-id.model_field”` in action inputs/outputs is recommended. It’s more stable than targeting like `“component-id_title.children”` because the internal implementation is more likely to be changed than the configuration API. (maybe we should mention something like this in the docs)

## For reviewers:
- Run the `scratch/app.py` and find (on the first page) the table we agreed about. This PR implements mappings from the "Additional output properties to support" column. Feel free to comment the `scratch/app.py` table for any additional mappings suggestions (that's the reason I put this table example)
- You're welcome to uncomment the `# TODO-REVIEWER-CHECK:` sections and play with different combinations. 

## Notice


huong-li-nguyen:
> 2. @huong-li-nguyen What’s the difference between:
> 
> * `html.Legend` , (used for title in checklist, radio_items)
> * `dbc.Label`, (mostly used for title in form components)
> * `html.H3`, (used for figure model titles)
> * `dcc.Markdown`, (used for figure model header/footer)
>   
>   1. Should I replace `html.Legend` with the `dbc.Label` in checklist and radio_items?
>   2. Is there any further improvement that should be added here in this PR.
>   3. Is there any further improvement that should be added in a separate PR?

1. No, don't replace it. We use `html.Legend` and `dbc.Label` on purpose; it's for accessibility reasons. You can read the article below, which will explain it well. Can you actually add that article as a code comment above html.Legend with a comment "don't replace with dbc.Label"? I feel like you guys have been trying to remove this so many times now haha 😄 : https://tetralogical.com/blog/2025/01/31/foundations-fieldset-and-legend/#:~:text=Using%20the%20and%20elements%20is%20an,a%20broad%20range%20of%20users.

These are just different typography elements.
- We use H2 for page and dashboard title
- We use H3 for all component titles, like figure, graph, table, aggrid
- We use html.Legend / dbc.Label for titles like controls
- We use p (which is dcc.Markdown) for simple text

All this ensures that we have the right typography hierarchy in place.

vizro-core/tests/unit/vizro/actions/test_abstract_action.py
```diff
@@ -248,6 +248,7 @@ class TestAbstractActionOutputs:
                 [Output("component_1", "property_1"), Output("component_2", "property_2")],
             ),
             (["known-model-id"], Output("known-model-id", "value")),
+            (["known-model-id.title"], Output("known-model-id_title", "children")),
```
petar-qb, line 251: Should we rename `known-model-id` to `dropdown-id` in tests so it become clear from where `"value"` and `"children"` properties come from?

petar-qb:
Hey @lingyielia 👋 vizro-ai docs build is failing due to not found 404 status for the `https://www.langchain.com/privacy-policy` link that appears in the docs. Should we replace it with another link?

vizro-core/tests/unit/vizro/actions/test_abstract_action.py
huong-li-nguyen: I actually like "known-model-id", because it makes the test case clear (We want to explicitly test for a model that is registered inside the model-manager). We also have a test case in there with "unknown-model-id". "dropdown-id" wouldn't be as clear to me, but that's my preference :)

huong-li-nguyen:
> https://www.langchain.com/privacy-policy

I think the page was down for a few hours, but it's working again :)

lingyielia:
Ah sorry I missed the question! thanks for checking @huong-li-nguyen  👍

huong-li-nguyen:
What a great `scratch_dev` app 🚀 This was the most fun I’ve had with a scratch dev app —super fun to play around with and great for reviewers! 😄 

This looks good to merge for me after fixing the remaining tests! Just a few small comments and questions:

* We didn’t update the `Table` yet, but just a reminder—we had planned to include `title` and `description` as output properties for `Container`, `Page`, and `Tabs` (where applicable). See this comment: https://github.com/mckinsey/vizro/pull/1153#issuecomment-2858136863
* Slightly related: I think we forgot to enable the `description` argument for `UserInput`, `TextArea`, and `Button`. There’s also a ticket floating around about enabling the tooltip for `Button`. If these were already enabled, I would’ve included them as output properties here too. Totally up to you if you want to add this in this PR or tackle it in a separate one—either works for me!
* Do we need additional tests for the `Table`/`AgGrid` `id` mapping? It’s just a mapping in the end, but feels like a key one, so might be worth checking.

Regarding your open questions:

1. In this case, I still wouldn’t render the title. If someone tries to update it via an action, I’d expect a Dash callback error stating that the `id` doesn’t exist. I wouldn’t add a default `html.Div` just for the title, since it clutters the HTML unnecessarily, especially when it might not be used at all.
2. The screenshots look good to me, but yep—unfortunately, you’ll need to replace them all. It would be great if you could do that, just so you’re familiar with the process too 🙂 Feel free to reach out to @l0uden if you’d like a refresher on how to do it! 👍

huong-li-nguyen (approved):

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -82,7 +88,23 @@ class AgGrid(VizroBaseModel):
...
-        return {"__default__": f"{self.id}.children"}
+        # TODO-AV2 E: Implement _action_trigger where makes sense.
+        #  For the AgGrid the mapping could look like: {"__default__": f"{self._input_component_id}.cellClicked"}
+        return {
+            "__default__": f"{self.id}.children",
+            "figure": f"{self.id}.children",
+            "title": f"{self.id}_title.children",
+            "header": f"{self.id}_header.children",
+            "footer": f"{self.id}_footer.children",
+            **{ag_grid_prop: f"{self._input_component_id}.{ag_grid_prop}" for ag_grid_prop in DAG_AG_GRID_PROPERTIES},
```
petar-qb, line 99: A question for the reviewers:

Should we move the `self._input_component_id` creation under the `model_post_init`? That would ensure its creation earlier which is always a good move. For example, in that case all tests under the `TestAttributesAgGrid` that require the `ag_grid.pre_build()` step could be moved to the `TestAgGridInstantiation` that doesn't.
```py
def model_post_init(self, context) -> None:
    super().model_post_init(context)
    self._input_component_id = self.figure._arguments.get("id",f"__input_{self.id}")`
```

vizro-core/tests/unit/vizro/actions/test_abstract_action.py
petar-qb: I refactored this so that `known` ag-grid is used in the tests (except the dropdown filter). The reason is that the AgGrid has default and non-default mappings for both actions outputs and inputs. So, we now only use one "known" component for all those tests.

vizro-core/src/vizro/models/_components/ag_grid.py
huong-li-nguyen: Moving the `_input_component_id` creation to `model_post_init` sounds good to me! 🤔  It'll make testing simpler since we won't need `pre_build()` just to get the ID, as you already said. We'll still keep the duplicate ID checks in `pre_build` since that's when we know all models are ready to be checked. So I can't think of anything that could go wrong, but maybe give it a try and see if anything unexpected comes up? :)

petar-qb:
@huong-li-nguyen I'm glad you enjoyed playing with the scratch example. More examples have been added is, so feel free to launch and review it again. 😃 

I added title/description support for: `Page`, `Container`, `Tabs` (`description` model field also added), `UserInput` and `TextArea`. However, I haven't added any mapping for the `Button`. Do you have any ideas of what could be added here? 

> Do we need additional tests for the Table/AgGrid id mapping? It’s just a mapping in the end, but feels like a key one, so might be worth checking.

I pushed some tests for the AgGrid id mapping. Could you check it?
@antonymilne and I decided not to add that kind of mapping for the `Table` as it's going to be deprecated.

huong-li-nguyen:
> @huong-li-nguyen I'm glad you enjoyed playing with the scratch example. More examples have been added is, so feel free to launch and review it again. 😃
> 
> I added title/description support for: `Page`, `Container`, `Tabs`, `UserInput` and `TextArea`. However, I haven't added any mapping for the `Button`. Do you have any ideas of what could be added here?

We were planning to enable `description` for Button as well, see this ticket: https://github.com/McK-Internal/vizro-internal/issues/1033 :) So you could enable that as well. In the _action_outputs mapping for Button I would have then expected `description` and `text`? `text` might already be supported.

vizro-core/src/vizro/models/_components/ag_grid.py
antonymilne: Great idea - this is a big improvement. As you say, we should generally do this stuff as earlier as possible, and now that pydantic v2 supports `model_post_init` it's definitely preferable to use that instead of `pre_build` wherever possible.

vizro-core/src/vizro/models/_components/tabs.py
```diff
@@ -26,22 +28,47 @@ class Tabs(VizroBaseModel):
...
+        BeforeValidator(coerce_str_to_tooltip),
+        Field(
+            default=None,
+            description="""Optional markdown string that adds an icon next to the title.
+            Hovering over the icon shows a tooltip with the provided description. Defaults to `None`.""",
+        ),
+    ]
+
+    @property
+    def _action_outputs(self) -> dict[str, _IdProperty]:
```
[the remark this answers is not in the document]
antonymilne: I think `title` would be a weird user expectation for `__default__` so I would just not set a `__default__` at all (good thing we chose to cover that case in our code in the end @huong-li-nguyen 😀).

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -97,6 +97,14 @@ class Container(VizroBaseModel):
         ]
     ]
 
+    @property
+    def _action_outputs(self) -> dict[str, _IdProperty]:
+        return {
+            "__default__": f"{self.id}.value",
+            "title": f"{self.id}_title.children",
+            **({"description": f"{self.description.id}.children"} if self.description else {}),
```
antonymilne, line 105: This comment applies in many places.

Now that I look at it again, I actually think this doesn't quite make sense because `Tooltip` does this:
```
dbc.Tooltip(id=self.id, children=dcc.Markdown(self.text))
```

So what you're doing here is replacing `dcc.Markdown` with plain text.

I feel like we should do one of the following:
* move `id=self.id` to the `dcc.Markdown` in the `Tooltip` (fine by me but maybe needs some other things to be changed?)
* put `_action_outputs` into the `Tooltip` model so you could do target `my-tooltip-id.text`, but this means users who want to change tooltips would need to do `description = vm.Tooltip(id="my-tooltip-id", ...)`

wdyt? @huong-li-nguyen please add your thoughts on this too 🙏

vizro-core/src/vizro/models/_components/form/_text_area.py
```diff
@@ -36,17 +48,25 @@ class TextArea(VizroBaseModel):
 
     @property
     def _action_outputs(self) -> dict[str, _IdProperty]:
-        return {"__default__": f"{self.id}.value"}
+        return {
+            "__default__": f"{self.id}.value",
+            "title": f"{self.id}_title.children",
```
antonymilne, line 53: Comment applies in many places. Should these be conditional on `if self.title` like you do with description to avoid trying to target a non-existent `html.Div`?

vizro-core/src/vizro/models/_components/form/_text_area.py
```diff
@@ -36,17 +48,25 @@ class TextArea(VizroBaseModel):
...
         return {"__default__": f"{self.id}.value"}
 
     @_log_call
     def build(self):
+        description = self.description.build().children if self.description else [None]
+
         return html.Div(
             children=[
-                dbc.Label(self.title, html_for=self.id) if self.title else None,
+                dbc.Label(children=[html.Div(self.title, id=f"{self.id}_title"), *description], html_for=self.id)
```
antonymilne, line 67: Comment applies in many places. 
```suggestion
                dbc.Label(children=[html.Span(self.title, id=f"{self.id}_title"), *description], html_for=self.id)
```

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -80,9 +86,30 @@ class AgGrid(VizroBaseModel):
     _input_component_id: str = PrivateAttr()
     _validate_figure = field_validator("figure", mode="before")(validate_captured_callable)
 
+    def model_post_init(self, context) -> None:
+        super().model_post_init(context)
+        self._input_component_id = self.figure._arguments.get("id", f"__input_{self.id}")
```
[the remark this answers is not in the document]
antonymilne: I like `_inner_component_id` best.

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -80,9 +86,30 @@ class AgGrid(VizroBaseModel):
...
+            **({"description": f"{self.description.id}.children"} if self.description else {}),
+            **{ag_grid_prop: f"{self._input_component_id}.{ag_grid_prop}" for ag_grid_prop in DAG_AG_GRID_PROPERTIES},
+        }
+
+    @property
+    def _action_inputs(self) -> dict[str, _IdProperty]:
+        return {
+            "__default__": f"{self.id}.children",
+            **{ag_grid_prop: f"{self._input_component_id}.{ag_grid_prop}" for ag_grid_prop in DAG_AG_GRID_PROPERTIES},
+        }
```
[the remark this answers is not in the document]
antonymilne: Yeah, it's because of the inner/outer id thing. Currently that's one of the most awkward and annoying bits of vizro so it's great to have this finally resolved.

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -80,9 +86,30 @@ class AgGrid(VizroBaseModel):
...
+            "header": f"{self.id}_header.children",
+            "footer": f"{self.id}_footer.children",
+            **({"description": f"{self.description.id}.children"} if self.description else {}),
+            **{ag_grid_prop: f"{self._input_component_id}.{ag_grid_prop}" for ag_grid_prop in DAG_AG_GRID_PROPERTIES},
+        }
+
+    @property
+    def _action_inputs(self) -> dict[str, _IdProperty]:
+        return {
+            "__default__": f"{self.id}.children",
```
antonymilne, line 110:
```suggestion
```

I think we shouldn't implement this because the object passed in really won't make sense.

vizro-core/src/vizro/models/_page.py
```diff
@@ -119,6 +120,14 @@ def _serialize_id(self, handler: SerializerFunctionWrapHandler, info: FieldSeria
             return result
         return result
 
+    @property
+    def _action_outputs(self) -> dict[str, _IdProperty]:
+        return {
+            "__default__": f"{self.id}.value",
```
antonymilne, line 126:
```suggestion
```

Like with `Container`.

antonymilne (commented):
Nice work! I think there's a few mistakes/things to change still but after that should all be good for a final ✅ 

Let me answer a few more questions fron you and @huong-li-nguyen here:

> P.S. Wherever it's possible, accessing/targeting “component-id.model_field” in action inputs/outputs is recommended. It’s more stable than targeting like “component-id_title.children” because the internal implementation is more likely to be changed than the configuration API. (maybe we should mention something like this in the docs)

Definitely yes when we document this we will explain this.

> * We didn’t update the `Table` yet, but just a reminder—we had planned to include `title` and `description` as output properties for `Container`, `Page`, and `Tabs` (where applicable). See this comment: https://github.com/mckinsey/vizro/pull/1153#issuecomment-2858136863. So this is still missing!
> 
>   - `Page.title` probably not because we use these as ids if not explicitly set, so might lead to conflicts? @antonymilne 

This won't lead to conflicts since a callback will just update the text on screen and not the underlying id or anything. It's maybe a weird of a weird thing to do but I don't have a problem with enabling it.

> What should we do if the title is initially empty and someone wants to change it? Currently, if the title is empty, we miss the title creation step. Should we modify this behaviour? P.S. Be aware of this recommendation: https://github.com/mckinsey/vizro/wiki/Developer-docs#htmldivhiddentrue-vs-none.

Agree with @huong-li-nguyen. I'd do like you did with `description` and not even add `title` to the `_action_outputs` in this case.

> Looks good to me 👍 One last thing: for the Button, I'd expose `text` and `description` as `_action_outputs`. Maybe hold off on implementing it just yet since it wasn’t originally scoped inside the Table—but I do think it makes sense 👍
>
> The tooltip around the Button is helpful for user feedback during actions. It was on the roadmap, but we didn't add it when we added `description` to other components.

We didn't add `description` to `Button` before because we weren't sure exactly how to do this, but if you have an idea then happy to add it. Let's save it for a separate PR though.

vizro-core/src/vizro/models/_components/container.py
huong-li-nguyen: Hmm, one question:
**1) move id=self.id to the dcc.Markdown in the Tooltip (fine by me but maybe needs some other things to be changed?)**
Agree with this one. This also fits into our paradigm that the reactive part should have the `id` 👍

**2) put _action_outputs into the Tooltip model so you could do target my-tooltip-id.text, but this means users who want to change tooltips would need to do description = vm.Tooltip(id="my-tooltip-id", ...)**
Generally also agree with the part of enabling `my-tooltip-id.text`. However, if someone wants to change the Page.description, do they then need to explicitly specify `Page.description = vm.Tooltip(xxx)`? If yes, I don't like it that much because in theory I would never actually configure the `Tooltip` and just want to update the `text` 🤔 . If we handle everything under the hood, then this also sounds good to me.

vizro-core/src/vizro/models/_components/form/_text_area.py
huong-li-nguyen: I like that!

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -97,6 +97,14 @@ class Container(VizroBaseModel):
         ]
     ]
 
+    @property
+    def _action_outputs(self) -> dict[str, _IdProperty]:
+        return {
+            "__default__": f"{self.id}.value",
```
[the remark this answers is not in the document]
petar-qb: Great catch! Added this accidentally..

vizro-core/src/vizro/models/_components/ag_grid.py
petar-qb: `__default__` maps to `f"{self.id}.children"`, just like `figure` does. It receives a `dag.AgGrid` object. If you check out the `"Figures - figure output"` scratch example, you'll see how naturally this mapping works. It's worth looking at the comment `# TODO-REVIEWER-CHECK: This is also valid output` added in this page in the example. 

Let me know if I'm missing something 😄

vizro-core/src/vizro/models/_components/container.py
petar-qb: Great catch @antonymilne! I really like the idea of ​​allowing `graph-id.description` to modify `dcc.Markdown.children`.

> move id=self.id to the dcc.Markdown in the Tooltip (fine by me but maybe needs some other things to be changed?)

Should we also move `self.extra` from the `dbc.Tooltip` to the `dcc.Markdown` then? It's weird that `id` and `extra` don't belong to the same component. However, it makes more sense for the `id` to live within the `dcc.Markdown` (because of `_action_outputs`), and for `extra` to live within the `dbc.Tooltip` (so that `is_open`, `hidden` and others can be applied properly). 

> but this means users who want to change tooltips would need to do description = vm.Tooltip(id="my-tooltip-id", ...)

Are you saying that `tooltip_id` and `tooltip_id.text` can only be targeted if `vm.Tooltip(id="tooltip_id", ...)` is explicitly defined this way? If so, that seems entirely expected to me.

vizro-core/src/vizro/models/_components/form/_text_area.py
petar-qb: Should I do it for all 13 occurrences 😆? So: graph, ag-grid, table, user-input, text-area, checklist, dropdown, radio-items, slider, range-slider, date-picker, tabs, container, and _dashboard(it displays page’s title).

vizro-core/src/vizro/models/_components/container.py
petar-qb: What do you think about adding the `id=f"{self._id}_text"` inside the `dcc.Markdown` object of the `vm.Tooltip` model?

In that case the `Tooltip._action_outputs` could be defined as:
  * `__default__`: `{self.id}-text.children`
  * `text`: `{self.id}-text.children`
  * `icon`: `{self.id}-icon.children`

Also the `description` mapping for other components could be changed in the following way:
`**({"description": f"{self.description.id}.children"} if self.description else {})` -> 
`**({"description": f"{self.description.id}-text.children"} if self.description else {}),`

In that case:
1. The mapping of `graph-id.description` would work correctly if `Graph.description` was defined as either `string` or `vm.Tooltip`,
2. `tooltip_id` would work
3. `tooltip_id.text` would work
4. `tooltip_id.text` would work

But, in that case, we can keep setting both: the `tooltip.id` and the `tooltip.extra` within the `dbc.Tooltip`.

vizro-core/changelog.d/20250514_164940_petar_pejovic_actions_shortcut_for_model_fields.md
```diff
@@ -0,0 +1,48 @@
...
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
```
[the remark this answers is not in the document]
petar-qb: Done in:
- changelog -> https://github.com/mckinsey/vizro/pull/1178/commits/e99b1257dd8071970194b2ffc4a6d91380e904bc
- docs: https://github.com/mckinsey/vizro/pull/1178/commits/bf21917e7fbddf2c3a55dd67158bdb0832251cdf

vizro-core/src/vizro/models/_components/form/_text_area.py
petar-qb: Done in:
- code -> https://github.com/mckinsey/vizro/pull/1178/commits/1e311aefc8391e10fe42b0930ba99aef41809e49
- tests -> https://github.com/mckinsey/vizro/pull/1178/commits/51d88515f9fbcc96025d32ff3d3be7ff7228553d

petar-qb:
Another quick question about the Button 😄 Should I add the action outputs mapping for "button-id.text". 
I probably don’t have to set the `__default__`. What do you think?

vizro-core/docs/pages/user-guides/tabs.md
```diff
@@ -138,4 +138,133 @@ To add [`Tabs`][vizro.models.Tabs] to your page, do the following:
 
         [![Tabs]][tabs]
 
+## Add a tooltip
+
+The `description` argument enables you to add helpful context to your tabs by displaying an info icon next to its title. Hovering over the icon shows a tooltip with your chosen text.
```
stichbury, line 143: Maybe nice to have a screengrab here. That way, people don't have to try out the whole dashboard below just to find out if it is what they think it is. They can get the sample look and feel from the static image.

stichbury (approved):
LGTM from a docs perspective. Thank you 🏆

vizro-core/changelog.d/20250514_164940_petar_pejovic_actions_shortcut_for_model_fields.md
```diff
@@ -0,0 +1,46 @@
...
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+### Added
+
+- Add `description` argument to `Tabs` to enable tooltips in the title. See this [user guide](https://vizro.readthedocs.io/en/stable/pages/user-guides/container/#add-a-tooltip) for an example. ([#1178](https://github.com/mckinsey/vizro/pull/1178))
```
antonymilne, line 21:
```suggestion
- Add `description` argument to `Tabs` to enable tooltips in the title. See the [user guide on tabs](https://vizro.readthedocs.io/en/stable/pages/user-guides/container/#add-a-tooltip) for an example. ([#1178](https://github.com/mckinsey/vizro/pull/1178))
```

vizro-core/docs/pages/user-guides/tabs.md
antonymilne: I would leave this as you have it is since it matches how we document it elsewhere and otherwise we would need to add the picture everywhere, which is more effort than it's worth 😅 We have the picture in a tab in the below example and I think that's enough.

vizro-core/docs/pages/user-guides/tabs.md
```diff
@@ -138,4 +138,133 @@ To add [`Tabs`][vizro.models.Tabs] to your page, do the following:
...
+
+You can provide markdown text as a string to use the default info icon or a [`Tooltip`][vizro.models.Tooltip] model to use any icon from the [Google Material Icons library](https://fonts.google.com/icons).
```
antonymilne, line 145:
```suggestion
You can provide [Markdown text](https://markdown-guide.readthedocs.io/)  as a string to use the default info icon or a [`Tooltip`][vizro.models.Tooltip] model to use any icon from the [Google Material Icons library](https://fonts.google.com/icons).
```

Please change everywhere this appears in docs 🙏

vizro-core/src/vizro/actions/_filter_interaction.py
```diff
@@ -61,7 +61,7 @@ def pre_build(self):
             raise ValueError(f"targets {invalid_targets} are not valid figures on the page.")
 
         # TODO: This check is temporarily disabled to avoid requiring a prior call to ag_grid or table pre_build.
-        #  Otherwise, their self._input_component_id may not be set, leading to an error when checking "modelID" in
+        #  Otherwise, their self._inner_component_id may not be set, leading to an error when checking "modelID" in
         #  triggered_model._filter_interaction_input. We should revisit this when reworking filter interaction
         #  to find a better way to reintegrate it.
```
antonymilne, line 66:
```suggestion
        #  to find a better way to reintegrate it.
        # Possibly now that self._inner_component_id is now set in model_post_init rather than pre_build
        # there's an easier solution here.
```

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -77,12 +83,33 @@ class AgGrid(VizroBaseModel):
...
 
+    def model_post_init(self, context) -> None:
+        super().model_post_init(context)
+        self._inner_component_id = self.figure._arguments.get("id", f"__input_{self.id}")
+
     @property
     def _action_outputs(self) -> dict[str, _IdProperty]:
-        return {"__default__": f"{self.id}.children"}
+        # TODO-AV2 E: Implement _action_trigger where makes sense.
+        #  For the AgGrid the mapping could look like: {"__default__": f"{self._inner_component_id}.cellClicked"}
```
antonymilne, line 96: Please can you move this somewhere outside the `_action_outputs` method? Otherwise it makes me think like something needs to change in this function.

vizro-core/src/vizro/models/_components/ag_grid.py
antonymilne: Let me check this!

vizro-core/src/vizro/models/_page.py
antonymilne: I think you missed this change.

antonymilne (approved):
This looks great, thank you! 🙏 Let me just answer some of the unresolved questions separately, but feel free to merge whenever you're happy with everything.

vizro-core/src/vizro/models/_components/form/_text_area.py
antonymilne: Yes 😀 Assuming the results of the screenshot tests look good.

Reason for the change is that div is block-level but span is inline: https://en.wikipedia.org/wiki/Div_and_span.

antonymilne:
> Another quick question about the Button 😄 Should I add the action outputs mapping for "button-id.text". I probably don’t have to set the `__default__`. What do you think?

Good idea - let's add `button-id.text` but nothing for `__default__`.

vizro-core/src/vizro/models/_components/ag_grid.py
antonymilne: Nope, still don't get it, sorry 😅 

What you say makes sense for `_action_outputs` but this is `_action_inputs` where I think it doesn't make sense?

vizro-core/src/vizro/models/_components/container.py
antonymilne: This is a great point @petar-qb 👍 I think your suggestion is absolutely the right thing to do, so please do go ahead with it. In general I think the rules should be:
* `self.extra` always goes with the same component that has `id=self.id`
* action mapping `__default__` doesn't need to point to the component with `id=self.id` where it doesn't make sense (like here)

As well as doing this for `dbc.Tooltip` we should check our other models for the same issue - it's exactly the same for `vm.Card`. So please could you:
* make the change for `vm.Tooltip` here
* make the same change for `vm.Card` and see if it breaks anything in our tests (CSS selectors might need to change) 
* check if there's any other instances like this - I think searching for cases where `defaults` doesn't contain `id` should catch them all
* add a note to the [developer docs](https://github.com/mckinsey/vizro/wiki/Developer-docs) explaining the rules

If it turns out that changes outside `vm.Tooltip` are more complicated/bigger than expected then just do the tooltip one here and we can do the other changes in a separate PR.

vizro-core/src/vizro/models/_components/ag_grid.py
petar-qb: Sorry, my bad.. I thought we're talking about the `_action_outputs`. I'm removing it.

PR #1179, opened 2025-05-15
[Bug] Fix info icon rendering without container title

nadijagraca:
## Description
Fixes a bug reported by Alexey where the info icon would render even when the container had no title.

## Notice


petar-qb:
I approved it, but actually I'm wondering should we rather raise an exception in this case instead of just not rendering it.

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -156,7 +156,7 @@ def _build_container(self):
 
     def _build_container_title(self):
         """Builds and returns the container title, including an optional icon and tooltip if collapsed."""
-        description = self.description.build().children if self.description else [None]
+        description = self.description.build().children if self.description is not None and self.title else [None]
 
         title_content = [html.Div([self.title, *description], className="inner-container-title")]
```
antonymilne, line 161: I think this was wrong before but actually may still be a bit wrong. Think of the case case that there's no title and no description and `self.collapsed is None`. Then this function will return `html.H3(html.Div())` when maybe we actually just want to return `None`?

Maybe we actually want to do is something like this:
```
title_content = [html.Div([self.title, *description], className="inner-container-title")] if self.title else []

...
return html.H3(...) if self.title_content else None.
```

I'm also not sure what this is meant to look like if you have no title but `self.collapse is not None`.

@huong-li-nguyen please can you check this? 🙏 If it looks fine on screen how it is then all good but I guess there might be some unwanted extra spaces there currently, I'm just not sure what the intended spacing and behaviour is.

antonymilne (commented):
I was about to hit Approve but then I realised actually I'm not sure about this 🤔 It may be ok but let's wait and see what @huong-li-nguyen thinks.

@petar-qb feel free to do the release without this.

> I approved it, but actually I'm wondering should we rather raise an exception in this case instead of just not rendering it.

I think this is the same as we do with all other components that have `description` and `title`. Raising an exception or maybe just warning is a good idea but I wouldn't do it in this PR since it would need to be done in lots of places.

vizro-core/src/vizro/models/_components/container.py
huong-li-nguyen: Great catch, @antonymilne! It doesn't make sense to allow `collapsed`  without a title, since the title is the part of the container that's always visible and includes the icon for expanding/collapsing.

So I think we should enforce the requirement that a `Container.title` is mandatory in two scenarios:

1. When it's used inside `Tabs` (which is already handled), and
2. When `collapsed` is set to `True`.

Something like the validator below should work:

```python
@model_validator(mode="before")
@classmethod
def _validate_title_if_collapsed(cls, values):
    collapsed = values.get("collapsed")
    title = values.get("title")
    if collapsed is not None and not title:
        raise ValueError("`Container` must have a `title` explicitly set when `collapsed` is not None.")
    return values
```

@antonymilne — quick reminder question (I always get this mixed up 😅): I know we generally prefer using `Annotated` along with `BeforeValidator` or `AfterValidator`, but for cross-field validations like this one, I think `Annotated` alone isn’t sufficient — is that right? Or is there a cleaner way to do it?

Once you add the validator, I think this should be right:

```python
...
return html.H3(...) if self.title else None
```

And then I agree with @antonymilne and @petar-qb - we should probably add a warning that a title needs to be provided for the info-icon to be visible 👍

huong-li-nguyen (changes requested):

vizro-core/src/vizro/models/_components/container.py
antonymilne: > So I think we should enforce the requirement that a `Container.title` is mandatory in two scenarios:
> 
> 1. When it's used inside `Tabs` (which is already handled), and
> 2. When `collapsed` is set to `True`.

Agree but remember the second condition should be `collapse is not None` because `collapsed = False` means "collapsible but closed by default" (it's right in the validator code you give).

> @antonymilne — quick reminder question (I always get this mixed up 😅): I know we generally prefer using Annotated along with `BeforeValidator` or `AfterValidator`, but for cross-field validations like this one, I think Annotated alone isn’t sufficient — is that right? Or is there a cleaner way to do it?

That's right, you can't use `Annotated` with cross-field validation like this. However, you should never use a before validator unless really need to, which in this case I'm pretty sure you don't. So it should be something like:
```
def _validate_title_if_collapsed(self) -> Self:
    if self.collapsed is not None and not self.title:
        raise ValueError("`Container` must have a `title` explicitly set when `collapsed` is not None.")
    return values
```

> And then I agree with @antonymilne and @petar-qb - we should probably add a warning that a title needs to be provided for the info-icon to be visible 👍

Agree but let's leave this for a separate PR since it will be a change across many places.

huong-li-nguyen: Ahh yes, that's what I meant with collapsed 👍

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -177,8 +183,12 @@ def _build_container_title(self):
...
-            className="container-title-collapse" if self.collapsed is not None else "container-title",
-            id=f"{self.id}_title",
+        return (
+            html.H3(
+                children=title_content,
+                className="container-title-collapse" if self.collapsed is not None else "container-title",
+                id=f"{self.id}_title",
+            )
+            if self.title
+            else None
```
huong-li-nguyen, line 193:
```suggestion
```

I think you can remove it here, given that you put the logic above :)

`self._build_container_title() if self.title else None`

huong-li-nguyen (approved):

PR #1184, opened 2025-05-19
[Feat] Add Dockerfile for Vizro MCP server

lingyielia:
## Description
It enables the Docker version of Vizro MCP server. Users can decide what's visible by only mounting the project relevant path.

## Screenshot
Screenshot 2025-05-19 at 02 05 53

## Notice


maxschulz-COL:
Hey, thanks so much for taking this on! And also for figuring out an easy way to mount external files!

QQ before a proper review: does this not (only) make sense if pushed to a central registry? I find it a little awkward for the PyPI version to be so hastle free, while for the Docker version one needs to actually build the image

lingyielia:
> Hey, thanks so much for taking this on! And also for figuring out an easy way to mount external files!
> 
> QQ before a proper review: does this not (only) make sense if pushed to a central registry? I find it a little awkward for the PyPI version to be so hastle free, while for the Docker version one needs to actually build the image

It's in general more secure. Under uvx approach, llm could propose any random path and `load_and_analyze_data` will try to run it. And for any code in the server, it's executed in a restricted space.

Plus, the implication on dependencies is different. uvx means uv as dependency and docker means docker as dependency.

maxschulz-COL:
> It's in general more secure. Under uvx approach, llm could propose any random path and `load_and_analyze_data` will try to run it. And for any code in the server, it's executed in a restricted space.
> 
> Plus, the implication on dependencies is different. uvx means uv as dependency and docker means docker as dependency.

I agree that Docker is in general more secure, but if we wanted, we could also limit the scope of the `read_xxx` functions, e.g. via an environment variable, shall we do that?

I am not terribly worried about any code execution, become it happens within the scope of fairly secure libraries, but yes, ultimately it is a little more secure. We should not forget this is mostly about the general trust of us as authors!

Regarding the dependencies, yes, agreed as well, but given that `uv` is already so ubiquitous (and open source) I think there is not so much of a difference to Docker...

vizro-mcp/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim
+
+WORKDIR /app
+
+COPY pyproject.toml hatch.toml uv.lock ./
```
maxschulz-COL, line 5: do we even need the pyproject toml and the hatch.toml?

vizro-mcp/src/vizro_mcp/__main__.py
maxschulz-COL, line 1: why is this needed?

vizro-mcp/README.md
```diff
@@ -116,6 +116,42 @@ Similarly, when using Cursor, after a short pause, you should see a green light
...
+{
+  "mcpServers": {
+    "vizro-mcp": {
+      "command": "docker",
+      "args": [
+        "run",
+        "-i",
+        "--rm",
+        "--mount",
+        "type=bind,src=<ABSOLUTE/HOST/PATH/TO/YOUR/DATA>,dst=<ABSOLUTE/CONTAINER/PATH>",
```
maxschulz-COL, line 145: Should we align the PyPI and Docker commands? An idea could be to have env variable in both cases?

vizro-mcp/README.md
```diff
@@ -116,6 +116,42 @@ Similarly, when using Cursor, after a short pause, you should see a green light
...
+### 🐳 Running with Docker
+
+You can run Vizro-MCP inside a Docker container for easy setup and isolation.
+
+#### Build the Docker image
+
+In the root of the `vizro-mcp` directory, build the Docker image with:
+
+```bash
+docker build -t vizro-mcp .
```
maxschulz-COL, line 128: I really think we should not have the local setup described, but maybe I am not seeing something. Can we not (for now) publish the image on our github?

vizro-mcp/README.md
```diff
@@ -116,6 +116,42 @@ Similarly, when using Cursor, after a short pause, you should see a green light
 
 </details>
 
+### 🐳 Running with Docker
```
maxschulz-COL, line 119: I think we need to rethink the overall structure now, this is not on the same level as the above. It is an alternative that can be used in either of the two examples. What do you think?

maxschulz-COL (commented):
thanks for taking this on, some preliminary comments!

vizro-mcp/Dockerfile
lingyielia: Removed hatch.toml. But a pyproject.toml in order to use the Project APIs (uv sync)

vizro-mcp/README.md
lingyielia: Can you explain bit more on how? 

```
...
"--mount",
"type=bind,src=<ABSOLUTE/HOST/PATH/TO/YOUR/DATA>,dst=<ABSOLUTE/CONTAINER/PATH>",
...
```
is a straight-forward command when mounting in docker. Shall we stick to what's looks most familiar to docker users, rather than having to unify PyPI and Docker commands

vizro-mcp/Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+RUN uv sync --frozen --no-install-project --no-dev --no-editable
```
antonymilne, line 6: I'm a bit confused. Is there a reason we don't just do the exact same `uvx` in this file like in the normal user instructions? Why do we do `uv sync` and then `uv run`?

vizro-mcp/src/vizro_mcp/__main__.py
```diff
@@ -0,0 +1,6 @@
+"""Entrypoint for running Vizro MCP with Docker."""
```
antonymilne, line 1: Why is this needed? 🤔

vizro-mcp/Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+
+COPY . /app
+
+ENTRYPOINT ["uv", "run", "python", "-m", "vizro_mcp"]
```
antonymilne, line 10: I don't understand why this isn't just `uvx vizro-mcp`?

vizro-mcp/README.md
lingyielia: This is here before it's available in places like Docker Catalog.
 
Packages could be a good alternative. But if we are aiming to publish in Docker Catalog, I don't want to spend time setup packages.

Any way, this command eventually should sit under `## Development or running from source` for local development.

vizro-mcp/src/vizro_mcp/__main__.py
lingyielia: it was used when directly running the vizro_mcp as a module. As suggested by @antonymilne, if I use `ENTRYPOINT ["uv", "run", "vizro-mcp"]`, which looks for pyproject.toml for entry point, then this is not needed

vizro-mcp/Dockerfile
lingyielia: as discussed, uvx doesn't work as expected. Stick to `RUN uv sync` still

vizro-mcp/src/vizro_mcp/__main__.py
lingyielia: not needed anymore

vizro-mcp/README.md
lingyielia: I now restructure it into


--------

### Setup Instructions
#### Configuration with `uvx`

<details>
Claude Desktop
</details>

<details>
Cursor
</details>

#### Configuration with `docker`
(without the Claude Desktop and Cursor details)

--------

Is that any clearer?

vizro-mcp/README.md
maxschulz-COL: Can we move it there then, and maybe have a tiny reference to it also in the main installation (like there is already for the developer setup)? I would not like to mix "consumer facing" and developer facing instructions

vizro-mcp/README.md
lingyielia: Updated as discussed.
- added multiple paths in example
- don't change the src and dst format

vizro-mcp/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim
+
+WORKDIR /app
+
+COPY pyproject.toml uv.lock ./
+RUN uv sync --frozen --no-install-project --no-dev --no-editable
+
+COPY . /app
+
+ENTRYPOINT ["uv", "run", "vizro-mcp"]
```
antonymilne, line 10:
```suggestion
ENTRYPOINT ["uv", "run", "--no-sync", "vizro-mcp"]
```

I think? Since the dependencies are already up to date so we can skip this stage here. Please do check though.

antonymilne (approved):
Looks good! Though I didn't actually try it out 😅 Maybe worth getting one other person to just give it a quick test.

vizro-mcp/Dockerfile
lingyielia: we need another sync here so that the actual vizro-mcp is installed. otherwise there is error

```
╰─ docker run -i --rm --mount type=bind,src=<some path>,dst=<some path> vizro-mcp
error: Failed to spawn: `vizro-mcp`
  Caused by: No such file or directory (os error 2)
```

lingyielia:
> Looks good! Though I didn't actually try it out 😅 Maybe worth getting one other person to just give it a quick test if @maxschulz-COL wasn't able to. Or maybe we just wait until it's suitable for a general consumer before doing that.
> 
> Completely agree with @maxschulz-COL's feedback above about keeping the general consumer instructions separate from the developer ones btw. The docs look really nice now 👍

Yes, I believe @maxschulz-COL run it 😄 And I'm creating another ticket to update instructions once the docker image is public

PR #1189, opened 2025-05-20
[Docs] Hide side navigation on index page

huong-li-nguyen:
## Description

- Hide side navigation on index page (we don't need it as there is only one item) to save space

**After**

Screenshot 2025-05-20 at 15 34 21

**Before**
Screenshot 2025-05-20 at 15 34 36

## Notice

### DIFF
--- a/vizro-core/docs/javascripts/extra.js
+++ b/vizro-core/docs/javascripts/extra.js
@@ -1,0 +1,10 @@
+document.addEventListener("DOMContentLoaded", function () {
+  const isHomepage = document.title.trim() === "Vizro";
+
+  if (isHomepage) {
+    const sidebar = document.querySelector(".md-sidebar--primary");
+    if (sidebar) {
+      sidebar.style.display = "none";
+    }
+  }
+});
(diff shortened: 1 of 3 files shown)


stichbury:
It's not currently working as planned 
image

I will adjust temporarily to make it wider still and let's see if it _ever_ pushes it to the bottom!

huong-li-nguyen:
Yes, I don't think it matters how wide it is 😄 It will always appear inside the left-navigation.

stichbury:
Yes! It is working on pages of docs e.g. https://vizro--1189.org.readthedocs.build/en/1189/pages/user-guides/graph/ is now at the bottom compared to https://vizro.readthedocs.io/en/stable/pages/user-guides/graph/#add-additional-text where it's right at the bottom of the ToC. So it does work for that, but not the front page.

IS there any way we can otherwise fiddle this and block the spot? Like, can we experiment with dropping our own image there just to see what happens? If it works, we could maybe put a white square in place temporarily.

I don't have an issue with the ads long-term...but for the TCUK submission, I'd rather keep the page as simple and non-complex for the reader as we can.

huong-li-nguyen:
@stichbury - in theory this works, but I noticed that you can't access the overlaying menu from the index page anymore then. So that might not be optimal 🤔 You can still access that overlaying menu from all other pages though.

Screenshot 2025-05-20 at 15 39 59

PR #1226, opened 2025-06-05
[Fix] Set height to auto for ag-root-wrapper-body in normal layout

vladimir-mck:
## Description
Closes [#1839](https://github.com/McK-Internal/vizro-internal/issues/1839)

**Change defaults inside `dash_ag_grid`:**
- Turn on `"domLayout": "autoHeight"`
- Set `"pagination": True`

**Context:**
By default, AgGrid sets its height to 400px. Overriding this with height: 100% via CSS can cause it to collapse to 0px in certain layout scenarios—especially inside flex containers where all elements have height: 100% but no parent has an explicit height. This change ensures AgGrid renders correctly in such cases.

Additionally, updating the domLayout property causes AgGrid to render all rows at once. While this may be needed for layout accuracy, it can significantly impact performance when dealing with thousands of rows. To mitigate this, pagination has been added to limit the number of rendered rows and maintain good performance.

## Notice


huong-li-nguyen (commented):
Hey @vladimir-mck - I am still checking some other examples, but already noticed this one. If you turn on pagination, it jumps and adds lots of space, and then jumps back.

```
import vizro.models as vm
from vizro import Vizro
from vizro.tables import dash_ag_grid


df = px.data.iris()

page = vm.Page(
    title="Page",
    layout=vm.Flex(),
    components=[
        vm.Container(
            variant="outlined", title="Container with AgGrid", components=[vm.AgGrid(figure=dash_ag_grid(df, dashGridOptions={"pagination": True}))]
        ),
    ],
)

if __name__ == "__main__":
```

https://github.com/user-attachments/assets/face47df-f74f-4d2f-bdb8-914720bb5916

huong-li-nguyen:
The rest seemed fine! It's just the weird behaviour with the pagination turned on. I think that's also why the screenshot test is failing at the moment! Previously the pagination was still on screen (in the test config).

vladimir-mck:
Updated css selector to catch only ag grid table containers.

Screenshot 2025-06-05 at 13 09 20
Screenshot 2025-06-09 at 10 56 31
Screenshot 2025-06-09 at 10 56 41

huong-li-nguyen:
Hey @vladimir-mck – the layout looks great now with `Flex` 🥳! However, when using `Grid`, the auto-height change seems to cause some issues. Could you take another look?

* The pagination is getting cut off and appears outside the container
* A double scrollbar is now showing up

Thanks!


```
page = vm.Page(
    title="Page",
    components=[
        vm.Container(
            variant="outlined", title="Container with AgGrid", components=[vm.AgGrid(figure=dash_ag_grid(df, dashGridOptions={"pagination": True}))]
        ),
    ],
)
```

**Before**
Screenshot 2025-06-09 at 15 35 39


**After**
Screenshot 2025-06-09 at 15 34 46

vizro-core/src/vizro/tables/_dash_ag_grid.py
```diff
@@ -89,6 +89,7 @@ def dash_ag_grid(data_frame: pd.DataFrame, **kwargs: Any) -> dag.AgGrid:
         "dashGridOptions": {
             "dataTypeDefinitions": _DATA_TYPE_DEFINITIONS,
             "animateRows": False,
+            "domLayout": "autoHeight",
```
huong-li-nguyen, line 92: @vladimir-mck – I added a scratch dev with both the `Table` and `AgGrid`, and previously the results were different, which is a bit confusing. It looks like with this change `"domLayout": "autoHeight"`, they now behave the same. In the default `Grid` layout, they still take on 100% of the height, and the bug we saw in the `FlexContainer` is also resolved 👍 So it seems like we can solve this bug, without having to change the default behaviour inside the Grid (still takes on 100% height there).

I agree that in other, more customized layouts, the height might not be optimal. But in those cases, users will need to manually set the height anyway, as you said. Currently waiting for the screenshots to render, but it seems like with this change it works! Can you check if I missed anything?

huong-li-nguyen (approved):
Please double-check my latest push and if there are any issues with it 😬 But i checked the screenshots and the examples, and they seem fine 👍 Let's discuss tomorrow!

huong-li-nguyen:
> This change looks just perfect! I tested different many layouts and all work as expected. I'm just wondering if we should announce the `pagination=True` by default change.. @huong-li-nguyen

We normally don't announce UX changes unless they are breaking anything for anyone. Even though in this case, I am a bit more open as it's not only a CSS change but a config inside the `dash_ag_grid` 🤔

maxschulz-COL:
> > This change looks just perfect! I tested different many layouts and all work as expected. I'm just wondering if we should announce the `pagination=True` by default change.. @huong-li-nguyen
> 
> We normally don't announce UX changes unless they are breaking anything for anyone. Even though in this case, I am a bit more open as it's not only a CSS change but a config inside the `dash_ag_grid` 🤔

I would definitely announce it, this is quite a fundamental user facing change. 

In general I think we only said that FE changes are not breaking, but that doesn't stop us from announcing it. In general a [changelog](https://keepachangelog.com/en/1.1.0/#who) should always communicate to a user what they can expect to change, maybe someone finds this important and would prefer not to upgrade! Or vice versa, someone is waiting for just this change, so they know what version to have as minimal version.

petar-qb:
I agree here with @maxschulz-COL. Let's announce it as it's quite a big FE change that some users will perhaps hit negatively.

vizro-core/changelog.d/20250605_132228_vladimir_nikolic_container_ag_grid.md
```diff
@@ -0,0 +1,58 @@
...
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+Changed AgGrid defaults to improve layout compatibility and performance:
+
+- Enabled "domLayout": "autoHeight" by default, allowing the grid to automatically adjust its height based on content. This prevents layout issues where the grid could collapse to 0px in flexible containers without an explicit parent height.
+- Enabled pagination by default to ensure good performance, as autoHeight causes all rows to be rendered at once. Pagination limits the number of visible rows and maintains responsiveness even with large datasets.
+
+These changes ensure AgGrid displays correctly in more layout scenarios and remains performant with large tables. ([#1226](https://github.com/mckinsey/vizro/pull/1226))
```
petar-qb: This changelog is quite detailed which is great for internal understanding, but it could be overwhelming for users. Hence, we aim to keep it as simple and focused as possible, by announcing only what directly affect our users.

For this PR it's only the `pagination=True` (by default) what's should be announced. (See the [comment](https://github.com/mckinsey/vizro/pull/1226#pullrequestreview-2935960025) and comments below).


So, we should update the https://vizro.readthedocs.io/en/stable/pages/user-guides/table/#enable-pagination to `...#disable-pagination`. And then update the changelog file to something like:

```suggestion
- Enable pagination for the `AgGrid` by default. See the [user guide on how to disable the pagination](https://vizro.readthedocs.io/en/stable/pages/user-guides/table/#disable-pagination). ([#1226](https://github.com/mckinsey/vizro/pull/1226))
```

vizro-core/changelog.d/20250605_132228_vladimir_nikolic_container_ag_grid.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Enable pagination for the `AgGrid` by default. See the [user guide on how to disable the pagination](https://vizro.readthedocs.io/en/stable/pages/user-guides/table/#disable-pagination). ([#1226](https://github.com/mckinsey/vizro/pull/1226))
```
petar-qb: Let's add the refactor the docs section `#enable-pagination` to `#disable-pagination` as described in the https://github.com/mckinsey/vizro/pull/1226#discussion_r2153959340.

vizro-core/changelog.d/20250605_132228_vladimir_nikolic_container_ag_grid.md
```diff
@@ -22,12 +22,22 @@ Uncomment the section that is right (remove the HTML comment wrapper).
...
 - A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
 
 -->
+
+Changed AgGrid defaults to improve layout compatibility and performance:
+
+- Enabled "domLayout": "autoHeight" by default, allowing the grid to automatically adjust its height based on content. This prevents layout issues where the grid could collapse to 0px in flexible containers without an explicit parent height.
+- Enabled pagination by default to ensure good performance, as autoHeight causes all rows to be rendered at once. Pagination limits the number of visible rows and maintains responsiveness even with large datasets.
+
+These changes ensure AgGrid displays correctly in more layout scenarios and remains performant with large tables. ([#1226](https://github.com/mckinsey/vizro/pull/1226))
```
maxschulz-COL:
```suggestion
Changed `dash_ag_grid` defaults to improve layout compatibility and performance ("domLayout": "autoHeight" and "pagination": `True`). ([#1226](https://github.com/mckinsey/vizro/pull/1226))
```

vladimir-mck: Updated

PR #1238, opened 2025-06-12
[Bugfix] Fixed dropdown height calculation in containers dropdown

nadijagraca:
## Description
closes https://github.com/McK-Internal/vizro-internal/issues/1846

Antony FYI: https://github.com/McK-Internal/vizro-internal/issues/1829
- we might want to reconsider moving this part of code to the selector, as part of your refactoring. 

Fixed dropdown height calculation when dropdown selector is used in containers 
## Notice

### DIFF
--- a/vizro-core/changelog.d/20250612_095418_nadija_ratkusic_graca_dropdown_height_in_containers.md
+++ b/vizro-core/changelog.d/20250612_095418_nadija_ratkusic_graca_dropdown_height_in_containers.md
@@ -37,7 +37,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 ### Fixed
 
-- Fixed dropdown height calculation in container controls. ([#1238](https://github.com/mckinsey/vizro/pull/1238))
+- Fix dynamically calculated row height for `Dropdown` when it's used within a container. ([#1238](https://github.com/mckinsey/vizro/pull/1238))
 
 
 <!--
--- a/vizro-core/changelog.d/20250612_095418_nadija_ratkusic_graca_dropdown_height_in_containers.md
+++ b/vizro-core/changelog.d/20250612_095418_nadija_ratkusic_graca_dropdown_height_in_containers.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Fixed dropdown height calculation in container controls. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+
+<!--
+### Security
+
+- A bullet item for the Security category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
--- a/vizro-core/changelog.d/20250612_095418_nadija_ratkusic_graca_dropdown_height_in_containers.md
+++ b/vizro-core/changelog.d/20250612_095418_nadija_ratkusic_graca_dropdown_height_in_containers.md
@@ -37,7 +37,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 ### Fixed
 
-- Fixed dropdown height calculation in container controls. ([#1](https://github.com/mckinsey/vizro/pull/1))
+- Fixed dropdown height calculation in container controls. ([#1238](https://github.com/mckinsey/vizro/pull/1238))
 
 
 <!--
--- a/vizro-core/changelog.d/20250612_095418_nadija_ratkusic_graca_dropdown_height_in_containers.md
+++ b/vizro-core/changelog.d/20250612_095418_nadija_ratkusic_graca_dropdown_height_in_containers.md
@@ -37,7 +37,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 
 ### Fixed
 
-- Fix dynamically calculated row height for `Dropdown` when it's used within a container. ([#1238](https://github.com/mckinsey/vizro/pull/1238))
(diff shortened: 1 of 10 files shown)


vizro-core/changelog.d/20250612_095418_nadija_ratkusic_graca_dropdown_height_in_containers.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Fixed dropdown height calculation in container controls. ([#1238](https://github.com/mckinsey/vizro/pull/1238))
```
petar-qb:
```suggestion
- Fix dynamically calculated row height for `Dropdown` when it's used within a container controls. ([#1238](https://github.com/mckinsey/vizro/pull/1238))
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -203,6 +200,9 @@ def pre_build(self):
             self.selector = cast(CategoricalSelectorType, self.selector)
             self.selector.options = self.selector.options or self._get_options(targeted_data)
 
+        # set defaults for container selectors
```
petar-qb: The comment looks irrelevant now. I would remove it.

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -62,3 +66,24 @@ def _build_inner_layout(layout, components):
...
+    """Calculates the height of the dropdown options based on the longest option."""
+    # 30 characters is roughly the number of "A" characters you can fit comfortably on a line in the page dropdown.
+    # "A" is representative of a slightly wider than average character:
+    # https://stackoverflow.com/questions/3949422/which-letter-of-the-english-alphabet-takes-up-most-pixels
+    # We look at the longest option to find number_of_lines it requires. Option height is the same for all options
+    # and needs 24px for each line + 8px padding.
+    list_of_labels = _get_list_of_labels(full_options)
+    max_length = max(len(str(option)) for option in list_of_labels)
+    number_of_lines = math.ceil(max_length / no_characters)
+    return 8 + 24 * number_of_lines
```
huong-li-nguyen: Would rename to `char_count` to be make it clearer

```suggestion
def _calculate_option_height(full_options: OptionsType, char_count: int = 30) -> int:
    """Calculates the height of the dropdown options based on the longest option."""
    # 30 characters is roughly the number of "A" characters you can fit comfortably on a line in the page dropdown (placed on the left-side). "A" is representative of a slightly wider than average character:
    # https://stackoverflow.com/questions/3949422/which-letter-of-the-english-alphabet-takes-up-most-pixels
    # We look at the longest option to find number_of_lines it requires. Option height is the same for all options
    # and needs 24px for each line + 8px padding.
    list_of_labels = _get_list_of_labels(full_options)
    max_length = max(len(str(option)) for option in list_of_labels)
    number_of_lines = math.ceil(max_length / char_count)
    return 8 + 24 * number_of_lines
```

vizro-core/src/vizro/models/_controls/_controls_utils.py
```diff
@@ -46,3 +47,8 @@ def set_container_control_default(control: ControlType) -> None:
     page = model_manager._get_model_page(control)
     if control not in page.controls and isinstance(control.selector, (Checklist, RadioItems)):
         control.selector.extra.setdefault("inline", True)
+
+    if control not in page.controls and isinstance(control.selector, Dropdown):
+        # 15 characters is roughly the number of "A" characters you can fit comfortably in the container dropdown line.
+        option_height = _calculate_option_height(full_options=control.selector.options, no_characters=15)
+        control.selector.extra.setdefault("optionHeight", option_height)
```
petar-qb:
```suggestion
    if control not in page.controls:
        if isinstance(control.selector, (Checklist, RadioItems)):
            control.selector.extra.setdefault("inline", True)
        elif isinstance(control.selector, Dropdown):
            # 15 characters is roughly the number of "A" characters you can fit comfortably in the container dropdown line.
            option_height = _calculate_option_height(full_options=control.selector.options, no_characters=15)
            control.selector.extra.setdefault("optionHeight", option_height)
```

huong-li-nguyen: I think I'd slightly rewrite this so you don't have to repeat `if control not in page.controls` everytime you add new logic in here, given that this always needs to be valid for the entire function.

```suggestion
    if control in page.controls:
        return

    selector = control.selector

    if isinstance(selector, (Checklist, RadioItems)):
        selector.extra.setdefault("inline", True)

    elif isinstance(selector, Dropdown):
        option_height = _calculate_option_height(full_options=selector.options, chart_count=15)
        selector.extra.setdefault("optionHeight", option_height)
```

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -62,3 +66,24 @@ def _build_inner_layout(layout, components):
...
+
+def _get_list_of_labels(full_options: OptionsType) -> Union[list[StrictBool], list[float], list[str], list[date]]:
+    """Returns a list of labels from the selector options provided."""
+    if all(isinstance(option, dict) for option in full_options):
+        return [option["label"] for option in full_options]  # type: ignore[index]
+    else:
+        return cast(Union[list[StrictBool], list[float], list[str], list[date]], full_options)
+
+
+def _calculate_option_height(full_options: OptionsType, no_characters: int = 30) -> int:
```
petar-qb: This function is used from Dropdown, Filter and Parameter so I'm totally fine with moving it here in the `_models_utils.py`. I just think we shouldn't define a separate `_get_list_of_labels` but do the label extraction within the `_calculate_option_height` directly.

petar-qb (approved):
Excellent fix!! 🥇 

Could you mention Antony's Issue here, so here's aware when he starts with the refactoring?

Could you also add a test where the height is automatically calculated and differs when the dropdown is defined within a Page and a Container?

vizro-core/src/vizro/models/_controls/_controls_utils.py
petar-qb: Li's suggestion is better -> https://github.com/mckinsey/vizro/pull/1238#discussion_r2142024238

huong-li-nguyen:
> Could you also add a test where the height is automatically calculated and differs when the dropdown is defined within a Page and a Container?

Just adding my thoughts before I forget — I’m not sure if adding separate cases for dropdowns inside a `Page` vs. a `Container` would fully cover it. What actually matters is whether the dropdown has the right fixed pixel width, since those "magic numbers" like `30` and `15` are based on specific pixel values.

For example, if we later change the dropdown width via CSS (like inside a `Container`), the test wouldn’t catch that — it would still assume `15`, but the rendered optionHeight would be off. 🤔

That’s why I was thinking it might make more sense to move the width into the `style` argument of the `Dropdown` itself. That way, it’s easier to test the actual applied width in combination with the magic character counts 😄 . 

But happy to leave it as is for now and just handle the differences in the test. We can always come back to improve the function later — just something to keep in mind since the test would break if those dropdown widths ever change. Not super likely, but good to be aware of.

vizro-core/src/vizro/models/_models_utils.py
petar-qb: Leave it as it though 😄 We'll get back to it in this https://github.com/McK-Internal/vizro-internal/issues/1554

vizro-core/changelog.d/20250612_095418_nadija_ratkusic_graca_dropdown_height_in_containers.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Fixed
+
+- Fix dynamically calculated row height for `Dropdown` when it's used within a container. ([#1238](https://github.com/mckinsey/vizro/pull/1238))
```
petar-qb: Sorry for commenting this again but maybe it's better like this: 
```suggestion
- Fix dynamically calculated row height for `Dropdown` when it's used within `Container.controls`. ([#1238](https://github.com/mckinsey/vizro/pull/1238))
```

Issue #541, opened 2024-06-21
Mobile version layout bugs


l0uden:
### Description

Here's some configurations where layout is not working as expected:

1. Table in one container, graph in second
<img width="294" alt="image" src="https://github.com/mckinsey/vizro/assets/35569332/f4a6c52f-72d0-4392-b678-340486a39cf5">

2. Table in one container, graph in second in horizontal orientation
<img width="843" alt="image" src="https://github.com/mckinsey/vizro/assets/35569332/ecde59a2-1c55-42bf-a40f-0ef1d117a9fa">

3. Two graphs in horizontal orientation plus card
<img width="841" alt="image" src="https://github.com/mckinsey/vizro/assets/35569332/1e165435-3a88-4714-9659-f873d7d634be">

4. Agrid title overlaps second container tab
<img width="295" alt="image" src="https://github.com/mckinsey/vizro/assets/35569332/f694f78f-88a2-42f6-ac78-d738a0facb1e">

5. Agreed is unreachable with another graph in horizontal orientation 
<img width="840" alt="image" src="https://github.com/mckinsey/vizro/assets/35569332/f8f94bf8-8d96-4d9c-90ab-978aa6f8df0d">

6. No graph displayed with two table on the same page
<img width="290" alt="image" src="https://github.com/mckinsey/vizro/assets/35569332/6a03b35f-171e-4597-970b-fec933359b1c">

7. No graph displayed with lots of components in one container
<img width="292" alt="image" src="https://github.com/mckinsey/vizro/assets/35569332/8d41f06a-ca49-453c-ae9e-41250070ac5a">

### Which package?

vizro

### Package version

0.1.17

### Python version

3.9

### OS

Mac, Linux

### How to Reproduce

Just run tests examples from `vizro-qa`

### Code of Conduct

- [X] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

vladimir-mck:
**Page 2 – Layout Behavior Explanation**

There are no bugs present on this page.

The layout behavior observed in landscape mode is expected and stems from the way CSS Grid distributes space among its children. The grid layout is defined using a configuration such as `layout=vm.Grid(grid=[[0], [2], [1]])`, which internally maps components to specific grid areas (e.g., `style="grid-area: 3 / 1 / 4 / 2;"`).

[...]

PR #1239, opened 2025-06-12
[Fix] Use `min-height: 100%` to fix Plotly graph rendering in tab panel

vladimir-mck:
## Description

Setting `height: 100%` on the tab panel causes Plotly to incorrectly trigger its responsive behavior, resulting in the graph rendering with 0 height. Using `min-height: 100%` instead ensures that Plotly has enough space to render the graph properly using default height or .

Closes [#1845](https://github.com/McK-Internal/vizro-internal/issues/1845)

## Notice


petar-qb:
@vladimir-mck Could you add a minimal reproducible example within the `scratch/app.py`?

huong-li-nguyen:
> * e2e vizro screenshots tests / test-e2e-vizro-screenshots for chrome (pull_request)

I haven’t had a chance to look at the code yet, but from the screenshots, it looks like there are a few instances where the chart disappears—especially when it's used inside tabs or similar components 😬 Happy to take another look once the screenshots are fixed! 👍 

Screenshot 2025-06-12 at 13 55 29

vizro-core/src/vizro/static/css/tabs.css
```diff
@@ -16,11 +16,23 @@
 
 /* The dbc component adds an additional div element to which we cannot assign a className.
 To ensure the dynamic height adjustment and prevent scrolling, the height must be specified for that div as below. */
+
+/* .tab-pane,
+.tab-pane > div:first-child, */
```
petar-qb: This looks like should be removed.

vizro-core/tests/e2e/screenshots/chrome_mobile/main_homepage_mobile.png
petar-qb, line 1: The other screenshots seem fine, but this one looks broken. Maybe I missed something. 
What do you think @huong-li-nguyen?

vizro-core/src/vizro/static/css/tabs.css
```diff
@@ -16,11 +16,23 @@
...
 .tab-pane,
 .tab-pane > div:first-child {
   height: 100%;
 }
 
+.flex-item > .tabs-container > .tab-content > .tab-pane {
+  height: 100%;
+}
```
petar-qb: Would it work and if so, would it be better to define this selector as:
```suggestion
.flex-item .tab-pane {
  height: 100%;
}
```

vizro-core/src/vizro/static/css/tabs.css
```diff
@@ -16,11 +16,23 @@
...
 To ensure the dynamic height adjustment and prevent scrolling, the height must be specified for that div as below. */
+
+/* .tab-pane,
+.tab-pane > div:first-child, */
 .tab-pane,
 .tab-pane > div:first-child {
   height: 100%;
 }
 
+.flex-item > .tabs-container > .tab-content > .tab-pane {
```
huong-li-nguyen: Could you just add a comment in there why we added this now? :)

vizro-core/tests/e2e/screenshots/chrome_mobile/main_homepage_mobile.png
huong-li-nguyen: I don’t think this issue is caused by this PR. The problem is that on mobile, the layout switches to `flex`, but the parent container remains a `grid-item` - since we set those when the dashboard is initialized. As a result, the cards end up taking 100% height and because we set `flex: 1`, this is now visible 🤔

@vladimir-mck — feel free to ignore this for now in this PR. Let’s address it as part of this issue instead: https://github.com/mckinsey/vizro/issues/541

huong-li-nguyen (approved):
This looks good to me 👍

vizro-core/tests/e2e/screenshots/chrome_mobile/main_homepage_mobile.png
vladimir-mck: @petar-qb  @huong-li-nguyen  In a previous PR I added min height 250px to dash graph and updated the mobile home page screenshot, but now I delete the min height and it reverted back to it's original look.

PR #1241, opened 2025-06-16
[Fix] `optionHeight` for dynamic dropdown in container

antonymilne:
## Description

A bugfix for a bugfix 😀 Initial bugfix has not been released yet hence no changelog entry here.

https://github.com/mckinsey/vizro/pull/1238 has a very minor problem that it doesn't handle dynamic data correctly. Try moving the "Change me to see bug slider" to a larger number and look how the long dropdown option behaves inside the container.

**Previously**: option height has `char_count = 15` set in pre-build but then switches to `30` on page load when control is in container (but works ok for control in page). This results in overflow:
image

**Now**: option height is recalculated for both control in container and control in page.
image

Aside from the bug, this is part of the work on https://github.com/McK-Internal/vizro-internal/issues/1829 that I was planning to do anyway. Luckily @nadijagraca's fix actually raises this great example which limits the possible solutions here about where and when we set `inline=True` for controls in Containers. Here's what I decided:
* logic for _how selector behaves when it's in a container_ depends on the individual selector and hence belongs inside the relevant selector model (`vm.RadioItems` etc.)
* logic for _is the selector in a container_ lives in `vm.Container` so there's no need to navigate model manager in any way e.g. we don't need to work out "am I in a container" from inside a `vm.Filter` or control. This information passes directly from `vm.Container` to `control.selector` without touching the intermediate `control` itself (unlike `vm.Filter._dynamic`).

FYI @maxschulz-COL hopefully these seems like a sensible approach still with your model manager considerations but we can change in future if it doesn't make sense any more and e.g. we want to navigate up the model hierarchy instead of down.

For additional context - in #1094 we said this:
image

## Notice


vizro-core/src/vizro/models/_components/form/checklist.py
```diff
@@ -73,6 +73,7 @@ class Checklist(VizroBaseModel):
     ]
 
     _dynamic: bool = PrivateAttr(False)
+    _in_container: bool = PrivateAttr(False)
```
antonymilne, line 76: This is a bit like `in_navbar` for [`dbc.DropdownMenu`](https://www.dash-bootstrap-components.com/docs/components/dropdown_menu/) with the major advantage that the user doesn't need to set it manually because we know whether the control is in a container or not.

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -124,7 +145,11 @@ def _action_inputs(self) -> dict[str, _IdProperty]:
 
     def __call__(self, options):
         full_options, default_value = get_options_and_default(options=options, multi=self.multi)
-        option_height = _calculate_option_height(full_options)
+        # 30 characters is roughly the number of "A" characters you can fit comfortably on a line in the page dropdown
+        # (placed on the left-side). 15 is half this width for when the dropdown is in a container's controls.
+        # "A" is representative of a slightly wider than average character:
+        # https://stackoverflow.com/questions/3949422/which-letter-of-the-english-alphabet-takes-up-most-pixels
+        option_height = _calculate_option_height(full_options, 15 if self._in_container else 30)
```
antonymilne, line 152: Note this is now recalculated with 15 or 30 every time the page is refreshed and `__call__` is executed. Previously it would only be set to 15 upfront and then switch back to 30 after that.

antonymilne:
@maxschulz-COL what has happened to the pycafe link? 🤔

huong-li-nguyen (approved):
LGTM - thanks! 👍

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -111,6 +111,11 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
             **({"description": f"{self.description.id}-text.children"} if self.description else {}),
         }
 
+    @_log_call
+    def pre_build(self):
+        for control in self.controls:
+            control.selector._in_container = True
```
maxschulz-COL, line 119: I cannot predict yet where things are going to go, but in the grand scheme of things, this seems quite natural to me, so happy to bet on this approach for now

maxschulz-COL (approved):
LGTM! THanks for the fix and I think we should go with the downward approach for now

vizro-core/src/vizro/models/_components/container.py
antonymilne: Also I just realised this introduces another `pre_build` ordering dependency. We had already enforced the correect order so it doesn't cause any new problems or need more additional logic but is just another example. See comments in https://github.com/mckinsey/vizro/pull/1241/commits/e7ab4624f4b38aa761d1a542253a02eb1681c746.

antonymilne:
@petar-qb Tests are all done now if you want to review again 🙂

vizro-core/src/vizro/models/_components/form/checklist.py
nadijagraca, line 76: This will make my life easier with the styling of checklist inside container for Select ALL  PR. 🙌

nadijagraca (approved):
Looks good! 🙌

PR #1243, opened 2025-06-17
[Feat] URL query parameters for controls

petar-qb:
Closes https://github.com/McK-Internal/vizro-internal/issues/1750

Fix for not losing external unknown URL parameters -> https://github.com/mckinsey/vizro/pull/1254
Closed Draft PR that used a different approach -> https://github.com/mckinsey/vizro/pull/1209

## For reviewers:
@maxschulz-COL It would be fantastic if you could review the test cases listed in `scratch_dev/app.py` below the `TODO-FOR-REVIEWER` section. Any feedback beyond that is, of course, very welcome but not required, as @antonymilne and @vladimir-mck have already reviewed and will be taking another pass. Thanks in advance!

## Description
By setting the `show_in_url=True` in Filter/Parameter, the control's selected value will be synced with the URL query parameter. 

## Screenshot
image

## Testing 
The most of the testing will be done with the new JS test and e2e tests. There's no much Python code changed in the PR and changes made in `_page.py` can't be tested with the unit tests.

### TODO:
- [ ] - align with @l0uden about the e2e test cases
- [ ] - align with @vladimir-mck about the JS tests

## Notice


vizro-core/changelog.d/20250617_102622_petar_pejovic_url_query_params_overwrite_persistence.md
```diff
@@ -0,0 +1,46 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+### Highlights ✨
+
+- Enable `show_in_url` inside `Filter` and `Parameter`. Set it to `True` to show the control's value as the URL query parameter. See the [user guide on filters](https://vizro.readthedocs.io/en/stable/pages/user-guides/filters/#show-in-url) for more information. ([#1243](https://github.com/mckinsey/vizro/pull/1243))
```
petar-qb: Add `filters/#show-in-url` and `parameters/#show-in-url` docs sections.

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,60 +1,274 @@
...
+    2.2 Change control values:
+        - Set controls to ["setosa", "versicolor"] & [3, 4.4]
+        - ✅ Confirm: URL updates accordingly. WyJzZXRvc2EiLCJ2ZXJzaWNvbG9yIl0 & WzMsNC40XQ
+
+    2.3 Refresh the page for couple of times:
+        - ✅ Confirm: Selected values persist after refresh (state is preserved from URL).
+
+    2.4 Copy and paste URL into a new tab:
+        - ✅ Confirm: Page loads with correct filter state applied (["setosa", "versicolor"], & [3, 4.4])
+            - ⚠️ RangeSlider number inputs above currently don't work correctly [2, 4.4] is set but should be [3, 4.4].
```
petar-qb, line 39: We can get back to this minor inconsistency when Dash bugfix is released as a part of this ticket -> https://github.com/McK-Internal/vizro-internal/issues/1545.

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,60 +1,274 @@
...
+        - http://localhost:8050/page_4?page_4_filter_sepal_width=WzMsNC40XQ&page_4_dfp=InZlcnNpY29sb3Ii
+            - ✅Results: ["ALL"], [3, 4.4], "versicolor"
+            - ✅Confirm that the `page_4_filter_species=WyJBTEwiXQ` is added to the URL.
+        (unknown query parameter ID)
+        - http://localhost:8050/page_4?UNKNOWN=InZlcnNpY29sb3Ii
+            - ✅Results: ["ALL"], [2.3, 4.4], "setosa"
+            - ✅Confirm that the `UNKNOWN` still exists in the URL.
+            - ✅Confirm that the URL contains all other controls.
+            - ✅Confirm that the changing the control neither page refresh does not remove UNKNOWN from the URL.
+            - ❗the UNKNOWN parameter is not removed from the URL when navigating to Page-1 and back to Page-4.
```
petar-qb, line 102: Try to solve this by overwriting the nav-links dynamically.

vizro-core/src/vizro/models/_components/form/range_slider.py
```diff
@@ -116,7 +116,7 @@ def __call__(self, min, max, current_value):
         output = [
             Output(f"{self.id}_start_value", "value"),
             Output(f"{self.id}_end_value", "value"),
-            Output(self.id, "value"),
+            Output(self.id, "value", allow_duplicate=True),
```
petar-qb, line 119: This code will be simplified by solving the https://github.com/McK-Internal/vizro-internal/issues/1545 ticket. Same for Slider clientside_callback

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -221,25 +222,25 @@ def pre_build(self):
     def build(self):
         # Cast is justified as the selector is set in pre_build and is not None.
         selector = cast(SelectorType, self.selector)
+
         selector_build_obj = selector.build()
-        # TODO: Align the (dynamic) object's return structure with the figure's components when the Dash bug is fixed.
-        #  This means returning an empty "html.Div(id=self.id, className=...)" as a placeholder from Filter.build().
-        #  Also, make selector.title visible when the filter is reloading.
-        if not self._dynamic:
-            return selector_build_obj
```
petar-qb: I think we should define the `filter.id` in the DOM anyway. This gives users a straightforward way to completely hide it using CSS, without writing awkward selectors like:
```css
div:has(> #hidden_selector_id) {
    display: none;
}
```

Maybe the only question is whether we should wrap a static Filter in `dcc.Loading` or just a regular `html.Div`. I don’t think `dcc.Loading` does any harm, even if the content is static. (same for vm.Parameter)

vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js
```diff
@@ -25,6 +25,8 @@ function gateway(
   ) {
     // If there's no more actions to execute, stop the loop perform.
     if (remaining_actions.length == 0) {
+      console.log("ALL ACTIONS EXECUTED.");
```
petar-qb: Remove all logs before merging.

vizro-core/src/vizro/static/js/models/dashboard.js
```diff
@@ -32,7 +32,7 @@ function update_graph_theme(figure, theme_selector_checked, vizro_themes) {
 function collapse_nav_panel(n_clicks, is_open) {
   if (!n_clicks) {
     /* Automatically collapses left-side if xs, s and m-devices are detected*/
-    if (window.innerWidth < 768 || window.innerHeight < 768) {
+    if (window.innerWidth < 68 || window.innerHeight < 68) {
```
petar-qb: Revert this before merging

petar-qb: Off-topic heads-up: FYI @huong-li-nguyen @antonymilne

I always make this change when working on a PR because dev tools are usually open while I'm testing, and I prefer the panel at the bottom 😄. I'm wondering do others find the hardcoded behaviour frustrating too. There were a very few users who asked for that already.

It’d be great to make the control panel collapsible, like we do with Container (and likely will with global-controls-panel/header/footer/etc.). I’d propose it now if there were a clear place (like a `ControlPanel` model) to define this config, but there isn’t yet.

vizro-core/examples/scratch_dev/app.py
petar-qb: The alternative approach: Keep these params in the dcc.Store (probably together with other data like global control values and so on). The code is currently written so that this kind of logic won’t be too hard to add.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -90,6 +90,7 @@ class Filter(VizroBaseModel):
         "If none are given then target all components on the page that use `column`.",
     )
     selector: Optional[SelectorType] = None
+    show_in_url: bool = False
```
antonymilne: Remember to add this to the docstring.

vizro-core/src/vizro/models/_controls/filter.py
antonymilne: Agree 👍 If it doesn't cause any problems then let's just always wrap in `dcc.Loading`.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -221,25 +222,25 @@ def pre_build(self):
...
-        if f"{selector.id}_end_value" in selector_build_obj:
-            selector_build_obj[f"{selector.id}_end_value"].className = "d-none"
+        if self._dynamic:
+            selector_build_obj[selector.id].className = "invisible"
+            if f"{selector.id}_start_value" in selector_build_obj:
+                selector_build_obj[f"{selector.id}_start_value"].className = "d-none"
+            if f"{selector.id}_end_value" in selector_build_obj:
+                selector_build_obj[f"{selector.id}_end_value"].className = "d-none"
 
+        # TODO: Align the (dynamic) object's return structure with the figure's components when the Dash bug is fixed.
```
antonymilne, line 257: Is this still relevant if static filters are also inside `dcc.Loading`?

vizro-core/src/vizro/models/_page.py
```diff
@@ -167,6 +167,31 @@ def pre_build(self):
                 )
             ]
 
+        # Define a clientside callback that syncs the URL query parameters with controls that have show_in_url=True.
```
antonymilne, line 170: Curious why you chose to do this in `Page.pre_build`. It's probably a good decision but did you consider and reject alternatives?

@maxschulz-COL this relates to .

I think the possible options here would be:
1. single dashboard-level callback (doesn't matter if defined in `pre_build` or `build` since it will only get run once). Maybe not possible until Dash releases `allow_optional=True`.
2. page-level callback defined in
  a.  `Page.pre_build`
  b. `Page.build`
3. control-level callback
  a.  `pre_build`
  b. `build`

Here you've done option 2a.

Options 2b and 3b will suffer from the problem @maxschulz-COL is looking at in https://github.com/McK-Internal/vizro-internal/issues/1864 so probably best to avoid.

3a actually feels cleanest to me but maybe doesn't work well if multiple controls are changed simultaneously.

vizro-core/src/vizro/models/_page.py
```diff
@@ -167,6 +167,31 @@ def pre_build(self):
...
+            control_ids_states = [State(control.id, "id") for control in url_controls]
+
+            clientside_callback(
+                ClientsideFunction(namespace="page", function_name="sync_url_query_params_and_controls"),
+                Output(f"{ON_PAGE_LOAD_ACTION_PREFIX}_trigger_{self.id}", "data"),
+                *selector_values_outputs,
+                *selector_values_inputs,
+                *control_ids_states,
+            )
+
```
antonymilne, line 200: Maybe worth a comment here on why we don't use `url.search` as either the `Input` or `Output` and use the `window.location.search` instead.

vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js
antonymilne: I'd actually quite like to leave some debug messages wherever you think they might be useful, just like we do with Python `logger.debug`.

I don't know whether `console.log` would be the right way to do that though, maybe @vladimir-mck can advise.

vizro-core/src/vizro/static/js/models/dashboard.js
antonymilne: I think this is probably like this because otherwise if you open a vizro dashboard on a small screen then you would just see the navigation which isn't great.

I think the problem with settings `collapsible=True/False` for the control panel would then be that maybe you want different behaviour depending on screen size, e.g. open by default on big screen and collapsed by default on small screens (basically the choice we've made for the user). So you'd need options like "always collapse on first page" or "collapse if screen size is less than x" etc. which sounds kind of complicated.

antonymilne (commented):
Some initial comments, now I'm going to play around with it properly.

vizro-core/src/vizro/models/_page.py
antonymilne: Here's a thought experiment that maybe justifies my instinct that 3a is maybe the better solution if it works 🤔 How would we extend the code to allow other things in the URL e.g. for a `Tabs` model we might want `show_in_url=True` to show the active tab in the URL.

With option 3a you would add a new callback to `Tabs.pre_build` that enables this. With options 1 and 2 it's more difficult because those are hardcoded to handle controls which have special behaviour like the control id is not the same as the selector id. Not impossible to do I'm sure, but it feels awkward because you'd need to handle different schemes in one single callback.

Definitely no need to change this for now, I'm very happy to keep to option 2a and we can always move in future if we need to. Just curious in your thoughts.

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  return { ids, values };
+}
+
+function sync_url_query_params_and_controls(...values_ids) {
+  // Control IDs are required due to Dash's limitations on clientside callback flexible signatures, so that we can:
+  //   1. Map url query parameters to control selector value outputs properly.
+  //   2. Map control selector value input to url query parameters properly.
+  // The Solution relies on the fact that the order of control IDs matches the order of the
```
antonymilne:
```suggestion
  // The solution relies on the fact that the order of control IDs matches the order of the
```

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -0,0 +1,104 @@
...
+
+    const { ids: decodedParamIds, values: decodedParamValues } =
+      decodeUrlParams(urlParams);
+
+    // Overwrite control selector input and output values with the values from the URL.
+    controlIds.forEach((controlId, i) => {
+      const index = decodedParamIds.indexOf(controlId);
+      // Update only if the control ID is found in the URL parameters.
+      if (index !== -1) {
+        controlValues[i] = decodedParamValues[index];
```
antonymilne: Why do we need to update `controlValues` here?

vizro-core/src/vizro/models/_page.py
antonymilne: I see this is explained for `Output` in the javascript function. Do we do `window.location.search` instead of `Input` just to match that?

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -0,0 +1,104 @@
...
+        outputSelectorValues[i] = decodedParamValues[index];
+      }
+    });
+  } else {
```
antonymilne, line 138: Should we do an early `return` in this `if` statement or put all the stuff that follows in the `else`? I don't think any of the following code after the `else` is relevant in the `isPageOpened` case other than the `return`?

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -0,0 +1,104 @@
+function encodeUrlParams(ids, values) {
+  return ids.map((id, i) => [
```
antonymilne: Some comments here and in the `decode` function would be nice. Maybe write the equivalent Python code which most people reading this would understand better 😅

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -0,0 +1,104 @@
...
+    const bytes = Uint8Array.from(str, (c) => c.charCodeAt(0));
+    return JSON.parse(new TextDecoder().decode(bytes));
+  }
+
+  const ids = [];
+  const values = [];
+
+  for (const [key, val] of params.entries()) {
+    ids.push(key);
+    // Silent the error if the value cannot be decoded. It's added for external query parameters.
```
antonymilne: Would it be easy to change the logic here a little bit so that we only try to `decodeParam` in the case that the id is something we've explicitly specified in `controlIds`? Instead of looping through the whole URL and trying to decode everything, just select the keys from `controlIds` and only decode those?

I'm just thinking of the edge case where someone has an external parameter that they actually want to be base64 encoded. It would be pretty annoying if we decoded that for them.

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -0,0 +1,104 @@
...
+  //   2. Map control selector value input to url query parameters properly.
+  // The Solution relies on the fact that the order of control IDs matches the order of the
+  // control selector value inputs and their corresponding outputs.
+
+  // Split control inputs and selector values that are in format:
+  // [selector-1-value, selector-2-value, selector-N-value, ..., control-1-id, control-2-id, control-N-id, ...]
+  const count = values_ids.length / 2;
+  let controlValues = values_ids.slice(0, count);
+  const controlIds = values_ids.slice(count);
+
```
antonymilne, line 102: I wonder if some of the code that follows would be a bit simpler if you re-formed the dictionary:
```
Object.fromEntries(controlIds.map((id, i) => [id, controlValues[i]]));
```

Might simplify the `forEach` stuff below a bit or might make it more complicated, I'm not sure...

antonymilne: Or maybe a `Map`? I don't know enough about JS data structures to comment really 😅

antonymilne (commented):
JS code reviewed and looks great as far as I can tell, but I defer to @vladimir-mck for a proper review there.

vizro-core/examples/scratch_dev/app.py
antonymilne: For now definitely let's just keep them in the URL I think.

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,39 +1,274 @@
...
+            - 🔒(expected) It only works with the latest and unreleased Dash version.
+
+    5.3. Apply filter interaction from AgGrid (by selecting some value from the "Sepal_length" column):
+        - ✅ Confirm: Graph updates
+
+    5.4 Copy and paste URL into a new tab:
+        - ✅ Confirm: Filters/parameters are applied from URL
+        - Note: Grid interaction is not currently reflected in the URL, New "interact" action will solve that.
+
+    5.5 Test order of URL parameters (Open each link in the new browser tab + Navigate to Page-1 and back to Page-4):
```
antonymilne, line 83: These tests under point 5 do not behave as you have documented - please could you check them?

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,39 +1,274 @@
...
 
-dashboard = vm.Dashboard(pages=[page])
+dashboard = vm.Dashboard(pages=[page_1, page_2, page_3, page_4])
 
 if __name__ == "__main__":
-    Vizro().build(dashboard).run()
+    app = Vizro().build(dashboard)
+
+    # To enable custom drill-through actio)
+    app.dash.layout.children.append(dcc.Location(id="vizro_url_callback_nav", refresh="callback-nav"))
```
antonymilne, line 353: Let's try and build this into Vizro properly next sprint as part of https://github.com/McK-Internal/vizro-internal/issues/1656.

vizro-core/changelog.d/20250617_102622_petar_pejovic_url_query_params_overwrite_persistence.md
antonymilne:
```suggestion
- The current state of your app can be shared by sharing its URL. Set `show_in_url=True` inside `Filter` and `Parameter` to include their values in the URL. See the [user guide on filters](https://vizro.readthedocs.io/en/stable/pages/user-guides/filters/#show-in-url) for more information. ([#1243](https://github.com/mckinsey/vizro/pull/1243))
```

antonymilne: Maybe also something higher-level in the docs about how Vizro apps have a shareable URL that works globally across pages. Not sure where it would go though 🤔

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -0,0 +1,104 @@
...
+      ),
+    )
+      .replace(/\+/g, "-")
+      .replace(/\//g, "_")
+      .replace(/=+$/, ""),
+  ]);
+}
+
+function decodeUrlParams(params) {
+  function decodeParam(encoded) {
```
vladimir-mck: Extract the function outside of `decodeUrlParams`, in the current implementation the function will be created every time `decodeUrlParams` is invoked. 

Also select the code block and use copilot -> Generate docs

Screenshot 2025-06-20 at 14 09 57

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -0,0 +1,104 @@
+function encodeUrlParams(ids, values) {
```
vladimir-mck: Add JsDocs here as well

vizro-core/src/vizro/static/js/models/page.js
vladimir-mck: Is it possible to get this part in an object or Map ?

vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js
petar-qb: The `console.debug` could be used instead, which is hide logs by default in Chrome dev tools. To see these messages, the `Verbose` log level must be enabled.

image

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: Yes as the comment just applies to the dynamic filter.

vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js
antonymilne: That sounds perfect to me 👍

vizro-core/src/vizro/models/_page.py
petar-qb: Yep, I added it to match the outputs and simplify input value handling. (added these reasons in the comment)

If I remember correctly (I'm not sure even it was just a few days ago 😅), it also helps with managing multiple clientside callback triggers. Using `dcc.Location` as an input could lead to multiple unintended triggers: a control change triggers the callback -> it updates the URL -> which triggers the callback again. This avoids that loop.

vizro-core/src/vizro/static/js/models/page.js
petar-qb: I did it just to reduce the number of lines by modifying the code directly. Alternatively, I can create a separate `newControlValues` list and use that instead. Let me know if you prefer that approach

In that case, `controlValues` would be replaced by `newControlValues` in this section.
```js
  // Update urlParams with updated control values
  const encodedControlsIdsValues = encodeUrlParams(controlIds, controlValues);
  encodedControlsIdsValues.forEach(([id, value]) => urlParams.set(id, value));
```

vizro-core/src/vizro/static/js/models/page.js
petar-qb: Actually, the code below needs to run in both cases, regardless of whether `isPageOpened` is true or not. The URL should be updated every time the function is triggered because it must reflect the current state in all scenarios, such as:
1. when a control is changed
2. When the page is opened with no (e.g. nav to page-x) or partially specified controls in the URL (e.g. the `drill-through` case or the `5.5:(missing species filter)` example from the scratch example)

vizro-core/src/vizro/static/js/models/page.js
petar-qb: Is the `b64_` prefix in the url param value the solution here?

antonymilne: That would be better than it is now but I think even better if we just don't touch external parameters at all and only loop through the ones we're responsible for. If that's difficult for some reason then it's not a big deal though.

vizro-core/src/vizro/static/js/models/page.js
petar-qb: Did it in -> https://github.com/mckinsey/vizro/pull/1243/commits/ce2053918a3d1229a5aef6b374ab1b9769c34a97

vizro-core/src/vizro/static/js/models/page.js
petar-qb: I refactored the code so it now only decodes URL parameter values that start with the `b64_` prefix. However, it still iterates over all `urlParams` rather than limiting iteration to just the `controlIds`. Narrowing the loop to iterate only through the`controlIds` would break the property where `encodeUrlParams(decodeUrlParams(urlParams))` is equal to `urlParams`. I consider this property important, as it helps maintain idempotency and reduces the chance of introducing subtle, error-prone behaviour in the codebase.

I'll let you know once the PR is ready for final review, and you can take another look then. Or feel free to take a look now if you’re interested.

petar-qb:
Hey @antonymilne 👋 

> empty value like ?filter_id= seems to work ok for dropdown (value is replaced by IiI which I guess corresponds to some null value or similar) but crashes the range slider ("Invalid argument value passed into RangeSlider")
similarly for manually entered invalid value like ?filter_id=invalid - gives empty dropdown but crashes range slider

With the changes in [this commit](https://github.com/mckinsey/vizro/pull/1243/commits/dfbdf6577ace25e61c5d0a962d6e91634c8c22e3), most cases seem to be handled now. The only one still not handled is when a valid base64-encoded and prefixed value like `b64_IkFMTCI` (which turns into `["ALL"]`) is passed to the range slider. This case is not easy to cover because the JS script doesn't know  about the valid range slider value (it actually doesn't know that the range slider is the output 😄). 

Other cases like invalid or non-prefixed values seem to be working correctly.

vizro-core/examples/scratch_dev/app.py
petar-qb: Ah.. I forgot to mention that it only works with the latest unreleased Dash version. With the latest Dash release, dynamic Checklist filter is always set to `["ALL"]` which is incorrect.

vizro-core/src/vizro/static/js/models/page.js
petar-qb: The `encodeUrlParams` and `decodeUrlParams` functions are refactored and the helper function `decodeParam` is removed. Documentation is added as suggested in [this comment](https://github.com/mckinsey/vizro/pull/1243#discussion_r2156625549).
Let me know if everything looks good now.

vizro-core/src/vizro/static/js/models/page.js
petar-qb: Answer is here -> https://github.com/mckinsey/vizro/pull/1243#discussion_r2163066268

vizro-core/src/vizro/static/js/models/page.js
antonymilne: Ah ok, this makes sense. My thinking was that it's relevant for case 1 obviously, and that case 2 would be handled by #1254.

But:
* partially specified controls in the URL are a good example of when this would be needed
* until we have #1254 this is obviously needed
* even if and when if we do #1254 this might still be useful to do because it means the drill-through doesn't need to pass all the URL params but just the single relevant one

antonymilne: Adding some comment about this would be nice 🙂

vizro-core/src/vizro/static/js/models/page.js
antonymilne: Fine by me, although I don't understand why this would break that `encodeUrlParams(decodeUrlParams(urlParams)) == urlParams` so I think we aren't quite on the same page here 🤔 

I'd expect something like this:
```
# <stuff> is the b64 encoded version of "plain_text"
# vizro_ parameters are things we control and use, unknown are external ones
urlParams = "vizro_1?=plain_text&vizro_2=b64_<stuff>&unknown_1=plain_text&unknown_2=b64_<stuff>"
# vizro b64_ gets decoded, anything unknown gets ignored
decode(urlParams) = {"vizro_1": "plain_text", "vizro_2": "plain_text", "unknown_1": "plain_text", "unknown_2": "b64_<stuff>"}
# vizro_2 gets encoded, anything unknown gets ignored
encode(decode(urlParams)) = "vizro_1?=plain_text&vizro_2=b64_<stuff>&unknown_1=plain_text&unknown_2=b64_<stuff>"
        == urlParams 
```

At the moment we don't have anything like `vizro_1` since the only thing we put in the URL are bae64 encoded controls, but maybe in the future we would have something encoded differently or not encoded at all (like `plain_text` here).

antonymilne:
> Hey @antonymilne 👋
> 
> > empty value like ?filter_id= seems to work ok for dropdown (value is replaced by IiI which I guess corresponds to some null value or similar) but crashes the range slider ("Invalid argument value passed into RangeSlider")
> > similarly for manually entered invalid value like ?filter_id=invalid - gives empty dropdown but crashes range slider
> 
> With the changes in [this commit](https://github.com/mckinsey/vizro/pull/1243/commits/dfbdf6577ace25e61c5d0a962d6e91634c8c22e3), most cases seem to be handled now. The only one still not handled is when a valid base64-encoded and prefixed value like `b64_IkFMTCI` (which turns into `["ALL"]`) is passed to the range slider. This case is not easy to cover because the JS script doesn't know about the valid range slider value (it actually doesn't know that the range slider is the output 😄).
> 
> Other cases like invalid or non-prefixed values seem to be working correctly.

Cool, the error `try/catch` definitely seems like the right thing to do here 👍 FYI it also covers the extremely edge and unimportant case that in the future we might have a plain-text parameters in the URL and the actual plain text value starts with `b64_` but _isn't_ actually meant to be a `b64` encoded string. e.g. for some reason it's plain text`b64_c64_d64`.

antonymilne:
Ah, one more thing I forgot to mention before: should issue either a warning or an error if you set `show_in_url=True` but haven't set the `id`? Possible reason:
* more important: a shareable URL won't work across server restarts if the control id randomly changes*
* less important: the URL looks weird with the automatically generated uuid control ids

*I just tried this out quickly and actually it seems to work ok because we fix a random seed. This solves the multi-process case but I'm actually a bit confused how/whether it solves app restart case and would feel happier if we could understand what's happening.

vizro-core/src/vizro/models/_page.py
petar-qb: As you said, most of the options are quite unfeasible before https://github.com/McK-Internal/vizro-internal/issues/1864 is done or before the `allow_optional=True` is released. 

Also, it can't work at the dashboard level because `allow_optional` will only be enabled for `State` and `Input`, but not for `Output`. We would need `allow_optional` for `Output` as well, because we update selector values based on the URL.

#### 3a. control-level pre_build 

This one looks really tempting, and yes, I agree that it won't be easy to add the support for the "Tab-in-url". 

However, I was like: It's easier to write the single orchestrator for handling all the precedences between: 
1. server returned value,
2. value from the URL,
3. dash persisted value,
4. value from the dcc.Store like the one that's used in Slider and RangeSlider,
5. upcoming vizro-store values

then to implement the choreography of clientside callbacks, and praying that it would work deterministically on every machine 😄. 

Another use case came on my mind that deflected me of even trying to implement this case: If two controls are changed at the same time, one callback would overwrite the url in one way and another callback would overwrite it in another way which would lead to the incorrect URL state.

I'm aware that we're trying to improve the actions loop and that this additioanl orchestration logic is probably the one step back, but I just didn't want to mess with the order of callback execution right now. When the persistence bug become released, some of code would be able to refactor (like update-slider-values mechanism) which will make easier to presume the order of callbacks. I really hope that we will then find the solution for the choreography of callbacks that will work deterministically and as expected, and that we will be able to get rid of any custom callback orchestration logic.

I'll link this comment to the new action loop PR, so we remember to try the `3a` approach and see if it works.

antonymilne: Good point about option 1 not being feasible since we'd need `allow_optional` for `Output` as well 👍 

> Another use case came on my mind that deflected me of even trying to implement this case: If two controls are changed at the same time, one callback would overwrite the url in one way and another callback would overwrite it in another way which would lead to the incorrect URL state.

Yes, I came across that problem too and it put me off this solution. Not sure whether there would be a good workaround for that or not 🤔 

> I'm aware that we're trying to improve the actions loop and that this additioanl orchestration logic is probably the one step back, but I just didn't want to mess with the order of callback execution right now. When the persistence bug become released, some of code would be able to refactor (like update-slider-values mechanism) which will make easier to presume the order of callbacks. I really hope that we will then find the solution for the choreography of callbacks that will work deterministically and as expected, and that we will be able to get rid of any custom callback orchestration logic.

Completely agree. I think 2a makes complete sense for now.

vizro-core/src/vizro/static/js/models/page.js
petar-qb: (EDITED)

I made some changes that you can find [here](https://github.com/mckinsey/vizro/pull/1243/files/03e7564c1d7357997f0ef446107ada636b89b8dd..fa53c3e14391adaccf3b2de66b9f3df053a069dc).

Now:

```py
input = {
    'vizro_1': 'raw_value',
    'vizro_2': 'b64_IjEyMyI',
    'foo': 'raw_value',
    'bar': 'b64_IjEyMyI',
    'baz': 'b64_invalid'
}
```

1. `decode(input)` -> throws the error as `applyOnKeys` is not provided.
2. `decode(input, ['vizro_1', 'vizro_2'])` -> {'vizro_2': '123'}
3. `decode(input, ['foo', 'bar', 'baz'])` -> {'bar', '123'}

So:
- Function only tries to decode and return the listed keys within the `applyOnKeys` input.
- Invalid base64 values (like 'baz') are not decoded/returned.
- The `encode` function is similarly implemented

petar-qb: I pushed the new code and edited the message above. The decode/encode function now only try to decode/encode vizro explicitly provided IDs. Decode also handles the incorrectly set url values.

vizro-core/examples/scratch_dev/app.py
petar-qb: Will be done as a part of -> https://github.com/mckinsey/vizro/pull/1254

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: Replaced `dcc.Loading` with a simple `html.Div` wrapper due to minor flickering that's somehow visible even for the static controls.

vizro-core/src/vizro/static/js/models/page.js
antonymilne: This looks perfect now, thanks for all the modifications 🙏

vizro-core/changelog.d/20250617_102622_petar_pejovic_url_query_params_overwrite_persistence.md
petar-qb: I haven't found any proper location so I added some docs within the `filter.md` and `parameter.md`. Here's the change -> https://github.com/mckinsey/vizro/pull/1243/commits/6e91220ecf48af5ca2b6324a43dd641122139dcc

vizro-core/src/vizro/static/js/models/page.js
petar-qb, line 1: Should we rewrite the code comments? This way, it looks a bit ugly 😄

vizro-core/changelog.d/20250617_102622_petar_pejovic_url_query_params_overwrite_persistence.md
```diff
@@ -0,0 +1,46 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+### Highlights ✨
+
+- The current state of your app can be shared by sharing its URL. Set `show_in_url=True` inside `Filter` and `Parameter` to include their values in the URL. See the [user guide on filters](https://vizro.readthedocs.io/en/stable/pages/user-guides/filters/#show-in-url) for more information. ([#1243](https://github.com/mckinsey/vizro/pull/1243))
```
antonymilne:
```suggestion
- The current state of a page in your app can be shared by sharing its URL. Set `show_in_url=True` in `Filter` and `Parameter` to include their values in the URL. See the [user guide on filters](https://vizro.readthedocs.io/en/stable/pages/user-guides/filters/#show-in-url) for more information. ([#1243](https://github.com/mckinsey/vizro/pull/1243))
```

Since we don't (yet?) share global app state, just that individual page.

vizro-core/docs/pages/user-guides/filters.md
```diff
@@ -266,7 +266,77 @@ Below is an advanced example where we only target one page component, and where
 
 To further customize selectors, see our [how-to-guide on creating custom components](custom-components.md).
 
+## Show in URL
+
+The `Filter` model now accepts an optional boolean argument `show_in_url` (default `False`). When `True`, the filter’s value is synchronized with the page URL as a query parameter, allowing the current control's state to be captured in the link. This makes it easy to share or bookmark specific `Page` settings.
```
antonymilne:
```suggestion
The `Filter` model accepts an optional argument `show_in_url`. When `show_in_url=True`, the filter’s value is synchronized with the page URL as a query parameter, allowing the current control's state to be captured in the link. This makes it easy to share or bookmark specific `Page` filter settings.
```

vizro-core/docs/pages/user-guides/filters.md
```diff
@@ -266,7 +266,77 @@ Below is an advanced example where we only target one page component, and where
...
+
+The URL query parameter uses the control’s id as its key, and the selected value, encoded in base64, as the URL query parameter value.
```
antonymilne:
```suggestion
The URL query parameter uses the control’s id as its key and the selected value, encoded in base64, as the URL query parameter value.
```

vizro-core/docs/pages/user-guides/filters.md
```diff
@@ -266,7 +266,77 @@ Below is an advanced example where we only target one page component, and where
...
+
+!!! warning
+
+    - PyCafe incompatibility: Vizro does not currently support external URL query parameters and they could be stripped or lost using the app. As a result, dashboards using this feature should not be embedded or deployed in PyCafe environments.
+
+    - Page-specific only: Only filters on the currently opened page are reflected in the URL. It is not possible to share or bookmark the state of multi-page dashboards.
```
antonymilne:
```suggestion
    - Page-specific only: Only filters on the currently opened page are reflected in the URL. It is not yet possible to share or bookmark the global state of a multi-page dashboard.
```

vizro-core/docs/pages/user-guides/filters.md
```diff
@@ -266,7 +266,77 @@ Below is an advanced example where we only target one page component, and where
...
+
+Here's an example of how to use this feature:
```
antonymilne:
```suggestion
```

vizro-core/docs/pages/user-guides/filters.md
```diff
@@ -266,7 +266,77 @@ Below is an advanced example where we only target one page component, and where
...
+
+!!! example "Filer in URL"
```
antonymilne:
```suggestion
!!! example "Filter in URL"
```

vizro-core/docs/pages/user-guides/filters.md
```diff
@@ -266,7 +266,77 @@ Below is an advanced example where we only target one page component, and where
...
 
+## Show in URL
+
+The `Filter` model now accepts an optional boolean argument `show_in_url` (default `False`). When `True`, the filter’s value is synchronized with the page URL as a query parameter, allowing the current control's state to be captured in the link. This makes it easy to share or bookmark specific `Page` settings.
+
+The URL query parameter uses the control’s id as its key, and the selected value, encoded in base64, as the URL query parameter value.
+
+!!! warning
+
+    - PyCafe incompatibility: Vizro does not currently support external URL query parameters and they could be stripped or lost using the app. As a result, dashboards using this feature should not be embedded or deployed in PyCafe environments.
```
antonymilne: Ah yes, I had forgotten that pycafe was a good reason that we might want to handle #1254 at some point. This is fine for now but a very good thing to realise 👍 

Can we link this to our guide to embedding/deploying on PyCafe? Just to give it a bit more context so a reader who doesn't know what we're talking about can find out more.

vizro-core/docs/pages/user-guides/parameters.md
```diff
@@ -189,5 +189,88 @@ Note that in the above example, one parameter affects multiple targets.
 
 If you use [dynamic data](data.md/#dynamic-data) that can be updated while the dashboard is running then you can pass parameters to the dynamic data function to alter the data loaded into your dashboard. For detailed instructions, refer to the section on [parametrized data loading](data.md/#parametrize-data-loading).
 
+## Show in URL
```
antonymilne, line 192: Similar suggestions to the comments on `filter.md` apply here too obviously.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -90,6 +92,7 @@ class Filter(VizroBaseModel):
         "If none are given then target all components on the page that use `column`.",
     )
     selector: Optional[SelectorType] = None
+    show_in_url: bool = False
```
antonymilne: This should have a `description` that matches the docstring.

antonymilne: (same goes for `Parameter` obviously)

vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js
```diff
@@ -25,6 +25,8 @@ function gateway(
   ) {
     // If there's no more actions to execute, stop the loop perform.
     if (remaining_actions.length == 0) {
+      console.debug("Actions loop: All actions executed.");
+      console.debug("");
```
antonymilne, line 29: Maybe:
```suggestion
```

Or add `\n` to the previous `debug` I guess?

vizro-core/src/vizro/static/js/models/page.js
antonymilne: I'd just put them in `/*` to do a multiline JS comment and then it's all good I think.

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -0,0 +1,156 @@
+// Python equivalent to the following JavaScript code:
+//def encode_url_params(decoded_map, apply_on_keys=None):
+//    encoded_map = {}
+//    for key, value in decoded_map.items():
+//        if key in apply_on_keys:
+//            json_str = json.dumps(value)
+//            encoded_bytes = base64.b64encode(json_str.encode("utf-8"))
+//            encoded_str = encoded_bytes.decode("utf-8") \
```
antonymilne: It would be worth putting a Python comment here that this is the same as what `urlsafe_b64encode` does.

Similarly for the decode function.

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -0,0 +1,156 @@
...
+        decodedMap.set(key, JSON.parse(json));
+      } catch (e) {
+        console.warn("Failed to decode URL parameter:", key, val);
+      }
+    }
+  }
+  return decodedMap;
+}
+
+function sync_url_query_params_and_controls(...values_ids) {
```
antonymilne, line 84: This function looks really great now, at least to someone with my knowledge of JS. Amazing work! All the comments are super useful, thank you for writing them 👍 🙏

vizro-core/examples/scratch_dev/app.py
petar-qb: will be done as a part of this ticket -> https://github.com/McK-Internal/vizro-internal/issues/1923

vizro-core/src/vizro/models/_components/form/range_slider.py
petar-qb: will be done as a part of this ticker -> https://github.com/McK-Internal/vizro-internal/issues/1545

petar-qb:
> Ah, one more thing I forgot to mention before: should issue either a warning or an error if you set show_in_url=True but haven't set the id? Possible reason:
>
> more important: a shareable URL won't work across server restarts if the control id randomly changes*
less important: the URL looks weird with the automatically generated uuid control ids
>
> *I just tried this out quickly and actually it seems to work ok because we fix a random seed. This solves the multi-process case but I'm actually a bit confused how/whether it solves app restart case and would feel happier if we could understand what's happening.

Yes, it should work ok because we fix the random seed. However, if another component is added to the dashboard, the bookmarked/shared link could stop working. So, I agree that some kind of warning would be great. Or we should rather raise an exception, so that users don't confuse when they see an url query parameter that's completely random (illegible key, and illegible value)

(edit)
> if you can confirm that the id is constant across app restarts (and hopefully explain to me why 😅) then all good, no need to do anything

As you said, we fixed the random seed. Also, the python interpreter will initialise Vizro models always in the same order (until the app.py is not changed), and this in a combination with the fixed seed makes that all random-id vizro component will have the same ID. When the app is restarted, the seed is set again to the fixed value, the initialisation happens again in the same order which lead to the same results. Disclaimer: that's how I currently see it and maybe there's something I missed 😄

vizro-core/src/vizro/static/js/actions/build_action_loop_callbacks.js
petar-qb: I tried several combinations like:
- console.debug("Actions loop: All actions executed.\n");
- console.debug(\`Actions loop: All actions executed.\n\`);
- console.debug("Actions loop: All actions executed.", \`\n\`);

but none of these worked.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -90,6 +92,13 @@ class Filter(VizroBaseModel):
         "If none are given then target all components on the page that use `column`.",
     )
     selector: Optional[SelectorType] = None
+    show_in_url: bool = Field(
```
petar-qb, line 95: Investigate raising a warning/exception if the ID is not explicitly specified. Look at the comment -> https://github.com/mckinsey/vizro/pull/1243#issuecomment-2999803418

Apply the same logic to Parameter as well.

petar-qb: FYI: @antonymilne

I removed the control IDs from the filters on page 2, then selected some values and bookmarked the resulting URL. Regardless of how many times I restarted the app, opening the bookmarked URL in a new or incognito tab, the bookmarked link consistently worked as expected.

However, after changing the order of the filters (page-2) in the Vizro configuration, the same URL no longer worked correctly. It attempted to assign ["setosa", "versicolor"] to a range_slider, and [3, 4.4] to a dropdown.

Here’s the link for reference: [page_2-pre_set_controls_link](http://127.0.0.1:8050/page_2?23a7711a-8133-2876-37eb-dcd9e87a1613=b64_WyJzZXRvc2EiLCJ2ZXJzaWNvbG9yIl0&1846d424-c17c-6279-23c6-612f48268673=b64_WzMsNC40XQ)

As expected, this behaviour remains consistent even when running the app with multiple workers (e.g., using Gunicorn with 4 workers).

Here's the code -> https://github.com/mckinsey/vizro/pull/1243/commits/aa79fc8b803386e3d2d0917163c63d28c329a1ca

petar-qb: I just realised that performing the check within `pre_build` isn't ideal for this warning, since the result doesn't rely on other components. A better approach would be to use either `BeforeValidator` (to capture the originally set ID before it gets overwritten by `VizroBaseModel` if that's possible ) or `model_post_init`.

Since we don’t want to modify the inputs nor raise an exception during this step, `model_post_init` seems like the most suitable place for this logic.

Done in -> https://github.com/mckinsey/vizro/pull/1243/commits/290bbde249553b47210b25bddd164878c63f7696

vizro-core/src/vizro/models/_controls/_controls_utils.py
```diff
@@ -40,3 +41,13 @@ def check_control_targets(control: ControlType) -> None:
...
+
+
+def warn_missing_id_for_url_control(control: ControlType) -> None:
+    if control.show_in_url and "id" not in control.model_fields_set:
+        warnings.warn(
+            "The `show_in_url=True` is set but no `id` was provided. "
+            "This can lead to unstable URLs, especially if the dashboard layout changes. "
+            "Set a fixed `id` to ensure links remain consistent and shareable.",
+            UserWarning,
+        )
```
antonymilne:
```suggestion
        warnings.warn(
            "`show_in_url=True` is set but no `id` was provided. "
            "Shareable URLs might be unreliable if your dashboard configuration changes in future. "
            "If you want to ensure that links continue working, set a fixed `id`.",
            UserWarning,
        )
```

vizro-core/src/vizro/models/_controls/filter.py
antonymilne: Definitely better to do this before pre-build to follow our usual strategy of doing the validation as early as possible.

It's a very minor thing, but I would suggest using a validator in preference to `model_post_init` though. This would be a bit more conventional even though we're not modifying values or raising an exception.
```
    @model_validator(mode='after')
    def check_id_set_for_url_control(self) -> Self:
        if self.show_in_url and "id" not in self.model_fields_set:
            ...
        return self
```

antonymilne (approved):

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -0,0 +1,164 @@
...
+  {'vizro_1': 'raw_value', 'vizro_2': 'b64_IjEyMyI', 'foo': 'raw_value', 'bar': 'b64_IjEyMyI', 'baz': 'b64_invalid'},
+  ['vizro_1', 'vizro_2']
+Example output (only vizro_2 is decoded):
+  {'vizro_1': 'raw_value', 'vizro_2': '123', 'foo': 'raw_value', 'bar': 'b64_IjEyMyI', 'baz': 'b64_invalid'}
+*/
+function decodeUrlParams(encodedMap, applyOnKeys) {
+  const decodedMap = new Map();
+  for (const [key, val] of encodedMap.entries()) {
+    if (val.startsWith("b64_") && applyOnKeys.includes(key)) {
+      try {
```
antonymilne, line 69: Do we make the key URL safe at some point? I think I tried it with URL unsafe characters and it worked ok but not sure why since I don't see us encoding stuff here 🤔

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -0,0 +1,164 @@
+/*
+Python equivalent to the following JavaScript code:
+def encode_url_params(decoded_map, apply_on_keys=None):
+    encoded_map = {}
+    for key, value in decoded_map.items():
+        if key in apply_on_keys:
+            # This manual base64 encoding could be simplified with base64.urlsafe_b64encode.
+            # It's kept here to match the javascript implementation.
+            json_str = json.dumps(value)
```
antonymilne:
```suggestion
            json_str = json.dumps(value, separators=(',', ':'))
```

vizro-core/changelog.d/20250617_102622_petar_pejovic_url_query_params_overwrite_persistence.md
```diff
@@ -0,0 +1,46 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+### Highlights ✨
+
+- The current state of a page in your app can be shared by sharing its URL. Set `show_in_url=True` in `Filter` and `Parameter` to include their values in the URL. See the [user guide on filters](https://vizro.readthedocs.io/en/stable/pages/user-guides/filters/#show-in-url) for more information. ([#1243](https://github.com/mckinsey/vizro/pull/1243))
```
maxschulz-COL:
```suggestion
- The current state of controls on a page in your app can be shared by sharing its URL. Set `show_in_url=True` in `Filter` and `Parameter` to include their values in the URL. See the [user guide on filters](https://vizro.readthedocs.io/en/stable/pages/user-guides/filters/#show-in-url) for more information. ([#1243](https://github.com/mckinsey/vizro/pull/1243))
```

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,141 +1,352 @@
-from time import sleep
+"""Dev app to try things out."""
```
maxschulz-COL, line 1: Out of curiosity, does this work with PyCafe? I tried the scratch app and it works, but of course one cant see the url...

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,141 +1,352 @@
...
+"""
+Test cases:
+1. When page is opened without the URL parameter:
+    1.1. The control should default to their persisted values (if persist in the storage) or initial server values.
+    1.2. If the control is show_in_url=True it should overwrite the URL.
+2. When control is changed:
+    2.1. The URL should be updated with the new control values.
+3. When page is opened with the URL parameter:
+    3.1. If the control is show_in_url=True, its should reflect the value from the URL.
+    3.2. 🔒Affected control should update the persistence storage.
```
maxschulz-COL, line 12: What is the persistence storage?

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,141 +1,352 @@
...
 
-@capture("figure")
-def custom_kpi_card(
-    data_frame: pd.DataFrame,
-) -> dbc.Card:
-    """Creates a custom KPI card."""
-    value = data_frame.info()
+# TODO-CHECK: 1. Drill-through to Page-3 - [WORKS]
+@capture("action")
+def custom_drill_through_action(clicked_point):
```
maxschulz-COL, line 147: Is there a plan to make this part of the official API, ie without a custom actions?

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,141 +1,352 @@
...
+    - Go back to Page-2
+    - Click the "versicolor" point on the graph
+
+    4.1 Check redirection:
+        - ✅ Confirm: Values are ["versicolor"] & 2
+        - ✅ Confirm: URL updates to Page-3 with filter set to InZlcnNpY29sb3Ii -> (["versicolor"])
+
+    4.2 Navigate to Page-1 and back to Page-3:
+    - ✅ Confirm: Values are still ["versicolor"] & 2
+        - 🔒(expected) It only works with the latest and unreleased Dash version.
```
maxschulz-COL, line 61: This does not work with the latest dash version for me:

```bash
Using Python 3.13.3
dash==3.1.1
dash-ag-grid==31.3.1
dash-bootstrap-components==2.0.3
dash-mantine-components==2.0.0
dash-testing-stub==0.0.2
```

When going back to Page 3, it flips back to ALL

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,141 +1,352 @@
...
 
 
-@capture("ag_grid")
-def custom_grid(data_frame: pd.DataFrame) -> dag.AgGrid:
-    defaults = {
-        "columnDefs": [{"field": col} for col in data_frame.columns],
-        "rowData": data_frame.to_dict("records"),
-    }
-    return dag.AgGrid(**defaults)
+def encode_to_base64(value):
```
maxschulz-COL, line 139: This should be a public util in any case I think

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,141 +1,352 @@
...
+
+    4.2 Navigate to Page-1 and back to Page-3:
+    - ✅ Confirm: Values are still ["versicolor"] & 2
+        - 🔒(expected) It only works with the latest and unreleased Dash version.
+
+
+5. Page-4: Dynamic Filters, DFP, and AgGrid Interaction
+    - Navigate to Page-4 (testbed for dynamic filter, DFP and AgGrid filter interaction)
+
+    5.1 Change:
```
maxschulz-COL, line 67: I also tested applying a URL even though the underlying Parameters do not have the option anymore. This works. I guess what happens is that the URL updates the controls, the controls then load the correct data, and thus any other control referring to that dynamic data is also correct.

Is there any race conditions here? Like does the order of URL parameter application matter?

And also, if I encoded URL parameters myself, could i create an invalid dashboard state?

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,141 +1,352 @@
...
+    5.5 Test order of URL parameters (Open each link in the new browser tab + Navigate to Page-1 and back to Page-4):
+    - 🔒(expected) It only works with the latest and unreleased Dash version.
+        Test cases:
+        (URL in the same order as control outputs)
+        - http://localhost:8050/page_4?page_4_filter_species=b64_WyJ2ZXJzaWNvbG9yIl0&page_4_filter_sepal_width=b64_WzMsNC40XQ&page_4_dfp=b64_InZlcnNpY29sb3Ii
+            - ✅Results: ["versicolor"], [3, 4.4], "versicolor"
+        (different order)
+        - http://localhost:8050/page_4?page_4_filter_sepal_width=b64_WzMsNC40XQ&page_4_filter_species=b64_WyJ2ZXJzaWNvbG9yIl0&page_4_dfp=b64_InZlcnNpY29sb3Ii
+            - ✅Results: ["versicolor"], [3, 4.4], "versicolor"
+        (missing species filter)
```
maxschulz-COL, line 92: Why does it not keep the previous species filter? Are we killing the entire storage by having a single URL parameter?

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,141 +1,352 @@
...
+        - http://localhost:8050/page_4?page_4_filter_sepal_width=b64_WzMsNC40XQ&page_4_dfp=b64_InZlcnNpY29sb3Ii
+            - ✅Results: ["ALL"], [3, 4.4], "versicolor"
+            - ✅Confirm that the `page_4_filter_species=WyJBTEwiXQ` is added to the URL.
+        (unknown query parameter ID)
+        - http://localhost:8050/page_4?UNKNOWN=InZlcnNpY29sb3Ii
+            - ✅Results: ["ALL"], [2.3, 4.4], "setosa"
+            - ✅Confirm that the `UNKNOWN` still exists in the URL.
+            - ✅Confirm that the URL contains all other controls.
+            - ✅Confirm that the changing the control neither page refresh does not remove UNKNOWN from the URL.
```
maxschulz-COL: For me it disappears when going to Page 1 and back

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,141 +1,352 @@
...
+            - ❗the UNKNOWN parameter is not removed from the URL when navigating to Page-1 and back to Page-4.
+        (unknown query parameter ID with value that is not base64 encoded)
+        - http://localhost:8050/page_4?UNKNOWN=asd
+            - ✅Confirm that the bug is not raised and that the page can be opened.
```
maxschulz-COL, line 105: Confirmed, but going back and fourth removes query param

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,141 +1,352 @@
...
+        Url should be updated to: b64_WzIsNC40XQ & b64_IkFMTCI
+    - (URL with non base64 encoded value)
+        - http://localhost:8050/page_2?page_2_filter_sepal_width=PLAIN_TEXT
+        - http://localhost:8050/page_2?page_2_filter_species=PLAIN_TEXT
+    - (URL with incorrect base64 encoded value)
+        - http://localhost:8050/page_2?page_2_filter_sepal_width=b64_INCORRECT
+        - http://localhost:8050/page_2?page_2_filter_species=b64_INCORRECT
+    - (URL with missing query parameter value)
+        - http://localhost:8050/page_2?page_2_filter_sepal_width=
+        - http://localhost:8050/page_2?page_2_filter_species=
```
maxschulz-COL, line 118: what about duplicate and contradicting URL parameters? From the above I gather that every URL parameter on a correct ID but that is wrong, will be deleted right? Every query parameter not matching an id on page will be kept but ignored, but currently on page change for me it then gets lost

vizro-core/docs/pages/user-guides/parameters.md
```diff
@@ -189,5 +189,86 @@ Note that in the above example, one parameter affects multiple targets.
...
+
+The `Parameter` model accepts an optional argument `show_in_url`. When `show_in_url=True`, the parameter’s value is synchronized with the page URL as a query parameter, allowing the current control's state to be captured in the link. This makes it easy to share or bookmark specific `Page` parameter settings.
+
+The URL query parameter uses the control’s id as its key and the selected value, encoded in base64, as the URL query parameter value.
+
+!!! warning
+
+    - [PyCafe](https://vizro.readthedocs.io/en/stable/pages/user-guides/run-deploy/#develop-in-pycafe) incompatibility: Vizro does not currently support external URL query parameters and they could be stripped or lost using the app. As a result, dashboards using this feature should not be embedded or deployed in PyCafe environments.
```
maxschulz-COL, line 200: I see, so whenever the URL is actually used beyond just sharing dashboard state, e.g. drill through, then it fails. This is pretty annoying, have you spoken to the PyCafe team about it, I think this is something we should resolve asap.

vizro-core/docs/pages/user-guides/parameters.md
```diff
@@ -189,5 +189,86 @@ Note that in the above example, one parameter affects multiple targets.
 
 If you use [dynamic data](data.md/#dynamic-data) that can be updated while the dashboard is running then you can pass parameters to the dynamic data function to alter the data loaded into your dashboard. For detailed instructions, refer to the section on [parametrized data loading](data.md/#parametrize-data-loading).
 
+## Show in URL
```
maxschulz-COL, line 192: Hm I am not sure I like this all that much here. I think in the past we have slowly deviated from our north star of the diataxis framework, where "How-to" guide are concise guides in how to get something done, to a more API like description of our models even in the guides. This is by no means about this PR, but for some reason it triggered me more just now. (Similar comments would apply to the `extra` argument description)

I think overall we should rather have a guide on URL parameters and drill throughs, link to that guide from here very shortly, and not duplicate twice what is essentially the same description.

Now if time is really pressing, maybe not as part of this PR, but I would actually try to do it, otherwise we will forget, and it will not get done. In the future we should also do the same for our various examples on `extra`...

vizro-core/docs/pages/user-guides/parameters.md
```diff
@@ -189,5 +189,86 @@ Note that in the above example, one parameter affects multiple targets.
...
+
+The `Parameter` model accepts an optional argument `show_in_url`. When `show_in_url=True`, the parameter’s value is synchronized with the page URL as a query parameter, allowing the current control's state to be captured in the link. This makes it easy to share or bookmark specific `Page` parameter settings.
+
+The URL query parameter uses the control’s id as its key and the selected value, encoded in base64, as the URL query parameter value.
+
+!!! warning
+
+    - [PyCafe](https://vizro.readthedocs.io/en/stable/pages/user-guides/run-deploy/#develop-in-pycafe) incompatibility: Vizro does not currently support external URL query parameters and they could be stripped or lost using the app. As a result, dashboards using this feature should not be embedded or deployed in PyCafe environments.
+
+    - Page-specific only: Only parameters on the currently opened page are reflected in the URL. It is not yet possible to share or bookmark the global state of a multi-page dashboard.
```
maxschulz-COL, line 202: In general I think if we have a central how to guide, then we would also have more space to show some of the more intricate amazing things that are possible now, such as drill through. This really deserves a place in the docs, but ideally as part of a how to guide that is dedicated to this

vizro-core/docs/assets/user_guides/control/filter_in_url.png
maxschulz-COL, line 1: Zoomed in Screenshot, the key info is hardly visible here :)

vizro-core/docs/pages/user-guides/parameters.md
maxschulz-COL: Same of course then for the `Filter`, which is kinda of the main point

vizro-core/src/vizro/models/_controls/_controls_utils.py
maxschulz-COL: Should we not forbid it alltogether?

vizro-core/src/vizro/models/_page.py
```diff
@@ -167,6 +167,37 @@ def pre_build(self):
                 )
             ]
 
+        # Define a clientside callback that syncs the URL query parameters with controls that have show_in_url=True.
+        url_controls = [
+            control
+            for control in cast(
+                Iterable[ControlType],
+                [*model_manager._get_models(Parameter, self), *model_manager._get_models(Filter, self)],
```
maxschulz-COL, line 175: Noted :)

maxschulz-COL: Hmmm, but should this not rather duck type for any field called `show_in_url`? Or is that too much of premature optimization?

vizro-core/src/vizro/static/js/models/page.js
maxschulz-COL, line 1: I only skimmed here as I think others are better qualified to take a look. I really didnt fully udnerstand though, so would be keen to go through in details with you

maxschulz-COL (approved):
⭐ ❤️ Great PR and even greater feature! ❤️ ⭐ 

This is really cool work, and was a joy to read!!! Very well done. I have lots of comments, so important, some not so much, but I  want to pre-approve this as I think this is on a very good path!

One main question: are we writing JS tests for the clientside callbacks? I remember that the two of us laid the foundation for this to be possible ages ago, does this mechanism still work?

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: Done in -> https://github.com/mckinsey/vizro/pull/1243/files/bd071b02530b708127ceab7a749453db409b0a6d..00032c724d75e55b9dcaf10afd7ba0a12c1c5d0a

vizro-core/src/vizro/models/_controls/_controls_utils.py
petar-qb: > Should we not forbid it alltogether?

This was also one of the options we considered. Here's the part of the conversation we had about it -> https://github.com/mckinsey/vizro/pull/1243#issuecomment-3007708855. We continued this discussion offline and decided to rather raise a warning because the shareable url will work across multiple dashboard server restarts which is important. Less important thing is that the url is illegible with the random UUID keys, but it's already illegible with the encoded values so that's not a big deal.

vizro-core/examples/scratch_dev/app.py
antonymilne: Very much yes. It's this ticket and I already have a PoC of it: https://github.com/McK-Internal/vizro-internal/issues/1656. All being well it should be done within the next few weeks.

vizro-core/examples/scratch_dev/app.py
antonymilne: This is going to be done in https://github.com/McK-Internal/vizro-internal/issues/1656 🙂

vizro-core/src/vizro/models/_page.py
antonymilne: I think too much premature optimisation because the code is really written just to handle controls at the moment and not a general model with `show_in_url`. e.g. if we wanted to include the active tab in the URL and added `show_in_url` field to that model  then it would take some further modification of the callback to get it properly working anyway.

Ideally the code would look for `model_manager._get_models(ControlType)` or similar rather than needing to explicitly call out `Filter` and `Parameter` but that's not possible.

vizro-core/docs/pages/user-guides/parameters.md
antonymilne: This definitely a good push and I generally agree 👍 But I would wait until we have implemented the drill-through feature properly to document it in this way. At the moment our "drill through guide" would not be great because the user has to write some weird looking custom code for it, but in a few weeks it will all be built in and make a lot more sense.

I'll add it to the Action track E ticket so we don't forget.

vizro-core/docs/pages/user-guides/parameters.md
antonymilne: 100% agree, but the intricate amazing things will be even better soon so I would wait until then to show them off properly and with more interesting examples etc.

vizro-core/examples/scratch_dev/app.py
petar-qb: It's the browser session storage.
image

vizro-core/examples/scratch_dev/app.py
petar-qb: It looks like it works on my machine. 

https://github.com/user-attachments/assets/0c7bae2e-40c1-4ab8-ae4d-af5ba13d4769

Can you double check this by adding `print(dash.__version__)` to make sure you are using the latest version of Dash?

vizro-core/examples/scratch_dev/app.py
petar-qb: I think it doesn't work properly as when you copy and paste the url to another browser tab, it shows the default instead of the selected control values. We will handle the Pycafe as the part of the https://github.com/McK-Internal/vizro-internal/issues/1923

vizro-core/examples/scratch_dev/app.py
petar-qb: > Is there any race conditions here? Like does the order of URL parameter application matter?

This is the great question!! 

The order of events is always the following:
1. `page.build` -> returns default values like ["ALL"] or [2, 4.4] for the page-2 from the example
2. Applying dash persistence -> This step changes default values to the values that were selected the last time a user has visited the page. e.g. ["ALL"] -> ["setosa", "versicolor"] 
3. sync_url_and_controls clientside callback is triggered that changes the URL query parameters if or/and updates the control values if the page is opened with already specified url params.
4. `on-page-load` is triggered. -> this one sends the correctly selected values from the controls to the server. Then the original data_frames and figure configurations are filtered/parametrised and returned from the server. With this step the process is finished because it can't happen currently that on-page-load changes the currently selected parameter value (which would trigger another request to the server).

> And also, if I encoded URL parameters myself, could i create an invalid dashboard state?

Have you tested the cases under the `5.6 Test incorrect URL parameters:`? They should ensure that the dashboard state can't be easily destroyed. 
image

(edited)
I saw you checked the cases under the `5.6.`. So, the answer is that it should be hard (but possible) to make the app to crush. It could happen is properly encoded  string like "species" is assigned through the URL to the range slider for example. This is really an edge case and can happen only if a user make a focused effort for that 😄

vizro-core/examples/scratch_dev/app.py
petar-qb: How do you mean? What's the steps order to reproduce this?

vizro-core/examples/scratch_dev/app.py
petar-qb: Ahh my mistake. It should be written like this (I'm fixing the comment): 

            - ✅ Confirm that the changing the control does not remove UNKNOWN from the URL.
            - ❗the UNKNOWN parameter is not removed from the URL when navigating to Page-1 and back to Page-4.    




We will solve the `- ❗the UNKNOWN parameter...` as the part of the  https://github.com/McK-Internal/vizro-internal/issues/1923

vizro-core/examples/scratch_dev/app.py
petar-qb: thanks for checking! that's expected to happen.

vizro-core/examples/scratch_dev/app.py
petar-qb: Great question! In case of duplicate url params, the last one will be applied and all its duplications from the url will be removed. 

> From the above I gather that every URL parameter on a correct ID but that is wrong, will be deleted right?

Correct! 

> Every query parameter not matching an id on page will be kept but ignored, but currently on page change for me it then gets lost

Also correct!

vizro-core/docs/pages/user-guides/parameters.md
petar-qb: We definitely have this on our minds (will be solved within the https://github.com/McK-Internal/vizro-internal/issues/1923). We postponed working on that as we didn't want to spend much time on that case right now. We'll get back to it as soon as we deliver the drill-through, the new intract action and others.

vizro-core/examples/scratch_dev/app.py
maxschulz-COL: It now works! Very interesting, as you see I was quite careful with the `dash` version... Hmm. Btw, just noticed that in debug mode it shows the dash version haha

image

Quite nifty, maybe something for us as well in the future. Btw @lingyielia, I think we should also make it easier to see which version of the server is running. Currently it is a bit tricky to find out

vizro-core/docs/pages/user-guides/parameters.md
maxschulz-COL: OK dokes, I think this is important, just to restress :)

vizro-core/examples/scratch_dev/app.py
maxschulz-COL: Ah I think I know what is happening. So when I enter that link on a tab that already had some filtering, then it does keep whatever was filtered. If it is in a new tab, it gets the default right?

vizro-core/src/vizro/models/_controls/_controls_utils.py
maxschulz-COL: Hm ok, I would still prefer the other way, but given on just restarting it is the same I am fine although I prefer the other approach

maxschulz-COL (approved):
Checked all comment, lgtm!

vizro-core/examples/scratch_dev/app.py
petar-qb: Correct! If the control is missing from the URL, it will default to the previously persisted value. However, if the page is opened in a new tab, that persisted value won't exist. In that case, the server will fall back to the default value.

vizro-core/src/vizro/static/js/models/page.js
petar-qb: The URL safe key is internally made in: `urlParams.set(id, value)`. The `urlParams` is the `URLSearchParams` object that does it internally. So the following control id:  `"id [_-+]"` has the following form in the URL `"id+%5B_-%2B%5D"`.

vizro-core/changelog.d/20250617_102622_petar_pejovic_url_query_params_overwrite_persistence.md
```diff
@@ -0,0 +1,46 @@
+<!--
```
petar-qb, line 1: Wait until `dash>=3.1.1` is in `pyproject.toml`, then proceed with merging the PR.

vizro-core/src/vizro/static/js/models/page.js
antonymilne: Ah that's great, thank you!

PR #1245, opened 2025-06-17
[CI] PyCafe docs links checker

l0uden:
## Description
Created a script which parses all pycafe link in the docs and checks if vizro is loading there with success.
As it runs for ~8 minutes, I've put it on schedule run for every Monday and added ability to run it manually.

## Notice


.github/workflows/pycafe-docs-links.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Check PyCafe docs links
+        run: |
+          hatch run python ../tools/pycafe/test_pycafe_predefined_links.py \
+          $(grep -hoEr 'https://py\.cafe[^])"]*' \
+          ../vizro-ai/docs/* docs/* |\
+          sort -u |\
+          grep -vE '^https://py\.cafe/?$' |\
+          grep -vE '^https://py\.cafe/docs/apps/vizro/?$' |\
+          grep -vE '^https://py\.cafe/logo\.png')
```
maxschulz-COL: I don't fully understand what this line does? How do you grap the links, and what links are those exectaly?

.github/workflows/pycafe-docs-links.yml
```diff
@@ -0,0 +1,54 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - name: Install Hatch
+        run: pip install hatch
+
+      - name: Install Playwright and browser
+        run: |
+          hatch run pip install playwright
```
maxschulz-COL: Could it be cleverer to integrate this dependency into a hatch environment, and also enable this to be run locally? If this is too much effort, then not, but just wanted to check if this was sensible

antonymilne: Agreed. If you make a new hatch environment `pycafe-docs-links` for it then you could do the next step with a [post-install script](https://hatch.pypa.io/1.13/config/environment/overview/#post-install).

Then you'd do `hatch run pycafe-docs-links:python ...`.

antonymilne: Then maybe also it would be a bit cleaner instead of having a single job that has `working-directory: vizro-core` and does `../vizro-ai/docs/* docs/*`, you'd define the hatch environment in both vizro-core and vizro-ai and have two jobs (or a matrix) that just do `vizro-ai` and `vizro-core` each.

.github/workflows/pycafe-docs-links.yml
antonymilne: Yes, some comments explaining this would be nice 🙏 Otherwise it's a bit cryptic.

tools/pycafe/test_pycafe_predefined_links.py
```diff
@@ -0,0 +1,16 @@
...
+
+    links = sys.argv[1:]
+
+    # Test the link
+    success = [
+        test_pycafe_link(url=link, wait_for_locator="#dashboard-container", wait_for_text=False) for link in links
+    ]
+
+    # Exit with appropriate status code
+    sys.exit(1 if False in success else 0)
```
antonymilne:
```suggestion
    exit_codes = (test_pycafe_link(url=link, wait_for_locator="#dashboard-container", wait_for_text=False) for link in links) 
    sys.exit(any(exit_codes))
```

tools/pycafe/pycafe_utils.py
```diff
@@ -226,3 +227,32 @@ def get_example_directories() -> dict[str, Optional[list[str]]]:
             "https://py.cafe/files/maartenbreddels/tokenizers-demo/tokenizers-0.20.2.dev0-cp312-cp312-pyodide_2024_0_wasm32.whl",
         ],
     }
+
+
+def test_pycafe_link(url: str, wait_for_text: [str, bool], wait_for_locator: [str, bool]):
```
antonymilne: This is not how type hints work! I'm surprised it doesn't throw a SyntaxError actually. What is `[str, bool]` meant to mean? `Union[str, bool]`? A return type of  `int` would be good too (let's change to exit codes).

tools/pycafe/test_pycafe_generated_links.py
```diff
@@ -78,7 +51,7 @@ def test_pycafe_link(url: str, wait_for_text: str):
     url_generated = generate_link(config, dev_directory, extra_requirements, use_latest_release=False)
 
     # Test the link
-    success = test_pycafe_link(url=url_generated, wait_for_text="Vizro Features")
+    success = test_pycafe_link(url=url_generated, wait_for_text="Tips Dashboard", wait_for_locator=False)
```
antonymilne: I'd change all this `success` stuff to `exit_code` to match the new exit code system. If you do a search for `success` then feel free to change all of them to match my proposal.

Then e.g. L68 could be removed completely because you could just put the `sys.exit` in the `if exit_code`.

tools/pycafe/pycafe_utils.py
```diff
@@ -226,3 +227,32 @@ def get_example_directories() -> dict[str, Optional[list[str]]]:
...
+
+            # Get the app frame and wait for title or element
+            frame = page.frame_locator("#app")
+            if wait_for_text:
+                frame.get_by_text(wait_for_text).wait_for(timeout=90000)
+            if wait_for_locator:
+                frame.locator(wait_for_locator).wait_for(timeout=90000)
+
+            print(f"✅ Successfully verified PyCafe link: {url}")  # noqa
+            return True
```
antonymilne: If we change to exit codes then this would match `sys.exit` elsewhere. At the moment it's pretty confusing because you have `success = True` and `failure = False` which is the opposite way round from exit codes 0/1.

Do add this to the docstring though so it's clear what we're doing.

```suggestion
            return 0
```

tools/pycafe/pycafe_utils.py
```diff
@@ -226,3 +227,32 @@ def get_example_directories() -> dict[str, Optional[list[str]]]:
...
+
+        except Exception as e:
+            print(f"❌ Failed to verify PyCafe link: {url}")  # noqa
+            print(f"Error: {str(e)}")  # noqa
+            return False
```
antonymilne:
```suggestion
            return 1
```

antonymilne (commented):
Generally looks good - I just left a few initial suggestions and can then re-review 🙂

tools/pycafe/pycafe_utils.py
```diff
@@ -229,7 +229,7 @@ def get_example_directories() -> dict[str, Optional[list[str]]]:
     }
 
 
-def test_pycafe_link(url: str, wait_for_text: [str, bool], wait_for_locator: [str, bool]):
+def test_pycafe_link(url: str, wait_for_text: Union[str, bool], wait_for_locator: Union[str, bool]):
```
maxschulz-COL: As A said, we could still do a return type

.github/workflows/pycafe-docs-links.yml
l0uden: Added detailed comments

.github/workflows/pycafe-docs-links.yml
l0uden: Created script to run it locally. 
Don't think there should be different envs for `vizro-ai` or `vizor-core`, because we want to check all repo links at once. The problem of loading pycafe dashboard can be caused not only by changing our code in appropriate folder.

tools/pycafe/test_pycafe_predefined_links.py
l0uden: Done

tools/pycafe/pycafe_utils.py
l0uden: Thanks, done

tools/pycafe/test_pycafe_generated_links.py
l0uden: Done

tools/pycafe/pycafe_utils.py
l0uden: Changed it

tools/pycafe/pycafe_utils.py
l0uden: Added

vizro-core/hatch.toml
```diff
@@ -51,6 +52,34 @@ prep-release = [
...
+  #
+  # grep -vE '^https://py.cafe/?$'
+  # grep -vE '^https://py.cafe/docs/apps/vizro/?$'
+  # grep -vE '^https://py.cafe/logo.png'
+  # here we're excluding this patterns from the grep
+  # -v: Invert match
+  # ^: Anchors the match to the start of the line.
+  # $: Anchors the match to the end of the line.
+  # /?: Matches zero or one / after hostname
+  """python ../tools/pycafe/test_pycafe_predefined_links.py \\
```
antonymilne: Is there any way to write this without the` \\`?

Maybe it works without them at all, or if you use `'''` rather than `"""` possibly you don't need them?

vizro-core/hatch.toml
```diff
@@ -51,6 +52,34 @@ prep-release = [
...
+  "playwright install --with-deps chromium",
+  # grep -hoEr 'https://py.cafe[^])\"]*'
+  # here in the '[]' we're deleting this '^])"'characters at the end of the links and stop reading link after it
+  # -h: Suppress filenames in the output
+  # -o: Print only the matched (non-empty) parts of a matching line.
+  # -E: Use regular expressions.
+  # -r: Recursively read all files under each directory.
+  #
+  # sort -u
+  # sort links and leave onlu unique ones
```
antonymilne:
```suggestion
  # sort links and leave only unique ones
```

vizro-core/hatch.toml
```diff
@@ -51,6 +52,34 @@ prep-release = [
...
+  # grep -vE '^https://py.cafe/docs/apps/vizro/?$'
+  # grep -vE '^https://py.cafe/logo.png'
+  # here we're excluding this patterns from the grep
+  # -v: Invert match
+  # ^: Anchors the match to the start of the line.
+  # $: Anchors the match to the end of the line.
+  # /?: Matches zero or one / after hostname
+  """python ../tools/pycafe/test_pycafe_predefined_links.py \\
+  $(grep -hoEr 'https://py.cafe[^])\"]*' \\
+  ../vizro-ai/docs/* docs/* |\\
```
antonymilne: I would still suggest breaking this into two jobs so that it's defined in `vizro-ai/hatch.toml` and `vizro-core/hatch.toml`. And then either have two github workflows, one for `vizro-ai` and one for `vizro-core`, or one workflow that does a matrix and switches between the two. wdyt?

vizro-core/hatch.toml
l0uden: Created separate script to use it in both `hatch.toml`, but line escapes necessary there.

vizro-core/hatch.toml
l0uden: Put it into docs env and split it between two jobs

vizro-core/hatch.toml
antonymilne: Nice, this looks like a good solution I think 👍

vizro-core/hatch.toml
antonymilne: Perfect, this looks very clean now 👍

antonymilne (approved):
This is great, thanks for making all the changes!

Just one thing to check: can you do a search for `success` to see if we do that anywhere else in our hatch scripts? If so then we should change those to exit codes too please.

Feel free to force-merge whenever you're ready since @maxschulz-COL already approved but github seems to have forgotten.

PR #1246, opened 2025-06-17
[Fix] Improve nav left side on mobile

vladimir-mck:
## Description

Update CSS to show only the navigation bar in mobile view when there is no content in the left side panel.

Closes [#1873](https://github.com/McK-Internal/vizro-internal/issues/1873)

## Screenshot

Screenshot 2025-06-17 at 13 38 27
Screenshot 2025-06-17 at 13 38 36

## Notice


huong-li-nguyen (commented):
Thanks for fixing so quickly! The only thing I noticed is that the nav-icons have different sizing. I think one has a zoom applied, and the other doesn't ? If we can just align these, then I think we are good to go 👍  Just double-check horizontal/vertical direction :)


Screenshot 2025-06-17 at 14 07 46
Screenshot 2025-06-17 at 14 07 29

maxschulz-COL:
Hey! This would be easier to review if it had a scratch dev attached. I guess in this case we could also take the dev example, but generally its quite fun to have a scratch dev :)

vladimir-mck:
@maxschulz-COL I pushed the scratch dev layout I used. When the page includes controls, it will occupy the full screen. If there are no controls, only the navigation bar icons will be displayed.

maxschulz-COL:
> @maxschulz-COL I pushed the scratch dev layout I used. When the page includes controls, it will occupy the full screen. If there are no controls, only the navigation bar icons will be displayed.

Ok thanks. For the controls panel I now see the difference thanks! What is the expected behaviour with icon navigation but no controls?

vladimir-mck:
> > @maxschulz-COL I pushed the scratch dev layout I used. When the page includes controls, it will occupy the full screen. If there are no controls, only the navigation bar icons will be displayed.
> 
> Ok thanks. For the controls panel I now see the difference thanks! What is the expected behaviour with icon navigation but no controls?

In that case it works the same as before, the nav bar is next to the main content.

huong-li-nguyen (approved):
Let's keep the zoom as is right now - I'll check with a designer whether we should remove the zoom on the left-side in the second scenario (navbar but no content)

PR #1250, opened 2025-06-18
[Tidy] Update model manager for pydantic v2

antonymilne:
## Description

**Before**: we had hardcoded which model fields we traversed when searching the model hierarchy in `model_manager`.
**Now**: in pydantic v2 there's a public way to get the model fields and so we no longer need to hardcode it.

This is just a small improvement I made while thinking about the model manager. Bigger and better changes to come in due course from @maxschulz-COL.

Relevant reviewers: @petar-qb @maxschulz-COL. Should be an easy one to approve hopefully.

## Notice


vizro-core/tests/unit/vizro/managers/test_model_manager.py
antonymilne, line 1: @petar-qb thanks to you there was a great set of tests here which makes me very confident I didn't break anything 🙌 🙏

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -88,35 +88,20 @@ def _get_models(
...
-        """Iterates through children of `model`.
-
-        Currently, this method looks only through certain fields (components, tabs, controls, actions, selector) and
-            their children so might miss some children models.
-        """
+        """Iterates through children of `model` with depth-first pre-order traversal."""
         from vizro.models import VizroBaseModel
 
         if isinstance(model, VizroBaseModel):
             yield model
```
petar-qb, line 95: Just curious: Would we get the post-order traversal if the `yield model` is moved at the end of the function?

petar-qb (approved):
Super simplification, added flexibility and improved maintainability.

vizro-core/src/vizro/managers/_model_manager.py
antonymilne: Yes exactly!

maxschulz-COL (approved):
Now this works as I expect it :) I am slightly confused why this worked at all before?? But also don't want to think about that now haha

PR #1254, opened 2025-06-19
[Fix] Handle outside unknown url param

petar-qb:
This is just a PoC that should be used for the discussion about the unknown params. The proper implementation will be done later.

## TODO:
- [ ] - Check that the url params works properly with PyCafe
- [ ] - Write tests
- [ ] - Update docs

## Description

Firstly, I tried to fix loosing the unknown url query params by overwriting all links in the `sync_url_query_params_and_controls` clientside callback from the `page.js`.  I stopped halfway because it seemed easier to implement it using `flask.g` on the server side.

#### Clientside approach:
I started the implementation by using [this code](https://github.com/mckinsey/vizro/pull/1209/files#diff-624948dfc409dad5986a2413b74f744ec8910022352b840aa4b3aa85f3dfa14cR240-R244). 
Then I found out that this could only work by replacing the `link.setAttribute("href", base + new_query_string);` with the `dash_clientside.set_props(link.id, {href: base + "?" + urlParams.toString()});`. In that case, we would need to assign the ID to every link on the server side. Other todos that are not finished are:
- Append only `unknown` query parameters. So, the params which keys doesn't exist in the currently opened page.
- Define the `sync_url_query_params_and_controls` to `all pages` even if they don't contain the `show_in_url` controls (consider moving this part of the code into another clientside callback to avoid defining it on every page)

#### Serverside approach:
This PR contains the code that solves it on the server side. However, we should consider what are the other places where the `_get_relative_path_with_unknown_url_params` should be called instead of the `dash.get_relative_path`. 

@antonymilne Could you let me know what this looks like to you and which approach you prefer?

## Notice

### DIFF
--- a/vizro-core/examples/scratch_dev/app.py
+++ b/vizro-core/examples/scratch_dev/app.py
@@ -98,13 +98,10 @@ TODO-FOR-REVIEWER: Manual testing steps for sync URL-controls:
             - ✅Confirm that the `UNKNOWN` still exists in the URL.
             - ✅Confirm that the URL contains all other controls.
             - ✅Confirm that the changing the control neither page refresh does not remove UNKNOWN from the URL.
-            - ❗ONLY FROM BRANCH: "feat/url-unknown-params":
-                - ✅Confirm that the UNKNOWN parameter persists in the URL when navigating to other pages.
+            - ✅the UNKNOWN parameter is not removed from the URL when navigating to Page-1 and back to Page-4.
         (unknown query parameter ID with value that is not base64 encoded)
         - http://localhost:8050/page_4?UNKNOWN=asd
             - ✅Confirm that the bug is not raised and that the page can be opened.
-            - ❗ONLY FROM BRANCH: "feat/url-unknown-params":
-                - ✅Confirm that the UNKNOWN parameter persists in the URL when navigating to other pages.
 """
 from dash import dcc
 import json


vizro-core/src/vizro/models/_navigation/accordion.py
```diff
@@ -93,7 +93,7 @@ def _create_nav_links(self, pages: list[ModelID]):
                     children=page.title,
```
petar-qb, line 93: If we decide to do it on the clienside, we should assign the ID here and in other similar places so it could be adjusted with the `dash_clientside.set_props` or `Output`.

antonymilne:
Thanks for raising the PR to demonstrate it! Let me just clarify two things:
1. I don't think we should interpret this as just "handle unknown URL params" but instead as "handle global app state in the URL". Doing this will enable a user to share the whole app state rather than just a single page, which could be a very useful feature. So it's not just about not losing parameters that come from outside vizro, it's about maintaining `filter_from_page_2=...` in the URL even when you're on `page_1`. In fact I don't care that much about losing parameters outside vizro, it's the cross-page sharing that's important really
2. if this proves tricky to implement we can definitely just go ahead and merge #1243 without figuring it out and then come back to it later.

I had also thought of doing this serverside approach but I don't think I really like it:
1. it doesn't feel as thorough/rigorous as replacing every link on the page since you need to remember to manually use `_get_relative_path_with_unknown_url_params` rather than `get_relative_path` everywhere. Think of a case where I have manually written `html.A` somewhere in a custom figure for example - it's an extra step I need to remember to wrap my URL correctly. I could have even written my links clientside e.g. @huong-li-nguyen was doing this to link text in a cell of an AgGrid, in which case the serverside `_get_relative_path_with_unknown_url_params` wouldn't even be accessible when I write the code for my link
2. when the page first builds the links will be correct but if you then change a control, the URL updates but the navigation links won't. So we'd need something clientside to do this anyway.

With that said, I also don't really like the clientside approach of doing `dash_clientside.set_props(link.id, {href: base + "?" + urlParams.toString()});` because as you say means giving everything an id, and even then we'll miss things as per point 1 above.

So my feeling is that the original `link.setAttribute("href", base + new_query_string);` approach would be best. I'm pretty sure I had a version of that which did work somewhere while I was working in #1209 although maybe the code isn't in any of those commits 😬 Do you know why it wasn't working here? Is there anything we can do to make it work?

antonymilne:
Now I look at this again, I actually think we shouldn't implement this straight away because it might limit how we do the `interact` action. I think there's two ways to do that through URL. Let's say:
* page 1 has `parameter_id=blue` set (in Dash persistence and URL for page 1)
* we're on page 2
* we want to navigate to page 1 and set `filter_id=setosa`
 
Then the two options are:
1. generate a link to `page_1?filter_id=setosa` (b64 encoded) i.e. just with the minimal information and rely on Dash persistence and our clientside callback to add `parameter_id=blue` to the URL
1. generate a link to `page_1?filter_id=setosa&parameter_id=blue` (b64 encoded) i.e. with all state included

I had assumed we would be doing option 2 here but @petar-qb's code in #1243 would actually work with option 1 because the URL is updated to fill in the missing pieces when you change page. Option 1 might be easier for us to implement with `interact` so let's go with it for now.

This means that:
* for now we just ignore unknown URL parameters so they get lost from links
* for now we don't enable global app-state sharing through URL
* we try to solve more of https://github.com/McK-Internal/vizro-internal/issues/1656 without worrying about global app URL
* we open a new ticket to maybe do this in future

wdyt @petar-qb?

petar-qb:
> So my feeling is that the original link.setAttribute("href", base + new_query_string); approach would be best. I'm pretty sure I had a version of that which did work somewhere while I was working in https://github.com/mckinsey/vizro/pull/1209 although maybe the code isn't in any of those commits 😬 Do you know why it wasn't working here? Is there anything we can do to make it work?

I spent some time trying to figure out why it wasn't working and how to fix it, but I couldn't. In the html DOM the href was changed correctly, but clicking on the navigation button took me to the original href path. I think the problem was that the React Dash component wasn't updated when we set it with `link.setAttribute("href", base + new_query_string);`. Then, when the link (nav button) is clicked, it takes me to the original href which still exists as a property of the React Dash component.

>This means that:
> * for now we just ignore unknown URL parameters so they get lost from links
> * for now we don't enable global app-state sharing through URL
>* we try to solve more of https://github.com/McK-Internal/vizro-internal/issues/1656 without worrying about global app URL
> * we open a new ticket to maybe do this in future

This looks like a plan! I completely agree we have to investigate the #1656 before we implement the global URL and the unknown URL params handling solution.

antonymilne:
> I spent some time trying to figure out why it wasn't working and how to fix it, but I couldn't. In the html DOM the href was changed correctly, but clicking on the navigation button took me to the original href path. I think the problem was that the React Dash component wasn't updated when we set it with `link.setAttribute("href", base + new_query_string);`. Then, when the link (nav button) is clicked, it takes me to the original href which still exists as a property of the React Dash component.

Interesting. I could have sworn that it worked ok for me when I tried it out but maybe I'm just imagining it 🤔 

> This looks like a plan! I completely agree we have to investigate the #1656 before we implement the global URL and the unknown URL params handling solution.

Cool, let's close this for now then. Please can you open a ticket with medium priority just to make sure we don't forget about it? 🙏

PR #1255, opened 2025-06-19
[Feat] Allow more flexible import paths of `CapturedCallable` (plus bypassing validation)

maxschulz-COL:
Closes https://github.com/McK-Internal/vizro-internal/issues/1880
Closes https://github.com/McK-Internal/vizro-internal/issues/1178

## Description
See issue

Outstanding ToDos:
- [x] write tests for other models, all figure models should be possible now, move tests
- [x] write test for when `callable_def` is not given
- [x] fix newly broken tests
- [x] write docs, and potentially security hint
- [x] check if we want to even go beyond `FIGURE`
- [x] think about whitelist or safe mode for import paths

## Notice

### DIFF
--- a/vizro-core/examples/scratch_dev/app.py
+++ b/vizro-core/examples/scratch_dev/app.py
@@ -55,7 +55,7 @@ dashboard_config = {
                 #     },
                 # },
                 # {"type": "ag_grid", "figure": {"_target_": "__main__.custom_grid", "data_frame": "iris"}},
-                {"type": "figure", "figure": {"_target_": "__main__.custom_kpi_card", "data_frame": "iris"}},
+                # {"type": "figure", "figure": {"_target_": "__main__.custom_kpi_card", "data_frame": "iris"}},
                 {
                     "type": "button",
                     "text": "Export data",


vizro-core/src/vizro/models/types.py
```diff
@@ -105,14 +106,61 @@ def validate_captured_callable(cls, value, info: ValidationInfo):
...
+        captured_callable_config=value, json_schema_extra=json_schema_extra, callable_defs=callable_defs
     )
 
 
+# This proxy class is used to allow undefined CapturedCallable's to be validated and serialized.
+# When trying to instatiate a dashboard from json, if we cannot import the fucntion, then the CapturedCallableProxy
+# is the instatiated object of choice if the `_target_` is in the `callable_defs` list of the validation context.
+# This allows us to instatiate dashboard objects wihtout needing to execute/import the function definiition. Useful
+# in the context of untrusted code generation (e.g. by LLMs).
+class CapturedCallableProxy:
```
antonymilne: Should we actually just make this a pydantic model? I guess then it wouldn't match `CapturedCallable` but presumably it would mean no need to write the schema/core_validation stuff manually.

vizro-core/src/vizro/models/types.py
```diff
@@ -299,10 +353,30 @@ def _parse_json(
...
-        except (AttributeError, ModuleNotFoundError) as exc:
-            raise ValueError(f"_target_={function_name} cannot be imported from {import_path}.") from exc
+
+        # Try multiple approaches to import the function - order matters here.
+        import_attempts = [
+            import_path + ":" + function_name,
+            function_name,
+        ]
+
+        for import_string in import_attempts:
```
antonymilne: Would this still do what we want if we rearranged to avoid doing `for-else`?

See what other reviewers think but my guess is 99% of people are confused by `for-else` (including me).

```
# put this upfront instead of in else so we don't even attempt import
if function_name in callable_defs:
    return CapturedCallableProxy(...)

# if there's only two options for import then I'd just do it explicitly rather than with a for loop
try:
    function: CapturedCallable = TypeAdapter(ImportString).validate_python(import_path + ":" + function_name)
except ValidationError:
    try:
        function: CapturedCallable = TypeAdapter(ImportString).validate_python(function_name)
    except ValidationError:
        raise ...

# or if you prefer for loop then ok but I would just put the whole return inside the try to avoid doing break
# probably this will confuse mypy (and other people) though because it's not clear when something gets returned
for import_string in import_attempts:
    try:
         function: CapturedCallable = TypeAdapter(ImportString).validate_python(import_string)
    except ValidationError:
        continue
    else:
         # All the other items in figure are the keyword arguments to pass into function.
        function_kwargs = captured_callable_config
        # It would seem natural to return cls(function, **function_kwargs) here, but the function is already decorated
        # with @capture, and so that would return a nested CapturedCallable.
        return function(**function_kwargs)

raise ...
```

vizro-core/src/vizro/models/types.py
```diff
@@ -299,10 +353,30 @@ def _parse_json(
...
+            try:
+                function: CapturedCallable = TypeAdapter(ImportString).validate_python(import_string)
+                break
+            except ValidationError:
+                continue
+        else:
+            if function_name in callable_defs:
+                return CapturedCallableProxy(
+                    function_name=function_name,
+                    call_args=captured_callable_config,
```
```suggestion
                    kwargs=captured_callable_config,
```

Probably?

vizro-core/src/vizro/models/types.py
```diff
@@ -299,10 +353,30 @@ def _parse_json(
...
+        for import_string in import_attempts:
+            try:
+                function: CapturedCallable = TypeAdapter(ImportString).validate_python(import_string)
+                break
+            except ValidationError:
+                continue
+        else:
+            if function_name in callable_defs:
+                return CapturedCallableProxy(
+                    function_name=function_name,
```
antonymilne: Given this could now be `something:function_name` probably we should rename the variable since it's not just `function_name` any more.
```suggestion
                    _target_=_target_,
```

Maybe? Weird name but it makes it clear exactly what this is. I'd rename from L295 onwards.

vizro-core/tests/unit/vizro/models/_components/test_graph.py
```diff
@@ -34,6 +36,87 @@ def standard_px_chart_with_title():
     return px.scatter(data_frame="gapminder", x="gdpPercap", y="lifeExp", title="Title")
```
antonymilne: I think these tests belong in `test_types` and `test_base` tests for `to_python`. See already existing `TestModelFieldJSONConfig` and we could also have something that tests a model there with `CapturedCallableProxy`. These tests aren't really anything to do with graph, more to do with `CapturedCallableProxy` itself. 

Edit: I wrote that under the assumption that we would be rolling out the same change to `vm.Action`, `vm.Grid` etc. but maybe that's not the case?

vizro-core/tests/unit/vizro/models/_components/test_graph.py
```diff
@@ -34,6 +36,87 @@ def standard_px_chart_with_title():
...
+                "y": "sepal_width",
+            },
+        }
+
+    return _graph_config_with_target
+
+
+@pytest.fixture
+def expected_code():
+    def _expected_code(target: str):
```
antonymilne: This is a super weird way to get three different expected results. Why not just this? No need for any fixtures or any switching logic. Did I miss something?
```
expected_code_1 = """...."""
expected_code_2 = """...."""
expected_code_3 = """...."""

  @pytest.mark.parametrize(
        "target,expected_code",
        [
            "scatter",expected_code_1,
            "tests.unit.vizro.models._components.test_graph.custom_scatter",,expected_code_2,
            "custom_bar",expected_code_3,
        ],
    )

```

vizro-core/tests/unit/vizro/models/_components/test_graph.py
```diff
@@ -110,6 +193,21 @@ class MyGraph(vm.Graph):
         assert my_graph.actions == []
 
 
+class TestGraphInstantiationJson:
+    @pytest.mark.parametrize(
+        "target",
+        [
+            "scatter",
+            "tests.unit.vizro.models._components.test_graph.custom_scatter",
```
antonymilne:
```suggestion
            f"{__name__}.test_graph.custom_scatter",
```
Feels nicer than harcdoding the file path. Then we can also have a case for something that we know will work and will test the generic import path case rather than the `from vizro.plotly.express` one:
```suggestion
            f"{__name__}.test_graph.custom_scatter",
            "vizro.plotly.express.scatter",
```

antonymilne: Actually what will the `"vizro.plotly.express.scatter"` case do? Will it find the source code for it and put it in? 🤔 I don't think it will work correctly.

vizro-core/src/vizro/models/types.py
```diff
@@ -105,14 +106,61 @@ def validate_captured_callable(cls, value, info: ValidationInfo):
     if isinstance(value, (export_data, filter_interaction)):
         return value
 
+    try:
+        # Downside of validation is that it gets raised multiple times if there are multiple graphs in the dashboard.
```
antonymilne: Why do we need to validate `callable_defs` at all? What happens if you don't do this? Why/how/when could someone specify it incorrectly?

vizro-core/src/vizro/_vizro.py
```diff
@@ -110,6 +113,11 @@ def build(self, dashboard: Dashboard):
         self._pre_build()
         self.dash.layout = dashboard.build()
 
+        # Check if there are undefined captured callables in the dashboard.
+        for model in list(model_manager._get_models(root_model=dashboard)):
+            if isinstance(model, Graph) and isinstance(model.figure, CapturedCallableProxy):
+                self._undefined_captured_callables.append(f"{model.figure.function_name}")
+
```
antonymilne:
```suggestion
        # Check if there are undefined captured callables in the dashboard.
        self._undefined_captured_callables = [model.figure.function_name for model in model_manager._get_models(root_model=dashboard, Graph) if isinstance(model.figure, CapturedCallableProxy)]

```

vizro-core/src/vizro/_vizro.py
```diff
@@ -36,6 +37,8 @@ def __init__(self, **kwargs):
                 [Dash documentation](https://dash.plotly.com/reference#dash.dash) for possible arguments.
 
         """
+        self._undefined_captured_callables: list[str] = []
```
antonymilne:
```suggestion
        self._undefined_captured_callables: set[str] = {}
```

Very minor suggestion.

antonymilne (commented):
ok, I think I'm now starting to understand the point of this a bit better but let's discuss again next week when I have a fresh head. Generally I think it looks good though, nice work!

Basically, in addition to allowing general `ImportString`, what we want is a system that:
* lets you specify _explicitly whitelisted_ functions that aren't importable but still passes validation of the dashboard
* in fact it can still complete `dashboard.build` successfully
* raises an error only when you try to run the dashboard

Now for some more questions:
1. Did I get that right above?
2. Why do we want to allow only explicitly whitelisted functions? i.e. what's the point of the check against `callable_defs` ? (I think I know the answer but want to check)
3. Do you think we should extend (a) `ImportString` or (b) `CapturedCallableProxy` to other places like `vm.Action` and `vm.Grid`? See https://github.com/McK-Internal/vizro-internal/issues/1178
4. Why doesn't `dashboard.build` fail? I would have thought there would be some references to `graph.figure["data_frame"]` etc. that wouldn't work if you had a `CapturedCallableProxy` rather than a real `CapturedCallable`.
5. Would it maybe be cleaner instead of creating a new `CapturedCallableProxy` to just extend `CapturedCallable` so that it can accept `function: str` which would then set a flag `self.unimportable = True`?

For me to think about next week:
1. check security of `ImportString` e.g. if you do `os.system`. How easy is it to execute arbitrary code?
2. I'm not sure I like the names `Proxy` and `callable_defs` so will try to suggest some alternatives

maxschulz-COL:
> ok, I think I'm now starting to understand the point of this a bit better but let's discuss again next week when I have a fresh head. Generally I think it looks good though, nice work!
> 
> Basically, in addition to allowing general `ImportString`, what we want is a system that:
> 
> * lets you specify _explicitly whitelisted_ functions that aren't importable but still passes validation of the dashboard
> * in fact it can still complete `dashboard.build` successfully
> * raises an error only when you try to run the dashboard
> 
> Now for some more questions:
> 
> 1. Did I get that right above?

Yes

> 2. Why do we want to allow only explicitly whitelisted functions? i.e. what's the point of the check against `callable_defs` ? (I think I know the answer but want to check)

Otherwise we loose a large part of the value of validation. In general validation _should_ fail, it should be a manual effort to circumvent that part, because it also makes your instantiated object less valuable. However, if you know what your're doing, it's a great feature.

> 3. Do you think we should extend (a) `ImportString` or (b) `CapturedCallableProxy` to other places like `vm.Action` and `vm.Grid`? See [Investigate the final solution for custom figure function in yaml and consider the announcement McK-Internal/vizro-internal#1178](https://github.com/McK-Internal/vizro-internal/issues/1178). Edit: now I see that (a) is already done everywhere because it's part of `CapturedCallable` and (b) is just done for `Graph`. Do you think that's right? e.g. question 5 would do it for all models. Note also that if `ImportString` works for `vm.Action` etc. then we could now write lots of our docs that currently say custom actions/figures/graphs/etc. aren't possible in yaml (no need to do it in this PR, it could be a big task).

It's a good question, so I think ultimately yes. I would have been personally fine to just do it for Graph for now, as this is where the need is most urgent. My advice would be, let's agree on the correct implementation, merge this PR, then tackle the other places. This would then quickly unblock a few things like the MCP server.

> 4. Why doesn't `dashboard.build` fail? I would have thought there would be some references to `graph.figure["data_frame"]` etc. that wouldn't work if you had a `CapturedCallableProxy` rather than a real `CapturedCallable`.

Oh wow, very well remembered. Yes, there is a part in the code which circumvents this. I think it's super important though that build doesn't fai;, because eventually I want to make build part of validation (not necessarily a validator, but something that can be done independent of starting the server)

> 5. Would it maybe be cleaner instead of creating a new `CapturedCallableProxy` to just extend `CapturedCallable` so that it can accept `function: str` which would then set a flag `self.unimportable = True`?

Hm, I thought about this for a while, but I decided against it, because they are really different things. One is essentially a string that can do nothing, the other is a proper python object. I found it cleaner to know what we are dealing with explicitly.

EDIT: however, that being said I also think there is some advantage of making it work in one object, as then a lot less code as to change (and be duplicated). Will try to play with this a bit 👍 

> For me to think about next week:
> 
> 1. check security of `ImportString` e.g. if you do `os.system`. How easy is it to execute arbitrary code?
> 2. I'm not sure I like the names `Proxy` and `callable_defs` so will try to suggest some alternatives

vizro-core/tests/unit/vizro/models/_components/test_graph.py
maxschulz-COL: I wanted to be in line with the other fixture, but you are probably right. Just did that now

vizro-core/src/vizro/models/types.py
maxschulz-COL: I removed this completely now, and actually reverted my decision not to use `CapturedCallable`. It is actually a lot easier to implement across figure models, so that is a good thing!

vizro-core/src/vizro/models/types.py
maxschulz-COL: I just wanted to avoid the nested `try` clause, but happy to go with whatever people want

vizro-core/tests/unit/vizro/models/_components/test_graph.py
maxschulz-COL: It works, see new test!

vizro-core/src/vizro/_vizro.py
```diff
@@ -114,10 +114,11 @@ def build(self, dashboard: Dashboard):
         self.dash.layout = dashboard.build()
 
         # Check if there are undefined captured callables in the dashboard.
-        for model in list(model_manager._get_models(root_model=dashboard)):
-            if isinstance(model, Graph) and isinstance(model.figure, CapturedCallableProxy):
-                self._undefined_captured_callables.append(f"{model.figure.function_name}")
-
+        self._undefined_captured_callables = {
```
antonymilne: Is there actually a reason we do this in `build` rather than just doing it in `run`? Then we can remove the `self._undefined_captured_callables` variable as a `self` variable altogether.

```
undefined_captured_callables =  (model
            for model in list(model_manager._get_models(root_model=dashboard, model_type=FIGURE_MODELS))
            if model.figure._prevent_run)

if undefined_captured_callables:
    raise ...
```

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,60 +1,154 @@
...
+#             controls=[vm.Filter(column="sepal_length")],
+#         )
+#     ],
+# )
 
-dashboard = vm.Dashboard(title="Test dashboard", pages=[page_1])
 
 if __name__ == "__main__":
+    dashboard = vm.Dashboard.model_validate(
+        dashboard_config, context={"callable_defs": [("custom_bar2", "graph"), ("weird_grid", "ag_grid")]}
```
antonymilne: Maybe a `NamedTuple` or `dataclass` would be nice now there's multiple properties you're passing? I don't really mind though.

antonymilne: Or even a pydantic model?! And then you wouldn't need to do the validation of `callable_defs` manually with `TypeAdapter` like you do now.

vizro-core/src/vizro/models/types.py
```diff
@@ -147,42 +159,49 @@ def __init__(self, function, /, *args, **kwargs):
         # it more difficult to handle variadic keyword arguments due to https://bugs.python.org/issue41745.
```
antonymilne: This comment should go inside the `if`.

vizro-core/src/vizro/models/types.py
```diff
@@ -232,6 +251,7 @@ def __setitem__(self, arg_name: str, value):
     def _arguments(self):
         # TODO: This is used twice: in _get_parametrized_config and in vm.Action and should be removed when those
         # references are removed.
+        # Addition MS: this seems to be also used in model_post_init in _components/ag_grid.py
         # TODO-AV2 B 1: try to subclass Mapping. Check if anything requires MutableMapping (used in Vizro AI tests
         #  and to set data_frame only?). Try to remove these by making special method for setting data_frame. Then
         # can remove as many uses of _arguments as possible and use .items() where suitable instead.
```
```suggestion
        # TODO-AV2 B 1: try to subclass Mapping. Check if anything requires MutableMapping (used in Vizro AI tests
        #  and to set data_frame only?). Try to remove these by making special method for setting data_frame. Then
        # can remove as many uses of _arguments as possible and use .items() where suitable instead.
```

antonymilne (commented):
> > 2. Why do we want to allow only explicitly whitelisted functions? i.e. what's the point of the check against `callable_defs` ? (I think I know the answer but want to check)
> 
> Otherwise we loose a large part of the value of validation. In general validation _should_ fail, it should be a manual effort to circumvent that part, because it also makes your instantiated object less valuable. However, if you know what your're doing, it's a great feature.

Makes sense. Just to suggest a possible alternative but not sure if it makes sense?
* allow `CapturedCallable` to take a `function: str` which does `self._prevent_run = True` as now but don't validate against a whitelist, instead allow _any_ str
* something in `Vizro.build` or `run` that loops through these `CapturedCallables` and runs some new `CapturedCallable.do_import` and basically re-configures them with the actual imported function. At this point if the function isn't available then the import will fail and raise an error

> > 3. Do you think we should extend (a) `ImportString` or (b) `CapturedCallableProxy` to other places like `vm.Action` and `vm.Grid`? See [Investigate the final solution for custom figure function in yaml and consider the announcement McK-Internal/vizro-internal#1178](https://github.com/McK-Internal/vizro-internal/issues/1178). Edit: now I see that (a) is already done everywhere because it's part of `CapturedCallable` and (b) is just done for `Graph`. Do you think that's right? e.g. question 5 would do it for all models. Note also that if `ImportString` works for `vm.Action` etc. then we could now write lots of our docs that currently say custom actions/figures/graphs/etc. aren't possible in yaml (no need to do it in this PR, it could be a big task).
> 
> It's a good question, so I think ultimately yes. I would have been personally fine to just do it for Graph for now, as this is where the need is most urgent. My advice would be, let's agree on the correct implementation, merge this PR, then tackle the other places. This would then quickly unblock a few things like the MCP server.

Agreed. For me the new implementation without `CapturedCallableProxy` looks a lot better I think, so even if we don't care about Action and AgGrid I would go for it.
 
> > 4. Why doesn't `dashboard.build` fail? I would have thought there would be some references to `graph.figure["data_frame"]` etc. that wouldn't work if you had a `CapturedCallableProxy` rather than a real `CapturedCallable`.
> 
> Oh wow, very well remembered. Yes, there is a part in the code which circumvents this. I think it's super important though that build doesn't fai;, because eventually I want to make build part of validation (not necessarily a validator, but something that can be done independent of starting the server)

Is this circumventing code still there now you put the `kwargs` into `__bound_arguments` in `CapturedCallable`? I can't see it.

> > 5. Would it maybe be cleaner instead of creating a new `CapturedCallableProxy` to just extend `CapturedCallable` so that it can accept `function: str` which would then set a flag `self.unimportable = True`?
> 
> Hm, I thought about this for a while, but I decided against it, because they are really different things. One is essentially a string that can do nothing, the other is a proper python object. I found it cleaner to know what we are dealing with explicitly.
> 
> EDIT: however, that being said I also think there is some advantage of making it work in one object, as then a lot less code as to change (and be duplicated). Will try to play with this a bit 👍

For me the code feels much more natural now. If you like it too then basically this PR looks good to me, I just still need to do these two things:
> check security of ImportString e.g. if you do os.system. How easy is it to execute arbitrary code?
> I'm not sure I like the names Proxy and callable_defs so will try to suggest some alternatives

maxschulz-COL:
> Makes sense. Just to suggest a possible alternative but not sure if it makes sense?
> 
> * allow `CapturedCallable` to take a `function: str` which does `self._prevent_run = True` as now but don't validate against a whitelist, instead allow _any_ str
> * something in `Vizro.build` or `run` that loops through these `CapturedCallables` and runs some new `CapturedCallable.do_import` and basically re-configures them with the actual imported function. At this point if the function isn't available then the import will fail and raise an error

I think it _may_ work generally, except for one thing atm: currently we need to know the `mode` to instantiate the CC. I am sure we could work around it, but atm without it the code fails due to some access to `sef._mode`

> > > 4. Why doesn't `dashboard.build` fail? I would have thought there would be some references to `graph.figure["data_frame"]` etc. that wouldn't work if you had a `CapturedCallableProxy` rather than a real `CapturedCallable`.
> > 
> > 
> > Oh wow, very well remembered. Yes, there is a part in the code which circumvents this. I think it's super important though that build doesn't fai;, because eventually I want to make build part of validation (not necessarily a validator, but something that can be done independent of starting the server)
> 
> Is this circumventing code still there now you put the `kwargs` into `__bound_arguments` in `CapturedCallable`? I can't see it.

Yes, after putting the `kwargs` there we could remove the lines before `data_frame = captured_callable["data_frame"]` in `_process_callable_data_frame`

> For me the code feels much more natural now. If you like it too then basically this PR looks good to me, I just still need to do these two things:
> 
> > check security of ImportString e.g. if you do os.system. How easy is it to execute arbitrary code?
> > I'm not sure I like the names Proxy and callable_defs so will try to suggest some alternatives

Great!

vizro-core/docs/pages/user-guides/extensions.md
```diff
@@ -88,3 +88,102 @@ For more information, see our documentation on [customizing CSS](custom-css.md)
...
+
+It is possible to refer to custom functions that are used as `CapturedCallable` by their import path in a `yaml`/`json` configuration of the dashboard.
+
+In the [documentation sections on custom charts, tables, figures and actions](#vizro-customizations), you will find examples on how the Vizro schema can be extended by using custom Python code. It is possible to refer to these custom functions in the `yaml`/`json` configuration by using the `_target_` key and the correct import path.
+
+!!! example "Custom charts with YAML config example"
+
+````
+```{.python pycafe-link  hl_lines="27 28"}
+import pandas as pd
```
petar-qb: Let's fix the docs here. Here's how it looks right now. 
image

vizro-core/src/vizro/models/types.py
petar-qb: This looks like another reason for defining the `callable_defs` as a dataclass or Pydantic model I guess.

vizro-core/src/vizro/models/types.py
```diff
@@ -147,42 +160,55 @@ def __init__(self, function, /, *args, **kwargs):
...
+            # Used in later validations of the captured callable.
+            self._mode = None
+            self._model_example = None
+            self._prevent_run = False
+        else:
+            self._prevent_run = True
+            self.__function = function
+            self.__bound_arguments = OrderedDict(kwargs)
+            self._mode = args[0]
+            self._model_example = None
```
petar-qb: A bit curious about the `self._mode = args[0]`. Do we assume that something like `"type": "graph",` will come as the first argument?

vizro-core/src/vizro/_vizro.py
maxschulz-COL: Yes, to have access to the dashboard object

vizro-core/src/vizro/models/types.py
maxschulz-COL: Changed it, I think this is more stable now

vizro-core/src/vizro/models/types.py
maxschulz-COL: Changed it to the above now

vizro-core/src/vizro/models/types.py
maxschulz-COL: It is now a `namedtuple`, but I did not make it a pydantic model to be more lightweight

vizro-core/src/vizro/models/_base.py
```diff
@@ -111,11 +111,17 @@ def _extract_captured_callable_source() -> set[str]:
...
-                for new in REPLACEMENT_STRINGS.values()
+            if (
+                isinstance(value, CapturedCallable)
+                and not any(
+                    # Check to see if the captured callable does use a cleaned module string, if yes then
+                    # we can assume that the source code can be imported via Vizro, and thus does not need to be defined
+                    value.__repr_clean__().startswith(new)
+                    for new in REPLACEMENT_STRINGS.values()
+                )
+                # If _function is a string, then we cannot import the code, and the user needs to use `callable_defs`
```
petar-qb:
```suggestion
                # If _function is a string, then we cannot import the code, and the user needs to use `allowed_undefined_captured_callables`
```

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,60 +1,147 @@
...
+        ]
+    },
+)
 
 if __name__ == "__main__":
-    app = Vizro().build(dashboard)
-    app.run()
+    # print(dashboard._to_python())
+    # config = dashboard.model_dump(exclude_unset=True)
+    # print(config)
```
petar-qb, line 140: I'm a bit unfamiliar with `dashboard.model_dump` and its `exclude_unset` configuration, so just want to say that the exception is raised when this line is uncommented along with the unknown ag-grid: `{"type": "ag_grid", "figure": {"_target_": "weird_grid", "data_frame": "iris"}}`. If it's intended, may you explain here why it happens and why `dashboard._to_python()` works but `dashboard.model_dump()` does not?

vizro-core/src/vizro/models/types.py
```diff
@@ -140,49 +168,62 @@ def __init__(self, function, /, *args, **kwargs):
...
+
+            # Used in later validations of the captured callable.
+            self._mode = None
+            self._model_example = None
+            self._prevent_run = False
+        else:
+            self._prevent_run = True
+            self.__function = function
+            mode = kwargs.pop("mode", None)  # we instantiate str proxy with mode as kwarg
+            self._mode = mode
```
petar-qb:
```suggestion
            self._mode = kwargs.pop("mode", None)  # we instantiate str proxy with mode as kwarg
```

vizro-core/src/vizro/models/types.py
```diff
@@ -299,10 +349,26 @@ def _parse_json(
             ) from exc
 
         import_path = json_schema_extra["import_path"]
+
+        # Check if we skip proper instantiation due to undefined function.
+        if function_name in (def_name for def_name, _ in allowed_undefined_captured_callables):
+            # Find the mode from allowed_undefined_captured_callables tuples
+            mode = next(mode for def_name, mode in allowed_undefined_captured_callables if def_name == function_name)
```
petar-qb: Sorry if it's obvious, but I'm wondering why do we need the `mode` inside the `allowed_undefined_captured_callables `? Can't we fetch the mode at the end of the the `CapturedCallable` constructor and then reuse here and in other places?

petar-qb: Is the reason actually because this piece of code is executed `before` the CapturedCallable constructor? If so, is there any improvement we are able to introduce now or in future (in that case a `# TODO` comment would be nice)?

vizro-core/src/vizro/models/types.py
```diff
@@ -96,7 +99,14 @@ class JsonSchemaExtraType(TypedDict):
     mode: str
 
 
-def validate_captured_callable(cls, value, info: ValidationInfo):
+class AllowedCapturedCallable(NamedTuple):
```
maxschulz-COL: here

vizro-core/src/vizro/_vizro.py
antonymilne: Ah yes, I see. Please can we move it so that we just do `self._dashboard = dashboard` here, remove `self._undefined_captured_callables` and put this code in `run` instead?

vizro-core/src/vizro/models/types.py
```diff
@@ -359,6 +425,9 @@ def __repr__(self):
 
     def __repr_clean__(self):
         """Alternative __repr__ method with cleaned module paths."""
+        if self._prevent_run:
+            args = ", ".join(f"{key}={value!r}" for key, value in self._arguments.items())
+            return f"{self._function}({args})"
         args = ", ".join(f"{key}={value!r}" for key, value in self._arguments.items())
         original_module_path = f"{self._function.__module__}"
         return f"{_clean_module_string(original_module_path)}{self._function.__name__}({args})"
```
petar-qb:
```suggestion
        args = ", ".join(f"{key}={value!r}" for key, value in self._arguments.items())
        if self._prevent_run:
            return f"{self._function}({args})"
        original_module_path = self._function.__module__
        return f"{_clean_module_string(original_module_path)}{self._function.__name__}({args})"
```

vizro-core/tests/unit/vizro/models/test_base.py
```diff
@@ -536,6 +538,69 @@ def chart(data_frame, hover_data: Optional[list[str]] = None):
...
+# data_manager["iris"] = ===> Fill in here <===
+
+
+########### Model code ############
+model = vm.Graph(
+    type="graph",
+    figure=custom_scatter(data_frame="iris", x="sepal_length", y="sepal_width"),
+)
+"""
+expected_code_pass_validation_via_json = """############ Imports ##############
```
petar-qb:
```suggestion
"""

expected_code_pass_validation_via_json = """############ Imports ##############
```

vizro-core/src/vizro/_vizro.py
```diff
@@ -131,6 +143,12 @@ def run(self, *args, **kwargs):  # if type annotated, mkdocstring stops seeing t
         data_manager._frozen_state = True
         model_manager._frozen_state = True
 
+        if self._undefined_captured_callables:
+            raise ValueError(
+                f"""Dashboard contains models with undefined CapturedCallable's: {self._undefined_captured_callables}.
+Provide a valid import path for the CapturedCallables in the dashboard config."""
```
antonymilne:
```suggestion
                f"""Dashboard contains models with undefined `CapturedCallable`s: {self._undefined_captured_callables}.
Provide a valid import path for these in your dashboard configuration."""
```

vizro-core/tests/unit/vizro/models/test_types.py
```diff
@@ -164,12 +164,27 @@ class ModelWithAction(VizroBaseModel):
...
     # The import_path here makes it possible to import the above function using getattr(import_path, _target_).
     function: SkipJsonSchema[CapturedCallable] = Field(json_schema_extra={"mode": "graph", "import_path": __name__})
     _validate_figure = field_validator("function", mode="before")(validate_captured_callable)
 
 
+class ModelWithInvalidModule(VizroBaseModel):
+    # The import_path doesn't exist. This lets us also simulate importing the function from a different module.
+    function: CapturedCallable = Field(json_schema_extra={"mode": "graph", "import_path": "invalid.module"})
+
+    _validate_figure = field_validator("function", mode="before")(validate_captured_callable)
```
petar-qb:
```suggestion
    function: CapturedCallable = Field(json_schema_extra={"mode": "graph", "import_path": "invalid.module"})
    _validate_figure = field_validator("function", mode="before")(validate_captured_callable)
```

vizro-core/src/vizro/_vizro.py
```diff
@@ -110,6 +114,14 @@ def build(self, dashboard: Dashboard):
         self._pre_build()
         self.dash.layout = dashboard.build()
 
+        # Check if there are undefined captured callables in the dashboard.
+        self._undefined_captured_callables = {
+            (model.figure._function, model.figure._mode)
```
antonymilne:
```suggestion
            model.figure
```

Probably looks fine in the error message just like this because it's now a namedtuple.

petar-qb (approved):
This looks really good! Great job @maxschulz-COL 🥇

vizro-core/src/vizro/models/types.py
```diff
@@ -140,49 +168,62 @@ def __init__(self, function, /, *args, **kwargs):
...
+            self._mode = None
+            self._model_example = None
+            self._prevent_run = False
+        else:
+            self._prevent_run = True
+            self.__function = function
+            mode = kwargs.pop("mode", None)  # we instantiate str proxy with mode as kwarg
+            self._mode = mode
+            self.__bound_arguments = OrderedDict(kwargs)
+            self._model_example = None
```
antonymilne: * Reordered to match `if` clause better
* Removed `OrderedDict` since it doesn't seem necessary since Python now preserves dictionary order but tell me if I missed something

```suggestion
            self.__function = function
            mode = kwargs.pop("mode", None)  # we instantiate str proxy with mode as kwarg
            self.__bound_arguments = kwargs
            self._mode = mode
            self._model_example = None
            self._prevent_run = True
```

vizro-core/src/vizro/models/types.py
```diff
@@ -299,10 +349,26 @@ def _parse_json(
             ) from exc
 
         import_path = json_schema_extra["import_path"]
+
+        # Check if we skip proper instantiation due to undefined function.
+        if function_name in (def_name for def_name, _ in allowed_undefined_captured_callables):
+            # Find the mode from allowed_undefined_captured_callables tuples
+            mode = next(mode for def_name, mode in allowed_undefined_captured_callables if def_name == function_name)
+            return CapturedCallable(function_name, mode=mode, **captured_callable_config)
```
antonymilne: Doing this loop/conditional the other way round would make much more sense:

```suggestion
        # Check if we skip proper instantiation due to undefined function.
        for captured_callable in allowed_undefined_captured_callables:
            if function_name == captured_callable.function_name:
                captured_callable = CapturedCallable(function_name, **captured_callable_config)
                captured_callable._mode = mode
                return captured_callable
```

vizro-core/src/vizro/models/types.py
```diff
@@ -299,10 +349,26 @@ def _parse_json(
...
+
+        # Try multiple approaches to import the function - order matters here.
```
```suggestion
```

vizro-core/src/vizro/models/types.py
```suggestion
class UndefinedCapturedCallable(NamedTuple):
```

Plus similar renaming throughout. Because the `allowed_undefined_capatures_callables` is really a list of `UndefinedCapturedCallable`s that are allowed rather than `AllowedCapturedCallable`s that are undefined.

I would also suggest we do `allow_undefined_captured_callable` that than `allowed` because it leaves open the possibility of `allow_undefined_captured_callable = True` if we want to do it in future, and it still makes sense as an argument name if it's just a list of allowed things.

vizro-core/src/vizro/models/types.py
```diff
@@ -96,7 +99,14 @@ class JsonSchemaExtraType(TypedDict):
     mode: str
 
 
-def validate_captured_callable(cls, value, info: ValidationInfo):
+class AllowedCapturedCallable(NamedTuple):
+    """Tuple to be used as validation context to allow undefined captured callables."""
+
+    function_name: str
```
antonymilne: Is this actually just a `function_name` or a possible import path?

vizro-core/src/vizro/models/types.py
```diff
@@ -105,14 +115,32 @@ def validate_captured_callable(cls, value, info: ValidationInfo):
...
+        ).validate_python(
+            info.context.get("allowed_undefined_captured_callables", []) if info.context is not None else []
+        )
+    except ValidationError:
+        raise ValueError(
+            """Invalid `allowed_undefined_captured_callables`.
+Must be a list of namedtuples with (function_name: str, mode: str).
+"""
+        )
+
```
antonymilne: I don't mind the namedtuple but to me it doesn't really make sense given that this code basically just does something you'd get for free with a pydantic model 🤔

vizro-core/tests/unit/vizro/models/test_base.py
```diff
@@ -592,6 +657,34 @@ def test_to_python_complete_dashboard(self, complete_dashboard):
...
+                "y": "sepal_width",
+            },
+        }
+
+        graph = vm.Graph.model_validate(
+            graph_config, context={"allowed_undefined_captured_callables": [("not_importable_scatter", "graph")]}
+        )
+        result = graph._to_python()
+        assert result == expected_code
+
```
antonymilne, line 690: Maybe another test here for what happens if you `vm.Graph.model_validate` on the `not_importable_scatter` without the context.

vizro-core/tests/unit/vizro/models/test_types.py
```diff
@@ -229,19 +244,32 @@ def test_all_arguments(self):
         assert isinstance(model.function, CapturedCallable)
         assert model.function() == 1 + 2 + 3 + 4
 
+    def test_all_arguments_different_import_path(self):
```
antonymilne:
```suggestion
    def test_different_import_path(self):
```

vizro-core/docs/pages/user-guides/extensions.md
```diff
@@ -88,3 +88,111 @@ For more information, see our documentation on [customizing CSS](custom-css.md)
 It is possible to create custom React.js components and add them directly to any Vizro dashboard so enabling you to code beneath both the Vizro and Dash layers and control React.js directly
 
 For more information, see the documentation on [using React.js components with Dash](https://dash.plotly.com/plugins)
+
+## Using custom functions in `yaml`/`json` configuration
```
antonymilne: I think section should go as a subsection of the `Vizro customizations` section.

antonymilne:
```suggestion
## Use custom functions in `yaml`/`json` configuration
```

vizro-core/docs/pages/user-guides/extensions.md
```diff
@@ -88,3 +88,111 @@ For more information, see our documentation on [customizing CSS](custom-css.md)
...
+
+!!! note "Exposing configuration to untrusted users"
+
+    Exposing Vizro configuration to untrusted users may pose a security risk. By choosing a specific import path for a custom function, it could be possible to inject malicious code that will be executed when the dashboard is run. Note that for this to work, the malicious code would need to be executed on import, and it would need to be part of a dependency that is available at validation time.
```
antonymilne:
```suggestion
    Exposing Vizro configuration to untrusted users may pose a security risk. A user with access to YAML/JSON configuration can potentially execute arbitrary Python code when the dashboard is run. The security of your `yaml`/`json` configuration should be regarded as equivalent to the security of your `app.py` file.
```

vizro-core/docs/pages/user-guides/extensions.md
```diff
@@ -88,3 +88,111 @@ For more information, see our documentation on [customizing CSS](custom-css.md)
...
+
+It is possible to refer to custom functions that are used as `CapturedCallable` by their import path in a `yaml`/`json` configuration of the dashboard.
+
+In the [documentation sections on custom charts, tables, figures and actions](#vizro-customizations), you will find examples on how the Vizro schema can be extended by using custom Python code. It is possible to refer to these custom functions in the `yaml`/`json` configuration by using the `_target_` key and the correct import path.
```
antonymilne:
```suggestion
It is possible to refer to custom functions for charts, tables, figures and actions by their import path in a `yaml`/`json` configuration of the dashboard. In the documentation on custom charts, tables, figures and actions, you will find examples of how the Vizro schema can be extended by using custom Python code. It is possible to refer to these custom functions in the `yaml`/`json` configuration by using the `_target_` key and the correct import path.
```

vizro-core/docs/pages/user-guides/extensions.md
```diff
@@ -88,3 +88,111 @@ For more information, see our documentation on [customizing CSS](custom-css.md)
...
+
+!!! example "Custom charts with YAML config example"
```
antonymilne: Please can we do this as multiple tabs that show all the files required rather than embedding a YAML string inside the Python code? I'd suggest:
* one file with our standard `app.py` for loading YAML 
* one file `custom_charts.py` with `bar` defined
* one `dashboard.yaml` file with `_target_: custom_charts.bar`

Note the yaml file shouldn't need any `"` marks anywhere I believe.

Then afterwards in a note or in a  `(1)!` thing like you do now, let's say that you could define the chart in `app.py` itself and refer to it as `__main__.bar`.

This way we're demonstrating the "normal" looking `custom_charts.bar` first and then the weird looking `__main__` afterwards.

Also would be worth linking to pydantic's docs on `ImportString` that explain the format of string expected.

vizro-core/docs/pages/user-guides/extensions.md
```diff
@@ -88,3 +88,111 @@ For more information, see our documentation on [customizing CSS](custom-css.md)
...
+    app = Vizro().build(dashboard)
+
+    if __name__ == "__main__":
+        app.run()
+    ```
+
+    1. Define the custom chart either in `app.py` or in a separate file.
+    1. Parse the YAML or JSON configuration that refers to the custom chart with the **correct import path**. If the custom chart is defined in the `app.py`, then use `__main__` as the import path.
+
+### Validating dashboards without executing `CapturedCallable` functions
```
antonymilne:
```suggestion
### Validate dashboards without defining `CapturedCallable` functions
```

vizro-core/docs/pages/user-guides/extensions.md
```diff
@@ -88,3 +88,111 @@ For more information, see our documentation on [customizing CSS](custom-css.md)
...
+
+It is possible to validate a dashboard configuration without executing the `CapturedCallable` functions.
```
antonymilne:
```suggestion
It is possible to validate a dashboard configuration without importing or executing some `CapturedCallable` functions.
```

vizro-core/docs/pages/user-guides/extensions.md
```diff
@@ -88,3 +88,111 @@ For more information, see our documentation on [customizing CSS](custom-css.md)
...
+
+You can use this method when you want to check if the dashboard configuration is valid, but you don't want to execute the custom functions until run-time, which may be in a sandboxed environment. This is useful when the custom functions are not available at validation time, or when they originate from untrusted sources (e.g. when a large language model is used to generate that code).
```
antonymilne:
```suggestion
You can use this method when you want to check if the dashboard configuration is valid, but you don't want to import or execute the custom functions until run-time, which may be in a sandboxed environment. This is useful when the custom functions are not available at validation time, or when they originate from untrusted sources (e.g. when a large language model is used to generate that code).
```

vizro-core/docs/pages/user-guides/extensions.md
```diff
@@ -88,3 +88,111 @@ For more information, see our documentation on [customizing CSS](custom-css.md)
...
+        context={
+            "allowed_undefined_captured_callables": [
+                AllowedCapturedCallable(function_name="llm_generated_grid", mode="ag_grid")
+            ]
+        },
+    )
+    app = Vizro().build(dashboard)  # (1)!
+    ```
+
+    1. Because this dashboard configuration contains a `CapturedCallable` function that is undefined, you will not be able to start the server. Hence `app.run()` will raise an error.
```
antonymilne: Something like this might be clearer? Assuming I understood the behaviour correctly.

```suggestion
    app = Vizro().build(dashboard)  # (1)!
    try:
        app.run()  # (2)!
    except Exception as exc:
        print(exc)
    ```

    1. The dashboard configuration contains a `CapturedCallable` function that is undefined but allowed by `allowed_undefined_captured_callables`. The app can still be built without raising any errors.
    2. However, it is not possible to run the app without undefined `CapturedCallable`s. This raises an error.
```

vizro-core/changelog.d/20250623_131556_maximilian_schulz_captured_callable_import_path.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Add option to validate yaml/json configurations without importing `CapturedCallable` functions referenced in the configuration. See [the docs on extensions](https://vizro.readthedocs.io/en/stable/pages/user-guides/extensions/) for more details. ([#1255](https://github.com/mckinsey/vizro/pull/1255))
```
antonymilne: I would split this into two points.

```suggestion
- yaml/json configurations can specify `_target_` functions with any import path rather than being restricted to built-in Vizro functions.

- Add option to validate yaml/json configurations without importing `CapturedCallable` functions. See [the user guide on extensions](https://vizro.readthedocs.io/en/stable/pages/user-guides/extensions/) for more details. ([#1255](https://github.com/mckinsey/vizro/pull/1255))
```

vizro-core/docs/pages/user-guides/extensions.md
```diff
@@ -88,3 +88,111 @@ For more information, see our documentation on [customizing CSS](custom-css.md)
...
+
+### Validating dashboards without executing `CapturedCallable` functions
+
+It is possible to validate a dashboard configuration without executing the `CapturedCallable` functions.
+
+You can use this method when you want to check if the dashboard configuration is valid, but you don't want to execute the custom functions until run-time, which may be in a sandboxed environment. This is useful when the custom functions are not available at validation time, or when they originate from untrusted sources (e.g. when a large language model is used to generate that code).
+
+!!! note When starting the server
+
+    If you validate your dashboard configuration with a `CapturedCallable` function that is undefined, you will not be able to start the server. Hence when running the dashboard, you will need to recreate a dashboard object with all the `CapturedCallable` functions defined.
```
antonymilne, line 116: FYI this is sort of why I was wondering whether in future we might have something that goes through the undefined captured callables and tries to define them by actually doing the import. Instead of having to redefine the whole dashboard configuration, you could just materialise the previously undefined parts.

vizro-core/examples/scratch_dev/app.py
antonymilne: I also found this while playing around something else recently. Not sure why but as soon as you set `exclude_unset=False` (the default) then even a very simple dashboard configuration can't be dumped.

maxschulz-COL: Very good catch @petar-qb! I can answer yours but not sure why Antony observed that in a very "simple" config.

`model_dump` itself does not fail, but the `print` function does. `print` uses an object's `repr`, while `to_python` uses `repr_clean` before, which then eliminates all things that could fail in `repr`.

I only tested `to_python`, but I think it's also important that `print` of a model dump doesn't fail! Let me fix

maxschulz-COL: Fixed in https://github.com/mckinsey/vizro/pull/1255/commits/f995862766162a8a5f70b1762ce5eed0ab557c46

vizro-core/src/vizro/models/types.py
maxschulz-COL: We're getting rid of the mode now, so very good comment indeed.

vizro-core/src/vizro/models/types.py
maxschulz-COL: Went for a slightly different approach that also fixes the `repr` method

vizro-core/src/vizro/models/types.py
maxschulz-COL: Mypy :)

vizro-core/src/vizro/models/types.py
maxschulz-COL: N/A

vizro-core/src/vizro/models/types.py
maxschulz-COL: N/A

vizro-core/src/vizro/models/types.py
maxschulz-COL: n/a

vizro-core/src/vizro/models/types.py
maxschulz-COL: n/a

vizro-core/docs/pages/user-guides/extensions.md
maxschulz-COL: Yes, that is a very good idea. Let me add a TODO to the code

vizro-core/src/vizro/models/types.py
maxschulz-COL: n/a anymore

vizro-core/tests/unit/vizro/models/test_base.py
maxschulz-COL: Yes, and we also need a test to check the `.run` failure

vizro-core/src/vizro/models/types.py
antonymilne: I think this part is still relevant though if you disagree then I'm happy how you have it:
> I would also suggest we do allow_undefined_captured_callable that than allowed because it leaves open the possibility of allow_undefined_captured_callable = True if we want to do it in future, and it still makes sense as an argument name if it's just a list of allowed things.

vizro-core/docs/pages/user-guides/extensions.md
maxschulz-COL: Ok will do although I did this on purpose because I like example that are simply copy past and run as is.

vizro-core/src/vizro/models/types.py
maxschulz-COL: ah im sorry i had overlooked the second paragraph

PR #1267, opened 2025-07-01
[Feat] Add `custom_header` for custom header content injection

huong-li-nguyen:
## Description

This PR introduces a new `custom_header` static method to the `Dashboard` class, enabling users to inject custom Dash components into a dedicated area of the dashboard header. The custom content is arranged in a horizontal flex box with an 8px gap and appears to the left of the theme switch.

Additionally, several dashboard header class names have been updated for clarity and consistency. These changes may affect any custom CSS targeting header elements.

Looking ahead, we could consider adding 1–3 additional customizable areas within the dashboard layout—such as a section below the controls. These areas would follow the same pattern, allowing users to override corresponding private methods to insert their own content. Overall, I don't expect we'll need more than 1–3 customizable containers, but we can wait and see where users might want to extend the layout.

Closes https://github.com/McK-Internal/vizro-internal/issues/1932

## Migration Notes
If you have custom CSS targeting the dashboard header or its sections, please update your selectors to match the new class names.

- **Class Name Changes:**  
    - The dasboard header is now `header` (was `page-header`)
    - The right section of the dashboard header is now `header-right` (contains custom-header and theme switch)
    - The left section of the dashboard header is now `header-left` (was `logo-and-title`)
    - The page title bar is now `page-header`

## Screenshot

custom_header_area

Example:
dashboard_custom_header

## Notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

    - I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
    - I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
    - I certify that the use of this contribution as authorized by the Apache 2.0 license does not violate the intellectual property rights of anyone else.
    - I have not referenced individuals, products or companies in any commits, directly or indirectly.
    - I have not added data or restricted code in any commits, directly or indirectly.

PR #1268, opened 2025-07-01
[Tidy] Refactor private dashboard layout methods

huong-li-nguyen:
## Description

This PR refactors the dashboard page layout logic by introducing three modular methods: `_inner_page`, `_outer_page`, and `_arrange_page`. These methods work together to build the dashboard’s page structure in a clearer way.

#### Key Changes

- **Introduced `_inner_page`**: Constructs and returns a dictionary of the main layout components for a dashboard page.
- **Introduced `_outer_page`**: Assembles the outer containers of the page using the components from `inner_page`, organizing them into higher-level layout containers.
- **Introduced `_arrange_page`**: Combines the outer containers from `outer_page` into the final Dash layout for rendering.
- Removed outer containers "left-sidebar" and "left-main" -> Use "nav-bar" and "nav-control-panel" instead

#### Layout Hierarchy and Relationships

**InnerPageContent:**

- header-left
- header-right
- page-header
- page-components
- nav-bar
- nav-control-panel

**OuterPageContent:**

- header
- right-side
- collapse-left-side,
- collapse-icon-outer

The structure of `InnerPageContentType` and `OuterPageContentType` is now mostly aligned, with a direct parent-child relationship, except for `collapse-left-side`, which acts as an additional wrapper required for the collapse functionality.

- **header**: Direct parent of `header-left` and `header-right`.
- **right-side**: Direct parent of `page-header` and `page-components`.
- **left-side**: Direct parent of `nav-bar` and `nav-control-panel` (created directly in `outer_page` and plugged into `collapse-left-side`).
- **collapse-left-side**: Direct parent of `left-side`, providing the collapsible sidebar functionality.
- **collapse-icon-outer**: A standalone, floating component for toggling the sidebar.

## Screenshot
See scratch-dev for a potential layout customization

Screenshot 2025-07-02 at 10 23 57

## Notice

### DIFF
--- a/vizro-core/changelog.d/20250701_130845_90609403+huong-li-nguyen_refactor_arrange_layout.md
+++ b/vizro-core/changelog.d/20250701_130845_90609403+huong-li-nguyen_refactor_arrange_layout.md
@@ -10,12 +10,11 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 - A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
 
 -->
-<!--
 ### Removed
 
-- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+- Simplified the `left-side` container by removing the extra `left-sidebar` and `left-main` containers. The left-side now directly uses `nav-bar` and `nav-control-panel` as its children. ([#1268](https://github.com/mckinsey/vizro/pull/1268))
 
--->
+
 <!--
 ### Added
 


vizro-core/examples/scratch_dev/app.py
```diff
@@ -5,20 +5,28 @@
...
+        collapsible_icon_div = outer_page_divs["collapsible-icon-div"]
+        right_side = outer_page_divs["right-side"]
+        d_header = outer_page_divs["d-header"]
+
+        page_main = html.Div(id="page-main", children=[d_header, right_side])
+        page_main_outer = html.Div(
+            children=[collapsible_left_side, collapsible_icon_div, page_main],
+            className="page-main-outer no-left" if _all_hidden(collapsible_icon_div) else "page-main-outer",
+        )
+        return page_main_outer
```
huong-li-nguyen: This change makes it much easier to customize the layout now, but users still need some basic CSS (see custom.css for required changes). In particular, they should understand how Flexbox works — otherwise, the direction might not behave as expected. They also need to handle cases where the left-side container isn’t present, as the layout depends on that.

The only way to avoid this would be to offer this alternative layout as a built-in, configurable option. That’s something we can consider down the line, but for now, I think this solution is good enough. I assume anyone interested in tweaking `_arrange_page_divs` likely has some CSS experience anyway — without that, it would be hard to adjust the container layout properly.

nadijagraca (approved):
I really like this refactor. Great job Li. 🚀 
I was able to make elementary horizontal navigation with few simple tweaks. 👍

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -74,6 +74,16 @@ def _all_hidden(components: Component | list[Component]):
     },
 )
 
+_OuterPageDivsType = TypedDict(
```
antonymilne: This one is also no longer just `Div` so should be renamed. As with my comments in #1267 it feels a bit "inconsistent scale" to have `collapsible-icon-div` on the same level as `right-side`. Is there any way we can fix that? Not a big problem if not though.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -283,53 +294,50 @@ def _get_page_divs(self, page: Page) -> _PageDivsType:
             ]
         )
 
-    def _arrange_page_divs(self, page_divs: _PageDivsType):
+    def _get_outer_page_divs(self, inner_page_divs: _InnerPageDivsType) -> _OuterPageDivsType:
```
antonymilne: When you want to make these public I'd go for something simpler like just `outer_page` as the function name - so no `get` or `divs`. e.g. we'd have `custom_header`, `outer_page`, `inner_page`, `arrange_page`.

antonymilne (approved):
My comments on #1267 apply here too 🙂 Basically looks great and whenever you're happy with it and everything feels consistent and nicely named, then we can make it public.

vizro-core/src/vizro/models/_dashboard.py
huong-li-nguyen: I've restructured things so the containers are mostly on the same level. But in this case, the `collapsible-icon-div` (now `collapse-icon-outer`) has to stay on the same level as `collapse-left-side` and `right-side`. This is needed to make sure the `left-side` collapse works properly in all situations, like when the `left-side` is there or not / when it's collapsed or not.

**This is the final split:**

**InnerPageContent:**

- header-left
- header-right
- page-header
- page-components
- nav-bar
- nav-control-panel

**OuterPageContent:**

- header
- right-side
- collapse-left-side,
- collapse-icon-outer

The structure of `InnerPageContent` and `OuterPageContent` is now mostly aligned as well, with a direct parent-child relationship, except for `left-side-collapse`, which acts as an additional wrapper needed for the collapse functionality. 

- `header`: Direct parent of `header-left `and `header-right`.  
- `right-side`: Direct parent of `page-header` and `page-components`.  
- `left-side`: Direct parent of `nav-bar` and `nav-control-panel` (It’s not stored anywhere; it’s created directly in `outer_page` and plugged into `collapse-left-side`)  
- `collapse-left-side`: Direct parent of `left-side.`  
- `collapse-icon-outer`: A standalone, floating component that doesn’t directly tie into the hierarchy (for restriction mentioned above)

huong-li-nguyen:
@antonymilne – I originally had `inner_page`, `outer_page`, and `arrange_page` as public methods, but while working on the docs, I realized it probably doesn’t make sense to expose them just yet. I’d love to chat more about this when you’re back, so for now I’m going to make them private. `custom_header` is public though from previous PR!

**Just to explain my thinking:**  
Right now, `_inner_page` and `_outer_page` are mainly responsible for building the standard layout containers (see the lists below). For `_inner_page`, I honestly couldn’t come up with any real use cases where someone would want to customize the order or structure of these inner containers—like, it always makes sense for the nav panel to be above the control panel, and for them to be stacked vertically. The same goes for the other sections, so I don’t see a reason to make `_inner_page` public at this point.

For `_outer_page`, it’s a similar story. The header and right-side containers only really make sense in their current order and direction. The only tweak I could imagine is moving the navbar into the header, but that would actually require a bigger redesign of the NavBar itself (like switching from icons to text/dropdowns), not just shuffling containers or updating CSS. So again, not much point in making `_outer_page` public for now.

As for `_arrange_page`, the only other layout I can think of is the one in `scratch-dev`, but I’m not sure that’s enough of a reason to expose it either. Since you’re away at the moment, I’ll just keep these as private for now. This change already helps a lot by cutting down on boilerplate in our recent demos, so it’s more than enough!

**Just for reference, here’s what’s in each content dict:**

**InnerPageContent:**
- header-left
- header-right
- page-header
- page-components
- nav-bar
- nav-control-panel

**OuterPageContent:**
- header
- right-side
- collapse-left-side
- collapse-icon-outer

PR #1272, opened 2025-07-03
[Fix] Improve various visual components

nadijagraca:
## Description
closes https://github.com/McK-Internal/vizro-internal/issues/1908

Control Panel
- [x] align the text of "select page" with the centre of the icon and the centre of the panel collapse button and the page title
- [x] replace "SELECT PAGE" with "Select Page" (ie. not all caps)

KPI cards
- [x] resolve why KPI card still too tall in some screen sizes (eg. it turns on the scroll bar in most laptop view, which doesn't look great for most CSTs, so it should probably be optimised to avoid this  - could not fully resolve this issue, but I removed the scroll bar from the regular kpi card. 

Icon bar
- [x] make icon bar narrower?
- [x] make sure the logo aligns vertically centrally with the icons in the left panel - the latest example from PMPx on 1st May 2025 didn't

## Notice


nadijagraca:
Logo file was added for testing purposes only, I will remove it before merging.

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -118,6 +118,11 @@
   height: calc(100vh - 64px);
 }
 
+#dashboard-title {
+  gap: 0;
+  padding-left: 0.5rem;
+}
+
```
huong-li-nguyen:
```suggestion
```

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -146,7 +151,6 @@
   align-items: center;
   display: flex;
   flex-direction: row;
-  gap: 8px;
```
huong-li-nguyen: Revert this change, I'll explain later 😬

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -56,7 +56,7 @@
   height: 60px;
   justify-content: space-between;
   min-height: 0;
-  padding: 0 24px;
+  padding: 0 24px 0 0;
```
huong-li-nguyen: This has to change to 20px because we made the icon-bar narrower. I'll explain later, but I would revert the 0 0 change here. 

```suggestion
  padding: 0 20px;
```

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -19,7 +19,7 @@
   flex-direction: column;
   gap: 1.5rem;
   overflow: auto;
-  padding: 24px 24px 0;
+  padding: 20px 24px 0;
```
huong-li-nguyen: I think it looks a bit off, if some side-padding has 20px and another 24px and then another 0. Would just align with 20px then, because of the navbar that is narrower now. 

```suggestion
  padding: 20px;
```

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
```diff
@@ -60,13 +60,16 @@ All the HEX values starting with --text-code are taken from the Github code high
...
 /* Move background color to BS */
 .accordion-item .nav-link:active, .accordion-item .nav-link.active,
 .accordion-item .nav-link:hover, .accordion-item .nav-link.hover {
   background: var(--surfaces-bg-card);
 }
 
+.accordion-button {
+  text-transform: none;
+}
+
```
huong-li-nguyen: For what do we need this? 🤔

vizro-core/src/vizro/static/css/figures.css
```diff
@@ -1,7 +1,7 @@
 .card-kpi {
   border-left: 4px solid var(--bs-secondary);
   height: 100%;
-  padding: 1rem;
+  padding: 0.5rem 1rem;
```
huong-li-nguyen: Not sure about this change, let's discuss 😬  If we want to make this smaller general, I would just do `padding: 0.5rem` then

huong-li-nguyen (changes requested):
Hey @nadijagraca,

Could you set up a quick call with J., you, and me? I think it'll be more efficient than me typing everything out here :) I know you just implemented the requested changes, but I don’t fully agree with all of them.

These changes are fine from my side:

- Align the "Select Page" text with the center of the icon and the panel collapse button
- Change "SELECT PAGE" to "Select Page" (no all caps)
- Make the icon bar narrower

Where I see an issue is with the logo/dashboard title adjustments. I briefly touched on this with J. a while ago, but essentially: the CSS you added optimizes the layout specifically for our current logo (with 4px padding and square proportions). But other users might upload logos with different padding or shapes, which would make the alignment look off. Also, if no logo is used, the dashboard title now looks misaligned.

**What was previously implemented:** We ensure the dashboard title aligns properly with the navbar icons and accordion. But it’s the user’s responsibility to upload and position their logo appropriately, since we can’t anticipate every logo style.

Also, if we narrow the icon bar, some paddings need to be adjusted as well to keep everything aligned — I can quickly show you what I mean on the call 👍

vizro-core/src/vizro/static/css/bootstrap_overwrites.css
nadijagraca: Apparently `accordion-button` has `text-transform: capitalize`, so this is needed to fix that. Let's discuss on the call.

huong-li-nguyen (commented):
Hey @nadijagraca,

I pushed a minor fix into your branch that I introduced. Hope you don't mind!

Overall, looks good, however, I think because we aligned the collapse-icon with the accordion-title, it is now misaligned with the page-title. I think reducing the padding of the accordion title could help here.

You would then also need to check the mobile version, because the position of the collapse-icon would probably have to change.
Screenshot 2025-07-07 at 19 14 25
Screenshot 2025-07-07 at 19 15 56

huong-li-nguyen (approved):
Hi @nadijagraca,

I’ve pushed a few updates. For the KPI cards, I reduced the padding slightly and decreased the font size—since it was actually bigger than the page title. From a hierarchy perspective, that didn’t feel quite right. These tweaks should also help reduce some of the scrolling 😊

I also made a few small alignment adjustments to the accordion title, collapse button, and page title to ensure everything lines up more precisely.

Feel free to merge after updating the screenshots 👍 🚀

PR #1276, opened 2025-07-09
[Tidy] Refactor Slider and RangeSlider models and JS

petar-qb:
## Notice


petar-qb:
Hey @huong-li-nguyen 😃 and thanks for testing the example.

This bad UX experience originates from the fact that, if not specified, the selector's `step` is automatically calculated. It ofter leads to the bad UX when min/max are float numbers (same on the `main` branch - see [here](https://github.com/mckinsey/vizro-qa/pull/137)). For example, if these values are `[4.3, 7.6]`, the `step` will be automatically set which will enable user to input only the following number through the text input `[4.3, 5.3, 6.3, 7.3]`. Other values will not be accepted. 

The workaround would be configuring the selector in a following way: `selector=vm.RangeSlider(step=0.01, marks=None)`.

What do you think, should we automatically set step and marks (from the Filter/Parameter.pre_build) when we find the min/max as floats? 🤔

huong-li-nguyen:
> What do you think, should we automatically set step and marks (from the Filter/Parameter.pre_build) when we find the min/max as floats? 🤔

yeah, I thought it would be something like this. Let’s leave it out for now, because I think we would have to investigate again. If the step size is that small by default it might look too crowded if someone turned on the marks. In the end, they can just use the handle which is more common anyway 👍

petar-qb:
I agree! Here's the ticket for fixing that -> https://github.com/McK-Internal/vizro-internal/issues/1971

vizro-core/src/vizro/static/js/models/range_slider.js
```diff
@@ -1,62 +1,26 @@
...
-  trigger_id = dash_clientside.callback_context.triggered;
-  if (trigger_id.length != 0) {
-    trigger_id =
-      dash_clientside.callback_context.triggered[0]["prop_id"].split(".")[0];
+  // slider component is triggered (slider_value)
+  if (trigger_id === slider_component_id) {
+    return [dash_clientside.no_update, slider_value[0], slider_value[1]];
   }
 
-  // text form component is the trigger
```
nadijagraca, line 22: Glad to see all of this gone. :) 🚀

nadijagraca (approved):
Looks good. I'm glad to see slider callbacks simplified. 🚀

PR #1280, opened 2025-07-11
[BUG] Fix unintended value clearing for dropdown filter that targets dynamic data_frame

petar-qb:
closes https://github.com/McK-Internal/vizro-internal/issues/1494
closes https://github.com/McK-Internal/vizro-internal/issues/1637

## Description

As detailed in one of the linked tickets, the issue stems from how `dcc.Dropdown` handles selected `values` that aren't present in its current `options`. The component clears any value entries that are not part of the current options, leading to unintended behaviour in specific scenarios during page load.

Here’s a breakdown of the problem. When the page is opened:
1. PB must return the `original default value` to avoid overwriting any persisted state.
2. Persistence is applied next, replacing the value from PB with the stored (currently selected) one.
3. OPL then executes with the goal of returning `updated dynamic options` and the `currently selected value`.

Problem scenario:
1. Initial state: `options = ["A"]`, `value = ["A"]`
2. User later selects: `value = ["A", "B", "C"]` after dynamic options become `["A", "B", "C"]`
3. User refreshes the page:
   1. PB returns `options = ["A"]`, `value = ["A"]`
   2. Persistence restores from `value = ["A"]` -> `value = ["A", "B", "C"]`
   3. `Issue`: dcc.Dropdown internally removes any values not in current options, so it removes `"B"` and `"C"` because PB only returned `"A"` in options.
   4. OPL receives and returns `value = ["A"]` (incorrect), `options = ["A", "B", "C"]`

Outcome:
Even though the correct options and value are available, internal dropdown implementation mistakenly drops part of the selection, resulting in only `["A"]` being shown instead of the expected `["A", "B", "C"]`.

This behaviour is only specific to the `dcc.Dropdown` component and is caused by the component's internal validation logic. See the relevant Dash PR that introduced this mechanism -> https://github.com/plotly/dash/pull/1970.

## For reviewers:

As it's already added into the scratch dev example, here are the steps for recreating the bug on the main branch:
```md
To reproduce the #1494 bug on the main:
1. Open the first page
2. Set the number of rows to 150 (bottom slider)
3. Select ["setosa", "versicolor"] in the multi-select dropdown
4. Select "versicolor" in the single-select dropdown
5. Refresh the page -> The value is unexpectedly reset to "setosa" in the multi-select dropdown and clear
   in the single-select dropdown. The value should remain the same as it was before the refresh.
```

```md
To reproduce the #1637 bug on the main:
1. Open the fourth page
2. Set the number of rows to 150 (bottom slider)
3. Select ["setosa", "versicolor"] in the multi-select dropdown
4. Refresh the page -> The dropdown value is incorrect, and there's also one more http request that happened after the OPL (see the browser devtools console).
```
## Notice


nadijagraca (approved):
All looks and works nicely. Great description, made my review so much easier. 🙌

vizro-ai/changelog.d/20250711_105938_petar_pejovic_fix_unintended_value_clearing_for_dynamic_dropdown_filter.md
```diff
@@ -0,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
```
huong-li-nguyen: I think you don't need this file :)

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -212,8 +199,54 @@ def _build_dynamic_placeholder(self):
...
+                ),
+                *hidden_select_all_dropdown,
+            ]
+        )
 
     @_log_call
     def build(self):
         return self._build_dynamic_placeholder() if self._dynamic else self.__call__(self.options)
+
+    def _define_clientside_callback(self):
```
huong-li-nguyen: Shall we call it similar to the function_name it calls?

```suggestion
    def _update_dropdown_select_all"(self):
```

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -212,8 +199,54 @@ def _build_dynamic_placeholder(self):
...
+            # Add the clientside callback as the callback has to be defined in the page.build process.
+            self._define_clientside_callback()
+            # hidden_select_all_dropdown is needed to ensure that clientside callback doesn't raise the no output error.
+            hidden_select_all_dropdown = [dcc.Dropdown(id=f"{self.id}_select_all", style={"display": "none"})]
+            placeholder_model = dcc.Checklist
+            placeholder_options = self.value
+        else:
+            hidden_select_all_dropdown = [None]
+            placeholder_model = dbc.RadioItems
+            placeholder_options = [self.value]  # type: ignore[assignment]
```
huong-li-nguyen, line 216: Is the solution with the placeholder models just a temporary one? Or is it something that's unlikely to be changed in dash and probably our permanent solution? Because on initial look, I was a bit confused why we had these placerholder models inside the `Dropdown` 🤔

huong-li-nguyen (approved):
I checked the example app, and the bug is fixed as you mentioned 👍 Thank you!

However, I am not sure if I fully grasp the underlying issue and whether these placeholder models are just a temporary solution until something is fixed on dash' side or if it's supposed to be our permanent solution?

On initial look, it seemed a bit odd to me to have these placeholder models inside the Dropdown, but I assume it's the only way? 🤔

vizro-core/src/vizro/models/_components/form/dropdown.py
petar-qb: This behaviour is unlikely to change in Dash, as this dropdown functionality is deliberately developed as a feature. While the permanent solution would be implementing the [Universal Vizro Placeholder Component](https://github.com/McK-Internal/vizro-internal/issues/1307), we ultimately chose not to pursue it due to its complexity.

The idea for the universal placeholder was originally inspired by this exact issue. We also wanted to increase the genericity of handling the dynamic placeholders as they differ in Vizro. However, given the implementation challenges, we decided not to proceed.

PR #1283, opened 2025-07-14
[Feat] Enable `show_select_all` argument in `vm.Checklist`

nadijagraca:
## Description
closes https://github.com/McK-Internal/vizro-internal/issues/1498

Enable users to remove 'Select All' option from checklist by specifying `show_select_all=False`

## Notice


vizro-core/src/vizro/models/_components/form/checklist.py
```diff
@@ -47,6 +48,7 @@ class Checklist(VizroBaseModel):
     title: str = Field(default="", description="Title to be displayed")
     # TODO: ideally description would have json_schema_input_type=Union[str, Tooltip] attached to the BeforeValidator,
     #  but this requires pydantic >= 2.9.
+    show_select_all: bool = Field(default=True, description="Boolean flag to show 'Select All' option.")
```
huong-li-nguyen:
```suggestion
    show_select_all: bool = Field(default=True, description="Whether to display the 'Select All' option that allows users to select or deselect all available options with a single click.")
```

vizro-core/src/vizro/models/_components/form/checklist.py
```diff
@@ -157,3 +152,20 @@ def _build_dynamic_placeholder(self):
     @_log_call
     def build(self):
         return self._build_dynamic_placeholder() if self._dynamic else self.__call__(self.options)
+
+    def _define_clientside_callback(self):
```
huong-li-nguyen: Ah, do we call it `_define_clientside_callback` because we call this for all selectors inside the page.build? I saw it in @petar-qb PR as well, but was a bit confused why we call it like this instead of the actual function_name

huong-li-nguyen: Ah, I think he has renamed it to the actual function name after my comment, so would be great if we do this here too 👍

vizro-core/src/vizro/models/_components/form/checklist.py
```diff
@@ -29,6 +29,7 @@ class Checklist(VizroBaseModel):
         options (OptionsType): See [`OptionsType`][vizro.models.types.OptionsType]. Defaults to `[]`.
         value (Optional[MultiValueType]): See [`MultiValueType`][vizro.models.types.MultiValueType]. Defaults to `None`.
         title (str): Title to be displayed. Defaults to `""`.
+        show_select_all (Optional[bool]): Optional flag to include 'Select All' option. Defaults to True.
```
huong-li-nguyen:
```suggestion
        show_select_all (bool): Whether to display the 'Select All' option that allows users to select or deselect all available options with a single click. Defaults to `True`.
```

vizro-core/src/vizro/models/_components/form/checklist.py
```diff
@@ -157,3 +156,20 @@ def _build_dynamic_placeholder(self):
     @_log_call
     def build(self):
         return self._build_dynamic_placeholder() if self._dynamic else self.__call__(self.options)
+
+    def _define_clientside_callback(self):
```
huong-li-nguyen:
```suggestion
    def _update_checklist_select_all(self):
```

PR #1284, opened 2025-07-14
[Tidy] Tidy icon naming to use Title Case

nadijagraca:
## Description

closes https://github.com/McK-Internal/vizro-internal/issues/1642

Small PR to tidy icon naming to use Title Case

## Notice


vizro-core/examples/dev/app.py
```diff
@@ -57,7 +57,7 @@
         data_frame=df_kpi,
         value_column="Actual",
         title="KPI with icon",
-        icon="shopping_cart",
+        icon="Shopping Cart",
```
huong-li-nguyen, line 60: I think the yaml examples would also need to be updated for all the examples you've updated :)

huong-li-nguyen (approved):
LGTM 👍 I think the only thing remaining is to update the yaml examples as well

huong-li-nguyen (changes requested):
Ahh, I just check double-checked again and I think there are more occurances e.g. look for "shopping_cart" in our repository. It's in many of our docs examples and tests as well.

huong-li-nguyen (approved):

huong-li-nguyen:
> * e2e vizro screenshots tests / test-e2e-vizro-screenshots for chrome (pull_request)

Something seems to be off with the screenshots, seems like it doesn't actually get rendered correctly with that typing 🤔 Could you double-check? In theory it should work, but if not, we might need to fix something first.

main_kpi_indicators_page

huong-li-nguyen:
> > * e2e vizro screenshots tests / test-e2e-vizro-screenshots for chrome (pull_request)
> 
> Something seems to be off with the screenshots, seems like it doesn't actually get rendered correctly with that typing 🤔 Could you double-check? In theory it should work, but if not, we might need to fix something first.
> 

Ah, we don't apply the same validation inside our kpi cards. So we need to add that into the kpi_card_reference and kpi_card first! Like this:

```

from vizro.models._models_utils import validate_icon

...


html.P(validate_icon(icon), className="material-symbols-outlined") if icon else None
```

PR #1290, opened 2025-07-17
[Fix] Add default column size to AgGrid

vladimir-mck:
## Description

Closes [#1986](https://github.com/McK-Internal/vizro-internal/issues/1986). 

The columnSize property was functional, but it was incorrectly passed via dashGridOptions. It should instead be applied directly to the dash_ag_grid component.

```  
vm.AgGrid(figure=dash_ag_grid(
    tips,
    columnSize = "sizeColumnsToFit"
 ))
```

I discovered that some of the column sizing options don't play well with flex on columns. To fix this, I removed the flex setting and instead set a default columnSize property to the AgGrid component. This delivers the same responsive behavior without conflicts.

**Technical details**

- Removed: flex from column definitions, which was causing layout issues in combination with columnSize.
- Added: columnSize prop directly on the Dash AG Grid component, ensuring proper column scaling per documented behavior (e.g., autoSize, sizeToFit, or responsiveSizeToFit) .

## Notice


vizro-core/changelog.d/20250717_113142_vladimir_nikolic_ag_grid_columnSize.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
```
huong-li-nguyen: Can you document this under Fixed?

vizro-core/src/vizro/tables/_dash_ag_grid.py
```diff
@@ -93,6 +92,8 @@ def dash_ag_grid(data_frame: pd.DataFrame, **kwargs: Any) -> dag.AgGrid:
             "pagination": True,
             "paginationPageSize": 20,
         },
+        "columnSize": "sizeToFit",
```
huong-li-nguyen, line 95: I didn't check yet, but did we double-check all of our previous bug-fixes and if all of these work as well? If yes, then great 🥳 

https://py.cafe/huong-li-nguyen/vizro-flex-grid-layouts

huong-li-nguyen (approved):
The dev example is wonderful! 🥳 This looks good to me - could you just:
- Update the changelog
- Double-check and update the screenshots
- Double-check the pyCafe examples and eventually previous bug-fixes we added for the Table whether these are all still gone

vizro-core/src/vizro/tables/_dash_ag_grid.py
vladimir-mck: Yep, I checked. I have these examples saved and check on them as well when I change something.

huong-li-nguyen: Amazing! 💯

huong-li-nguyen:
Would be amazing if you could create docs on this next! :) https://github.com/McK-Internal/vizro-internal/issues/1993
It can be in a separate PR though 👍

PR #1292, opened 2025-07-22
[Tidy] Remove global font and custom font overrides

huong-li-nguyen:
## Description

- Remove the global font-family definition from `index.css`, as it is now handled by the Bootstrap theme for the body.
- Clean up `code.css` by removing unnecessary font-family overrides, which are no longer required.

**Context:**
With the Bootstrap theme now managing the font-family for the body, the global font definition in `index.css` is redundant. As a result, we only need to explicitly define font-families for components that use a different native font-family. In this case, adjustments were luckily only required for the `AgGrid` and `Table`.

The primary benefit of this change is improved code clarity and maintainability. Previously, we had to include numerous overrides to ensure embedded code adhered to a monospace font, as the global font-family was overriding it. With the global font removed, these overrides are no longer necessary.

Screenshots only had to be updated because the Plotly icons seem to look different, but it has nothing to do with the font:
main_ag_grid_page_difference_from_main

## Notice


vizro-core/src/vizro/static/css/aggrid.css
```diff
@@ -1,5 +1,6 @@
 .ag-theme-quartz.ag-theme-vizro,
 .ag-theme-quartz-dark.ag-theme-vizro {
+  --ag-font-family: inter, sans-serif, arial, serif;
```
antonymilne, line 3: Very minor but the capitalisation here is different from in `table.css`.

antonymilne (approved):
Ahhh yes, the number one annoyance that I always notice and no one else cares about! Amazing work 😀

vizro-core/src/vizro/static/css/aggrid.css
huong-li-nguyen: Yes, for some reason, the linter wants it lower-case letter here 🤔 I've kept it lower-case now

PR #1300, opened 2025-07-24
[Tidy] Fix flex layout

vladimir-mck:
## Description

Close [#2002](https://github.com/McK-Internal/vizro-internal/issues/2002)

This PR resolves a layout issue caused by flex growth. 

- Removed flex: 1 from flex items to prevent them from expanding and overflowing their container.
- Aligned the flex row to flex-start to ensure consistent and predictable item positioning.

## Notice

### DIFF
--- a/vizro-core/src/vizro/static/css/tabs.css
+++ b/vizro-core/src/vizro/static/css/tabs.css
@@ -37,6 +37,10 @@ To ensure the dynamic height adjustment and prevent scrolling, the height must b
   height: 100%;
 }
 
+.tab-pane > div > .container-fluid .flex-row {
+  align-content: flex-start;
+}
+
 .inner-tabs-title {
   align-items: center;
   display: flex;
--- a/vizro-core/changelog.d/20250724_145237_vladimir_nikolic_flex_layout.md
+++ b/vizro-core/changelog.d/20250724_145237_vladimir_nikolic_flex_layout.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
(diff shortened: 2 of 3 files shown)


vizro-core/src/vizro/static/css/tabs.css
```diff
@@ -37,6 +37,10 @@ To ensure the dynamic height adjustment and prevent scrolling, the height must b
   height: 100%;
 }
 
+.tab-pane > div > .container-fluid .flex-row {
+  align-content: flex-start;
+}
+
```
huong-li-nguyen: Shall we rather do below? So it's always valid, not only when it's a flex-container inside tabs?
```suggestion
.container-fluid .flex-row {
  align-content: flex-start;
}

```

vizro-core/examples/scratch_dev/app.py
```diff
@@ -2,335 +2,428 @@
 # check out https://github.com/mckinsey/vizro for more info about Vizro
 # and checkout https://vizro.readthedocs.io/en/stable/ for documentation.
 
-import pandas as pd
```
huong-li-nguyen, line 5: Could you double-check the example `Flex - row - table`? I think that one still looks a bit off. Optimally, it should look visually similar to `Flex - row - aggrid`. But let me know if this is not possible 🤔

huong-li-nguyen (commented):
Have one last example to check, but otherwise this looks a lot better already! 🚀 

Don't forget to add the below to the scratch dev for the integration tests to pass 👍 

`if __name__ == "__main__":`

vizro-core/examples/scratch_dev/app.py
vladimir-mck: Flex item width is based on it's content.  They behave the same on wide screens.  

See the second screenshot has empty space to the right. 

Screenshot 2025-07-25 at 10 57 05


Flex row ag grid:
Screenshot 2025-07-25 at 10 56 44

huong-li-nguyen: Ah, I think the difference then comes from the padding. Each AgGrid cell and header have a padding of `padding: 0 0.75rem`. Can we do the same for the Table? So add a padding of `0 0.75rem` to each cell and header cell and then make sure it's aligned nicely. Then I think the screenshot above should look nicer. Because at the moment, it looks off, because some columns are left/right aligned and it's not clear where a column starts/ends.

PR #1301, opened 2025-07-24
[Tidy] Remove actions loop

antonymilne:
## Description

Closes https://github.com/McK-Internal/vizro-internal/issues/1613.

An extremely long-awaited PR! This hopefully completes the last major piece reworking the actions system 🤞 After this PR, there will still be further improvements and new features, but all the _core architecture_ should be complete.

## Key change

The actions loop and `ActionsChain` model have been **completely removed**. These have done a sterling job for us and we should be hugely grateful to @petar-qb for all the hard work he spent engineering them. However, they have now been replaced by a mechanism which is much simpler and will be easier for users to understand and integrate with Dash callbacks.

For all users this change will for now be unnoticed because there's no apparent difference in behaviour. The power that the changes bring will become apparent soon e.g. with the interact functionality.

`actions = [action_1, action_2]` works exactly the same as before, i.e. `action_2` will run only after `action_1` has completed. This is called an _actions chain_. What's different now is:
* there's no underlying `ActionsChain` model governing this. So `actions` is now just `list[ActionsType]` rather than something involving a private type
* the mechanism that makes an action chain execute correctly has been hugely simplified from the old actions loop system. in fact it's so simple that there's no need for any external actions loop system at all any more - it's all written in `actions.py` instead
* you can now trigger the initial action in an actions chain from a pure Dash callback, which will then trigger the remaining actions in that chain as per normal Dash callbacks. Same as before: actions _after_ the initial one in an actions chain still can't be triggered outside the action chain.

Of course, 99% of the time currently we just have a single action in the chain, like `actions = [action_1]`. It remains to be seen if we eventually remove the concept of an actions chain completely (would be breaking change) or how/whether we allow parallel as well as serial execution.

## New actions chain mechanism

Public documentation for users will follow shortly in a separate PR, but here's a quick guide for developers. Firstly, forget everything you know about actions, actions chain, actions loop, etc. Here's what happens now.

One Vizro action wraps one Dash callback. This callback is defined in `Action._define_callback` as `action_callback`. In terms of callback execution there are only _three_ differences between Vizro actions and Dash callbacks: actions chain,  one trigger, guard.

**1. Actions chain** 

Vizro provides a mechanism for a user to chain arbitrary multiple actions together without needing to write for themselves extra machinery or dependencies between actions. Writing `actions = [action_1, action_2]` means "run callback for `action_1` followed by callback for `action_2`". 

Behind the scenes, Vizro does this by creating a dummy Dash input/output. In the above example this would be `dcc.Store(id="action_1_finished)")` which would be an `Output` of `action_1` and `Input` for `action_2`.

**2. One trigger**

The Dash callback for a Vizro action can have only a single trigger (Dash `Input`). You can still enter other information into the callback using Vizro `inputs` (Dash `State`), but these can't trigger the callback.

The trigger for a Vizro action is:
* for the first action in a chain given by `action_triggers["__default__"]`. In future a user would be able to configure this to other properties (e.g. hover vs. click on a cell in an AgGrid)
* for all subsequent actions in the chain, the action is triggered by the completion of the preceding action using the `dcc.Store(*_finished)` dummy object

In practice, this means a pure Dash callback can trigger the start of an actions chain (which would then trigger execution of the rest of the actions chain) but cannot trigger execution halfway through an actions chain.

Pure Dash callbacks can be triggered by any action regardless of its position in the chain.

**3. Guard**

There is a one-to-one mapping between Vizro action and a Dash callback, but each actions chain also has an extra callback. This is defined in `Action._defined_callback` as `guard_actions_chain`. The purpose of this is a workaround for Dash behaviour whereby `prevent_initial_call` doesn't actually prevent firing of a callback when a new component is created. Getting this to work correctly is a small technical detail that you can basically ignore but was the hardest part to get right.

## Other code changes

* `_action_triggers` written for all relevant models and used in the `make_actions_chain` validator. This replaces the old `ActionsChain` validator
* `Action.build` is now called `Action._define_callback` and doesn't return any Dash components
* Slight rewrite of `filter_interaction` helper functions now that we don't need to loop through `ActionChains` to match up models

## How to test

@petar-qb I've changed logging levels and used `logging.critical` just as a hack to make manual testing easier. When we're finished manually testing we should change revert that and change the `guard_action_callback` to clientside.

TODO:
- [x] update custom-components.md @antonymilne 
- [x] consider changing AgGrid trigger to `selectedRows` or just don't set for now and wait for following PR to do. Need `filter_interaction` to work though @antonymilne
- [x] resolve remaining `TODO NOW` @petar-qb including fixing URL parameters
- [x] manual testing @petar-qb @maxschulz-COL. To include different components, pages without on-page-load
- [x] check `vizro_download, vizro_url` still all work correctly (I didn't test anything here at all) @petar-qb  
- [x] review #1334 and #1328 and merge @petar-qb 
- [x] fix tests and write new ones @antonymilne or @petar-qb 
- [x] go through Max's review @antonymilne 
- [ ] update our developer docs @petar-qb 
- [ ] add new js unit tests @petar-qb @vladimir-mck 

## Notice


vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -323,13 +387,11 @@ def build(self) -> html.Div:
...
-        def callback_wrapper(external: Union[list[Any], dict[str, Any]], internal: dict[str, Any]) -> dict[str, Any]:
+        def action_callback(external: Union[list[Any], dict[str, Any]], internal: dict[str, Any]) -> dict[str, Any]:
             return_value = self._action_callback_function(inputs=external, outputs=callback_outputs.get("external"))
             if "external" in callback_outputs:
-                return {"internal": {"action_finished": None}, "external": return_value}
-            return {"internal": {"action_finished": None}}
-
-        return html.Div(id=f"{self.id}_action_model_components_div", children=self._dash_components, hidden=True)
+                return {"internal": {"action_finished": time.time()}, "external": return_value}
+            return {"internal": {"action_finished": time.time()}}
```
antonymilne, line 397: N.B. the `_define_callback` (was `build`) method no longer returns `self._dash_components`. These are obtained directly in `Page.build`.

vizro-core/src/vizro/models/_base.py
```diff
@@ -73,20 +70,13 @@ def _dict_to_python(object: Any) -> str:
     if isinstance(object, dict) and "__vizro_model__" in object:
         __vizro_model__ = object.pop("__vizro_model__")
```
antonymilne, line 72: FYI @maxschulz-COL no special case for handling actions chain any more 🎉

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -78,3 +80,55 @@ def warn_description_without_title(description, info: ValidationInfo):
...
+    # Convert any built in actions written in the legacy style vm.Action(function=filter_interaction(...)) or
+    # vm.Action(function=export_data(...)) to the new style filter_interaction(...) or export_data(...).
+    # We need to delete the old action models from the model manager so they don't get built. After that,
+    # built in actions are always handled in the new way.
+    for action in self.actions:
+        if isinstance(action.function, (export_data, filter_interaction)):
+            del model_manager[action.id]
+            converted_actions.append(action.function)
+        else:
+            converted_actions.append(action)
```
antonymilne, line 107: This part is copied across from the old `_actions_chain._set_actions` validator.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -51,6 +48,18 @@ class _BaseAction(VizroBaseModel):
     function: ClassVar[Callable[..., Any]]
     outputs: ClassVar[Union[list[str], dict[str, str]]]
 
+    # These are set in the make_actions_chain validator (same for both Action and _AbstractAction).
+    # In the future a user would probably be able to specify something here that would look up a key in
+    # _action_triggers or just a full _IdProperty. The _first_in_chain and _prevent_initial_call_of_guard would remain
+    # private though. These are required for correct functioning of the actions chain guard.
+    _trigger: _IdProperty = PrivateAttr()
```
antonymilne, line 57: Key change: all actions now have `_trigger` property. This would look like e.g. `graph_id.cellClicked`.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -285,24 +298,75 @@ def _action_callback_function(
...
-    def build(self) -> html.Div:
-        """Builds a callback for the Action model and returns required components for the callback.
-
-        Returns:
-            Div containing a list of required components (e.g. dcc.Download) for the Action model
-
-        """
-        # TODO: after sorting out model manager and pre-build order, lots of this should probably move to happen
-        #  some time before the build phase.
+    def _define_callback(self):
```
antonymilne, line 324: Key change: this function contains the actions chain mechanism: `guard_action_chain` and `action_callback`.

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -82,22 +80,18 @@ class AgGrid(VizroBaseModel):
...
     _validate_figure = field_validator("figure", mode="before")(validate_captured_callable)
+    _make_actions_chain = model_validator(mode="after")(make_actions_chain)
 
     def model_post_init(self, context) -> None:
         super().model_post_init(context)
         self._inner_component_id = self.figure._arguments.get("id", f"__input_{self.id}")
 
-    # TODO-AV2 E: Implement _action_trigger where makes sense.
-    #  For the AgGrid the mapping could look like: {"__default__": f"{self._inner_component_id}.cellClicked"}
+    @property
```
antonymilne, line 93: Key change: `_action_triggers` now defined throughout and used in the `make_actions_chain` validator. This replaces the old `AfterValidator(_action_validator_factory("cellClicked"))` mechanism with something much more robust and extensible.

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -126,7 +120,7 @@ def __call__(self, **kwargs):
             kwargs["data_frame"] = data_manager[self["data_frame"]].load()
         figure = self.figure(**kwargs)
         figure.id = self._inner_component_id
-        return figure
+        return html.Div([figure, dcc.Store(id=f"{self._inner_component_id}_guard_actions_chain", data=True)])
```
antonymilne, line 124: Key change: we return `*_guard_actions_chain` to operate the `guard_actions_chain` callback mechanism.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -158,6 +160,11 @@ def build(self):
         for page in self.pages:
             page.build()  # TODO: ideally remove, but necessary to register slider callbacks
 
+        # Define callbacks when the dashboard is built but not every time the page is changed.
```
antonymilne, line 163: Previously these callbacks were defined by code that lived inside `ActionLoop._create_app_callbacks()`. Now it's done explicitly here but effectively still defined in the same place.

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -78,3 +80,55 @@ def warn_description_without_title(description, info: ValidationInfo):
             UserWarning,
         )
     return description
+
+
+def make_actions_chain(self) -> Self:
```
antonymilne: Key change: this validator handles the triggers to link together an actions chain.

vizro-core/src/vizro/models/_page.py
```diff
@@ -214,9 +206,28 @@ def build(self) -> _PageBuildType:
...
 
+        # Components that are required to make action chains function correctly:
+        #   - {action.id}_guarded_trigger for the first action in a chain so that guard_action_chain can prevent
+        #     undesired triggering (workaround for Dash prevent_initial_call=True behaviour)
+        #   - {action.id}_finished for completion of an action callback to trigger the next action in the chain
+        #   - action._dash_components for particular actions (e.g. dcc.Download for export_data) - hopefully will be
+        #     removed in future
+        # These components are recreated on every page rather than going at the global dashboard level so that we do
+        # not accidentally trigger callbacks (workaround for Dash prevent_initial_call=True behaviour).
+        action_components = []
```
antonymilne: Key change: these components are what's needed for actions chains to work. Previously this sort of thing was done in `ActionsLoop`.

vizro-core/schemas/0.1.44.json
maxschulz-COL, line 1: This makes me smile :)

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -82,22 +80,18 @@ class AgGrid(VizroBaseModel):
...
-        list[ActionType],
-        AfterValidator(_action_validator_factory("cellClicked")),
-        PlainSerializer(lambda x: x[0].actions),
-        Field(default=[]),  # TODO[MS]: here and elsewhere: do we need to validate default here?
-    ]
-
+    actions: list[ActionType] = []
     _inner_component_id: str = PrivateAttr()
     _validate_figure = field_validator("figure", mode="before")(validate_captured_callable)
+    _make_actions_chain = model_validator(mode="after")(make_actions_chain)
```
maxschulz-COL: Just as a general comment. When working on the mkdocstring pydantic implementation, I noticed that this re-use pattern is not official and for example not caught by the extension. Since we are not using the extension yet, it is not a problem, but I guess it shows a little bit the fragility of doing it like that.

Since these are model validators, there are no strong alternatives apart from basically being a bit more verbose as shown in https://github.com/mckinsey/vizro/pull/1307/files#r2251067711

My preference would be slightly for the more verbose version just in order to do things "officially" and having less trouble with external plugins

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -78,3 +79,55 @@ def warn_description_without_title(description, info: ValidationInfo):
             UserWarning,
         )
     return description
+
+
+def make_actions_chain(self):
```
maxschulz-COL, line 84: Maybe a completely random question, but without having digested it as a whole: was not one of the advantages previously that we grouped all callbacks into a single http request in order to be more efficient? Is that still the case now with all the callbacks triggering one after the other?

vizro-core/src/vizro/models/_components/ag_grid.py
maxschulz-COL: In fact this looks to me like we may even want to define it on the BaseModel no?

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -74,6 +67,12 @@ class Button(VizroBaseModel):
         ]
     ]
 
+    _make_actions_chain = model_validator(mode="after")(make_actions_chain)
+
+    @property
```
maxschulz-COL, line 74: Is it clever to work with a mapping here? This would mean that we need to manually define in the future what other things would go in here, and the user selects via the key right?

Of course this makes my schema heart very happy, because the trigger of a model could be a well defined `enum`, but maybe on the other hand we would be unhappy to maintain such as thing?

vizro-core/src/vizro/models/_components/form/checklist.py
```diff
@@ -144,6 +142,7 @@ def __call__(self, options):
                     ],
                     className="checklist-inline" if self._in_container else None,
                 ),
+                dcc.Store(id=f"{self.id}_guard_actions_chain", data=True) if self._dynamic else None,
```
maxschulz-COL: Is that not something we could automate more? Does it matter where the the layout we having this given the ID?

vizro-core/src/vizro/models/_components/form/switch.py
```diff
@@ -75,6 +68,12 @@ class Switch(VizroBaseModel):
     _dynamic: bool = PrivateAttr(False)
     _in_container: bool = PrivateAttr(False)
 
+    _make_actions_chain = model_validator(mode="after")(make_actions_chain)
+
+    @property
+    def _action_triggers(self) -> dict[str, _IdProperty]:
```
maxschulz-COL, line 76: Is there e.g. no guard action for this model?

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -109,6 +102,11 @@ class Graph(VizroBaseModel):
     ]
 
     _validate_figure = field_validator("figure", mode="before")(validate_captured_callable)
+    _make_actions_chain = model_validator(mode="after")(make_actions_chain)
+
+    @property
```
maxschulz-COL, line 108: Again the question regarding guard, will only mention things once if I detect them, but all my comments of course apply to all models

vizro-core/src/vizro/models/_dashboard.py
maxschulz-COL: I like that 👍

vizro-core/src/vizro/models/types.py
```diff
@@ -693,9 +693,6 @@ class OptionsDictType(TypedDict):
 # JSONSchema should be skipped for private actions that are not part of the public API.
 # In addition, `_filter` doesn't have a well defined schema due the Callables,
 # so if we were to include it, the JSONSchema would need to be defined.
-# TODO: Note that atm ActionType violates our (and pydantic's) convention that the type of the model ensures
```
maxschulz-COL, line 696: ❤️

maxschulz-COL (commented):
Went through the code once to get a quick overall understanding. All in all this is very exciting and immediately a lot better to understand - well done. I will now spend some time building more complex examples and testing things that are a little off the beaten track.

maxschulz-COL:
An overall thing I observed, not sure if related to this PR, but it happens in `scratch_app`: debug mode hot reloading does not work for newly added Pages. Need to restart the server.

vizro-core/src/vizro/models/_components/ag_grid.py
antonymilne: Ah, interesting. Just to check I understand, the difference is:
```
# what we do now
_x = model/field_validator(...)(x)

# what we should do to work better with mkdocstrings
@model/field_validator(...)
def _x():
    x()
```

> In fact this looks to me like we may even want to define it on the BaseModel no?

It's only relevant for models that have the `actions` field defined, and there's a couple of special case exceptions at the moment, so I think probably best to leave it as manually copy and pasted for now unless you feel strongly about it.

vizro-core/src/vizro/models/_models_utils.py
antonymilne: The number of HTTP requests will be the same now as before. We have never grouped callbacks into a single HTTP request before, so before and after this change a chain of n callbacks does  n HTTP requests.

The big efficiency you're probably thinking of is that when we load the page, there is a single action/callback that executes (on_page_load). So we don't do e.g. a callback to apply filter 1 and then another callback to apply filter 2 - they're grouped together into a single on_page_load callback. This PR doesn't change anything about that either.

vizro-core/src/vizro/models/_components/button.py
antonymilne: Yes indeed - and this is actually how the (to be documented) action inputs and outputs also work. See https://github.com/mckinsey/vizro/pull/1153 and https://github.com/mckinsey/vizro/pull/1178. (FYI this system is what helped us finally fix the awful AgGrid/Table underlying input id thing.)

e.g. for a graph with id `graph_id`, you can use as output:
- `graph_id` - will lookup the `__default__` property, which is `figure`
- `graph_id.figure` -  the same
- `graph_id.header` - shortcut to the relevant div that contains the header
- `graph_id.arbitrary_dash_property` - will still work even if it's not defined in the `action_outputs` mapping

For trigger I'm imagining something similar, e.g. for graph you maybe have:
- `graph_id` - will lookup the `__default__` property, which is `clickData`
- `graph_id.click` - same but more user-friendly/less Dash-specific notation than `clickData`
- `graph_id.clickData` - same since reverts to underlying Dash property
- `graph_id.hover` - same but more user-friendyl/less Dash-specific notation than `hoverData`
- `graph_id.hoverData` - same since reverts to underlying Dash property

This way hopefully users can switch intuitively between doing e.g. a click interaction vs. a hover interaction for both Graph and AgGrid without needing to know that in one case the underlying property is `clickData` and the other is `selectedRows` or whatever. To a user they'd both be `graph/table_id.click`, or if `click` is the default then it's just `graph/table_id`.

I think it will be fine to maintain this ongoing because there aren't that many shortcut properties to expose. It's going to be a bit like an enum but also for "advanced" users it's not strictly limited since you can still specify an arbitrary Dash component id and property.

vizro-core/src/vizro/models/_components/form/checklist.py
antonymilne: To workaround the Dash bug/feature about `prevent_initial_call` it is absolutely crucial that this component is created at the same time as the checklist itself. So unfortunately it can't go anywhere else.

vizro-core/src/vizro/models/_components/form/switch.py
antonymilne: Yes, good spot! This is because `Switch` can't be dynamic and so there's no need for the guard component to be created. The guard is only relevant to prevent unwanted Dash callbacks in the case that a component is recreated on the page, which it can't be here.

vizro-core/src/vizro/models/_components/graph.py
antonymilne: This is correct but a very good question so let me add a comment to the code explaining why there's not one in a few places! The key difference is that after the page build has completed, `dcc.Graph` doesn't ever get recreated - we just update the `graph_id.figure`. This is unlike what happens with AgGrid/table, where the underlying `dag.AgGrid` object itself does get recreated.

The weird Dash bug/feature `prevent_initial_call` behaviour only occurs in the case that a component gets recreated, so the guard is only needed in those cases.

antonymilne:
> An overall thing I observed, not sure if related to this PR, but it happens in `scratch_app`: debug mode hot reloading does not work for newly added Pages. Need to restart the server.

Interesting! @petar-qb please could you check this when you do your testing? My guess is that this PR hasn't changed anything here - I don't see how it could have anyway 🤔

petar-qb:
> An overall thing I observed, not sure if related to this PR, but it happens in `scratch_app`: debug mode hot reloading does not work for newly added Pages. Need to restart the server.

Hey @maxschulz-COL 👋 Have you removed the `use_reloader=False` from the server run? Adding a new page is hot reloaded when I commented it out.

https://github.com/user-attachments/assets/e2c81c12-9965-4c97-8144-9c4ef9507b04

vizro-core/src/vizro/models/_components/ag_grid.py
maxschulz-COL: Yes indeed, that is the difference.

vizro-core/src/vizro/models/_models_utils.py
maxschulz-COL: Ah gotcha, yes that is what i was wondering

antonymilne:
@maxschulz-COL thank you for your very careful review! There were lots of interesting things in your app and I've made a note of them in relevant future tickets. You didn't find anything unexpected which is good news 👍 

A few comments and responses as FYI:
* for the pivot table (great example, we will have something like this in the docs when it's more natively built in) you labelled the underlying selectors of your filters so you could target them with `outputs=["filter_company_selector.value", "filter_product_selector.value"]`. Soon you shouldn't need to do that, you should just be able to use the underlying filter_id like `outputs="filter_company.value"` or maybe even just `outputs="filter_company"`.
* several of your examples with a store could be simplified with [`Patch`](https://dash.plotly.com/partial-properties). Basically instead of taking in the current store as `State`, modifying it and then outputting it, there's no need to take it in as `State`. You can instead just return `Patch` with the new entry and it will merge it into the current store
* in the todo app you have a repeating pattern that basically does `update_store` and then `update_ui_from_store`. This is very similar to something we're thinking about on the chatbot @lingyielia. I think we will eventually need some general functionality or at least advice for this but FYI there's another interesting pattern to consider here: write a Dash callback that does `update_ui_from_store` (can probably be clientside). Then instead of explicitly chaining `update_store` and `update_ui_from_store` into an actions chain, you can just do `update_store` by itself and leave it to Dash to implicitly trigger `update_ui_from_store`. This way the store acts as a sort of source of truth/local database, and there's a callback whose job it is to keep the UI in sync with it that you never need to call manually. There are various pros and cons to this compared to the explicit actions chain approach (all undocumented now I know...) In the chatbot case I think it won't work well due to streaming requirements but for something like the todo app it might work well.
* pattern-matching callbacks: realistically I think we are a long way off supporting these, although I wouldn't say we'll never do it. I've considered changing everything we do to pattern-matching a few times and am glad I didn't so far, but for these dynamics cases like the todo app it is definitely the only good way to do it.

maxschulz-COL:
> * for the pivot table (great example, we will have something like this in the docs when it's more natively built in) you labelled the underlying selectors of your filters so you could target them with `outputs=["filter_company_selector.value", "filter_product_selector.value"]`. Soon you shouldn't need to do that, you should just be able to use the underlying filter_id like `outputs="filter_company.value"` or maybe even just `outputs="filter_company"`.

Amazing, that will make for a good example :) Btw, this is the sort of thing where some for of data lineage (is that even the right term) would be amazing. What I mean is: we _know_ the the source DF is this, and that it was pivoted to be displayed here, thus double filters make sense on the original DF. I believe PBI has some of that functionality - not sure if it is realistic that we do

> * several of your examples with a store could be simplified with [`Patch`](https://dash.plotly.com/partial-properties). Basically instead of taking in the current store as `State`, modifying it and then outputting it, there's no need to take it in as `State`. You can instead just return `Patch` with the new entry and it will merge it into the current store

Ah great, yes that is good to know.

> * in the todo app you have a repeating pattern that basically does `update_store` and then `update_ui_from_store`. This is very similar to something we're thinking about on the chatbot @lingyielia. I think we will eventually need some general functionality or at least advice for this but FYI there's another interesting pattern to consider here: write a Dash callback that does `update_ui_from_store` (can probably be clientside). Then instead of explicitly chaining `update_store` and `update_ui_from_store` into an actions chain, you can just do `update_store` by itself and leave it to Dash to implicitly trigger `update_ui_from_store`. This way the store acts as a sort of source of truth/local database, and there's a callback whose job it is to keep the UI in sync with it that you never need to call manually. There are various pros and cons to this compared to the explicit actions chain approach (all undocumented now I know...) In the chatbot case I think it won't work well due to streaming requirements but for something like the todo app it might work well.

Also good to know. I sometimes also didn't look for optimization on purpose, because I just wanted to play with as many ways as possible to get something done, but that idea sounds like a good thing. Definitely keen for a "local DB"

> * pattern-matching callbacks: realistically I think we are a long way off supporting these, although I wouldn't say we'll never do it. I've considered changing everything we do to pattern-matching a few times and am glad I didn't so far, but for these dynamics cases like the todo app it is definitely the only good way to do it.

:( but understandable. Tbh I think this almost seems like an antipattern in general. I wonder if instead we would like to invest our time in making single isolated react component _very easily_ pluggable (much easier as currently in dash). Then people could (maybe with help of some MCP or what not) Vibe code such self container components without worry, and beause they are isolated, there is very little risk to put them in the app. Then people could interfere with these component via the usual action system. I think that is probably the key when have dynamic page content

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -63,7 +63,14 @@ class _BaseAction(VizroBaseModel):
 
     @property
     def _dash_components(self) -> list[Component]:
-        raise NotImplementedError
+        # Users cannot add Dash components using vm.Action.
```
```suggestion
"""Internal components needed to run the actions chain system:
    - {action.id}_finished for completion of an action callback to trigger the next action in the chain
    - {action.id}_guarded_trigger for the first action in a chain so that guard_action_chain callback
     prevent undesired triggering (workaround for Dash prevent_initial_call=True behavior)

In theory, subclasses can add additional components to the list, as done in export_data, but this should not be generally encouraged. In future it might not be possible.
"""
```

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -63,7 +63,14 @@ class _BaseAction(VizroBaseModel):
...
+        dash_components: list[Component] = []
+
+        if self._first_in_chain:
+            dash_components.append(dcc.Store(id=f"{self.id}_guarded_trigger"))
+        dash_components.append(dcc.Store(id=f"{self.id}_finished"))
```
```suggestion
        dash_components = [dcc.Store(id=f"{self.id}_finished")]
        if self._first_in_chain:
            dash_components.append(dcc.Store(id=f"{self.id}_guarded_trigger"))
```

vizro-core/src/vizro/models/_page.py
```diff
@@ -209,21 +214,15 @@ def build(self) -> _PageBuildType:
...
         # not accidentally trigger callbacks (workaround for Dash prevent_initial_call=True behavior).
         action_components = []
-
-        # TODO NOW: should this just go through this page's actions or across whole dashboard? Probably doesn't
-        #  matter much apart from if we want to do cross-page actions.
-        for action in cast(Iterable[_BaseAction], model_manager._get_models(_BaseAction)):
-            if action._first_in_chain:
-                action_components.append(dcc.Store(id=f"{action.id}_guarded_trigger"))
-            action_components.append(dcc.Store(id=f"{action.id}_finished"))
+        for action in cast(Iterable[_BaseAction], model_manager._get_models(_BaseAction, root_model=self)):
```
antonymilne: Maybe neater now the code is so much simpler? Or maybe harder to understand depending on your opinion. Up to you.

```
action_components = list(itertools.chain.from_iterable(action._dash_components for action in cast(Iterable[_BaseAction], model_manager._get_models(_BaseAction, root_model=self))))
```

vizro-core/src/vizro/models/_page.py
```diff
@@ -209,21 +214,15 @@ def build(self) -> _PageBuildType:
         components_container.id = "page-components"
 
         # Components that are required to make action chains function correctly:
-        #   - {action.id}_guarded_trigger for the first action in a chain so that guard_action_chain can prevent
-        #     undesired triggering (workaround for Dash prevent_initial_call=True behavior)
+        #   - {action.id}_guarded_trigger for the first action in a chain so that guard_action_chain callback
```
antonymilne: Let's delete lines 217-221. I've moved it to the `_dash_components` property.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -139,13 +139,24 @@ def __call__(self, target_to_data_frame: dict[ModelID, pd.DataFrame], current_va
...
-            self.selector = cast(CategoricalSelectorType, self.selector)
-            return self.selector(options=self._get_options(targeted_data, current_value))
-        elif isinstance(self.selector, SELECTORS["numerical"] + SELECTORS["temporal"]):
-            self.selector = cast(NumericalTemporalSelectorType, self.selector)
+        # Cast is justified as the selector is set in pre_build and is not None.
+        selector = cast(SelectorType, self.selector)
+
+        if isinstance(selector, SELECTORS["categorical"]):
+            selector = cast(CategoricalSelectorType, selector)
+            selector_build_obj = selector(options=self._get_options(targeted_data, current_value))
```
antonymilne: Maybe? I was a bit confused by the name because there was no `build` there.

```suggestion
            selector_call_obj = selector(options=self._get_options(targeted_data, current_value))
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -139,13 +139,24 @@ def __call__(self, target_to_data_frame: dict[ModelID, pd.DataFrame], current_va
...
+        elif isinstance(selector, SELECTORS["numerical"] + SELECTORS["temporal"]):
+            selector = cast(NumericalTemporalSelectorType, selector)
             _min, _max = self._get_min_max(targeted_data, current_value)
-            return self.selector(min=_min, max=_max)
+            selector_build_obj = selector(min=_min, max=_max)
+
+        if self._dynamic:
+            if f"{selector.id}_guard_actions_chain" in selector_build_obj:
```
antonymilne: Why would this happen? Haven't we removed it from all selectors?

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -139,13 +139,24 @@ def __call__(self, target_to_data_frame: dict[ModelID, pd.DataFrame], current_va
...
+                selector_build_obj[f"{selector.id}_guard_actions_chain"].data = True
+            else:
+                selector_build_obj.children.append(dcc.Store(id=f"{selector.id}_guard_actions_chain", data=True))
```
antonymilne: Does this rely on `selector_build_obj` being a `html.Div`?

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -139,13 +139,24 @@ def __call__(self, target_to_data_frame: dict[ModelID, pd.DataFrame], current_va
...
+        if isinstance(selector, SELECTORS["categorical"]):
+            selector = cast(CategoricalSelectorType, selector)
+            selector_build_obj = selector(options=self._get_options(targeted_data, current_value))
+        elif isinstance(selector, SELECTORS["numerical"] + SELECTORS["temporal"]):
+            selector = cast(NumericalTemporalSelectorType, selector)
             _min, _max = self._get_min_max(targeted_data, current_value)
-            return self.selector(min=_min, max=_max)
+            selector_build_obj = selector(min=_min, max=_max)
+
+        if self._dynamic:
```
antonymilne: A small comment here about why we're inserting `guard_actions_chain` would be good.

vizro-core/src/vizro/models/_controls/parameter.py
```diff
@@ -101,7 +101,17 @@ def pre_build(self):
 
     @_log_call
     def build(self):
-        return html.Div(id=self.id, children=self.selector.build())
+        selector_build_obj = html.Div(id=self.id, children=self.selector.build())
```
antonymilne: This is a bit confusing because in filter the `selector_build_obj` is the `self.selector.build()` itself rather than the parent `html.Div`.
```suggestion
        selector_build_obj = self.selector.build()
        
        if self.show_in_url:
            selector_build_obj.children.append(...) # just one child and now it matches filter
            
        return html.Div(id=self.id, children=selector_build_obj)
```

vizro-core/src/vizro/models/_components/graph.py
antonymilne: @petar-qb please could you add a comment somewhere suitable (maybe `vm.Graph`?) about why that model doesn't have the guard component while AgGrid and table do.

antonymilne (commented):
Partial review, will comment more in a second...

vizro-core/src/vizro/static/js/models/action.js
```diff
@@ -0,0 +1,52 @@
...
+function guard_action_chain(value, created, trigger_component_id) {
+  if (created === true) {
+    // Case 1: Guard component has data = true.
+    // This means that the trigger component was just created,
+    // so we must skip running the actions chain - it’s not a genuine trigger.
+    console.debug(
+      `Not running actions chain (guard is True) -> Trigger component: ${trigger_component_id}`,
+    );
+
+    // Set the guard component's data to false so that future triggers from this component are treated as genuine.
```
antonymilne, line 20: Maybe worth putting some of the old comments from the Python version here? Since potentially when Dash releases `Output(allow_optional=True)` we'd change this.
```
                 # We must use set_props here rather than using Output(component_guard_id, "data") because the
                    # component might not exist. This is allowed for a state with allow_optional=True but not for an
                    # Output.
                    set_props(component_guard_id, {"data": False})
                    # This must be dash.no_update rather than PreventUpdate or set_props won't work.
                    return dash.no_update
```

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -156,44 +155,22 @@ function sync_url_query_params_and_controls(...values_ids) {
...
-    const selectorValue = outputSelectorValues[i - 1];
-
+  // After this clientside callback, the "guard_action_chain" callback may run.
+  // If the selector value is updated based on the URL parameters,
+  // set its values and the selector’s guard flag to **true**.
+  // This ensures triggering the guard action chain callback
+  // and prevents unnecessary actions from being triggered by the value change.
+  for (let i = 0; i < selectorIds.length; i++) {
+    const selectorId = selectorIds[i];
+    const selectorValue = outputSelectorValues[i];
```
antonymilne: This is probably slightly more idiomatic javascript:

```suggestion
selectorIds.forEach((selectorId, i) => {
  const selectorValue = outputSelectorValues[i];
```

antonymilne (commented):
AMAZING WORK! 🎉 The resolutions of all the TODOs look great to me. Moving the actions components to `_dash_components` is much better - that was a really good idea 👍 And congratulations on fixing all the URL parameter stuff. Sorry I left you with such a big task there - I really thought it was just going to be a quick and easy thing 😅

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: Ah, I see! I thought the guard component created in `Filter.build` would still exist, but since the entire filter object gets recreated, we can just skip that deletion step 😄

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: It relies on `selector_build_obj` being any object that has the `children` property. It's clearly wrong so I did this -> https://github.com/mckinsey/vizro/pull/1301/commits/95ca08226ff1817c0e4539466669117f1a52a6a7

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -78,3 +79,55 @@ def warn_description_without_title(description, info: ValidationInfo):
...
+    """Creates actions chain from a list of actions.
+
+    Ideally this would have been implemented as an AfterValidator for the actions field, but we need access to
+    action_triggers class property, which needs the parent model instance. Hence, it must be done as a model validator.
```
antonymilne:
```suggestion
    action_triggers property, which needs the parent model instance. Hence, it must be done as a model validator.
```

vizro-core/tests/unit/vizro/models/_components/form/test_checklist.py
```diff
@@ -125,10 +125,10 @@ def test_create_checklist_invalid_value_format(self):
         with pytest.raises(ValidationError, match="Input should be a valid list"):
             Checklist(value="A", options=["A", "B", "C"])
 
-    def test_set_action_via_validator(self, identity_action_function):
-        checklist = Checklist(actions=[Action(function=identity_action_function())])
-        actions_chain = checklist.actions[0]
-        assert actions_chain.trigger.component_property == "value"
+    def test_checklist_trigger(self, identity_action_function):
+        checklist = Checklist(id="checklist-id", actions=[Action(function=identity_action_function())])
+        action = checklist.actions[0]
```
antonymilne: Very minor preference but easy to do a regex find and replace... Doesn't really matter in tests, but I usually prefer this in source code.

```suggestion
        [action] = checklist.actions
```

vizro-core/tests/unit/vizro/models/_components/test_ag_grid.py
```diff
@@ -152,13 +152,13 @@ def test_underlying_id_is_auto_generated(self, standard_ag_grid):
         ag_grid = vm.AgGrid(id="text_ag_grid", figure=standard_ag_grid)
         ag_grid.pre_build()
         # ag_grid() is the same as ag_grid.__call__()
-        assert ag_grid().id == "__input_text_ag_grid"
+        assert "__input_text_ag_grid" in ag_grid()
```
antonymilne: Possibly these tests should be `assert_component_equal` ones instead now? You could probably make this stricter and clearer without much effort just with `assert_component_equal(keys_to_strip=STRIP_ALL)`. I'm not too bothered though, up to you.

vizro-core/tests/unit/vizro/models/conftest.py
```diff
@@ -16,3 +20,24 @@ def _identity_action_function(arg=None):
 @pytest.fixture(params=[vm.Container, vm.Page])
 def model_with_layout(request):
     return request.param
+
+
+@pytest.fixture
+def mock_class_model_with_actions():
```
antonymilne: Do we need to do this with a `fixture`? And should we put it directly in `test_models_utils.py` instead? Is it used in any other tests outside that file?

vizro-core/tests/unit/vizro/actions/test_export_data.py
```diff
@@ -259,11 +281,17 @@ def test_export_data_outputs(self, config_for_testing_all_components_with_action
         ],
         indirect=True,
     )
-    def test_export_data_dash_components(
-        self, config_for_testing_all_components_with_actions, expected_dash_components
+    def test_export_data_first_in_chain_dash_components(
```
antonymilne, line 284: Why is this `first_in_chain`? Do we have another case for `second_in_chain`? If so then do we really want to test both of those here? It seems excessive to me and we don't do it for other actions.

vizro-core/tests/unit/vizro/models/_components/form/test_checklist.py
petar-qb: See it.. I changed it here. There are a few other places where we can do something similar.

1. action.py | 271
`callback_outputs = callback_outputs[0]` -> `[callback_outputs] = callback_outputs`
2. action.py | 348
`trigger_component_id = self._trigger.split(".")[0]` -> `[trigger_component_id, _] = self._trigger.split(".")`
3. _controls_utils.py | 15
`target_id = target.split(".")[0]` -> `[target_id, _] = target.split(".")`
4. _grid.py | 210
`self._component_grid_lines = _get_grid_lines(self.grid)[0]` -> `[self._component_grid_lines, *_] = _get_grid_lines(self.grid)`
5. nav_link.py | 51
`first_page_id = all_page_ids[0]` -> `[first_page_id, *_] = all_page_ids`

I haven't tested any of these new suggestions, but if you like some of them, let me know so I can add.

vizro-core/tests/unit/vizro/actions/test_export_data.py
petar-qb: Below this test, there's `test_export_data_not_first_in_chain_dash_components` test that tests `export_data._dash_components`. I made separate test for this here as `export_data` implements its own `_dash_components`.

vizro-core/schemas/0.1.44.json
petar-qb, line 1: Should I revert these changes?

vizro-core/changelog.d/20250806_115509_90609403+huong-li-nguyen_enable_single_action.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Enable passing a single action to the `actions` argument e.g. `actions=export_data()`. ([#1328](https://github.com/mckinsey/vizro/pull/1328))
```
petar-qb: Should we announce the single output for custom actions now, or wait and include it with other improvements once the actions docs are updated?

vizro-core/src/vizro/models/_components/form/_text_area.py
```diff
@@ -40,12 +38,11 @@ class TextArea(VizroBaseModel):
...
     ]
     placeholder: str = Field(default="", description="Default text to display in input field")
-    actions: list[ActionType] = []
 
-    # Reused validators
     # TODO: Before making public, consider how actions should be triggered and what the default property should be
     # See comment thread: https://github.com/mckinsey/vizro/pull/298#discussion_r1478137654
-    _set_actions = _action_validator_factory("value")
+    # This would mean creating _action_triggers and using make_actions_chain.
+    actions: ActionsType = []
```
petar-qb, line 45: Should we create the trigger now? 
Same for other similar components.

vizro-core/changelog.d/20250806_115509_90609403+huong-li-nguyen_enable_single_action.md
antonymilne: Actually I'm going to remove this entry from the changelog because it doesn't make sense until we've announced you can do `actions=export_data()` instead of `vm.Action(function=export_data())`.

The outputs one I'll leave though because it's such a minor change.

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -21,8 +22,7 @@
 )
 
 if TYPE_CHECKING:
-    from vizro.models import VizroBaseModel
-    from vizro.models.types import ActionType
+    pass
```
antonymilne: We should remove this.

vizro-core/src/vizro/models/_components/form/_text_area.py
antonymilne: Not yet - let's leave it until we finally make these public (hopefully not too long...)

vizro-core/tests/unit/vizro/models/test_types.py
```diff
@@ -388,3 +389,79 @@ def test_graph_template_crash(set_pio_default_template):
...
+        """Test that _coerce_to_list correctly handles various input types."""
+        result = _coerce_to_list(input_value)
+        assert result == expected_output
+
+
+class TestCoerceActionsAndOutputsType:
+    @pytest.mark.parametrize(
+        "actions_input",
+        [
+            Action(function=export_data()),  # Single action
```
antonymilne: Split out legacy tests

vizro-core/changelog.d/20250806_115509_90609403+huong-li-nguyen_enable_single_action.md
antonymilne: Actually I just changed this example to use `vm.Action` and now I think it's ok.

vizro-core/tests/unit/vizro/models/_components/form/test_checklist.py
antonymilne: I made these changes to the places where we don't need to do `*_` 🙂  There were a few left in tests too.

PR #1317, opened 2025-07-31
[Feat] Rework of Model Manager Part 1

maxschulz-COL:
## Description
Still in draft, more to follow

## Notice


vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -34,38 +34,51 @@ class DuplicateIDError(ValueError):
...
+        # AM rough notes: need to handle CapturedCallable in future too? Which doesn't have .id
+        # AM rough notes: forward_attrs=True looks nice for simplifying syntax.
+        self.__dashboard_tree = TypedTree(calc_data_id=lambda tree, data: data.id)
         self._frozen_state = False
 
+    # AM rough notes: probably better to just instantiate model manager in Vizro and pass dashboard in init rather
+    # than doing it as separate method here.
+    # Ideal next checkpoint future state: work out whether to put this __init__.
+    def _set_dashboard(self, dashboard):
+        self.__dashboard_tree.clear()
```
antonymilne: What's this line for?

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -34,38 +34,51 @@ class DuplicateIDError(ValueError):
 
 class ModelManager:
     def __init__(self):
-        self.__models: dict[ModelID, VizroBaseModel] = {}
+        # AM rough notes: need to handle CapturedCallable in future too? Which doesn't have .id
+        # AM rough notes: forward_attrs=True looks nice for simplifying syntax.
+        self.__dashboard_tree = TypedTree(calc_data_id=lambda tree, data: data.id)
```
antonymilne:
```suggestion
        # AM rough notes: importantly the nodes are just pointers to real objects so not duplicated in memory.
        # assert d.first_child().data.pages[0] is d["Test page"].data
        self.__dashboard_tree = TypedTree(calc_data_id=lambda tree, data: data.id)
```

antonymilne: Just putting in some rough notes that we can tidy into proper comments eventually.

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -34,38 +34,51 @@ class DuplicateIDError(ValueError):
...
+    # @_state_modifier
+    # def __delitem__(self, model_id: ModelID):
+    #     # Only required to handle legacy actions and could be removed when those are no longer needed.
+    #     del self.__models[model_id]
+
+    # Ideal next checkpoint future state: this still exists but uses self.__dashboard_tree. DONE
     def __getitem__(self, model_id: ModelID) -> VizroBaseModel:
         # Do we need to return deepcopy(self.__models[model_id]) to avoid adjusting element by accident?
-        return self.__models[model_id]
+        return self.__dashboard_tree.find(data_id=model_id).data
```
antonymilne: Note to myself from before - check this:
```
# AM rough notes: smart lookup in nutree allows lookup by id and model itself and probably good to use here.
# But there's unexpected behaviour/maybe a bug in nutree where you set Tree(calc_data_id) upfront.
# Probably easy to fix.
# Check performance of different look ups and whether to use find or find_first.
```

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -34,38 +34,51 @@ class DuplicateIDError(ValueError):
...
     def __iter__(self) -> Generator[ModelID, None, None]:
         """Iterates through all models.
 
         Note this yields model IDs rather key/value pairs to match the interface for a dictionary.
         """
         # TODO: should this yield models rather than model IDs? Should model_manager be more like set with a special
         #  lookup by model ID or more like dictionary?
-        yield from self.__models
+        # yield from self.__models
+        yield from self.__dashboard_tree.iterator(method=IterMethod.PRE_ORDER)
```
antonymilne: Slightly more Pythonic.

```suggestion
        yield from iter(self.__dashboard_tree)
```

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -80,39 +93,69 @@ def _get_models(
...
-            for child in model.values():
-                yield from self.__get_model_children(child)
+            for child_model in model.values():
+                self.__populate_tree(child_model, parent, field_name)
         elif isinstance(model, Collection) and not isinstance(model, str):
-            for child in model:
-                yield from self.__get_model_children(child)
+            for child_model in model:
+                self.__populate_tree(child_model, parent, field_name)
+        # TODO[MS]: does there not need to be a case for not being a model? I guess that should
```
antonymilne: This case would just be the same as `else: pass` so we just ignore it I think?

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -80,39 +93,69 @@ def _get_models(
...
+                yield model
 
-    def __get_model_children(self, model: Model) -> Generator[Model, None, None]:
+    def __populate_tree(
+        self, model: Union[Model, Mapping[str, Model], Collection[Model]], parent=None, field_name=None
+    ):
         """Iterates through children of `model` with depth-first pre-order traversal."""
+        # AM rough notes: this is basically copied and pasted from __get_model_children and then modified to populate
+        # dashboard tree instead of yielding.
+        # MS: This misses items created in pre-build (like selectors) and "behind the scene" items like ActionsChain
```
antonymilne: These are now added by `_from_dict_in_build`, right? Is there anything else still missing from `__dashboard_tree` after `pre_build` has completed?

Would be worth checking we haven't missed anything here by trying it on some complicated example dashboards.

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -80,39 +93,69 @@ def _get_models(
...
+
+        while current_node.parent:
+            current_node = current_node.parent
+            if isinstance(current_node.data, Page):
+                return current_node.data
+
+        raise ValueError(f"Model with id='{model.id}' is not contained within any Page")
 
     @staticmethod
     def _generate_id() -> ModelID:
```
antonymilne, line 151: Rough notes from before for me to refer back to:

```
Keep get_models etc. centralised into MM for now but then maybe move to Dashboard model later.

Might be useful to have property in every model to enable you to get to its Node in the tree or just a reference to the whole dashboard tree? Basically this inserts model manager into every model so no need for global. 

---

Rough thoughts on how we might handle _is_container property. Could keep as property and/or:
key question is given model (or model.id), how to find corresponding node in the tree?

Dropdown()._nutree_node.up(2)
model_manager[dropdown_id].up(2)

Storing just parent as property in all model's seems like a worse version of storing the corresponding node:
dropdown._nutree_node
dropdown.)nutree_tree[dropdown.id]
```

vizro-core/src/vizro/_vizro.py
```diff
@@ -165,22 +172,28 @@ def _pre_build():
...
-            # defined in the filter's pre_build method. Also, the calculation of the data_frame Parameter targets
-            # depends on the filter targets, so they should be pre-built after the filters as well.
-            # It's also essential for filters to be pre-built before Container.pre_build runs or otherwise
-            # control.selector won't be set.
-            filter.pre_build()
-        for model_id in set(model_manager):
-            model = model_manager[model_id]
-            if hasattr(model, "pre_build") and not isinstance(model, Filter):
-                model.pre_build()
+        # MS: I believe that this currently holds only for things that are created below in the tree.
```
antonymilne: Hmm, it would be good to check this with some basic toy example (just nutree outside Vizro). What happens if you add a node to the tree during iteration - does the new node get visited or not? Does it depend on the traversal order?

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -80,39 +93,69 @@ def _get_models(
...
+            raise ValueError(f"Model with id='{model.id}' not found in dashboard tree")
+
+        while current_node.parent:
+            current_node = current_node.parent
+            if isinstance(current_node.data, Page):
+                return current_node.data
+
+        raise ValueError(f"Model with id='{model.id}' is not contained within any Page")
 
     @staticmethod
```
antonymilne, line 150: Let's move this directly into the `VizroBaseModel` validator unless there's a good reason not to do so?

vizro-core/src/vizro/models/_base.py
```diff
@@ -223,9 +224,15 @@ class VizroBaseModel(BaseModel):
         ),
     ]
 
-    @_log_call
-    def model_post_init(self, context: Any) -> None:
-        model_manager[self.id] = self
+    # TODO[MS]: Make more official access to the tree
+    @classmethod
+    def _from_dict_in_build(cls, parent_id: str, field_name: str, data: dict[str, Any]) -> Self:
```
antonymilne, line 349: I like the idea of something like this to add something to the dashboard tree on the fly but not sure about including the `pre_build` here rather than some alternative like:
* calling manually
* doing another iteration through the dashboard tree (probably not a good idea though)

Two more rough ideas:
* if we put the tree in all models this might become simpler
* can we do it automatically through a pydantic validator when the field is set?

antonymilne (commented):
Still looking through it and thinking about everything, but my initial quick impression is that it seems good 👍

vizro-core/src/vizro/managers/_model_manager.py
maxschulz-COL: This is a remnant of a temporary state from when I was re-building several times. It is however interesting to think about when and if we should clear the tree. As mentioned previously, currently one cannot build the tree twice one after the other - but it may be interesting to at least partially rebuild things.

vizro-core/src/vizro/managers/_model_manager.py
maxschulz-COL: I think so far it works very well, and tbh I caught most hidden things by trying out the `dev` example. Will try the others too now.

maxschulz-COL: Visual vocabulary works great too - let's see the integration tests

vizro-core/src/vizro/_vizro.py
maxschulz-COL: Made a little example, so to answer your questions:
- yes it visits the node if it is due to visit that node in principle still, it does not if not
- yes it does depend on traversal order

```python
import uuid

from nutree import IterMethod, Tree, TypedTree


class Department:
    def __init__(self, name: str):
        self.guid = uuid.uuid4()
        self.name = name

    def __str__(self):
        return f"Department<{self.name}>"


class Person:
    def __init__(self, name: str, age: int):
        self.guid = uuid.uuid4()
        self.name = name
        self.age = age

    def __str__(self):
        return f"Person<{self.name} ({self.age})>"


development_dep = Department("Development")
test__dep = Department("Test")
marketing_dep = Department("Marketing")

alice = Person("Alice", 25)
bob = Person("Bob", 35)
claire = Person("Claire", 45)
dave = Person("Dave", 55)

tree = Tree("Organization", calc_data_id=lambda tree, data: str(data.name))

dev_node = tree.add(development_dep)
test_node = dev_node.add(test__dep)
mkt_node = tree.add(marketing_dep)

tree.add(alice)
dev_node.add(bob)
test_node.add(claire)
mkt_node.add(dave)


print("==========Starting tree=================")
visited = False
tree.print(repr="{node.data}")
print("--------------------------------")
for node in tree.iterator(method=IterMethod.LEVEL_ORDER):
    # print(node.data.name)
    if node.data.name == "Bob":
        tree.find(data_id="Alice").add_child(Person("Alice 2", 25))
        # tree.print(repr="{node.data}")
    if node.data.name == "Alice 2":
        visited = True
if visited:
    print("We visited the created node")
else:
    print("We did not visit the created node")
print("==========Ending tree=================")
tree.print(repr="{node.data}")
```

So in the above:
- pre and post order: it visits the created node
- level order: it does not

### Conclusion
To me this is intuitive to understand, which is good. The question is whether we are happy with the fact that this would make it possible to create a double `pre-build` (combined with the alternative init) when adding a model upwards or sideways. This is so far never the case, and in my eyes probably shouldnt. Wdyt?

vizro-core/src/vizro/managers/_model_manager.py
maxschulz-COL: You mean the ID generation?

vizro-core/src/vizro/models/_base.py
maxschulz-COL: Hmm interesting. The are my thoughts on the options:

1. Calling manually --> this seems somewhat prone to error. You gain potentially some flexibility in when this is called, but it can also easily be forgotten , and then we have lots of individual pre-builds scattered again across the codebase. So far we really only have one really
2. Doing another iteration --> i thought about that a lot, and it seems sort of viable, but brings its own set of problems: 
  - If we do a blind second iteration, this does lots of duplicate pre-builds, which I am not too keen about.
  - We could do a tree diff, and just run over the newly created models, which is probably the most acceptable method?

Pydantic validator: I don't think that's possible, because it defeats the purpose of pre-build. Pre-build is the point in time when the dashboard structure is finalized, so it cannot and should not be a field validator. That being said, all things that can be validators should be!!

As previously mentioned, I am quite keen to reduce the pre-build burden and shift at least some things to a model validator in the dashboard model, but will not be able to do that with all pre-build functionality. It will actually sharpen what a pre-build is even more, so that's quite nice!

vizro-core/src/vizro/managers/_model_manager.py
maxschulz-COL: @antonymilne What do you mean by this? Also according to their docs `find` is and alias for `find_first`

vizro-core/examples/dev/app.py
maxschulz-COL, line 1: clean before merge

vizro-core/src/vizro/managers/_model_manager.py
antonymilne: Yes exactly.

vizro-core/src/vizro/_vizro.py
antonymilne: Very clear example, thank you! I played around with it myself and agree it makes sense which is good.

Just to make sure we're on the same page: currently all cases where model X `pre_build` creates a new model Y, we know that Y would be a child of X? In which case your example code looks like this:
```
    if node.data.name == "Alice":
        node.add_child(Person("Alice 2", 25))
```

and the results are that post-order traversal (goes up the tree) doesn't visit the new node while pre-order traversal (goes down the tree) does. Agreed?

antonymilne: @maxschulz-COL said he agrees ⭐

vizro-core/src/vizro/managers/_model_manager.py
antonymilne: This seems like a good idea I think and with the new `model_validate` approach we're nearly there!

Let's say that instead of every model having `self._tree`, they have `self._model_manager` (or just `self._models` maybe) which itself contains the `tree`. This is the _same tree object shared across every model_. But the `model_manager` instance itself is probably different because it should probably contain a reference to the current model in which it's defined - otherwise you end up doing `self._model_manager.method(self)` all the time. Or maybe we don't even need to store the `tree` at all because you can just lookup using `node.tree`. Probably `model_manager` shouldn't expose `tree` publicly but leave it as an implementation detail. Either way, the important thing is that it's the same tree shared everywhere during and after `model_validate`.

Then whenever we currently access `model_manager` as a global from inside a model instance we can instead do `self._model_manager`. Remains to be seen what to do with the access from outside a model instance, but maybe there aren't actually any important ones?

If you make a new model during runtime it's outside the `build_tree` context so doesn't get added to the tree or modify the model manager - this is good. You can still set the model_manager in the new model instance if you need it with `new_model._model_manager = sef._model_manager` from the caller.

maxschulz-COL:
Superseeded by https://github.com/mckinsey/vizro/pull/1367/files

vizro-core/src/vizro/models/_base.py
```diff
@@ -233,9 +224,135 @@ class VizroBaseModel(BaseModel):
         ),
     ]
 
-    @_log_call
-    def model_post_init(self, context: Any) -> None:
-        model_manager[self.id] = self
+    # def __init_subclass__(cls, **kwargs):
+    #     super().__init_subclass__(**kwargs)
+    #     # print("==== INIT SUBCLASS ====")
+    #     field_names = [f for f in cls.__annotations__.keys() if (not f.startswith("_") and f != "type")]
```
maxschulz-COL, line 230: This is important to not forget

PR #1325, opened 2025-08-05
[Draft] [Docs] Update docs front page and nav to simplify it

stichbury:
## Description
Closes https://github.com/McK-Internal/vizro-internal/issues/2017

Decluttering to improve nav and front page 

## Notice

### DIFF
--- a/vizro-core/docs/index.md
+++ b/vizro-core/docs/index.md
@@ -8,39 +8,39 @@ Vizro is an open-source Python-based toolkit. You can use it to build data visua
 
     ---
 
-    [:octicons-arrow-right-24: Quickstart tutorial](pages/tutorials/first-dashboard.md)
+    [Quickstart tutorial](pages/tutorials/first-dashboard.md)
 
-    [:octicons-arrow-right-24: Explore Vizro tutorial](pages/tutorials/explore-components.md)
+    [Explore Vizro tutorial](pages/tutorials/explore-components.md)
 
-    [:octicons-arrow-right-24: FAQs](pages/explanation/faq.md)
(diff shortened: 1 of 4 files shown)


stichbury:
> This looks good to me 🚀 – just a couple of small comments/questions:

Thank you for the feedback!

> * We're calling it the `Quickstart tutorial`, but in the left-hand navigation it's listed as _"A first dashboard"_. Should we rename that to _"Quickstart"_ for consistency? That would match how we've named _"Explore Vizro tutorial"_.

I agree. I've made the change to "Quickstart" on the menu, the page and the filename.

> * On the homepage, clicking _"Fundamentals"_ currently redirects to _Dashboard_ instead of the first page within the _Fundamentals_ section. Was that intentional? Up until now, we've always linked to the first page of a section, so just wanted to double-check 🙂

The first page of that section is the install guide, which we link to as part of setup. It didn't make sense to me to link to that when the "Fundamentals" is generally more about using Vizro's features. It's a deliberate choice.

> * Not sure how I feel about the collapsible accordions in the navigation 🤔 They make the "Overview" pages feel a bit redundant, since those are also meant to serve as landing pages to access the subpages. What do you think? Is it fine as is, or does it feel like too much clicking around?

I know, I'm really up in the air about it too. I am not sure if extra clicking is that accessible, but I do like the tidiness of the shorter ToC.

When I discussed with @antonymilne we wondered about splitting the "How to guides" into 2 separate sections in the menu and thus reducing the ToC for each. But I think we agreed to make changes after the actions docs are done. 

Let's see what other reviewers think of this approach. I'll preview it in standup so we can discuss.

PR #1328, opened 2025-08-06
[Tidy] Enable users to pass a single action to the `actions` argument

huong-li-nguyen:
Closes https://github.com/McK-Internal/vizro-internal/issues/2037

## Description

Enable users to pass a single action directly to the actions argument instead of requiring a list:

- Added `_coerce_actions_type()` function that converts single actions to lists
- Updated `actions` field across all components with new type `ActionsType`
- Updated documentation and component docstrings
- Added unit tests for `_coerce_actions_type`

**Usage**

```
# New: Single action syntax
vm.Button(actions=export_data())

# Existing: List syntax (still supported)  
vm.Button(actions=[export_data()])
```

**Approach**

> Two possible ways to implement. It remains to be seen which was is best. Potentially @maxschulz-COL has an opinion here.
> 
> change actions: list[ActionsType] to actions: Union[list[ActionsType], ActionType]
> keep actions: list[ActionsType] and convert action into [action] in a validator (BeforeValidator I guess)

I went with the `BeforeValidator` instead of a `Union` type after testing both approaches:

✅ It follows existing patterns in the codebase like `coerce_pages_type` and `coerce_str_to_tooltip`, so it's more in line with how we already handle this kind of input.

✅ It keeps things simpler overall – while the `Union` type approach makes the schema clearer, it broke a few other validators during runtime (like `make_actions_chain`) that expect `actions` to always be a list. Supporting both types would've meant updating multiple places to handle both cases, which felt harder to maintain long-term.

But let me know if you disagree and prefer the Union type approach.

## Notice

### DIFF
--- a/vizro-core/src/vizro/models/types.py
+++ b/vizro-core/src/vizro/models/types.py
@@ -86,11 +86,18 @@ def _clean_module_string(module_string: str) -> str:
     return ""
 
 
-def _coerce_actions_type(actions: Any) -> Any:
-    """Converts a single action into a list of actions."""
-    if isinstance(actions, list):
-        return actions
-    return [actions]
+def _coerce_to_list(value: Any) -> Any:
+    """Converts a single item into a list unless it's already a list or dict (relevant for action outputs).
+
+    Args:
+        value: The value to potentially coerce to a list
+
+    Returns:
+        The original value if it's a list or dict, otherwise a list containing the value
+    """
+    if isinstance(value, (list, dict)):
+        return value
+    return [value]
 
 
 # Used to describe _DashboardReadyFigure, so we can keep CapturedCallable generic rather than referring to
@@ -717,10 +724,17 @@ ActionType = Annotated[
 
 # TODO: ideally actions would have json_schema_input_type=Union[list[ActionType], ActionType] attached to
 # the BeforeValidator, but this requires pydantic >= 2.9.
-ActionsType = Annotated[list[ActionType], BeforeValidator(_coerce_actions_type), Field(default=[])]
+ActionsType = Annotated[list[ActionType], BeforeValidator(_coerce_to_list), Field(default=[])]
 """List of actions that can be triggered by a component. Accepts either a single
 [`ActionType`][vizro.models.types.ActionType] or a list of [`ActionType`][vizro.models.types.ActionType].
 Defaults to `[]`."""
+
+# TODO: ideally outputs would have json_schema_input_type=Union[list[str], dict[str, str], str] attached to
+# the BeforeValidator, but this requires pydantic >= 2.9.
+OutputsType = Annotated[Union[list[str], dict[str, str]], BeforeValidator(_coerce_to_list), Field(default=[])]
+"""List or dictionary of outputs modified by the action function. Accepts either a single string,
+a list of strings, or a dictionary mapping strings to strings. Each output can be specified as
+`<model_id>` or `<model_id>.<argument_name>` or `<component_id>.<property>`. Defaults to `[]`."""
 
 # Extra type groups used for mypy casting
 FigureWithFilterInteractionType = Union["Graph", "Table", "AgGrid"]
--- a/vizro-core/changelog.d/20250811_102650_90609403+huong-li-nguyen_enable_single_actions_output.md
+++ b/vizro-core/changelog.d/20250811_102650_90609403+huong-li-nguyen_enable_single_actions_output.md
@@ -1,0 +1,45 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Highlights ✨
+
+- A bullet item for the Highlights ✨ category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+### Added
+
+- Enable single string outputs in `Action` model. You can now specify `outputs="component_id.property"` instead of `outputs=["component_id.property"]` for a single output. ([#1334](https://github.com/mckinsey/vizro/pull/1334))
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Security
+
+- A bullet item for the Security category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
--- a/vizro-core/examples/scratch_dev/app.py
+++ b/vizro-core/examples/scratch_dev/app.py
@@ -422,7 +422,7 @@ page_14 = vm.Page(
             id="page_14_button",
             actions=[
                 vm.Action(
-                    function=capture("action")(lambda x: x)("page_14_button.n_clicks"), outputs=["page_14_button.text"]
+                    function=capture("action")(lambda x: x)("page_14_button.n_clicks"), outputs="page_14_button.text"
                 )
             ],
         ),
@@ -442,7 +442,7 @@ page_15 = vm.Page(
             actions=[
                 vm.Action(
                     function=capture("action")(lambda x: radio_items_options[int(x) % 3])("page_15_button.n_clicks"),
-                    outputs=["page_15_checklist.value"],
+                    outputs="page_15_checklist.value",
                 )
             ],
         ),
@@ -452,7 +452,7 @@ page_15 = vm.Page(
             value=radio_items_options[0],
             actions=[
                 vm.Action(
-                    function=capture("action")(lambda x: x)("page_15_checklist.value"), outputs=["page_15_card.text"]
+                    function=capture("action")(lambda x: x)("page_15_checklist.value"), outputs="page_15_card.text"
                 )
             ],
         ),
@@ -487,22 +487,22 @@ page_16 = vm.Page(
                 vm.Button(
                     id="page_16_button_download",
                     text="Export data!",
-                    actions=[vm.Action(function=my_custom_export(), outputs=["vizro_download.data"])],
+                    actions=vm.Action(function=my_custom_export(), outputs="vizro_download.data"),
                 ),
                 vm.Button(
                     id="copy_page_16_button_download",
                     text="Copy Export data!",
-                    actions=[vm.Action(function=my_custom_export(), outputs=["vizro_download.data"])],
+                    actions=vm.Action(function=my_custom_export(), outputs="vizro_download.data"),
                 ),
                 vm.Button(
                     id="page_16_button_location",
                     text="Go to page 17!",
-                    actions=[vm.Action(function=my_custom_location(), outputs=["vizro_url.href"])],
+                    actions=vm.Action(function=my_custom_location(), outputs="vizro_url.href"),
                 ),
                 vm.Button(
                     id="copy_page_16_button_location",
                     text="Copy Go to page 17!",
-                    actions=[vm.Action(function=my_custom_location(), outputs=["vizro_url.href"])],
+                    actions=vm.Action(function=my_custom_location(), outputs="vizro_url.href"),
                 ),
             ],
         ),
@@ -518,7 +518,7 @@ page_16 = vm.Page(
     prevent_initial_call=True,
 )
 def export_callback(_):
-    return dcc.send_data_frame(iris.to_csv, "mydf.csv")
+    return dcc.send_data_frame(df_gapminder.to_csv, "mydf.csv")
 
 
 @callback(
@@ -527,7 +527,7 @@ def export_callback(_):
     prevent_initial_call=True,
 )
 def export_callback(_):
-    return dcc.send_data_frame(iris.to_csv, "mydf.csv")
+    return dcc.send_data_frame(df_gapminder.to_csv, "mydf.csv")
 
 
 # --- Location ---
--- a/vizro-core/schemas/0.1.45.dev0.json
+++ b/vizro-core/schemas/0.1.45.dev0.json
@@ -34,7 +34,7 @@
     },
     "Action": {
       "additionalProperties": false,
-      "description": "Action to be inserted into `actions` of relevant component.\n\nArgs:\n    function (CapturedCallable): Action function.\n    inputs (list[str]): List of inputs provided to the action function. Each input can be specified as `<model_id>`\n        or `<model_id>.<argument_name>` or `<component_id>.<property>`. Defaults to `[]`.\n    outputs (Union[list[str], dict[str, str]]): List or dictionary of outputs modified by the action function. Each\n        output can be specified as `<model_id>` or `<model_id>.<argument_name>` or `<component_id>.<property>`.\n        Defaults to `[]`.",
+      "description": "Action to be inserted into `actions` of relevant component.\n\nArgs:\n    function (CapturedCallable): Action function.\n    inputs (list[str]): List of inputs provided to the action function. Each input can be specified as `<model_id>`\n        or `<model_id>.<argument_name>` or `<component_id>.<property>`. Defaults to `[]`.\n    outputs (OutputsType): See [`OutputsType`][vizro.models.types.OutputsType].",
       "properties": {
         "id": {
           "default": "",
@@ -58,7 +58,6 @@
             }
           ],
           "default": [],
-          "description": "List or dictionary of outputs modified by the action function. Each output can be specified as\n            `<model_id>` or `<model_id>.<argument_name>` or `<component_id>.<property>`. Defaults to `[]`.",
           "title": "Outputs"
         },
         "type": {
--- a/vizro-core/src/vizro/actions/_abstract_action.py
+++ b/vizro-core/src/vizro/actions/_abstract_action.py
@@ -4,11 +4,10 @@ import abc
 import inspect
 from typing import Union
 
-from dash import Output
 from pydantic import TypeAdapter
 
 from vizro.models._action._action import _BaseAction
-from vizro.models.types import _IdOrIdProperty
+from vizro.models.types import OutputsType, _IdOrIdProperty
 
 
 # TODO-AV2 D 5: make public.
@@ -53,7 +52,7 @@ class _AbstractAction(_BaseAction, abc.ABC):
 
     @property
     @abc.abstractmethod
-    def outputs(self) -> Union[list[_IdOrIdProperty], dict[str, _IdOrIdProperty]]:  # type: ignore[override]
+    def outputs(self) -> Union[_IdOrIdProperty, list[_IdOrIdProperty], dict[str, _IdOrIdProperty]]:  # type: ignore[override]
         """Must be defined by concrete action, even if there's no output.
 
         This should return a dictionary of the form `{"key": "dropdown.value"}`, where the key corresponds to the key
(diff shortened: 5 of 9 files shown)


vizro-core/docs/pages/user-guides/custom-components.md
```diff
@@ -387,11 +387,11 @@ Add the custom action `open_offcanvas` as a `function` argument inside the [`Act
 
 As mentioned above, custom components can trigger actions. To enable the custom component to trigger the action, add the `actions` field and specify which property triggers the actions:
 
-1. **Add the `actions` argument to your custom component**. The type of the `actions` argument is `list[ActionType]`.
```
huong-li-nguyen, line 390: Only updated the docs here @antonymilne , because I wasn't sure how it fits into your overall rework of the docs

antonymilne: Cool, thanks! I'll do the other updates in my PR.

vizro-ai/examples/dashboard_ui/components.py
```diff
@@ -6,9 +6,9 @@
 import dash_bootstrap_components as dbc
```
antonymilne: This dashboard has pinned `vizro==0.1.28` so let's not change it.

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -131,3 +132,10 @@ def make_actions_chain(self):
     # until the pydantic bug is fixed. See https://github.com/pydantic/pydantic/issues/6597.
     self.__dict__["actions"] = converted_actions
     return self
+
+
+def coerce_actions_type(actions: Union[list[ActionType], ActionType]) -> list[ActionType]:
+    """Converts a single action into a list of actions."""
+    if isinstance(actions, list):
```
antonymilne: Very minor change...

```suggestion
    if isinstance(actions, Iterable):  # import from collections.abc
```

vizro-core/tests/unit/vizro/models/test_models_utils.py
```diff
@@ -53,3 +54,19 @@ def test_warns_if_description_and_no_title(self):
         info = MockValidationInfo(data={"title": ""})
         with pytest.warns(UserWarning, match="description.*title.*missing or empty"):
             warn_description_without_title("description", info)
+
+    def test_coerce_actions_type(self):
+        """Test that coerce_actions_type converts single actions to lists and leaves lists unchanged."""
+        # Single action -> list
+        single_action = vm.Action(function=export_data())
+        result = coerce_actions_type(single_action)
+        assert isinstance(result, list) and len(result) == 1 and result[0] == single_action
```
antonymilne:
```suggestion
        assert result == [single_action]
```

Does exactly the same 😀

vizro-core/tests/unit/vizro/models/test_models_utils.py
```diff
@@ -53,3 +54,19 @@ def test_warns_if_description_and_no_title(self):
...
+
+        # List of actions -> unchanged
+        action_list = [vm.Action(function=export_data())]
+        result = coerce_actions_type(action_list)
+        assert result == action_list
+
+        # Integration: single action works with components
+        button = vm.Button(actions=export_data())
+        assert isinstance(button.actions, list) and len(button.actions) == 1
```
antonymilne:
```suggestion
        assert button.actions == ...
```

vizro-core/tests/unit/vizro/models/test_models_utils.py
```diff
@@ -53,3 +54,19 @@ def test_warns_if_description_and_no_title(self):
         info = MockValidationInfo(data={"title": ""})
         with pytest.warns(UserWarning, match="description.*title.*missing or empty"):
             warn_description_without_title("description", info)
+
+    def test_coerce_actions_type(self):
```
antonymilne: Minor preference to split this into two tests:
* one to test the `assert result == [...]` - can be parametrized over two different values of  inputs but not need to parameterise expected output since it's the same for both
* one to test the integration with `vm.Button`

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -131,3 +132,10 @@ def make_actions_chain(self):
     # until the pydantic bug is fixed. See https://github.com/pydantic/pydantic/issues/6597.
     self.__dict__["actions"] = converted_actions
     return self
+
+
+def coerce_actions_type(actions: Union[list[ActionType], ActionType]) -> list[ActionType]:
```
antonymilne: When it comes to completing https://github.com/McK-Internal/vizro-internal/issues/2037 you _might_ (not sure if it will work since that case is a bit more complicated) be able to reuse this validator in which case it would become more than just `coerce_actions_type` but some other `coerce_to_list`. Definitely leave it like you have it for  now though

antonymilne (commented):
Nice! When I discussed this briefly with @maxschulz-COL earlier this week we had a vague preference for the other approach, but reading your justification here for why you chose this approach it completely makes sense. I forgot that dealing with `Union` types is actually a complete pain because it means all code that subsequently relies on that field has to deal with multiple types so that's a big advantage for this approach.

@maxschulz-COL - is this approach ok do you think?

Two more comments for both you and @maxschulz-COL:
1. if we go for the non-`Union` approach then we should add that bit of code that informs that both list and singleton are valid inputs in the JSON schema. Like we already have this:
   ```
    ```
2. something I've considered doing before but never actually done, and maybe it would make sense here. What would happen if we attached the validator to a type in `types.py` so we could reuse it everywhere without copying and pasting? Does it work with our API docs? Like this:
   ```
    # define once
    ActionsType = Annotated[list[ActionType], BeforeValidator(coerce_actions_type), Field(default=[])]

    # use in models like this:
    actions: ActionsType
    ````

vizro-ai/examples/dashboard_ui/components.py
huong-li-nguyen: Do we actually plan to update the Vizro version here at some point? @lingyielia
There have been so many changes already, and it's hard to keep track of what needs to be taken over at this point.

If we do plan to upgrade at some point, let's start documenting which changes still need to be carried over.
If we don't plan to upgrade this version at all, then there's no need to worry about it.

Just a few updates I know we haven't applied here yet:

* A bunch of CSS changes (e.g. renamed/removed `selected` class) - I updated them in all other demos but not this one, but let @lingyielia know
* Replacing `CodeClipboard` with the `dmc` version
* Reviewing the custom components - we should be able to remove or simplify some now (e.g. `Switch`, `Flex`, etc. are part of Vizro now)
* Typing updates from this PR
* Probably a few more things I'm forgetting 😅

antonymilne: Personally I would leave this alone or at least rate it as a very low priority to update.

vizro-core/src/vizro/models/_models_utils.py
huong-li-nguyen: mypy doesn't like that unless we also change the type annotations or cast the type 😄

Do you prefer below over the current version?

```
def coerce_actions_type(actions: Union[Iterable[ActionType], ActionType]) -> Iterable[ActionType]:
    """Converts a single action into a list of actions."""
    if isinstance(actions, Iterable):
        return actions
    return [actions]
```
 
or we add a cast to a list:

```
def coerce_actions_type(actions: Union[list[ActionType], ActionType]) -> list[ActionType]:
    """Converts a single action into a list of actions."""
    if isinstance(actions, Iterable):
        return cast(list[ActionType], actions)
    return [actions]
```

vizro-ai/examples/dashboard_ui/components.py
huong-li-nguyen: Also I just saw @maxschulz-COL comment on our old thread: https://github.com/mckinsey/vizro/pull/1104#discussion_r2035031063

He did like my initial idea of just upgrading the version here and taking over the changes, but just not syncronising to pyCafe. Shall we do that?

vizro-core/src/vizro/models/_models_utils.py
huong-li-nguyen: Noted :)

huong-li-nguyen:
> 2. something I've considered doing before but never actually done, and maybe it would make sense here. What would happen if we attached the validator to a type in `types.py` so we could reuse it everywhere without copying and pasting? Does it work with our API docs? Like this:
>    ```
>     # define once
>     ActionsType = Annotated[list[ActionType], BeforeValidator(coerce_actions_type), Field(default=[])]
>    
>     # use in models like this:
>     actions: ActionsType
>    ```

I just tried it out on the `Page` model here: https://github.com/mckinsey/vizro/pull/1328/commits/3a82d8ed149320467a2f24be342718cca0c434cf and the API docs would look like this with clickable references. Looks fine to me, but the docs build is broken now due to circular imports. Not sure if we can fix that 🤔 What do you think? 


Screenshot 2025-08-07 at 11 25 09

Screenshot 2025-08-07 at 11 24 56

vizro-core/src/vizro/models/_models_utils.py
antonymilne: Yes, I have a preference for this:
```
def coerce_actions_type(actions: Union[Iterable[ActionType], ActionType]) -> Iterable[ActionType]:
    """Converts a single action into a list of actions."""
    if isinstance(actions, Iterable):
        return actions
    return [actions]
```

The reason for this is not likely to be relevant here, hence why it's a minor point, but I think is a good principle to follow where we can do so easily: pydantic does automatic coercion of iterables to list by itself, e.g.
```
class X(BaseModel):
   field: list[int]

x = X(field=(1, 2, 3))
assert x.field == [1, 2, 3]
```

So in this case with actions, in theory someone could enter `(action_1, action_2)` which is valid configuration and be a valid input to `coerce_actions_type`. This would pass through our `coerce_actions_type` unchanged and then be turned into a list by pydantic's own field validation.

antonymilne:
> > 2. something I've considered doing before but never actually done, and maybe it would make sense here. What would happen if we attached the validator to a type in `types.py` so we could reuse it everywhere without copying and pasting? Does it work with our API docs? Like this:
> >    ```
> >     # define once
> >     ActionsType = Annotated[list[ActionType], BeforeValidator(coerce_actions_type), Field(default=[])]
> >    
> >     # use in models like this:
> >     actions: ActionsType
> >    ```
> 
> I just tried it out on the `Page` model here: [3a82d8e](https://github.com/mckinsey/vizro/commit/3a82d8ed149320467a2f24be342718cca0c434cf) and the API docs would look like this with clickable references. Looks fine to me, but the docs build is broken now due to circular imports. Not sure if we can fix that 🤔 What do you think?

How about if you move the `coerce_actions_type` into `types.py` itself? Give it a leading `_` so it doesn't appear in API docs.

huong-li-nguyen:
> Move `coerce_actions_type` to `types`

That would solve the docs build, but it then doesn't work anymore. Pydantic complains that it is "unable to extract tag using discriminator `_get_action_discriminator()`. I've implemented it in this PR, so you can see, but I'll revert after: https://github.com/mckinsey/vizro/pull/1328/commits/96b20d1fba4517ad886ccf1747bd530cab304c5e

I think it might be problematic that we have to define it in this order 🤔 


```
ActionType = Annotated[...]

def _coerce_actions_type(actions: Union[Iterable[ActionType], ActionType]) -> Iterable[ActionType]:
    ....

ActionsType = Annotated[list[ActionType], BeforeValidator(_coerce_actions_type), Field(default=[])]
```

vizro-ai/examples/dashboard_ui/components.py
lingyielia: For better long term consistency and maintenance, shall we sunset/remove this example from our open source codebase? I doubt anyone is still using it, and we are not able to apply any newer version to PyCafe as demo purpose.

huong-li-nguyen: Yes, even better 😄

vizro-core/src/vizro/models/_models_utils.py
antonymilne: Ahah so this seemed like a great idea but was actually really not... It's what caused this problem: https://github.com/mckinsey/vizro/pull/1328#issuecomment-3163604968.

I completely forgot that a pydantic model is itself iterable, so when you do this:
```
actions=export_data()
```

then our validator passes the `isinstance(actions, Iterable)` check and passes through `actions` just like that, without putting it into a list, which makes the discriminated union get very confused.

Sorry about that! I think the correct version is actually this:
```
from vizro.models import Action # probably maybe need to put this inside the coerce_actions_type function to avoid circular imports

def _coerce_actions_type(actions: Any) -> Any:
    """Converts a single action into a list of actions."""
    if isinstance(actions, Action):
        return [actions]
    return actions
```

If that still doesn't work then just go back to the list version you originally had, sorry!

antonymilne:
Just so you don't miss it, I think this is the resolution: https://github.com/mckinsey/vizro/pull/1328#discussion_r2260607506

vizro-ai/examples/dashboard_ui/components.py
maxschulz-COL: Let's do

vizro-core/src/vizro/models/types.py
```diff
@@ -707,6 +709,18 @@ class OptionsDictType(TypedDict):
...
 
+
+def _coerce_actions_type(actions: Union[Iterable[ActionType], ActionType]) -> Iterable[ActionType]:
+    """Converts a single action into a list of actions."""
+    if isinstance(actions, Iterable):
+        return actions
+    return [actions]
+
+
+ActionsType = Annotated[list[ActionType], BeforeValidator(_coerce_actions_type), Field(default=[])]
```
maxschulz-COL: This is exactly how we should use types, very nice indeed!

I would actually only add the comment about the JSON schema here, because this is where it would eventually be implemented now that we have this common type. Also avoids a bit of clutter elsewhere

maxschulz-COL: Idea: could we add to the `pyproject.toml` behind the pydantic dependency a comment that says something like:
"when updating this to >=2.7, search code for XXX to find all occurences of code that is now ready to change"

So that we actually remember doing this when we finally update. Whatever XXX becomes should find this comment, and the other comments regarding the Tooltip I believe

vizro-core/docs/pages/user-guides/custom-components.md
```diff
@@ -387,11 +387,11 @@ Add the custom action `open_offcanvas` as a `function` argument inside the [`Act
...
+1. **Add the `actions` argument to your custom component**. While the type annotation is `list[ActionType]`, the `BeforeValidator(coerce_actions_type)` ensures you can provide either a single `ActionType` or a list of actions.
```
maxschulz-COL: wait but this is not true anymore right?

vizro-core/src/vizro/models/_models_utils.py
maxschulz-COL: Oh boy, so much to consider, but I guess a good learning is to sometimes not try to do everything at once. E.g. here I guess we really only want to put the single action into a list, else trust entirely pydantic to coerce or fail right?

vizro-core/src/vizro/models/types.py
maxschulz-COL: P.S. Just saw that this causes trouble with the build, so hopefully we can remove this issue because besides that I think this is a great idea!

antonymilne: Nice, thanks for the comments @maxschulz-COL! I agree with everything you say here.

vizro-core/src/vizro/models/_models_utils.py
antonymilne: Yes agreed! Basically before-validators should be minimally invasive and leave as much to pydantic as possible. In this case what that means is we only want to interfere with the special case that the user provided a single action. Everything else we pass through unchanged and let pydantic handle it, whether that works or fails.

vizro-core/docs/pages/user-guides/custom-components.md
huong-li-nguyen: I didn't update everything yet, because we were just trying out the other approach last minute 😄  Will do that after getting it to work

vizro-core/src/vizro/models/_models_utils.py
huong-li-nguyen: Had to go back to the list version, otherwise it wouldn't work: https://github.com/mckinsey/vizro/pull/1328/commits/8462f5cfa8572ef2de241731d08275dabe7a090b

vizro-core/src/vizro/models/types.py
huong-li-nguyen: Done

vizro-core/docs/pages/user-guides/custom-components.md
huong-li-nguyen: Done

vizro-ai/examples/dashboard_ui/components.py
huong-li-nguyen: See https://github.com/mckinsey/vizro/pull/1332

vizro-core/src/vizro/models/types.py
```diff
@@ -707,6 +715,13 @@ class OptionsDictType(TypedDict):
 """Discriminated union. Type of action: [`Action`][vizro.models.Action], [`export_data`][vizro.models.export_data] or [
 `filter_interaction`][vizro.models.filter_interaction]."""
 
+# TODO: ideally actions would have json_schema_input_type=Union[list[ActionType], ActionType] attached to
+# the BeforeValidator, but this requires pydantic >= 2.9.
+ActionsType = Annotated[list[ActionType], BeforeValidator(_coerce_actions_type), Field(default=[])]
```
maxschulz-COL: I almost overlooked the (s) and got confused. Maybe rename to make it clearer? Like `ActionListType` or something?

vizro-core/src/vizro/models/_models_utils.py
antonymilne: Oh that's weird, what was wrong with the `isinstance(actions, Action)` version? I'm fine to leave with the `list` version anyway.

vizro-core/pyproject.toml
```diff
@@ -21,7 +21,10 @@ dependencies = [
   "dash_mantine_components>=1",  # 1.0.0 needed to support dash>=3.0.0
   "pandas>=2",
   "plotly>=5.24.0",
-  "pydantic>=2.7.0",  # must be <= the version in pre-commit mypy hook, sync manually
+  # Must be <= the version in pre-commit mypy hook, sync manually
+  # When updating this to >=2.9, search code for `but this requires pydantic >= 2.9` to find all
+  # occurrences of code that is now ready to change
+  "pydantic>=2.7.0",  #
```
antonymilne:
```suggestion
  "pydantic>=2.7.0",
```

vizro-core/src/vizro/models/_models_utils.py
antonymilne, line 1: Should this file exist?

vizro-core/src/vizro/models/types.py
antonymilne: I wouldn't go for `ActionListType` because in future we might have `ActionChainType` or similar and then it will get more confusing I think. Agree there's a high potential for missing the `s` with this naming but overall I think `ActionsType`, as it is now, is best.

antonymilne (approved):
Looks great, thank you for trying out all the different approaches! 🙌 

Please do check with @petar-qb before hitting merge just in case it disrupts something he's work on right now.

petar-qb:
> Please do check with @petar-qb before hitting merge just in case it disrupts something he's work on right now.

I believe I’ve fixed all the bugs on the remove-actions-loop branch and am now finalising the PR. I expect to wrap everything up by the end of the day.

@huong-li-nguyen, would it work for you to hold off merging until then?

huong-li-nguyen:
> > Please do check with @petar-qb before hitting merge just in case it disrupts something he's work on right now.
> 
> I believe I’ve fixed all the bugs on the remove-actions-loop branch and am now finalising the PR. I expect to wrap everything up by the end of the day.
> 
> @huong-li-nguyen, would it work for you to hold off merging until then?

Yes, sure. No rush :) Can merge whenever

vizro-core/src/vizro/models/_models_utils.py
huong-li-nguyen: I think that version is too restrictive as it doesn't account for actions like `export_data`, this example would e.g. fail at the moment (which I think is a very likely one), as `export_data` is an `_AbstractAction`.

```
page = vm.Page(
    title="My first dashboard",
    layout=vm.Flex(),
    components=[
        vm.AgGrid(figure=dash_ag_grid(df_gapminder)),
        vm.Button(actions=export_data()),
    ],
)
```

vizro-core/src/vizro/models/_models_utils.py
huong-li-nguyen: Not sure why it is displayed the way it is displayed 😄 I added no changes, but yes, this file should exist as it contains other important shared functions (independent of this PR) like `make_actions_chain` etc.

vizro-core/changelog.d/20250806_115509_90609403+huong-li-nguyen_enable_single_action.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
```
antonymilne: I would add this to the changelog.

huong-li-nguyen: Added :)

vizro-core/src/vizro/models/_models_utils.py
antonymilne: Ahhh yes, of course. We would have had to do `isinstance(actions, _BaseAction)` or something instead 🤦 

Oh well, leave with the `list` version because otherwise you wouldn't be able to reuse the same validator in #1334 anyway... so probably after all that, the way you initially had it was best after all! Sorry for the time wasted on this 😬

petar-qb (approved):
Looks great!

Really glad to see the `BeforeValidator(_coerce_actions_type)` approach won (it was my preference too). I also enjoyed following the discussion in this PR. I always end up learning something new from these.

PR #1334, opened 2025-08-11
[Tidy] Enable single string outputs in `Action` model

huong-li-nguyen:
## Description

Closes https://github.com/McK-Internal/vizro-internal/issues/2050

Users can now specify `outputs="component_id"` instead of requiring `outputs=["component_id"]` for single outputs in both `Action` and `_AbstractAction`

- Introduced `OutputsType` similar to `ActionsType`, see https://github.com/mckinsey/vizro/pull/1328 for reference
- Enhanced `_BaseAction._transformed_outputs` and `_AbstractAction._transformed_outputs`: Added coercion logic using `_coerce_to_list` to handle single strings
- Added unit tests for single string outputs

## Screenshot
Screenshot 2025-08-11 at 10 34 12

## Notice

### DIFF
--- a/vizro-core/src/vizro/actions/_abstract_action.py
+++ b/vizro-core/src/vizro/actions/_abstract_action.py
@@ -4,7 +4,6 @@ import abc
 import inspect
 from typing import Union
 
-from dash.development.base_component import Component
 from pydantic import TypeAdapter
 
 from vizro.models._action._action import _BaseAction


vizro-core/changelog.d/20250811_102650_90609403+huong-li-nguyen_enable_single_actions_output.md
```diff
@@ -0,0 +1,45 @@
...
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+### Added
+
+- Enable single string outputs in `Action` model. You can now specify `outputs="component_id"` instead of requiring `outputs=["component_id"]` for single outputs. ([#1334](https://github.com/mckinsey/vizro/pull/1334))
```
antonymilne:
```suggestion
- Enable single string outputs in `Action` model. You can now specify `outputs="component_id.property"` instead of `outputs=["component_id.property"]` for single outputs. ([#1334](https://github.com/mckinsey/vizro/pull/1334))
```

The `component_id` without `property` documentation is still WIP so I wouldn't mention it yet.

vizro-core/src/vizro/models/types.py
```diff
@@ -86,11 +86,18 @@ def _clean_module_string(module_string: str) -> str:
     return ""
 
 
-def _coerce_actions_type(actions: Any) -> Any:
-    """Converts a single action into a list of actions."""
-    if isinstance(actions, list):
-        return actions
-    return [actions]
+def _coerce_to_list(value: Any) -> Any:
+    """Converts a single item into a list unless it's already a list or dict.
```
antonymilne: Might be worth explaining in this comment why we exclude the dict case since it's just for `outputs`.

vizro-core/src/vizro/actions/_abstract_action.py
```diff
@@ -73,9 +73,10 @@ def outputs(self) -> Union[list[_IdOrIdProperty], dict[str, _IdOrIdProperty]]:
 
     @property
     def _transformed_outputs(self) -> Union[list[Output], dict[str, Output]]:
-        # Action.outputs is already validated by pydantic as list[str] or dict[str, str], but for
+        # Action.outputs is already validated by pydantic as str, list[str] or dict[str, str] via OutputsType, but for
         # _AbstractAction.outputs we need to do the validation manually with TypeAdapter.
-        TypeAdapter(Union[list[str], dict[str, str]]).validate_python(self.outputs)
+        coerced_outputs = _coerce_to_list(self.outputs)
+        TypeAdapter(Union[list[str], dict[str, str]]).validate_python(coerced_outputs)
```
antonymilne: Can we use the `OutputsType` type directly here? That would hopefully be a big advantage of defining it, not sure if it will work though.

```
TypeAdapter(OutputsType, self.outputs)
```

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -225,22 +227,23 @@ def _transformed_inputs(self) -> Union[list[State], dict[str, Union[State, Contr
...
+        Converts self.outputs (string, list of strings or dictionary of strings where each string is in the format
         '<component_id>.<property>' or '<component_id>') and converts into Dash Output objects.
-        For example, ['my_graph.figure'] becomes [Output('my_graph', 'figure', allow_duplicate=True)].
+        For example, 'my_graph.figure' or ['my_graph.figure'] becomes [Output('my_graph', 'figure', allow_duplicate=True)].
 
         Returns:
             Union[list[Output], dict[str, Output]]: A list of Output objects if self.outputs is a list of strings,
             or a dictionary mapping keys to Output objects if self.outputs is a dictionary of strings.
         """
+        coerced_outputs = _coerce_to_list(self.outputs)
```
antonymilne: Wait, don't we know by this point that `self.outputs` is already `list[str]` or `dict[str, str]`? Why do we need to handle the `str` case again?

vizro-core/tests/unit/vizro/models/test_types.py
```diff
@@ -411,3 +435,31 @@ def test_coerce_actions_type_integration(self):
...
+            ({"key1": "output1"}, {"key1": "output1"}),
+            ({"key1": "output1", "key2": "output2"}, {"key1": "output1", "key2": "output2"}),
+        ],
+    )
+    def test_coerce_outputs_type(self, output, expected_output):
+        """Test that _coerce_to_list works correctly for Action.outputs (preserves lists and dicts)."""
+        result = _coerce_to_list(output)
+        assert result == expected_output
+
+    def test_coerce_outputs_type_integration(self):
```
antonymilne: Please can we parameterise this into three cases with `expected_output`?

antonymilne (commented):
Generally looks really good 🙏 Just a few initial comment 🙂

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: Might be missing something but inside the `_AbstractAction.transformed_outputs` we're calling the `return super()._transformed_outputs` from the `BaseAction` and while it's guaranteed that it's either a list or dict on the `Action.outputs` and `_AbstractAction.transformed_outputs` because we apply the `_coerce_to_list` there, it is not guaranteed on the `BaseAction.outputs`? 🤔

vizro-core/src/vizro/actions/_abstract_action.py
huong-li-nguyen: Seems to work :)

vizro-core/src/vizro/models/_action/_action.py
antonymilne: Oh I think I see. Basically we could do `TypeAdapter(OutputsType, self.outputs)` to do the conversion of `"output"` to `["output"]` at this point, but then the output of that is effectively lost and wouldn't be consumed in `transformed_outputs`?

I guess what we want to do here is `self.outputs = TypeAdapter(OutputsType, self.outputs)` so we can guarantee that in `_transformed_outputs` we don't need to do any more type conversions. But that won't work because it would be setting a property 🤔 

I think we can work this out, but first let me know if we're on the same page about what the problem is here!

huong-li-nguyen: Yes, exactly. So the transformation gets lost because we need to return `super()._transformed_outputs` inside the `_AbstractAction._transformed_outputs`, hence I thought we need the transformation inside the BaseClass as well. Returning the coerced values directly instead of `super()._transformed_outputs` led to a bunch of failing unit tests, as we can't set the property as you've mentioned. So we would get this: `AttributeError: 'outputs' is a ClassVar of `_action_with_mock_outputs` and cannot be set on an instance.`

Hence, I added the conversion here as well 🤔

antonymilne: I think this should be a good fix: [c4fa9e9](https://github.com/mckinsey/vizro/pull/1334/commits/c4fa9e9c38b8e466f43d6b96b422e759c9b5b824). It seems to pass tests for `AbstractAction` and `Action` but I didn't manually test at all.

huong-li-nguyen: Ah, amazing 🥳 I think that should do it then!

vizro-core/changelog.d/20250811_102650_90609403+huong-li-nguyen_enable_single_actions_output.md
```diff
@@ -0,0 +1,45 @@
...
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+### Added
+
+- Enable single string outputs in `Action` model. You can now specify `outputs="component_id.property"` instead of `outputs=["component_id.property"]` for single outputs. ([#1334](https://github.com/mckinsey/vizro/pull/1334))
```
antonymilne:
```suggestion
- Enable single string outputs in `Action` model. You can now specify `outputs="component_id.property"` instead of `outputs=["component_id.property"]` for a single output. ([#1334](https://github.com/mckinsey/vizro/pull/1334))
```

antonymilne (approved):
Looks perfect to me ⭐ 🌟 

@petar-qb - please review and merge this and #1328 whenever you're ready.

PR #1339, opened 2025-08-13
[Refactor] Refactor navigation to decouple title and model ID

maxschulz-COL:
## Description
Closes https://github.com/McK-Internal/vizro-internal/issues/2079

ToDo
- [x] Write docs
- [x] Write tests that checks for reference by ID not by title
- [x] Write tests that check new model ID functionality

## Notice


vizro-core/docs/pages/user-guides/navigation.md
```diff
@@ -68,7 +68,7 @@ By default, if the `navigation` argument is not specified, Vizro creates a navig
 
 ## Include a subset of pages
 
-To include only some of your dashboard pages in your navigation then list them in the `pages` argument of the `Navigation` model. To refer to a page inside the `Navigation` model, you should always use the page's `id`, which [defaults to the page `title`](pages.md#customize-the-page-url), we specify `pages = ["My first page", "My second page"]` rather than `pages=[page_1, page_2]`.
+To include only some of your dashboard pages in your navigation then list them in the `pages` argument of the `Navigation` model. To refer to a page inside the `Navigation` model, you should always use the page's `id` or `title`. In case you have duplicate titles, you should use the `id` to refer to the page.
```
stichbury:
```suggestion
To include only some of the dashboard pages in the navigation, list them in the `pages` argument of the `Navigation` model. 

To refer to a page inside the `Navigation` model, you should always use the page's `id` or `title`. If you have duplicate titles, use the `id` to refer to the page.
```

stichbury: > To include only some of the dashboard pages in the navigation, list them in the `pages` argument of the `Navigation` model. 

What is the default, if I don't list them? Nothing or all pages? Probably should make this clear.

> To refer to a page inside the `Navigation` model, you should always use the page's `id` or `title`. If you have duplicate titles, use the `id` to refer to the page.

Is this not always the case? Why highlight it here?

stichbury (approved):
LGTM, one minor comment.

vizro-core/changelog.d/20250813_132042_maximilian_schulz_decouple_title_id.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Model ID is now independently set from the page title. Referring to Pages in Navigation can be done by ID or title. ([#1339](https://github.com/mckinsey/vizro/pull/1339))
```
antonymilne:
```suggestion
- Page ID is now set independently from the page title. You can refer to a page in navigation by ID or title. ([#1339](https://github.com/mckinsey/vizro/pull/1339))
```

vizro-core/src/vizro/models/_base.py
```diff
@@ -209,24 +214,30 @@ def _split_types(type_annotation: type[Any]) -> type[Any]:
...
-def set_id(id: ModelID) -> ModelID:
-    return id or model_manager._generate_id()
+def _generate_id() -> ModelID:
+    return str(uuid.UUID(int=rd.getrandbits(128)))
+
+
+def set_id(value: Any) -> Any:
+    if value is None:
+        return _generate_id()
+    return value
```
antonymilne:
```suggestion
def set_id(value: Any) -> Any:
    if value is None:
        return str(uuid.UUID(int=rd.getrandbits(128)))
    return value
```

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -114,12 +116,42 @@ def _get_model_page(self, model: Model) -> Page:  # type: ignore[return]
...
+        Returns:
+            Tuple of (page_by_id, pages_by_title) dictionaries.
+        """
+        from vizro.models import Page
+
+        page_by_id: dict[ModelID, Page] = {}
+        pages_by_title: dict[str, list[Page]] = {}
+        for page in self._get_models(Page):
+            page_by_id[page.id] = page
+            pages_by_title.setdefault(page.title, []).append(page)
```
antonymilne: Just a bit more conventional.

```suggestion
        pages_by_title: dict[str, list[Page]] = defaultdict(list)
        for page in self._get_models(Page):
            page_by_id[page.id] = page
            pages_by_title[page.title].append(page)
```

vizro-core/tests/unit/vizro/models/test_base.py
```diff
@@ -252,6 +252,31 @@ def _serialize_id(self, nxt: SerializerFunctionWrapHandler, info: FieldSerializa
...
+    def test_model_id_set(self):
+        model = Model(id="model_id")
+        assert model.id == "model_id"
+
+    def test_model_id_set_to_none(self):
+        model = Model(id=None)
+        assert model.id is not None
+        assert isinstance(model.id, str)
+
+    def test_model_id_set_to_empty_string(self):
```
antonymilne, line 240: FYI I am going to introduce some limitations on `id` that will make empty string and certain characters invalid. No one should set an empty string as an id or stuff will break in reality 😅 But this test is fine anyway.

vizro-core/tests/unit/vizro/models/test_base.py
```diff
@@ -252,6 +252,31 @@ def _serialize_id(self, nxt: SerializerFunctionWrapHandler, info: FieldSerializa
         return result
 
 
+class TestModelID:
+    def test_model_id_default(self):
+        model = Model()
+        assert model.id is not None
+        assert isinstance(model.id, str)
```
antonymilne: A bit stricter - we can test if it's a valid UUID and not just string. Similarly for `test_model_id_set_to_none`.

```suggestion
        assert str(uuid.UUID(model.id)) == model.id
```

vizro-core/tests/integration/test_navigation.py
```diff
@@ -34,74 +34,81 @@ def dashboard_result(request):
...
     ),
     (
         lambda: vm.Navigation(),
-        lambda: vm.Navigation(nav_selector=vm.Accordion(pages=["Page 1", "Page 2"])),
+        lambda: vm.Navigation(nav_selector=vm.Accordion(pages=["page_1", "page_2"])),
     ),
     (
         lambda: vm.Navigation(nav_selector=vm.Accordion()),
-        lambda: vm.Navigation(nav_selector=vm.Accordion(pages=["Page 1", "Page 2"])),
+        lambda: vm.Navigation(nav_selector=vm.Accordion(pages=["page_1", "page_2"])),
```
antonymilne, line 45: Is it worth mixing these cases so we test id and title like `["page_1", "Page 2"]`? Like you did in unit tests.

vizro-core/tests/integration/test_navigation.py
```diff
@@ -34,74 +34,81 @@ def dashboard_result(request):
...
+        lambda: vm.Navigation(nav_selector=vm.NavBar(pages=["page_1"])),
     ),
     (
         lambda: vm.Navigation(pages=["Page 1"], nav_selector=vm.NavBar(pages=["Page 2"])),
-        lambda: vm.Navigation(nav_selector=vm.NavBar(pages=["Page 2"])),
+        lambda: vm.Navigation(nav_selector=vm.NavBar(pages=["page_2"])),
     ),
     (
-        lambda: vm.Navigation(pages=["Page 1"], nav_selector=vm.NavBar()),
+        lambda: vm.Navigation(pages=["Page 1"], nav_selector=vm.NavBar()),  # Isn't this identical to above?
```
antonymilne: It doesn't look the same to me?

vizro-core/src/vizro/models/_navigation/nav_bar.py
```diff
@@ -41,8 +42,14 @@ class NavBar(VizroBaseModel):
 
     @_log_call
     def pre_build(self):
+        from vizro.models import Page
+
         self.items = self.items or [
-            NavLink(label=group_title, pages=pages) for group_title, pages in self.pages.items()
+            NavLink(
```
antonymilne, line 49: A comment here would be good.

vizro-core/src/vizro/models/_base.py
antonymilne, line 1: Can we remove the special code you had in `_to_python` now?

vizro-core/src/vizro/models/_page.py
```diff
@@ -42,6 +41,15 @@
...
+# MS: how dangerous to rely on MM in validation? Probably this will immediately move to pre-build in next PR
+def _check_for_duplicate_path(path: str) -> bool:
+    registered_paths = [page.path for page in cast(Iterable[Page], model_manager._get_models(Page))]
+
+    if path in registered_paths:
+        return True
+    return False
+
+
 def set_path(path: str, info: ValidationInfo) -> str:
```
antonymilne: Ahhh yes, now I remember one of the reasons we implemented everything like this before. Just thinking out loud here to make sure we're on the same page:
* we need page path to be unique, hence derived from `id`
* we want page path to be readable and not just a random string, hence derived from `title`

Old solution: `title` gives `id` which gives `path`, so guaranteed to be unique and also readable. If there are multiple pages with the same `title` then raise an error saying you need to set unique id. In navigation refer to page by `id`.

New solution: derive `path` from `title` directly. If there are multiple pages with the same `title` then raise an error saying you need to set unique id. In navigation refer to page by `title` or `id`.

Is that all correct?

vizro-core/src/vizro/models/_page.py
```diff
@@ -52,8 +60,15 @@ def clean_path(path: str, allowed_characters: str) -> str:
...
     # Allow "/" in path if provided by user, otherwise turn page id into suitable URL path (not allowing "/")
     if path:
+        if _check_for_duplicate_path(path):
+            raise ValueError(f"Path {path} is already used by another page.")
         return clean_path(path, "-_/")
-    return clean_path(info.data["id"], "-_")
+
+    # Try title first, then fall back to id if title path is duplicate
+    # MS: There is a small breaking change (I think) when once chooses a path that is the same as the home page,
+    # this case would work before, but not anymore.
```
antonymilne: I'm pretty confused here... 😅 It's probably right but I don't quite get it. Let's discuss.

vizro-core/src/vizro/models/_page.py
```diff
@@ -42,6 +41,15 @@
 _PageBuildType = TypedDict("_PageBuildType", {"control-panel": html.Div, "page-components": html.Div})
 
 
+# MS: how dangerous to rely on MM in validation? Probably this will immediately move to pre-build in next PR
```
antonymilne: Yeah, definitely not great... But given we'll fix it very soon it's ok.

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -114,12 +116,42 @@ def _get_model_page(self, model: Model) -> Page:  # type: ignore[return]
...
-        return str(uuid.UUID(int=rd.getrandbits(128)))
+    def _get_page_lookup_maps(self) -> tuple[dict[ModelID, Page], dict[str, list[Page]]]:
+        """Build lookup maps for pages by ID and title.
+
+        Returns:
+            Tuple of (page_by_id, pages_by_title) dictionaries.
+        """
+        from vizro.models import Page
+
+        page_by_id: dict[ModelID, Page] = {}
```
antonymilne:
```suggestion
        pages_by_id: dict[ModelID, Page] = {}
```

Used in `navigation_utils.py` also.

vizro-core/src/vizro/models/_navigation/_navigation_utils.py
```diff
@@ -1,31 +1,103 @@
...
+    # First check if it's a valid page ID
+    if page_ref in page_by_id:
+        return page_ref
+
+    # Try to find by title
+    if page_ref in pages_by_title:
+        matching_pages = pages_by_title[page_ref]
+        if len(matching_pages) > 1:
+            page_ids = [p.id for p in matching_pages]
+            raise ValueError(_AMBIGUOUS_PAGE_ERROR_MSG.format(page_ref=page_ref, page_ids=page_ids))
```
antonymilne: Isn't this error handled in the model manager now?

vizro-core/src/vizro/managers/_model_manager.py
antonymilne: I think the data structures here could be simplified a bit which would make everything clearer. Let's discuss this afternoon.

vizro-core/tests/integration/test_navigation.py
maxschulz-COL: Hm not quite sure if I follow - I thought that was what I was doing. Remember the expected case is always the IDs, but the reference can be both, I split that into title and ID wherever possible

vizro-core/tests/integration/test_navigation.py
maxschulz-COL: To line 81?

vizro-core/src/vizro/models/_base.py
maxschulz-COL: Very well spotted, but I believe the special code is in the page, as we do not need to take care of the case of same id anymore 🚀

maxschulz-COL: I also removed some code in the `test_base` file - have a look if you agree

vizro-core/docs/pages/user-guides/navigation.md
maxschulz-COL: To point 1: All pages in dashboard are taken

To point 2: what is always the case? It is special that you can refer to a page title, usually one can only refer to models by ID

maxschulz-COL: To point 1: this is explained in the section above

vizro-core/src/vizro/models/_base.py
```diff
@@ -243,7 +250,7 @@ def model_post_init(self, context: Any) -> None:
     # This was like pydantic's own __exclude_fields__ but this is not possible in V2 due to the non-recursive nature of
     # the model_dump method. Now this serializer allows to add the model name to the dictionary when serializing the
     # model if called with context {"add_name": True}.
-    # Excluding specific fields is now done via overwriting this serializer (see e.g. Page model).
+    # Excluding specific fields is now done via overwriting this serializer.
```
antonymilne: I just checked and we don't do this anywhere now 🎉 

```suggestion
    # Excluding specific fields could be done via overwriting this serializer, but we don't currently do this anywhere.
```

Or just delete the comment completely now.

vizro-core/src/vizro/models/_base.py
antonymilne: Ah yes, looks good. Definitely a good sign we're making an improvement here.

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -114,9 +108,16 @@ def _get_model_page(self, model: Model) -> Page:  # type: ignore[return]
             if model in self.__get_model_children(page):  # type: ignore[operator]
                 return page
 
-    @staticmethod
-    def _generate_id() -> ModelID:
-        return str(uuid.UUID(int=rd.getrandbits(128)))
+    def _get_page_lookup_maps(self) -> dict[str, list[ModelID]]:
```
antonymilne: This is so small now that I'd actually just move it into the point it's used rather than keeping it in `model_manager`.

Note if you want to keep it here then the function should be renamed e.g. `_get_page_titles`.

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -114,9 +108,16 @@ def _get_model_page(self, model: Model) -> Page:  # type: ignore[return]
...
+        """Build lookup map for pages by title."""
+        from vizro.models import Page
+
+        title_to_ids: dict[str, list[ModelID]] = {}
+
+        for page in self._get_models(Page):
+            title_to_ids.setdefault(page.title, []).append(page.id)
```
antonymilne: This would still be better as a `defaultdict`:

```suggestion
        title_to_ids: dict[str, list[ModelID]] = defaultdict(list)

        for page in self._get_models(Page):
            title_to_ids[page.title].append(page.id)
```

vizro-core/src/vizro/models/_page.py
```diff
@@ -42,6 +38,15 @@
 _PageBuildType = TypedDict("_PageBuildType", {"control-panel": html.Div, "page-components": html.Div})
 
 
+# TODO[MS]: Probably this will immediately move to pre-build in next PR
+def _check_for_duplicate_path(path: str) -> bool:
+    registered_paths = [page.path for page in cast(Iterable[Page], model_manager._get_models(Page))]
+
+    if path in registered_paths:
+        return True
+    return False
```
antonymilne:
```suggestion
    return path in registered_paths
```

Or even more simply:

```
def _check_for_duplicate_path(path: str) -> bool:
    return any(path == page.path for page in cast(Iterable[Page], model_manager._get_models(Page)))
```

vizro-core/src/vizro/models/_page.py
```diff
@@ -51,9 +56,20 @@ def clean_path(path: str, allowed_characters: str) -> str:
         return path if path.startswith("/") else "/" + path
 
     # Allow "/" in path if provided by user, otherwise turn page id into suitable URL path (not allowing "/")
+    # maybe remove?
     if path:
+        if _check_for_duplicate_path(path):
+            raise ValueError(f"Path {path} is already used by another page.")
```
antonymilne:
```suggestion
            raise ValueError(f"Path {path} cannot be used by more than one page.")
```

vizro-core/src/vizro/models/_page.py
```diff
@@ -51,9 +56,20 @@ def clean_path(path: str, allowed_characters: str) -> str:
...
         return clean_path(path, "-_/")
-    return clean_path(info.data["id"], "-_")
+
+    # Changes to previous iteration before PR 1339:
+    #
+    # - if non duplicated title and explicit id: BEFORE: used id             AFTER: title
+    # - if non duplicated title and no id:       BEFORE: used title (via id) AFTER: title
+    # - if duplicated title and explicit id:     BEFORE: used id             AFTER: id
+    # - if duplicated title and no id:           BEFORE: error (via id)      AFTER: id (rand gen) or err (nav defined)
+    path = clean_path(info.data["title"], "-_") if "title" in info.data else "invalid"  # pydantic validation will catch
```
antonymilne: Please can you remind me why do we need the `else` here? Field ordering means this validator should always run after `title` has been set, and it's a compulsory field?

vizro-core/src/vizro/models/_navigation/_navigation_utils.py
```diff
@@ -1,31 +1,90 @@
...
+# introduced with handling things this way
 def _validate_pages(pages: NavPagesType) -> NavPagesType:
-    """Reusable validator to check if provided Page IDs exist as registered pages."""
-    from vizro.models import Page
+    """Reusable validator to check if provided Page titles exist as registered pages."""
+    title_to_ids = model_manager._get_page_lookup_maps()
+
+    # Check if pages is empty (either an empty list or a dict with empty values)
+    if (isinstance(pages, dict) and not any(pages.values())) or (isinstance(pages, list) and not pages):
+        raise ValueError(_EMPTY_PAGES_ERROR_MSG)
```
antonymilne:
```suggestion
```

Given you have the `_EMPTY_PAGES_ERROR_MSG` variable and the below if/else I would just put this in the relevant blocks to save you having to repeat the if/else logic on L66.

```
# Process pages based on structure (dict or list)
    if isinstance(pages, dict):
        if not any(pages.values()):
            raise ...
    else:
        if not pages:
            raise ...
```

vizro-core/src/vizro/models/_navigation/_navigation_utils.py
```diff
@@ -1,31 +1,90 @@
...
-    if unknown_pages := [page for page in pages_as_list if page not in registered_pages]:
-        raise ValueError(f"Unknown page ID {unknown_pages} provided to argument 'pages'.")
-    return pages
+    if unknown_pages:
+        available_titles = list(title_to_ids.keys())
+        raise ValueError(
+            _UNKNOWN_PAGES_ERROR_MSG.format(unknown_pages=unknown_pages, available_titles=available_titles)
+        )
+    return_value = validated_dict if isinstance(pages, dict) else validated_list
+    return return_value
```
antonymilne: I suggest renaming variables above to `validated_pages` in both cases, then you can just `return validated_pages` here without needing to do another if/else on L86. The `unknown_pages` is already being reused nicely so would be good to do the same for `validated_dict/list`.

vizro-core/src/vizro/models/_navigation/_navigation_utils.py
```diff
@@ -1,31 +1,90 @@
...
+
+def _resolve_list_of_page_references(
+    group_pages: list[str], title_to_ids: dict[str, list[ModelID]]
+) -> tuple[list[str], list[ModelID]]:
+    """Resolve a list of page references to page IDs."""
+    unknown_pages: list[str] = []
+    validated_list: list[ModelID] = []
+
+    for page_ref in group_pages:
+        resolved_id = _resolve_page_reference(page_ref, title_to_ids)
```
antonymilne:
```suggestion
        resolved_ref = _resolve_page_reference(page_ref, title_to_ids)
```

vizro-core/src/vizro/models/_navigation/_navigation_utils.py
```diff
@@ -1,31 +1,90 @@
...
 from typing_extensions import TypedDict
 
 from vizro.managers import model_manager
-from vizro.models.types import NavPagesType
+from vizro.models.types import ModelID, NavPagesType
 
+# Error message constants
+_AMBIGUOUS_PAGE_ERROR_MSG = (
+    "Ambiguous page reference '{page_ref}': Multiple pages found with title '{page_ref}'. "
+    "Page IDs: {page_ids}. Page titles must be unique to avoid ambiguity."
```
antonymilne:
```suggestion
    "Page IDs: {page_ids}. You should refer to these pages by ID to avoid ambiguity."
```

vizro-core/src/vizro/models/_navigation/nav_bar.py
antonymilne: Please can you remind me why this changed?

vizro-core/src/vizro/models/_base.py
```diff
@@ -209,24 +214,26 @@ def _split_types(type_annotation: type[Any]) -> type[Any]:
...
 
     """
 
     id: Annotated[
         ModelID,
-        AfterValidator(set_id),
+        BeforeValidator(set_id),
         Field(
-            default="",
+            default=None,
```
antonymilne: 💡 Even better, why don't we just do this as a `default_factory=str(uuid.UUID(int=rd.getrandbits(128)))`?

Maybe this would even help with the next stage if default values are somehow filled in before model validators? I'm not sure... If it doesn't help then probably not worth making the change.

In hindsight I wonder why we didn't always do this with `default_factory` 🤔  Possibly there's other bits of code across our models where we're setting default values through a validator that would actually be much better off as `default_factory`.

vizro-core/src/vizro/models/_page.py
```diff
@@ -51,9 +56,20 @@ def clean_path(path: str, allowed_characters: str) -> str:
...
+    return path if not _check_for_duplicate_path(path) else clean_path(info.data["id"], "-_")
```
antonymilne: Sorry but reading through it again, this function still feels wrong to me :| Let me try to explain why on a call but I would suggest we end up with something like this:

```
if path:
    path = clean_path(path, "-_/")
elif "id" in self.model_fields_set:
    path = clean_path(info.data["id"], "-_")
else:
    path = clean_path(info.data["title"], "-_")
    
if any(path == page.path for page in cast(Iterable[Page], model_manager._get_models(Page))):
    raise ValueError

return path
```

Current system feels to me:
* confusing/unexpected behaviour
* unstable/inconsistent wrt ordering which feels wrong and I think has real consequences
* misses check for duplicate paths with `id` (very minor point)

There's an advantage of just working in case that navigation isn't explicitly specified and no ids are specified but I don't think I care about that enough to outweight the cons.

Ultimately the check here should possibly not go in `Pages.pre_build` at all but instead in `Dashboard.pre_build`?

vizro-core/src/vizro/models/_base.py
antonymilne: I did a super quick test of this and in principle I think it should work since `default_factory` gets run before the earliest validator! I guess that's obvious because of the existence of the whole `validate_default` option which wouldn't make sense otherwise, but good to know.

maxschulz-COL: Passed all but one test - you now cannot do `id=None`. Fine by me

vizro-core/src/vizro/models/_navigation/_navigation_utils.py
maxschulz-COL: The `unknown_pages` is a list in both cases - mypy really doesn't like this branching..

maxschulz-COL: So I kept it as is with a small renaming

vizro-core/src/vizro/models/_navigation/nav_bar.py
maxschulz-COL: So before when `id==title` we would have a nicely readable automatic Tooltip. Without this change this is now the ID, which is not as nice as the title

maxschulz-COL: The deeper reason is as follow:
- first `coerce_pages_type` always creates a `Mapping`, where the keys are the elements of the list
- if the elements of the list are model IDs, then the key will be a model ID
- we _always_ have the dictionary mapping to a model ID, but before the model ID was either the title, or a chosen (hence nice) ID

I still need to see how this plays out in the approach discussed yesterday, ie whether there is a chance to have non self-selected model IDs here

maxschulz-COL: Ok, now having implemented the other approach the situation is clear:

- either the user provides `pages_as_list` to the `Navbar` --> `coerce_pages_type` will make a mapping, and the keys will be the elements of that list --> they become the Tooltip
- if the user doesn't provide anything, that list will be the Model IDs --> Model IDs become tooltip

Now we have a failing test, because in the test we provide (in order to test referring to both ID and title) a title and an ID. Before that would have set the ID to the title, so both tooltips would have been title. Now its a title and a page ID.

Question: do we want the Tooltip to show the list content, or rather the titles of the pages the list refers to? Or something else?

maxschulz-COL: Add comment and check for page

vizro-core/src/vizro/models/_navigation/accordion.py
```diff
@@ -14,7 +14,7 @@
 from vizro.models.types import ModelID
 
 
-def coerce_pages_type(pages):
+def coerce_pages_type(pages: Any) -> Any:
     if isinstance(pages, Mapping):
```
antonymilne: Following some recent lessons learnt from @huong-li-nguyen's work on converting `actions = action` to `actions = [action]`...

```suggestion
def coerce_pages_type(pages: Any) -> dict:
    if isinstance(pages, dict):
```

vizro-core/src/vizro/models/_navigation/accordion.py
```diff
@@ -25,7 +25,8 @@ class Accordion(VizroBaseModel):
 
     Args:
         type (Literal["accordion"]): Defaults to `"accordion"`.
-        pages (dict[str, list[ModelID]]): Mapping from name of a pages group to a list of page IDs. Defaults to `{}`.
+        pages (dict[str, list[ModelID]]): Mapping from name of a pages group to a list of page IDs/titles.
```
antonymilne, line 27: This type hint show now be `dict[str, list[str]]` or maybe `dict[str, list[Union[ModelID, str]]]` (kind of meaningless given that `ModelID` is just `str` but maybe a bit clearer? I don't know).

This would need changing on L37 and there's probably a few other type hints in navigation which should be updated - if you just search for `ModelID` then you should find them all.

vizro-core/src/vizro/models/_navigation/nav_bar.py
```diff
@@ -1,31 +1,36 @@
...
+from vizro.managers import model_manager
 from vizro.models import VizroBaseModel
 from vizro.models._models_utils import _log_call
 from vizro.models._navigation._navigation_utils import _NavBuildType, _validate_pages
 from vizro.models._navigation.nav_link import NavLink
 from vizro.models.types import ModelID
 
 
-def coerce_pages_type(pages: Union[list[str], dict[str, list[str]]]) -> dict[str, list[str]]:
+def coerce_pages_type(pages: Any) -> Any:
```
antonymilne, line 17: Let's change these to `isinstance(pages, dict) ` and `isinstance(pages, list)` just to be consistent with other validators.

vizro-core/src/vizro/models/_navigation/nav_bar.py
```diff
@@ -35,14 +40,24 @@ class NavBar(VizroBaseModel):
...
     items: list[NavLink] = []
 
     @_log_call
     def pre_build(self):
+        from vizro.models import Page
+
         self.items = self.items or [
-            NavLink(label=group_title, pages=pages) for group_title, pages in self.pages.items()
+            NavLink(
+                # If the group title is a model ID, then we prefer to have the title of that page be used
```
antonymilne:
```suggestion
                # If the group title is a page ID (as is the case if you do `NavBar(pages=["page_1_id", "page_2_id"])`, then we prefer to have the title rather than id of that page be used
```

vizro-core/src/vizro/models/_navigation/nav_bar.py
```diff
@@ -35,14 +40,24 @@ class NavBar(VizroBaseModel):
...
+                # as Tooltip label
+                label=cast(Page, model_manager[group_title]).title
+                if group_title in [page.id for page in model_manager._get_models(model_type=Page)]
```
antonymilne, line 53: Don't change it here but this is the sort of thing I really hope will look nicer after our forthcoming changes to the model manager!

vizro-core/src/vizro/models/_page.py
```diff
@@ -90,20 +80,33 @@ class Page(VizroBaseModel):
...
+            if not self.id == page.id and new_path == page.path:
+                raise ValueError(f"Path {new_path} cannot be used by more than one page.")
+
+        # We should do self.path = new_path but this leads to a recursion error. The below is a workaround
+        # until the pydantic bug is fixed. See https://github.com/pydantic/pydantic/issues/6597.
+        self.__dict__["path"] = new_path
+
+        return self
+
+    # get rid of this either way
```
antonymilne: This needs doing still. I believe it's still in `VizroBaseModel` too and shouldn't be.

vizro-core/src/vizro/models/_page.py
```diff
@@ -90,20 +80,33 @@ class Page(VizroBaseModel):
...
-        values.setdefault("id", values["title"])
-        return values
-
+    # This should ideally be a field validator, but we need access to the model_fields_set
+    @model_validator(mode="after")
+    def validate_path(self):
+        fields_set = self.model_fields_set
+        if self.path:
+            new_path = clean_path(self.path, "-_/")
+        elif "id" in fields_set:
```
antonymilne:
```suggestion
        elif "id" in self.model_fields_set:
```

Very minor style comment - there's no point putting this in a variable.

vizro-core/tests/unit/vizro/models/_navigation/test_accordion.py
```diff
@@ -22,14 +22,18 @@ def test_mandatory_only(self):
         assert hasattr(accordion, "id")
         assert accordion.pages == {}
 
-    def test_mandatory_and_optional(self, pages_as_dict):
+    def test_mandatory_and_optional(self, pages_as_dict, prebuilt_two_page_dashboard):
```
antonymilne: I feel like all the navigation unit tests would be clearer if we just modified `prebuilt_two_page_dashboard` to explicitly assign `id` in the definition of `vm.Page`. That way the tests would be more readable and you could do `{"Group": ["page_1", "page_2"]}` etc.

Ideally I think all the references to `page_1.id` in navigation unit tests would then be replaced with the explicit `"page_1"` etc.

antonymilne: Oh, now I see that you deliberately assigned `id` to `page_1` but not to `page_2`. I would suggest just assigning explicit id to both because we already have tests that check a model ID is correctly generated if it's not explicitly assigned, so I don't think we need to cover that again here.

vizro-core/tests/integration/test_navigation.py
antonymilne: Ah yes I see now - you were doing it right already 👍

vizro-core/tests/integration/test_navigation.py
antonymilne: Hmmm yes, not sure why there's two tests here. Let's remove this one and leave the one on L81.

vizro-core/docs/pages/user-guides/navigation.md
```diff
@@ -68,7 +68,7 @@ By default, if the `navigation` argument is not specified, Vizro creates a navig
 
 ## Include a subset of pages
 
-To include only some of your dashboard pages in your navigation then list them in the `pages` argument of the `Navigation` model. To refer to a page inside the `Navigation` model, you should always use the page's `id`, which [defaults to the page `title`](pages.md#customize-the-page-url), we specify `pages = ["My first page", "My second page"]` rather than `pages=[page_1, page_2]`.
+To include only some of the dashboard pages in the navigation, list them in the `pages` argument of the `Navigation` model. To refer to a page inside the `Navigation` model, you should always use the page's `id` or `title`. If you have duplicate titles, use the `id` to refer to the page.
```
antonymilne, line 71: Do we actually say anywhere in our docs that page title don't need to be unique? If not then might be worth adding a small note to the guide on pages to point that out.

antonymilne (approved):
A few final bits to tidy but after that we're finally good to merge! Good job persevering with this - I know it was painful at times, but I think we're in much better shape now 🙌

vizro-core/src/vizro/models/_navigation/accordion.py
maxschulz-COL: No, the type hint shows the guaranteed types after validation, and that is `ModelID` - hence see my comment next to the model type.

maxschulz-COL: And also see my comment next to `NavPagesType`

vizro-core/tests/unit/vizro/models/_navigation/test_accordion.py
maxschulz-COL: That's gonna be a bit of work, but for you I'll do it ;)

maxschulz-COL: My original intention was to make tests more comprehensive to cover both cases

vizro-core/src/vizro/models/_page.py
maxschulz-COL: Removed this one, good catch, but what about the other? Not sure what is left in `VizroBaseMOdel`

vizro-core/src/vizro/models/_navigation/accordion.py
antonymilne: Ah yes, sorry, I forgot that!

vizro-core/src/vizro/models/_page.py
antonymilne: Isn't `model_post_init` still in `VizroBaseModel`?

maxschulz-COL: Yes, but I believe we need that one!

```
    @_log_call
    def model_post_init(self, context: Any) -> None:
        model_manager[self.id] = self
```

vizro-core/docs/pages/user-guides/navigation.md
maxschulz-COL: added it to the page docs

petar-qb:
@maxschulz-COL Due to the lack of PR and issue description, it is quite difficult to understand key points of this PR. Would you mind sharing more about this PR in one of the next learning sessions?

maxschulz-COL:
> @maxschulz-COL Due to the lack of PR and issue description, it is quite difficult to understand key points of this PR. Would you mind sharing more about this PR in one of the next learning sessions?

Point taken - I update the issue: https://github.com/McK-Internal/vizro-internal/issues/2079

That plus the new docs should hopefully explain it? IN the end it is a very small conceptual change with a large chain of consequences :)

PR #1341, opened 2025-08-13
[Feat] Add icon argument to `vm.Button`

nadijagraca:
## Description

closes https://github.com/McK-Internal/vizro-internal/issues/2069

This PR adds `icon` argument to `vm.Button`. 

Expected behaviour:
- if you set both `icon` and `text`, icon will appear to the left of text and the button will render as a rectangle just like now
- if you set only `icon`, and `text=""`, the button will render as a circle
- if you set `description` and `icon` but not `text` then info icon will not be displayed. The description text should appear as a tooltip on hovering over the button/icon circle. 

## Screenshot
Screenshot 2025-08-13 at 09 43 30

## Notice


vizro-core/changelog.d/20250730_124911_nadija_ratkusic_graca_update_dbc_links.md
```diff
@@ -16,12 +16,12 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 - A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
 
 -->
-<!--
+
 ### Added
 
-- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+- Enabled `icon` argument in `vm.Button` which enables users to add any icon from [Google Material Icons library](https://fonts.google.com/icons) ([#1](https://github.com/mckinsey/vizro/pull/1))
```
huong-li-nguyen:
```suggestion
- Add `icon` argument to `vm.Button`, allowing any icon from the [Google Material Icons library](https://fonts.google.com/icons) to be displayed inside the `vm.Button`. ([#1341](https://github.com/mckinsey/vizro/pull/1341))
```

vizro-core/changelog.d/20250813_150653_nadija_ratkusic_graca_add_icon_argument.md
```diff
@@ -0,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
```
huong-li-nguyen, line 2: One changelog file should be removed :)

vizro-core/tests/unit/vizro/actions/conftest.py
```diff
@@ -67,7 +67,7 @@ def managers_one_page_two_graphs_one_button(box_chart, scatter_chart):
         components=[
             vm.Graph(id="box_chart", figure=box_chart),
             vm.Graph(id="scatter_chart", figure=scatter_chart),
-            vm.Button(id="button"),
+            vm.Button(id="button", text="test"),
```
huong-li-nguyen: The tests shouldn't need to change as we discussed, at least any changes related to the `text` argument :)

vizro-core/tests/unit/vizro/actions/test_export_data.py
```diff
@@ -49,7 +49,7 @@ def target_data_filtered_pop(request, gapminder_2007):
 @pytest.fixture
 def managers_one_page_without_graphs_one_button():
     """Instantiates a simple model_manager and data_manager with a page, and no graphs."""
-    vm.Page(id="test_page", title="My first dashboard", components=[vm.Button(id="button")])
+    vm.Page(id="test_page", title="My first dashboard", components=[vm.Button(id="button", text="Click me!")])
```
huong-li-nguyen:
```suggestion
    vm.Page(id="test_page", title="My first dashboard", components=[vm.Button(id="button")])
```

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -34,7 +43,14 @@ class Button(VizroBaseModel):
...
     type: Literal["button"] = "button"
-    text: Annotated[str, Field(default="Click me!", description="Text to be displayed on button.", min_length=1)]
+    icon: Annotated[
+        str,
+        AfterValidator(validate_icon),
+        Field(description="Icon name from Google Material icons library.", default=""),
+    ]
+    text: Annotated[
+        str, AfterValidator(validate_text), Field(description="Text to be displayed on button.", default="Click me!")
+    ]
```
huong-li-nguyen: Would keep the same order as in the docstrings above. So depending on which order is required for the validation, I would either change the order here or inside the docstrings

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -8,18 +8,27 @@
...
+
+
 class Button(VizroBaseModel):
     """Component provided to `Page` to trigger any defined `action` in `Page`.
 
     Args:
         type (Literal["button"]): Defaults to `"button"`.
         text (str): Text to be displayed on button. Needs to have at least 1 character. Defaults to `"Click me!"`.
         href (str): URL (relative or absolute) to navigate to. Defaults to `""`.
+        icon (str): Icon name from [Google Material icons library](https://fonts.google.com/icons).
```
huong-li-nguyen:
```suggestion
        icon (str): Icon name from [Google Material icons library](https://fonts.google.com/icons). Defaults to `""`.
```

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -84,16 +100,32 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
         defaults = {
             "id": self.id,
-            "children": html.Span([self.text, *description], className="button-text"),
+            "children": html.Span([*icon, self.text, *description], className="button-text"),
             "href": get_relative_path(self.href) if self.href.startswith("/") else self.href,
             "target": "_top",
             # dbc.Button includes `btn btn-primary` as a class by default and appends any class names provided.
             # To prevent unnecessary class chaining, the button's style variant should be specified using `color`.
             "color": variants[self.variant],
+            "class_name": "circular_button" if self.icon and not self.text else "",
```
huong-li-nguyen:
```suggestion
            "class_name": "btn-circular" if self.icon and not self.text else "",
```

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -293,3 +293,7 @@ Current logic:
   flex-direction: row;
   gap: 12px;
 }
+
+.circular_button {
+  border-radius: 50%;
+}
```
huong-li-nguyen:
```suggestion

.btn-circular {
  border-radius: 50%;
}
```

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -84,16 +100,32 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
+        icon = (
+            html.Span(self.icon, id=f"{self.id}-icon", className="material-symbols-outlined tooltip-icon")
+            if self.icon
+            else None,
+        )
 
         defaults = {
             "id": self.id,
-            "children": html.Span([self.text, *description], className="button-text"),
+            "children": html.Span([*icon, self.text, *description], className="button-text"),
```
huong-li-nguyen: Not related to this PR, but could we rename this as well? To be consistent with bootstrap naming :) 

```suggestion
            "children": html.Span([*icon, self.text, *description], className="btn-text"),
```

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -8,18 +8,27 @@
 
 from vizro.models import Tooltip, VizroBaseModel
 from vizro.models._action._actions_chain import _action_validator_factory
-from vizro.models._models_utils import _log_call
+from vizro.models._models_utils import _log_call, validate_icon
 from vizro.models._tooltip import coerce_str_to_tooltip
 from vizro.models.types import ActionType, _IdProperty
 
 
+def validate_text(text, info):
```
huong-li-nguyen:
```suggestion
def validate_text(text, info: ValidationInfo):
```

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -8,18 +8,27 @@
...
+    icon = info.data.get("icon")
+
+    if text or icon:
+        return text
+    raise ValueError("Please provide either the text or icon argument.")
```
huong-li-nguyen: I think this direction is easier to understand

```suggestion
    if not text and not icon:
        raise ValueError("Please provide either the text or icon argument.")
    
    return text
```

vizro-core/src/vizro/static/css/layout.css
huong-li-nguyen: I don't think the circular buttons look correct at the moment, it doesn't look like a circle but more like an oval shape? I would expect it to look exactly like our collapse-icon :) So you might have to specify the same height / width as well, so it's really squared first.

Screenshot 2025-08-14 at 11 07 14

vizro-core/docs/pages/user-guides/button.md
```diff
@@ -240,6 +240,82 @@ You can provide [Markdown text](https://markdown-guide.readthedocs.io/) as a str
...
+        ```{.python pycafe-link hl_lines="11 24"}
+        import vizro.models as vm
+        from vizro import Vizro
+
+        page = vm.Page(
+            title="Buttons with an icon",
+            layout=vm.Flex(),
+            components=[
+                vm.Button(
+                    text="Visit link to learn more!",
```
huong-li-nguyen:
```suggestion
                    text="View Data Source",
```

vizro-core/docs/pages/user-guides/button.md
```diff
@@ -240,6 +240,82 @@ You can provide [Markdown text](https://markdown-guide.readthedocs.io/) as a str
...
+                        y="sepal_length",
+                        color="species",
+                        size="petal_length",
+                    ),
+                ),
+                vm.Button(
+                    icon="download",
+                    description="Download the data!",
+                    variant="outlined",
+                ),
```
huong-li-nguyen: I'd probably place this button in the top right corner of the chart. Are there any other quick actions that could go there? A single circular button on its own doesn't fully show the benefit of using a circular style 🤔 — especially since we don't have other quick in-built actions yet.

I think the concept would make more sense if, for example, we had three circular buttons grouped in the top right for quick actions like: refresh data, download data, or take a screenshot of the chart.

Given we don't have those built in yet, I'd leave it as a single circular button in the top right for now, though it does look a little odd to me 😄 Curious to hear what @antonymilne and you think

vizro-core/docs/pages/user-guides/button.md
```diff
@@ -240,6 +240,82 @@ You can provide [Markdown text](https://markdown-guide.readthedocs.io/) as a str
...
+
+        ```{.python pycafe-link hl_lines="11 24"}
+        import vizro.models as vm
+        from vizro import Vizro
+
+        page = vm.Page(
+            title="Buttons with an icon",
+            layout=vm.Flex(),
+            components=[
+                vm.Button(
```
huong-li-nguyen, line 261: This one i'd place below the chart

vizro-core/src/vizro/static/css/layout.css
huong-li-nguyen: For reference, this is how round the collapse-icon looks like:

Screenshot 2025-08-14 at 11 28 18

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -84,16 +100,32 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
         return dbc.Button(**(defaults | self.extra))
+
+    def _build_description(self):
+        if not self.description:
+            return [None]
+
+        description = self.description.build().children
+        if not self.text:
+            description = [description[1]]
+            description[0].target = f"{self.id}-icon"
```
huong-li-nguyen: I forgot, what's inside description[1] and description[0]? 😄 Can we assign them to variables with meaningful names?

stichbury (commented):
LGTM (the docs changes are only a small % of the work but all good, and the button itself looks awesome).

vizro-core/docs/pages/user-guides/button.md
nadijagraca: I would leave a simple example of how it looks like for the docs, and maybe add more interesting example in the dev dashboard, where we could showcase the full benefit of circular buttons.

huong-li-nguyen:
After this will be merged, your screenshot tests should also pass :) https://github.com/mckinsey/vizro/pull/1342
It's currently broken on main, but I'll fix it in a sec.

vizro-core/changelog.d/20250730_124911_nadija_ratkusic_graca_update_dbc_links.md
```diff
@@ -16,6 +16,7 @@ Uncomment the section that is right (remove the HTML comment wrapper).
 - A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
 
 -->
+
```
huong-li-nguyen, line 19: Revert change here?

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -2,22 +2,32 @@
...
+    return text
+
+
 class Button(VizroBaseModel):
     """Component provided to `Page` to trigger any defined `action` in `Page`.
 
     Args:
         type (Literal["button"]): Defaults to `"button"`.
+        icon (str): Icon name from [Google Material icons library](https://fonts.google.com/icons). Defaults to `""`.
         text (str): Text to be displayed on button. Needs to have at least 1 character. Defaults to `"Click me!"`.
```
huong-li-nguyen:
```suggestion
        text (str): Text to be displayed on button. Defaults to `"Click me!"`.
```

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -84,16 +101,35 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
         return dbc.Button(**(defaults | self.extra))
+
+    def _build_description(self):
+        if not self.description:
+            return [None]
+
+        description = self.description.build().children
+        if not self.text:
+            description_tooltip = description[1]
+            # if text="" tooltip icon is not rendered, in which case the target for the tooltip moves to the button icon
```
huong-li-nguyen: Maybe move this into the docstring and quickly explain all relevant info inside the docstring of the method :) 

```suggestion
            # if text="" tooltip icon is not rendered, in which case the target for the tooltip moves to the button icon
```

vizro-core/tests/unit/vizro/actions/conftest.py
```diff
@@ -86,7 +86,7 @@ def managers_one_page_two_graphs_one_table_one_aggrid_one_button(
             vm.Graph(id="scatter_chart", figure=scatter_chart),
             vm.Table(id="vizro_table", figure=dash_data_table_with_id),
             vm.AgGrid(id="ag_grid", figure=ag_grid_with_id),
-            vm.Button(id="button"),
+            vm.Button(id="button", text="test"),
```
huong-li-nguyen:
```suggestion
            vm.Button(id="button"),
```

vizro-core/tests/unit/vizro/models/_components/test_button.py
```diff
@@ -142,10 +151,32 @@ def test_button_build_with_description(self):
...
+                class_name="",
             ),
         )
+
+    @pytest.mark.parametrize(
+        "icon, text, class_name",
+        [("home", "Test", ""), ("home", "", "btn-circular")],
+    )
+    def test_button_build_with_icon(self, icon, text, class_name):
+        button = vm.Button(id="button_id", icon=icon, text=text).build()
```
huong-li-nguyen, line 168: I think we also have to add the use cases with description right? So if no text is provided, but an icon and a description etc. So I guess all possible combinations between text, icon and description and their builds

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -293,3 +293,9 @@ Current logic:
   flex-direction: row;
   gap: 12px;
 }
+
+.btn-circular {
+  border-radius: 50%;
+  height: 32px;
+  width: 32px;
```
huong-li-nguyen:
```suggestion
  height: 28px;
  width: 28px;
```
Screenshot 2025-08-14 at 16 01 00


I think they looked a bit too big. Could you also make sure the collapse-icon has height/width of 28px then. Because optimally these two should look the same in terms of dimension :)

vizro-core/changelog.d/20250730_124911_nadija_ratkusic_graca_update_dbc_links.md
nadijagraca: Yes, this is a changelog from another PR (where I updated the dcs links) and I added text here by mistake. Reverted the changes here, and added it to the correct changelog file.

vizro-core/tests/unit/vizro/models/_components/test_button.py
nadijagraca: Added a separate test as it would be too cluttered to put all options in one test.

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -2,23 +2,33 @@
...
+from vizro.models._models_utils import _log_call, validate_icon
 from vizro.models._tooltip import coerce_str_to_tooltip
 from vizro.models.types import ActionType, _IdProperty
 
 
+def validate_text(text, info: ValidationInfo):
+    icon = info.data.get("icon")
+
+    if not text and not icon:
+        raise ValueError("Please provide either the text or icon argument.")
```
huong-li-nguyen:
```suggestion
        raise ValueError("Please provide either the `text` or `icon` argument.")
```

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -293,3 +293,9 @@ Current logic:
   flex-direction: row;
   gap: 12px;
 }
+
+.btn-circular {
+  border-radius: 50%;
+  height: 28px;
+  width: 28px;
```
huong-li-nguyen, line 301: I think you need to apply the same height/width to the collapse-icon. The last time I checked it was on 26px I think

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -84,16 +101,37 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
             # To prevent unnecessary class chaining, the button's style variant should be specified using `color`.
             "color": variants[self.variant],
+            "class_name": "btn-circular" if self.icon and not self.text else "",
         }
 
         return dbc.Button(**(defaults | self.extra))
+
+    def _build_description(self):
+        """Builds and returns the description elements. If text='' tooltip icon is not rendered and tooltip target
+        is reassigned to the button icon.
```
huong-li-nguyen:
```suggestion
        """Conditionally returns the tooltip based on the provided `text` and icon arguments. If text='', the tooltip icon is omitted, and the tooltip text is shown when hovering over the button icon. Otherwise, the tooltip icon is displayed.
```

vizro-core/tests/unit/vizro/models/_components/test_button.py
```diff
@@ -68,55 +70,62 @@ def test_button_build(self):
...
         )
-        assert_component_equal(button, expected)
 
     @pytest.mark.parametrize(
         "variant, expected_color",
         [("plain", "link"), ("filled", "primary"), ("outlined", "secondary")],
     )
     def test_button_with_variant(self, variant, expected_color):
-        result = vm.Button(variant=variant).build()
+        result = vm.Button(variant=variant, text="Click me!").build()
```
huong-li-nguyen:
```suggestion
        result = vm.Button(variant=variant).build()
```

vizro-core/tests/unit/vizro/models/_components/test_button.py
```diff
@@ -142,10 +151,64 @@ def test_button_build_with_description(self):
...
+            button,
+            dbc.Button(
+                id="button_id",
+                children=html.Span([expected_icon, "", *expected_description], className="btn-text"),
+                target="_top",
+                href="",
+                color="primary",
+                class_name="btn-circular",
+            ),
+            keys_to_strip={"id"},
```
huong-li-nguyen: I think they still pass even without? And if not, then I guess there is no need to provide `ids` in the config if they get ignored anyway

```suggestion
```

vizro-core/tests/unit/vizro/models/_components/test_button.py
```diff
@@ -142,10 +151,64 @@ def test_button_build_with_description(self):
...
+            button,
+            dbc.Button(
+                id="button_id",
+                children=html.Span([expected_icon, f"{text}", None], className="btn-text"),
+                target="_top",
+                href="",
+                color="primary",
+                class_name=class_name,
             ),
+            keys_to_strip={"id"},
```
huong-li-nguyen:
```suggestion

```

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -2,23 +2,33 @@
...
 
 from vizro.models import Tooltip, VizroBaseModel
 from vizro.models._action._actions_chain import _action_validator_factory
-from vizro.models._models_utils import _log_call
+from vizro.models._models_utils import _log_call, validate_icon
 from vizro.models._tooltip import coerce_str_to_tooltip
 from vizro.models.types import ActionType, _IdProperty
 
 
+def validate_text(text, info: ValidationInfo):
```
maxschulz-COL: Since this is an after validator, we should definitely add typing

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -2,23 +2,33 @@
...
+    icon = info.data.get("icon")
+
+    if not text and not icon:
```
maxschulz-COL: Is there a specific reason you chose a `field_validator` here? My gut feeling tells me that a model validator is better suited here just to avoid depending on the order of the fields?

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -34,7 +44,14 @@ class Button(VizroBaseModel):
     """
 
     type: Literal["button"] = "button"
-    text: Annotated[str, Field(default="Click me!", description="Text to be displayed on button.", min_length=1)]
+    icon: Annotated[
+        str,
+        AfterValidator(validate_icon),
+        Field(description="Icon name from Google Material icons library.", default=""),
```
maxschulz-COL, line 39: Is there a reason we have a default of `""`? Does it not make more sense to have a default of `None`?

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -84,16 +101,39 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
             "target": "_top",
             # dbc.Button includes `btn btn-primary` as a class by default and appends any class names provided.
             # To prevent unnecessary class chaining, the button's style variant should be specified using `color`.
             "color": variants[self.variant],
+            "class_name": "btn-circular" if self.icon and not self.text else "",
         }
 
         return dbc.Button(**(defaults | self.extra))
+
+    def _build_description(self):
```
maxschulz-COL: Could we add typing here

maxschulz-COL (commented):
Not done yet, but away for a bit so posting these already

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -34,7 +44,14 @@ class Button(VizroBaseModel):
     """
 
     type: Literal["button"] = "button"
-    text: Annotated[str, Field(default="Click me!", description="Text to be displayed on button.", min_length=1)]
+    icon: Annotated[
+        str,
+        AfterValidator(validate_icon),
+        Field(description="Icon name from Google Material icons library.", default=""),
+    ]
+    text: Annotated[
```
maxschulz-COL: We seem to have gotten rid of the `min_length` - but there is still a comment below about it:

>         # AfterValidator(warn_description_without_title) is not needed here because 'text' is mandatory and
>         # must have at least one character.

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -84,16 +101,39 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
+
+        If text='', the tooltip icon is omitted, and the tooltip text is shown when hovering over the button icon.
+        Otherwise, the tooltip icon is displayed.
+        """
+        if not self.description:
+            return [None]
+
+        description = self.description.build().children
+        if not self.text:
+            description_tooltip = description[1]
```
maxschulz-COL: Hm, could we hint or at least add a comment what sort of data structure we are expecting? Are we sure it is always going to be `[1]`?

maxschulz-COL (commented):
⭐ I really like this addition ⭐ 

Can we make the code a bit more clear in terms of typing, but then I think good to go! I only read through the Python part given that Li already manually tested everything

vizro-core/src/vizro/models/_components/button.py
nadijagraca: I've updated the comment to reflect current state.

vizro-core/src/vizro/models/_components/button.py
nadijagraca: I've update the docstring and added a comment. It can be a bit confusing as our model is called Tooltip as well as the underlying component that we are trying to access here (dbc.Tooltip). The position of the dbc.Tooltip should always be [1], if not we have tests that will alert us of any future changes.

vizro-core/src/vizro/models/_components/button.py
nadijagraca: There is no particular reason. We have a default `""` for navlink icon as well, and I just wanted to stay consistent. If you have a strong preference for `None` I can change it.

vizro-core/src/vizro/models/_components/button.py
maxschulz-COL: This also includes return types

vizro-core/src/vizro/models/_components/button.py
maxschulz-COL: I see there was something added below, do we still need the check here then?

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -84,16 +108,40 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
             "target": "_top",
             # dbc.Button includes `btn btn-primary` as a class by default and appends any class names provided.
             # To prevent unnecessary class chaining, the button's style variant should be specified using `color`.
             "color": variants[self.variant],
+            "class_name": "btn-circular" if self.icon and not self.text else "",
         }
 
         return dbc.Button(**(defaults | self.extra))
+
+    def _build_description(self) -> list[Union[None, dbc.Tooltip, html.Span]]:
```
maxschulz-COL:
```suggestion
    def _build_description(self) -> list[Optional[Union[dbc.Tooltip, html.Span]]:
```

What you wrote is actually the preferred method I believe, but for consistency sake I think we should still use `Optional`

vizro-core/src/vizro/models/_components/button.py
maxschulz-COL: If possible, and if typing actually allows it,  I think for all icon cases, `Optional[str] = None` is the clearer choice. To me, an empty string is a confusing value that really should be a null. But let's check with @huong-li-nguyen or @antonymilne what they think

huong-li-nguyen: We actually don't just use this approach for `navlink`, but for all optional string fields such as `title`, `header`, `footer`, etc.

Originally, we defined them as `Optional[str] = None`, but after some back-and-forth on what the default should be for optional strings, we eventually decided on `str = ""`. Not sure if this decision is now outdated with Pydantic v2. I don't remember all the reasoning, but one key point was definitely that we wanted to avoid handling `None` in all the validators 😬. I’m sure @antonymilne can provide more context on that 👍 So for now, I'd leave it as is.

maxschulz-COL: We discussed this yesterday and both preferred `Optional[str] = None` - however, it is maybe wise not to do this quickly and some more thought needs to be given to the consequences on the schema. So I would also be fine to leave as is

maxschulz-COL: I thought about this some more and I think this is indeed more complicated than expected. Let's leave as is for now.

vizro-core/docs/pages/user-guides/button.md
```diff
@@ -240,6 +240,82 @@ You can provide [Markdown text](https://markdown-guide.readthedocs.io/) as a str
...
+        ```{.python pycafe-link hl_lines="9 25"}
+        import vizro.models as vm
+        from vizro import Vizro
+
+        page = vm.Page(
+            title="Buttons with an icon",
+            layout=vm.Flex(),
+            components=[
+                vm.Button(
+                    icon="download",
```
antonymilne:
```suggestion
                    icon="Download",
```

Minor preference that we should always refer to Google Material icons by e.g. "Shopping Cart" rather than "shopping_cart".

vizro-core/docs/pages/user-guides/button.md
```diff
@@ -240,6 +240,82 @@ You can provide [Markdown text](https://markdown-guide.readthedocs.io/) as a str
...
+                        x="sepal_width",
+                        y="sepal_length",
+                        color="species",
+                        size="petal_length",
+                    ),
+                ),
+                vm.Button(
+                    text="View Data Source",
+                    href="https://www.kaggle.com/datasets/uciml/iris",
+                    icon="link",
```
antonymilne:
```suggestion
                    icon="Link",
```

vizro-core/docs/pages/user-guides/button.md
```diff
@@ -240,6 +240,82 @@ You can provide [Markdown text](https://markdown-guide.readthedocs.io/) as a str
...
+        ```
+
+    === "app.yaml"
+
+        ```{.yaml hl_lines="7 18"}
+        pages:
+          - title: Buttons with an icon
+            components:
+              - type: button
+                icon: download
```
antonymilne:
```suggestion
                icon: Download
```

vizro-core/docs/pages/user-guides/button.md
```diff
@@ -240,6 +240,82 @@ You can provide [Markdown text](https://markdown-guide.readthedocs.io/) as a str
...
+                  x: sepal_width
+                  y: sepal_length
+                  color: species
+                  size: petal_length
+                  data_frame: iris
+                type: graph
+              - type: button
+                text: Visit link to learn more!
+                href: https://www.kaggle.com/datasets/uciml/iris
+                icon: link
```
antonymilne:
```suggestion
                icon: Link
```

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -74,6 +80,13 @@ class Button(VizroBaseModel):
         ]
     ]
 
+    @model_validator(mode="after")
+    def validate_text(self):
+        if not self.text and not self.icon:
+            raise ValueError("Please provide either the `text` or `icon` argument.")
```
antonymilne:
```suggestion
            raise ValueError("You must provide either the `text` or `icon` argument.")
```

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -84,16 +97,40 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
         defaults = {
             "id": self.id,
-            "children": html.Span([self.text, *description], className="button-text"),
+            "children": html.Span([*icon, self.text, *description], className="btn-text"),
             "href": get_relative_path(self.href) if self.href.startswith("/") else self.href,
             "target": "_top",
             # dbc.Button includes `btn btn-primary` as a class by default and appends any class names provided.
             # To prevent unnecessary class chaining, the button's style variant should be specified using `color`.
             "color": variants[self.variant],
+            "class_name": "btn-circular" if self.icon and not self.text else "",
```
antonymilne:
```suggestion
            "class_name": "btn-circular" if not self.text else "",
```

Should do the same thing I think? Given that you must provide at least one of `icon` and `text`.

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -84,16 +97,40 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
+        If text='', the tooltip icon is omitted, and the tooltip text is shown when hovering over the button icon.
+        Otherwise, the tooltip icon is displayed.
+        """
+        if not self.description:
+            return [None]
+
+        description = self.description.build().children
+        if not self.text:
+            # retrieves the dbc.Tooltip instance from the vm.Tooltip model to update its tooltip target
+            tooltip_component = description[1]
```
antonymilne:
```suggestion
            tooltip_component = description[description.id]
```

I think should work the same and would be more robust than referencing by position in a list.

This is generally nice code though, I like it 👍

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -84,16 +97,40 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
+
+        If text='', the tooltip icon is omitted, and the tooltip text is shown when hovering over the button icon.
+        Otherwise, the tooltip icon is displayed.
+        """
+        if not self.description:
+            return [None]
+
+        description = self.description.build().children
+        if not self.text:
+            # retrieves the dbc.Tooltip instance from the vm.Tooltip model to update its tooltip target
```
antonymilne:
```suggestion
            # When there's no text, we don't display the tooltip icon.
            # Instead we update the tooltip target to the button's icon.
```

vizro-core/tests/unit/vizro/models/_components/test_button.py
```diff
@@ -142,10 +151,62 @@ def test_button_build_with_description(self):
...
+    )
+    def test_button_build_with_icon(self, icon, text, class_name):
+        button = vm.Button(id="button_id", icon=icon, text=text).build()
+
+        expected_icon = html.Span(f"{icon}", id="button_id-icon", className="material-symbols-outlined tooltip-icon")
+        assert_component_equal(
+            button,
+            dbc.Button(
+                id="button_id",
+                children=html.Span([expected_icon, f"{text}", None], className="btn-text"),
```
antonymilne:
```suggestion
                children=html.Span([expected_icon, text, None], className="btn-text"),
```

vizro-core/changelog.d/20250813_150653_nadija_ratkusic_graca_add_icon_argument.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Add `icon` argument to `vm.Button`, allowing any icon from the [Google Material Icons library](https://fonts.google.com/icons) to be displayed inside the `vm.Button`. ([#1341](https://github.com/mckinsey/vizro/pull/1341))
```
antonymilne:
```suggestion
- Add `icon` argument to `vm.Button`, enabling any icon from the [Google Material Icons library](https://fonts.google.com/icons) to be displayed inside the `vm.Button`. This also enables circular buttons with only an icon. ([#1341](https://github.com/mckinsey/vizro/pull/1341))
```

vizro-core/docs/pages/user-guides/button.md
antonymilne: Yeah, agreed with @nadijagraca. I think this is ok just for a quick docs example. Once we have some more built in actions like you suggest then I agree we should come up with some better examples and place it more suitably.

vizro-core/src/vizro/models/_components/button.py
antonymilne: Agreed. I haven't heard the new complications, but as it stands my general preference now is probably `Optional[str] = None` like @maxschulz-COL suggested. But let's not change that now - there's a whole bunch of places across our codebase where we could change this another time. So leave as you have it here @nadijagraca.

vizro-core/src/vizro/models/_components/button.py
nadijagraca: Unfortunatelly, we cannot access it by id as `description.children` is a list. 

Even if we access the whole build object (below), description.id does not exist as `self.id` is assigned to a dbc.Tooltip component. With this set up we cannot know the `id` of the tooltip component. Currently indexing is the best option.  
```
return html.Div(
            [
                html.Span(self.icon, id=f"{self.id}-icon" ...),
                dbc.Tooltip(**(defaults | self.extra)),
            ]
        )
```
Do you want me to add a ticket to assign `self.id` of `description` to the outer `html.Div`?
In that case we could use your suggestion and access the tooltip component by 
`tooltip_component = description[f"{description.id}-tooltip"]`

antonymilne: Oh I see, this is what we should be doing I think? We don't want the `.children` here upfront because that prevents us from looking up by `id`. The `dbc.Tooltip` should have the right `id` assigned to it since we put `"id": self.id` in the `defaults` dictionary?
```
        description = self.description.build()

        if not self.text:
            tooltip_component = description[description.id]
            tooltip_component.target = f"{self.id}-icon"
            return [tooltip_component]

        return description.children
```

nadijagraca: Removing `children` would not help in this case, as `id` is unknown at this point. description is a html.Div without id attribute and I'm getting a `AttributeError: 'Div' object has no attribute 'id'`. 
In order to access `dbc.Tooltip` by `id` we have to go back to `vm.Tooltip` implementation and assign `self.id` not to the tooltip but to the outer div, and assign `"id": f"{self.id}-tooltip",`

And then we could do `tooltip_component = description[f"{description.id}-tooltip"]`

antonymilne: Sorry, this is what I meant:
```
        description_build_obj = self.description.build()

        if not self.text:
            tooltip_component = description_build_obj[self.description.id]
            tooltip_component.target = f"{self.id}-icon"
            return [tooltip_component]

        return description_build_obj.children
```

Unless I made another mistaking writing this out. Basically this is what I'm trying to do. It should definitely work if we just write it correctly 😅 

```
description = Tooltip(text="hi", icon="info")
description_build_obj = description.build()
print(description_build_obj[description.id]) # same as description_build_obj.childrenj[1]
```

nadijagraca: Oh sorry, I must be blind. I see now.

PR #1347, opened 2025-08-19
[Feat] Added new `collapse_expand_containers` action for collapsing/expanding containers

nadijagraca:
## Description

closes https://github.com/McK-Internal/vizro-internal/issues/2048

Added new `collapse_expand_containers` action which enables users to control collapsibility of the containers on a button click. 
#### Actions arguments:
- `collapse`: list of container ids to collapse
- `expand`: list of container ids to expand

#### Validation rules:
- at least one list needs to be defined, either `collapse` or `expand`
- `collapse` and `expand` cannot have any overlap, meaning no container id can appear in both `collapse` and `expand`
- ids provided in the lists must be container ids and the containers in question need to be collapsible.

## Notice


vizro-core/src/vizro/actions/_collapse_expand_container.py
```diff
@@ -0,0 +1,62 @@
...
+from pydantic import Field, model_validator
+
+from vizro.actions._abstract_action import _AbstractAction
+from vizro.managers import model_manager
+from vizro.models._models_utils import _log_call
+from vizro.models.types import ModelID
+
+
+class collapse_expand_containers(_AbstractAction):
+    """Exports visible data of target charts/components.
```
maxschulz-COL: ?

vizro-core/src/vizro/actions/_collapse_expand_container.py
```diff
@@ -0,0 +1,62 @@
...
+    type: Literal["collapse_expand_containers"] = "collapse_expand_containers"
+    collapse: list[ModelID] = Field(default=[], description="List of collapsible container ids to collapse.")
+    expand: list[ModelID] = Field(default=[], description="List of collapsible container ids to expand.")
+    #  toggle would need multiple state as input. Maybe possible in future but not now. Add new ticket for this
+    # toggle: list[ModelID]
+
+    @model_validator(mode="after")
+    def validate_collapse_or_expand_present(self):
+        if not self.collapse and not self.expand:
+            raise ValueError("Please provide either the `collapse` or `expand` argument.")
```
maxschulz-COL:
```suggestion
            raise ValueError("Either the `collapse` or `expand` list must contain at least a single element.")
```

vizro-core/src/vizro/actions/_collapse_expand_container.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if invalid_targets:
+            raise ValueError(f"Invalid component IDs found: {invalid_targets}")
+
+    def function(self) -> dict[ModelID, Any]:
+        """Collapse or expand containers on page."""
+        return dict.fromkeys(self.collapse, True) | dict.fromkeys(self.expand, False)
+
+    @property
+    def outputs(self):  # type: ignore[override]
```
maxschulz-COL, line 57: just curious why mypy complains here?

maxschulz-COL (commented):
Looks good to me! Will we add docs?

vizro-core/changelog.d/20250819_170759_nadija_ratkusic_graca_collapse_expand_containers_action.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Enabled `collapse_expand_containers` action to control collapsible containers by specifying lists of container ids to collapse or expand. ([#1](https://github.com/mckinsey/vizro/pull/1))
```
petar-qb:
```suggestion
- Enabled `collapse_expand_containers` action to control collapsible containers by specifying lists of container ids to collapse or expand. ([#1347](https://github.com/mckinsey/vizro/pull/1347))
```

vizro-core/src/vizro/actions/_collapse_expand_container.py
```diff
@@ -0,0 +1,62 @@
...
+    def validate_collapse_or_expand_present(self):
+        if not self.collapse and not self.expand:
+            raise ValueError("Please provide either the `collapse` or `expand` argument.")
+        return self
+
+    @model_validator(mode="after")
+    def validate_collapse_and_expand_overlap(self):
+        if set(self.collapse) & set(self.expand):
+            raise ValueError("Collapse and expand lists cannot contain the same elements!")
+        return self
```
petar-qb: Let's merge these two model_validators in a single one. (I also added [Max's suggestion](https://github.com/mckinsey/vizro/pull/1347/files#r2293082841) below)

```suggestion
    @model_validator(mode="after")
    def validate_collapse_and_expand(self):
        if not self.collapse and not self.expand:
            raise ValueError("Either the `collapse` or `expand` list must contain at least a single element.")  
        if set(self.collapse) & set(self.expand):
            raise ValueError("Collapse and expand lists cannot contain the same elements!")
        return self
```

vizro-core/src/vizro/actions/_collapse_expand_container.py
```diff
@@ -0,0 +1,62 @@
...
+            for model in model_manager._get_models(Container, root_model=model_manager._get_model_page(self))
+            if model.collapsed is not None
+        ]
+
+        invalid_targets = [
+            container_id for container_id in user_container_ids if container_id not in collapsible_container_ids
+        ]
+
+        if invalid_targets:
+            raise ValueError(f"Invalid component IDs found: {invalid_targets}")
```
petar-qb: Would something like this work?

```suggestion
        page_collapsible_container_ids = {
            model.id
            for model in model_manager._get_models(Container, root_model=model_manager._get_model_page(self))
            if model.collapsed is not None
        }

        invalid_ids = set(self.collapse + self.expand) - page_collapsible_container_ids
        if invalid_ids:
            raise ValueError(
                f"Invalid component IDs found: {invalid_ids}."
                " Action's collapse and expand IDs must be collapsible containers on the same page as the action."
            )
```

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -100,7 +100,7 @@ def make_actions_chain(self):
     # We need to delete the old action models from the model manager so they don't get built. After that,
     # built in actions are always handled in the new way.
     for action in self.actions:
-        if isinstance(action.function, (export_data, filter_interaction)):
+        if isinstance(action.function, (export_data, filter_interaction, collapse_expand_containers)):
```
petar-qb: @nadijagraca could you double check that the `collapse_expand_containers ` is referenced in all needed places. Just search for "export_data" in the repo and double check that I didn't miss anything. 

Could you also ensure that the order here (and in other similar places) is alphabetical. So: `(collapse_expand_containers, export_data, filter_interaction)`.

P.S. We would need (some other time so maybe you can add a TODO here) to add a new type like `BuiltInActionsType = union[collapse_expand_containers, export_data, filter_interaction)` and then reuse it all over the code. That would simplify adding a new built-in action in future.

vizro-core/tests/unit/vizro/actions/test_collapse_expand_container_action.py
```diff
@@ -0,0 +1,65 @@
...
+        with pytest.raises(ValueError, match="Please provide either the `collapse` or `expand` argument."):
+            model_manager["button"].actions = [collapse_expand_containers(id="test_action")]
+
+    def test_pre_build_invalid_collapse_expand(self):
+        # Add action to relevant component
+        model_manager["button"].actions = [collapse_expand_containers(id="test_action", collapse=["test_page"])]
+        action = model_manager["test_action"]
+
+        with pytest.raises(ValueError, match=r"Invalid component IDs found: \['test_page'\]"):
+            action.pre_build()
```
petar-qb:
```suggestion
        model_manager["button"].actions = [collapse_expand_containers(id="test_action", collapse=["invalid_target_id"])]
        action = model_manager["test_action"]

        with pytest.raises(ValueError, match=r"Invalid component IDs found: \['invalid_target_id'\]"):
            action.pre_build()
```

vizro-core/tests/unit/vizro/actions/test_collapse_expand_container_action.py
```diff
@@ -0,0 +1,65 @@
+import pytest
+
+from vizro.actions import collapse_expand_containers
+from vizro.managers import model_manager
+
+
+@pytest.mark.usefixtures("managers_one_page_two_containers")
+class TestCollapseExpandContainersActionPreBuild:
```
petar-qb, line 58: Let's create the new `class TestCollapseExpandContainersActionInstantiation` and model_validator tests there.

petar-qb (approved):
Really great work @nadijagraca! 

Excited to see how this PR becomes the launchpad for even more actions ahead. 🚀

vizro-core/src/vizro/actions/_collapse_expand_container.py
```diff
@@ -0,0 +1,62 @@
...
+    Args:
+        collapse (list[ModelID]): List of collapsible container ids to collapse. Defaults to `[]`.
+        expand (list[ModelID]): List of collapsible container ids to expand. Defaults to `[]`.
+    """
+
+    type: Literal["collapse_expand_containers"] = "collapse_expand_containers"
+    collapse: list[ModelID] = Field(default=[], description="List of collapsible container ids to collapse.")
+    expand: list[ModelID] = Field(default=[], description="List of collapsible container ids to expand.")
+    #  toggle would need multiple state as input. Maybe possible in future but not now. Add new ticket for this
+    # toggle: list[ModelID]
```
petar-qb:
```suggestion
    # TODO-AV2 D 5: toggle would need multiple state as input. Make that built-in actions can deal with arbitrary number of inputs.
    # toggle: list[ModelID]
```

vizro-core/src/vizro/actions/_collapse_expand_container.py
nadijagraca: Without the ignore you get `Cannot override writeable attribute with read-only property` error. 
I don't think its a real issue, but rather a mypy specific issue when dealing with combination of `@property` and `@abc.abstractmethod` decorators.

vizro-core/changelog.d/20250819_170759_nadija_ratkusic_graca_collapse_expand_containers_action.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Enabled `collapse_expand_containers` action to control collapsible containers by specifying lists of container ids to collapse or expand. ([#1347](https://github.com/mckinsey/vizro/pull/1347))
```
antonymilne:
```suggestion
- Add `collapse_expand_containers` action to collapse and expand collapsible containers. ([#1347](https://github.com/mckinsey/vizro/pull/1347))
```

vizro-core/src/vizro/actions/_collapse_expand_container.py
```diff
@@ -0,0 +1,58 @@
...
+        collapse (list[ModelID]): List of collapsible container ids to collapse. Defaults to `[]`.
+        expand (list[ModelID]): List of collapsible container ids to expand. Defaults to `[]`.
+    """
+
+    type: Literal["collapse_expand_containers"] = "collapse_expand_containers"
+    collapse: list[ModelID] = Field(default=[], description="List of collapsible container ids to collapse.")
+    expand: list[ModelID] = Field(default=[], description="List of collapsible container ids to expand.")
+    # TODO-AV2 D 5: toggle would need multiple state as input. Make that built-in actions can deal with arbitrary
+    #  number of inputs.
+    # toggle: list[ModelID]
```
antonymilne, line 24:
```suggestion
```

I made a whole new ticket for this now so no need to keep the comment here.

vizro-core/src/vizro/actions/_collapse_expand_container.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @model_validator(mode="after")
+    def validate_collapse_and_expand(self):
+        if not self.collapse and not self.expand:
+            raise ValueError("Either the `collapse` or `expand` list must contain at least a single element.")
+        if set(self.collapse) & set(self.expand):
+            raise ValueError("Collapse and expand lists cannot contain the same elements!")
```
antonymilne:
```suggestion
        if (overlap := set(self.collapse) & set(self.expand)):
            raise ValueError(f"`collapse` and `expand` cannot both contain the same IDs {overlap}.")
```

Just to be a bit more explicit with the error message and identify the duplicated IDs.

vizro-core/src/vizro/actions/_collapse_expand_container.py
```diff
@@ -0,0 +1,58 @@
...
+    def pre_build(self):
+        from vizro.models import Container
+
+        page_collapsible_container_ids = [
+            model.id
+            for model in model_manager._get_models(Container, root_model=model_manager._get_model_page(self))
+            if model.collapsed is not None
+        ]
+
+        invalid_ids = set(self.collapse + self.expand) - set(page_collapsible_container_ids)
```
antonymilne:
```suggestion
        invalid_ids = set(self.collapse) | set(self.expand) - page_collapsible_container_ids
```

vizro-core/src/vizro/actions/_collapse_expand_container.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @_log_call
+    def pre_build(self):
+        from vizro.models import Container
+
+        page_collapsible_container_ids = [
+            model.id
+            for model in model_manager._get_models(Container, root_model=model_manager._get_model_page(self))
+            if model.collapsed is not None
+        ]
```
antonymilne:
```suggestion
        page_collapsible_container_ids = {
            model.id
            for model in model_manager._get_models(Container, root_model=model_manager._get_model_page(self))
            if model.collapsed is not None
        }
```

vizro-core/src/vizro/actions/_collapse_expand_container.py
```diff
@@ -0,0 +1,58 @@
...
+
+        if invalid_ids:
+            raise ValueError(
+                f"Invalid component IDs found: {invalid_ids}."
+                " Action's collapse and expand IDs must be collapsible containers on the same page as the action."
+            )
+
+    def function(self) -> dict[ModelID, Any]:
+        """Collapse or expand containers on page."""
+        return dict.fromkeys(self.collapse, True) | dict.fromkeys(self.expand, False)
```
antonymilne, line 54: Surely this is the wrong way round? Note the output property is `is_open`, which should presumably be `True` in the case that we're _expanding_?

vizro-core/src/vizro/actions/_collapse_expand_container.py
antonymilne: Yes indeed. We actually have a ticket to see if there's a better way of handling this but it's not very important.

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -100,7 +100,7 @@ def make_actions_chain(self):
     # We need to delete the old action models from the model manager so they don't get built. After that,
     # built in actions are always handled in the new way.
     for action in self.actions:
-        if isinstance(action.function, (export_data, filter_interaction)):
+        if isinstance(action.function, (collapse_expand_containers, export_data, filter_interaction)):
```
antonymilne, line 103: No need to add it here for the same reason.

vizro-core/src/vizro/models/types.py
```diff
@@ -118,9 +118,9 @@ def validate_captured_callable(cls, value: Any, info: ValidationInfo):
     """Reusable validator for the `figure` argument of Figure like models."""
     # Bypass validation so that legacy vm.Action(function=filter_interaction(...)) and
     # vm.Action(function=export_data(...)) work.
-    from vizro.actions import export_data, filter_interaction
+    from vizro.actions import collapse_expand_containers, export_data, filter_interaction
 
-    if isinstance(value, (export_data, filter_interaction)):
+    if isinstance(value, (collapse_expand_containers, export_data, filter_interaction)):
```
antonymilne: No need to add it here for the same reason.

vizro-core/src/vizro/models/types.py
```diff
@@ -394,11 +394,11 @@ def _check_type(
         cls, captured_callable: CapturedCallable, json_schema_extra: JsonSchemaExtraType
     ) -> CapturedCallable:
         """Checks captured_callable is right type and mode."""
-        from vizro.actions import export_data, filter_interaction
+        from vizro.actions import collapse_expand_containers, export_data, filter_interaction
 
         # Bypass validation so that legacy {"function": {"_target_": "filter_interaction"}} and
         # {"function": {"_target_": "export_data"}} work.
-        if isinstance(captured_callable, (export_data, filter_interaction)):
+        if isinstance(captured_callable, (collapse_expand_containers, export_data, filter_interaction)):
```
antonymilne: No need to add it here for the same reason.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -421,7 +421,7 @@ class Action(_BaseAction):
     # now specifies a user defined action in YAML (ok if not possible initially since it's not already) - could just
     # enable class-based one? Presumably import_path is no longer relevant though.
     function: Annotated[  # type: ignore[misc, assignment]
-        SkipJsonSchema[Union[CapturedCallable, export_data, filter_interaction]],
+        SkipJsonSchema[Union[CapturedCallable, collapse_expand_containers, export_data, filter_interaction]],
```
antonymilne: No need to add it here - this is only needed for legacy actions that existed before the new `_AbstractAction` system. We never expect a user to do `actions=vm.Action(function=collapse_expand_containers)` - they would always be using the new system `actions=collapse_expand_containers()` from the start.

vizro-core/tests/unit/vizro/actions/conftest.py
```diff
@@ -103,3 +103,32 @@ def managers_one_page_one_graph_with_dict_param_input(scatter_matrix_chart):
         ],
     )
     Vizro._pre_build()
+
```
antonymilne: If this is only used in the `collapse_expand_containers` test then let's move it to that file rather than putting it in conftest.

vizro-core/tests/unit/vizro/actions/test_collapse_expand_container_action.py
```diff
@@ -0,0 +1,75 @@
+import pytest
+
+from vizro.actions import collapse_expand_containers
+from vizro.managers import model_manager
+
+
+@pytest.mark.usefixtures("managers_one_page_two_containers")
+class TestCollapseExpandContainersActionInstantiation:
+    def test_invalid_duplicate_collapse_expand(self):
+        with pytest.raises(ValueError, match="Collapse and expand lists cannot contain the same elements!"):
```
antonymilne: This would mention the explicit ID that's duplicated with the suggestion I made.

antonymilne (commented):
Nice work! This looks generally good other than the fact that I think the collapse/expand is actually the wrong way round - see my comment on that.

The one thing I wonder whether we can fix is that this doesn't work quite the same way as if I manually expand/collapse the container. If you take a look at the arrows then you'll see what I mean - we don't get the nice rotation animation, but more importantly it seems that the arrow also can end up being in the direction since it doesn't change at all when we expand/collapse with this action!

Let's talk about how we might be able to solve this 🙂 I wonder whether we need to rework out clientside callback a bit of target something other than `is_open` here instead (like `n_clicks` to simulate actually clicking on the container?). Edit: looking at it quickly, we might actually need to wait until we've completed https://github.com/McK-Internal/vizro-internal/issues/2114 before we can do this correctly unless we change how our current clientside callback works 😬 

btw don't worry about docs for now - I have an idea for all these new built-in actions, so don't write anything for it yet.

vizro-core/src/vizro/actions/_collapse_expand_container.py
nadijagraca: Hmm I'm confused now.
`collapsed=True` means the container's `is_open=False` as we have `is_open=not self.collapsed,`. So if we want to collapse container we need to change its `is_open=True` to `is_open=False`. 
if I put True in expand, nothing happens.

petar-qb:
Blocked by https://github.com/McK-Internal/vizro-internal/issues/2108

PR #1350, opened 2025-08-20
[Feature] Add global progress indicator functionality

vladimir-mck:
## Description

Add global progress indicator functionality

The progress indicator has been added to the dashboard as an optional property that displays a spinning icon when actions are running.

**Changes:**
- Added `ProgressIndicator` component in `src/vizro/models/_components/`
- Added `show_progress_indicator` boolean property to `Dashboard` model (defaults to `False`)
- Progress indicator appears in the header next to the theme switcher when enabled

**Testing:**
To test the indicator, set `show_progress_indicator=True` in your Dashboard configuration, open browser console → Network tab, and enable throttling to simulate slow network conditions.

```python
dashboard = vm.Dashboard(
    pages=[...],
    show_progress_indicator=True
)```


## Screenshot

<img width="176" height="61" alt="Screenshot 2025-08-20 at 15 57 48" src="https://github.com/user-attachments/assets/3e772404-fcd4-4172-a12e-503fdce7aa8b" />


## Notice

### CODE REVIEW

File: vizro-core/src/vizro/models/_dashboard.py [hunk 1]
```diff
@@ -108,12 +110,18 @@ class Dashboard(VizroBaseModel):
 
     pages: list[Page]
     theme: Literal["vizro_dark", "vizro_light"] = Field(
-        default="vizro_dark", description="Theme to be applied across dashboard. Defaults to `vizro_dark`."
+        default="vizro_dark",
```
maxschulz-COL, line 111: Just curious, how does this formatting come about? Do you have different local linter settings?

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -127,6 +135,10 @@ class Dashboard(VizroBaseModel):
             tags. Defaults to `None`.""",
         ),
     ]
+    show_progress_indicator: bool = Field(
+        default=False,
```
maxschulz-COL, line 137: Why `False`, is that not always super useful?

vizro-core/src/vizro/models/_dashboard.py
maxschulz-COL: Because it makes it super hard to see what has actually changed 😂

huong-li-nguyen (commented):
So cool to see this working! Great job figuring this out @vladimir-mck and @petar-qb 🚀 I haven't reviewed the code yet since I'll wait for the merge conflicts to be resolved first, but my initial thoughts are:

- For now, I'd keep it non-configurable - basically always on, without an option to turn it off (similar to the theme switch). We're still figuring out how to make the theme-switch configurable, so things might change. So I'd suggest leaving it as always-on for now and removing the argument to turn it off.
- Could we add a loading delay (e.g. only show the spinner if the action takes longer than ~300ms or another reasonable threshold)? Right now it flashes very briefly on every page load, which feels more like visual noise than useful feedback 😬

vizro-core/src/vizro/models/_dashboard.py
vladimir-mck: It's the Black Formatter extension on VSCode.

maxschulz-COL: Could you change you line-width to 120 pls 🙏 ?

maxschulz-COL: This PR has 72 changed file because of it haha

maxschulz-COL: (btw we us `ruff` and not `black` so might be good to keep that in sync - there is a ruff extension for VS Code)

vladimir-mck: Thanks, I'll install ruff now. 

Now I have 23 changes after merging... I'll reopen a new PR

petar-qb:
Thanks @huong-li-nguyen 

> For now, I'd keep it non-configurable - basically always on, without an option to turn it off (similar to the theme switch). We're still figuring out how to make the theme-switch configurable, so things might change. So I'd suggest leaving it as always-on for now and removing the argument to turn it off.

A totally agree! I would also hardcode it and then make it as opt-out if we find it useful for our users. 

> Could we add a loading delay (e.g. only show the spinner if the action takes longer than ~300ms or another reasonable threshold)? Right now it flashes very briefly on every page load, which feels more like visual noise than useful feedback 😬

If we want to enable that, then the implementation should be done as the second approach from [this comment](https://github.com/McK-Internal/vizro-internal/issues/1399#issuecomment-3183852965). 

I while ago we had similar conversation about setting the `delay_show` and `delay_hide` to the `dcc.Loading` that wraps our Graph, AgGrid, Table, Figure, and dynamic Filter. We eventually decided not to add these artificial dcc.loading features as we don't exactly know how long it would take the server to respond in different users dashboards and their setups. For example, if we set `delay_show=300ms` and the server answers in `400ms`, the spinner will blink (for ~100ms). In that case it's better not to have the `delay_show=300ms` set, so that the spinner spins for natural `~400ms`.

I find `delay` properties useful when the specific dashboard application is built (in vizro or in dash), as in that case the dashboard creator knows exactly the processes (which one is slow/fast). However, we don't know how long it will take different action to finish for different users. 

Now, I'm thinking about some combination of `delay_show` and `delay_hide`. Maybe by setting `300ms` on both which would result that loading spinner won't spin if the action takes less than `300ms`, but if it takes more than 300ms, then the spinner will spin for at least `300ms` (even if the action has finished earlier and even if the figures are already loaded).

What do you think guys, would it be worth checking?

vladimir-mck:
Closing this and reopening a new [PR](https://github.com/mckinsey/vizro/pull/1352) for better visibility

PR #1352, opened 2025-08-21
[Feat] Add global progress indicator functionality

vladimir-mck:
closes https://github.com/McK-Internal/vizro-internal/issues/1399

## Description

Add global progress indicator functionality

The progress indicator has been added to the dashboard that displays a spinning icon when actions are running.

**Changes:**
- Progress indicator appears in the dashboard or page header next to the theme switcher when action is running.

#### Progress indicator in the dashboard header
https://github.com/user-attachments/assets/e80631dd-5757-403f-8ce7-eced7c7dec25

#### Progress indicator in the page header
https://github.com/user-attachments/assets/92ee444b-b8b9-4af8-8eba-f633d77e1bd4

## Notice


maxschulz-COL:
Much better haha 👍

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -127,6 +127,10 @@ class Dashboard(VizroBaseModel):
             tags. Defaults to `None`.""",
         ),
     ]
+    show_progress_indicator: bool = Field(
```
vladimir-mck, line 130: @maxschulz-COL I changed it to True but I'm considering to remove this options. WDYT?

petar-qb:
Moving the comment from the duplicate and closed PR so to continue discussing it here:

---
 
Thanks @huong-li-nguyen 

> For now, I'd keep it non-configurable - basically always on, without an option to turn it off (similar to the theme switch). We're still figuring out how to make the theme-switch configurable, so things might change. So I'd suggest leaving it as always-on for now and removing the argument to turn it off.

A totally agree! I would also hardcode it and then make it as opt-out if we find it useful for our users. 

> Could we add a loading delay (e.g. only show the spinner if the action takes longer than ~300ms or another reasonable threshold)? Right now it flashes very briefly on every page load, which feels more like visual noise than useful feedback 😬

If we want to enable that, then the implementation should be done as the second approach from [this comment](https://github.com/McK-Internal/vizro-internal/issues/1399#issuecomment-3183852965). 

I while ago we had similar conversation about setting the `delay_show` and `delay_hide` to the `dcc.Loading` that wraps our Graph, AgGrid, Table, Figure, and dynamic Filter. We eventually decided not to add these artificial dcc.loading features as we don't exactly know how long it would take the server to respond in different users dashboards and their setups. For example, if we set `delay_show=300ms` and the server answers in `400ms`, the spinner will blink (for ~100ms). In that case it's better not to have the `delay_show=300ms` set, so that the spinner spins for natural `~400ms`.

I find `delay` properties useful when the specific dashboard application is built (in vizro or in dash), as in that case the dashboard creator knows exactly the processes (which one is slow/fast). However, we don't know how long it will take different action to finish for different users. 

Now, I'm thinking about some combination of `delay_show` and `delay_hide`. Maybe by setting `300ms` on both which would result that loading spinner won't spin if the action takes less than `300ms`, but if it takes more than 300ms, then the spinner will spin for at least `300ms` (even if the action has finished earlier and even if the figures are already loaded).

What do you think guys, would it be worth checking?

huong-li-nguyen:
> Now, I'm thinking about some combination of `delay_show` and `delay_hide`. Maybe by setting `300ms` on both which would result that loading spinner won't spin if the action takes less than `300ms`, but if it takes more than 300ms, then the spinner will spin for at least `300ms` (even if the action has finished earlier and even if the figures are already loaded).
> 
> What do you think guys, would it be worth checking?

Yes, I like that approach 🌟 I think your suggestion with delay_show=300ms and delay_hide=300ms could work and is definitely worth trying. We'd avoid flicker on fast actions while also ensuring that, once the spinner does appear, it stays visible for at least 300ms.

This would mean:

- Actions <300ms → no spinner (no flicker).
- Actions around 400ms → spinner shows for ~300ms (instead of a blink) 
- Actions >600ms → spinner shows naturally for the full duration.

The 300ms are currently made up, would be great if you could check what makes sense @vladimir-mck :)

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -279,6 +283,21 @@ def _inner_page(self, page: Page) -> _InnerPageContentType:
...
+            className="material-symbols-outlined progress_indicator",
+            children="forward_media",
+            id="global-progress-indicator",
+        )
+
+        # Only include progress indicator if enabled
+        header_controls_children = []
+        if self.show_progress_indicator:
+            header_controls_children.append(progress_indicator)
+        header_controls_children.append(settings)
```
huong-li-nguyen: This is actually what the `settings` container is for! I like your naming of `header_controls` more though, so I would refactor this part of the code to only have header_controls and not the settings container anymore. Double-check the CSS files and docs of any mentions of settings then, and just replace with header-controls :) So you can just add the `dbc.Switch` directly to the header_controls_children.

huong-li-nguyen (approved):
This looks good to me 👍  Thanks for incorporating the feedback :)

petar-qb:
This PR looks good to me, but since I also contributed some changes, it wouldn’t make much sense for me to approve it.
Would one of you be able to take a look instead? @nadijagraca, @maxschulz-COL, @antonymilne

huong-li-nguyen:
> This PR looks good to me, but since I also contributed some changes, it wouldn’t make much sense for me to approve it. Would one of you be able to take a look instead? @nadijagraca, @maxschulz-COL, @antonymilne

I think we can also merge just with our two approvals @petar-qb . It's such a visual minor change, that doesn't interfere with anything. If the others have time at some point, they can take a look and we can make changes in a separate PR. But for the time being, I think it's fine if we merge if you are also fine with the current changes 👍

PR #1357, opened 2025-08-21
[Docs] New tutorial on custom actions

antonymilne:
## Description

**Big things to look at:**
1. [New tutorial on custom actions](https://vizro--1357.org.readthedocs.build/en/1357/pages/tutorials/custom-actions-tutorial/) is finished - reviewed by @huong-li-nguyen ✅ 
2. Completely re-written user guides on [custom actions](https://vizro--1357.org.readthedocs.build/en/1357/pages/user-guides/custom-actions/) and [built-in actions](https://vizro--1357.org.readthedocs.build/en/1357/pages/user-guides/actions/). ✅ reviewed by @maxschulz-COL @petar-qb 
3. [New explanation of actions](https://vizro--1357.org.readthedocs.build/en/1357/pages/explanation/actions-explanation/). reviewed by @huong-li-nguyen ✅ 
4. Rewritten some of [custom components](https://vizro--1357.org.readthedocs.build/en/1357/pages/user-guides/custom-components/). Needs a review and screenshots from @petar-qb ✅ 
5. Changelog now contains many features we've added over the last few months but hadn't yet documented

All the examples here are still quite "toy" ones. We need some more real world ones with drill-through, "configure and run model" etc. still. These will come in due course but not this PR.

## Notes/things to add to dev guidance/cover in team learning

For all our "extensions" I think our terminology should be:
* "built-in" - not pre-defined
* "custom" ("user-defined" also ok) or just no word at all - like "write your own action" implies it's not built-in.

Other bits:
* `vizro.actions` is `va`
* New narrative cross-links
* Deprecation and experimental appears in API 

## Notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

    - I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
    - I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
    - I certify that the use of this contribution as authorized by the Apache 2.0 license does not violate the intellectual property rights of anyone else.
    - I have not referenced individuals, products or companies in any commits, directly or indirectly.
    - I have not added data or restricted code in any commits, directly or indirectly.

PR #1366, opened 2025-08-29
[Feat] Add trigger/outputs/inputs action properties to filter/parameter and add _trigger as action's builtin argument

petar-qb:
closes [#2082](https://github.com/McK-Internal/vizro-internal/issues/2082)
closes [#2119](https://github.com/McK-Internal/vizro-internal/issues/2119)

useful reference - [PR](https://github.com/mckinsey/vizro/pull/1369) merged into this branch

## Description
This PR adds `_action_trigger`, `_action_outputs` and `_action_inputs` to `vm.Filter` and `vm.Parameter`.

It enables Vizro users to input/output control selector property in an action without specifying selector's ID.
Shortcut example:
- `"filter-id.options"` is transformed to `"selector-checklist.options"`,
- `"parameter-id"` is transformed to `"selector-range_slider_id.value"`,

Also the `_trigger` built-in keyword argument is added in actions. This should give a Dash `State` that corresponds to the trigger of the first action in the actions chain. Ultimate aim here is that we can do `update_control` action without needing to specify the source figure id.

## For Reviewers:
Open the scratch app.py and take a look into the homepage (Table of contents, and then explore example pages under  the `**NEW** Controls` and `**NEW** _trigger` nav accordion.

## Notice


vizro-core/examples/scratch_dev/app.py
petar-qb, line 1: @antonymilne  Do you think we should add e2e tests for model action properties? If so, we could reuse most of this dashboard.

vizro-core/src/vizro/actions/_on_page_load.py
```diff
@@ -34,4 +36,7 @@ def function(self, _controls: _Controls) -> dict[ModelID, Any]:
 
     @property
     def outputs(self):  # type: ignore[override]
-        return {target: target for target in self.targets}
+        return {
+            target: f"{target}.selector" if isinstance(model_manager[target], vm.Filter) else target
+            for target in self.targets
+        }
```
petar-qb, line 43: Currently, only `on-page-load` and `data-frame parameter` actions can output `vm.Filter`. All other actions return figures.
P.S. `vm.Parameter` as action's output isn’t supported yet.

vizro-core/src/vizro/models/_controls/parameter.py
```diff
@@ -99,6 +130,14 @@ def pre_build(self):
...
         self._set_actions()
 
+        # A set of properties unique to selector (inner object) that are not present in html.Div (outer build wrapper).
+        # Creates _action_outputs and _action_inputs for accessing selector's properties via the outer vm.Parameter ID.
+        # Example: "parameter-id.options" is transformed to "checklist.options".
+        if selector_inner_component_class := getattr(self.selector, "_inner_component_class", None):
+            self._selector_properties = set(selector_inner_component_class().available_properties) - set(
+                html.Div().available_properties
+            )
+
```
petar-qb: File: `parameter.py`

@antonymilne Should we move this chunk of code somewhere else since a parameter’s selector is known during the initialisation? This is not the case for the Filter as its selector can be created in `filter.pre_build`.

vizro-core/src/vizro/models/_components/form/checklist.py
```diff
@@ -83,6 +83,7 @@ class Checklist(VizroBaseModel):
 
     _dynamic: bool = PrivateAttr(False)
     _in_container: bool = PrivateAttr(False)
+    _inner_component_class: DashComponentClass = PrivateAttr(dbc.Checklist)
```
petar-qb: There were three approaches for solving this:
1. `_inner_component: Component = PrivateAttr(dbc.Checklist())`
    - [Used only in filter/parameter when fetching its available properties.
    - [Hard to maintain as introduces the duplication (if underlying component class is changed).
2.  `_inner_component_class: DashComponentClass = PrivateAttr(dcc.Dropdown)`
    - Used in filter/parameter.
    - Used in models build/call methods.
    - Less readable, easier to maintain as reduces the duplication.
3.  `_inner_component: list[str] = PrivateAttr(dbc.RadioItems().available_properties)`
    - Used in filter/parameter.
    - More readable, harder to maintain as introduces the duplication.

The first option falls somewhere between the second and third options. I’m not a fan of it because, even though it’s only used in one place (filter/parameter), it’s not fully prepared. Currently, it’s only used to extract available properties, but the properties themselves (as in the third option) aren’t prepared in advance. So in practice, the first option feels like a less complete version of the third option.

The choice between the second and third options mostly comes down to code style (readability versus maintainability). I’ve evaluated it from the maintainability perspective, which is why I lean toward the current version.

@antonymilne, if you prefer a different approach, I can adjust it easily. This version reflects my slight preference.

vizro-core/changelog.d/20250829_112650_petar_pejovic_controls_action_properties.md
petar-qb, line 1: Should I announce it in the changelog file or it will be added as a part of the following PR? -> https://github.com/mckinsey/vizro/pull/1357/files#diff-f8063a1167dd2e1e105917b07fca95e3d7c0ad56ae78bed7a398885443fb2662R7-R20

vizro-core/src/vizro/models/_controls/_controls_utils.py
```diff
@@ -33,6 +33,8 @@ def _get_control_parent(control: ControlType) -> Optional[VizroBaseModel]:
 
     # Otherwise, return the Container that contains the control.
     page_containers = model_manager._get_models(model_type=Container, root_model=page)
+
+    # TODO: Fix the StopIteration Exception that happen when a control is added within the Page.components
```
petar-qb: The following code raises the StopIteration exception:
```py
import vizro.models as vm
from vizro import Vizro

df = px.data.iris()

vm.Page.add_type("components", vm.Filter)

page = vm.Page(
    title="Test Page",
    components=[
        vm.Filter(column="species"),
        vm.Graph(figure=px.scatter(df, x="sepal_width", y="sepal_length")),
    ],
)

if __name__ == "__main__":
```

@maxschulz-COL @antonymilne 

Should we:
- Allow users to add a `vm.Filter` directly to page components and fix this bug, or
- Forbid it (as it can be added only within the Page/Container controls) and provide a clearer exception message?

Also, do we want to address this separately, or include it as part of the new `model_manager` PR? (See the TODO above the function.)

vizro-core/src/vizro/models/_components/form/checklist.py
antonymilne: I haven't reviewed this property yet but just have a quick thought off the top of my head, it might not make any sense or be any better than the above alternatives...

What about if we do `self.selector.build()[selector.id]` to get the actual object? It means a kind of unnecessary `build` operation but that's probably ok. It's guaranteed to not get out of sync with the underlying component. Would it still work with dynamic filters?

petar-qb: That was my initial try, but many `"Duplication callback outputs"` runtime exceptions discouraged me to continue with this approach. The exceptions occurred due to multiple creation of client-side callbacks in selectors' build method.

antonymilne: Ahhh hmmm yes I see. If we didn't create the clientside callbacks in the build methods then would this otherwise be a good approach do you think? Maybe just comment out the clientside callbacks for now to see if there's any other problems with it.

petar-qb: It works with `self.selector.build()[selector.id]` approach when clientside callbacks in `dropdown`, `checklist`, `slider` and `range_slider` are commented out. What do you think I should do next?

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -118,10 +119,41 @@ def check_id_set_for_url_control(self):
...
+            "__default__": f"{selector.id}.value",
+            "selector": f"{self.id}.children",
+            **({"title": f"{selector.id}_title.children"} if selector.title else {}),
+            **({"description": f"{selector.description.id}-text.children"} if selector.description else {}),
+            **(
+                {selector_prop: f"{selector.id}.{selector_prop}" for selector_prop in self._selector_properties}
+                if self._selector_properties
+                else {}
+            ),
+        }
```
antonymilne: Now that I look at this again i think we could achieve the same more easily with this? Similarly for `Parameter`.

```suggestion
        return {
            "selector": f"{self.id}.children",
             **self.selector._actions_outputs,
              **(
                {selector_prop: f"{selector.id}.{selector_prop}" for selector_prop in self._selector_properties}
                if self._selector_properties
                else {}
            ),
        }
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -118,10 +119,41 @@ def check_id_set_for_url_control(self):
...
+
+    @property
+    def _action_triggers(self) -> dict[str, _IdProperty]:
+        # Note this relies on the fact that filters are pre-built upfront in Vizro._pre_build. Otherwise,
+        # control.selector might not be set. Cast is justified as the selector is set in pre_build and is not None.
+        selector = cast(SelectorType, self.selector)
+        return {"__default__": f"{selector.id}.value"}
```
antonymilne:
```suggestion
        return self.selector._action_triggers
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -118,10 +119,41 @@ def check_id_set_for_url_control(self):
...
+
+    @property
+    def _action_inputs(self) -> dict[str, _IdProperty]:
+        # Note this relies on the fact that filters are pre-built upfront in Vizro._pre_build. Otherwise,
+        # control.selector might not be set. Cast is justified as the selector is set in pre_build and is not None.
+        selector = cast(SelectorType, self.selector)
+        return {
+            "__default__": f"{selector.id}.value",
```
antonymilne:
```suggestion
            **self.selector._action_inputs,
```

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -249,6 +281,14 @@ def pre_build(self):
                 ),
             ]
 
+        # A set of properties unique to selector (inner object) that are not present in html.Div (outer build wrapper).
+        # Creates _action_outputs and _action_inputs for accessing inner selector properties via the outer vm.Filter ID.
+        # Example: "filter-id.options" is transformed to "checklist.options".
```
antonymilne:
```suggestion
        # Creates _action_outputs and _action_inputs for forwarding properties to the underlying selector.
        # Example: "filter-id.options" is forwarded to "checklist.options".
```

vizro-core/src/vizro/models/_controls/parameter.py
antonymilne: I'd leave it here for simplicity to match `Filter` but with a comment that it doesn't really need to be here.

vizro-core/src/vizro/models/_controls/filter.py
antonymilne: The tests are good how they are though since they're nice and explicit, so don't change those!

vizro-core/src/vizro/actions/_on_page_load.py
antonymilne: It would be good to add a comment explaining this here and in the `_parameter_action`. Also would be good to say why we need to specify `selector` because the `__default__` would otherwise alter that `value`.

vizro-core/src/vizro/models/_components/form/checklist.py
antonymilne: I'm not a huge fan of it (maybe slight preference for option 3 over 2 personally), but it's ok! Let's keep as you have it for now and make a new ticket to see if we can move to moving the clientside callbacks and doing `self.selector.build()[selector.id]` in future.

vizro-core/src/vizro/models/types.py
```diff
@@ -623,6 +624,10 @@ def wrapped(*args, **kwargs) -> CapturedCallable:
         )
 
 
+# Used for type annotations in form components.
+DashComponentClass: TypeAlias = type[Component]
```
```suggestion
_DashComponentClass: TypeAlias = type[Component]
```

This should definitely be private, and please check this doesn't show in the API docs.

With that said, I don't really care if it exists at all. If it's possible to do this without the type hint at all then we can just do that:
```
_inner_component_class = PrivateAttr(dcc.Dropdown)
```

vizro-core/src/vizro/models/_controls/_controls_utils.py
antonymilne: Just to check, this isn't a new bug introduced by this PR, right?

It's very minor and I think we shouldn't do anything about it. Even if it's a weird thing to do, I don't think we should explicitly ban adding `Filter` to `Page.components`. Probably this will fix itself as part of the model manager rework. I wouldn't even put in the `TODO` to fix it since it's so minor, and if it remains as a bug in the future and someone finds it then we can just worry about it then.

vizro-core/examples/scratch_dev/app.py
antonymilne: I'd say yes but it's low priority, so as and when @l0uden has capacity for it.

vizro-core/src/vizro/models/_controls/_controls_utils.py
petar-qb: > Just to check, this isn't a new bug introduced by this PR, right?

Correct. I was able to reproduce it on main as well.

> Probably this will fix itself as part of the model manager rework.

Exactly 👍 that’s why I tagged Max in the comment above.

I’ll go ahead and remove the TODO. Let’s leave it as is for now and see if anyone else runs into the bug.

vizro-core/examples/scratch_dev/app.py
petar-qb: Added on the backlog as the "Medium" priority ticket. -> https://github.com/McK-Internal/vizro-internal/issues/2140

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: Works like a charm. I like it 👍 👍

vizro-core/changelog.d/20250829_112650_petar_pejovic_controls_action_properties.md
antonymilne: I'll add it to my PR.

vizro-core/src/vizro/models/types.py
petar-qb: The `mypy` does not complain about `_inner_component_class = PrivateAttr(dcc.Dropdown)`. 👍

vizro-core/src/vizro/models/_components/form/checklist.py
petar-qb: Here's the ticket -> https://github.com/McK-Internal/vizro-internal/issues/2142

petar-qb: > I'm not a huge fan of it (maybe slight preference for option 3 over 2 personally),

After taking a break and coming back with fresh eyes after several days, I realised I actually prefer the third option too, so I implemented it. 😄

vizro-core/changelog.d/20250829_112650_petar_pejovic_controls_action_properties.md
petar-qb: Then let's add the explanation for the `_trigger` argument too

vizro-core/changelog.d/20250829_112650_petar_pejovic_controls_action_properties.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
```
huong-li-nguyen, line 22: Is this worth mentioning here? Or will it be done by @antonymilne PR?

vizro-core/changelog.d/20250901_091405_petar_pejovic_add_action_trigger_builtin_argument.md
```diff
@@ -0,0 +1,48 @@
+<!--
```
huong-li-nguyen: I guess you can remove one of the changelog files?

vizro-core/src/vizro/actions/_on_page_load.py
```diff
@@ -34,4 +36,8 @@ def function(self, _controls: _Controls) -> dict[ModelID, Any]:
 
     @property
     def outputs(self):  # type: ignore[override]
-        return {target: target for target in self.targets}
+        # Add ".selector" for vm.Filter outputs as filter's __default__ action output would alter the selector value.
```
huong-li-nguyen: Maybe rephrase this comment - I initially found it hard to understand and had to read through your conversation to understand why it was here. Maybe something like:

"Additional logic for vm.Filter added, as otherwise the filter's __default__ action output would alter the selector value"

vizro-core/src/vizro/actions/_parameter_action.py
```diff
@@ -42,4 +44,8 @@ def function(self, _controls: _Controls) -> dict[ModelID, Any]:
 
     @property
     def outputs(self):  # type: ignore[override]
-        return {target: target for target in self._target_ids}
+        # Add ".selector" for vm.Filter outputs as filter's __default__ action output would alter the selector value.
```
huong-li-nguyen: Same comment on rephrasing here

vizro-core/src/vizro/models/_components/form/switch.py
```diff
@@ -66,7 +66,7 @@ class Switch(VizroBaseModel):
     ]
 
     _dynamic: bool = PrivateAttr(False)
-    _in_container: bool = PrivateAttr(False)
```
huong-li-nguyen, line 69: Why do we remove the _in_container property? We didn't do so e.g. for the radio_items

vizro-core/changelog.d/20250829_112650_petar_pejovic_controls_action_properties.md
petar-qb: It will be done in Antony's PR.

vizro-core/src/vizro/models/_components/form/switch.py
petar-qb: `_in_container` is used in `dropdown`, `checklist`, and `radio_items` for inline layouts and calculating dropdown height. It's irrelevant for the `switch` and other selectors (e.g.numerical, temporal). That line of code was added there by mistake so I removed it.

PR #1373, opened 2025-09-01
[Feat] Implement `set_control`

antonymilne:
Closes https://github.com/McK-Internal/vizro-internal/issues/1656
Closes https://github.com/McK-Internal/vizro-internal/issues/2081

## Also added and then removed - mechanism for experimental features

See https://github.com/mckinsey/vizro/pull/1373#issuecomment-3302760157. This was removed as the feature is not experimental but built-in properly. In future when we have experimental features we might want to use the decorator or the namespace, still not decided.

Old text: We now have a way to do experimental features! I opted not to put it in a namespace `experimental` but instead use decorator `experimental` (a similar approach is used in pandas among others). This raises a warning when a user uses the function and also appears highlighted in our docs as experimental. The functionality is basically copy and pasted from our existing `deprecated` decorator which does similarly (both the decorator itself and the docs extension that is added in #1357).

## Notice


vizro-core/changelog.d/20250901_210343_antony.milne_experimental_update_controls.md
petar-qb, line 1: Should we announce it as it's in the experimental phase? 
Should we announce the `AgGrid` default trigger change and changes made in `dash_ag_grid`?

vizro-core/src/vizro/_vizro.py
```diff
@@ -168,6 +168,9 @@ def _pre_build():
         # Any models that are created during the pre-build process *will not* themselves have pre_build run on them.
         # In future may add a second pre_build loop after the first one.
 
+        # for dashboard in cast(Iterable[Dashboard], model_manager._get_models(Dashboard)):
+        #     dashboard.pre_build()
+
```
petar-qb: TODO PP: 
```suggestion
```

vizro-core/src/vizro/_vizro_utils.py
petar-qb, line 1: @antonymilne there are many `mypy` errors in this file for the added section. Should we extract this section to the new file that begins with `# type: ignore`?

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -0,0 +1,159 @@
...
+I'm not sure what the right syntax would be for targeting multiple controls?
+
+Copied from https://github.com/McK-Internal/vizro-internal/issues/1939:
+It should be possible to target multiple controls in one click
+(think e.g. clicking on a 2D heatmap, see Max example in mckinsey/vizro#1301).
+So we need some way to do e.g. target="country_filter", lookup="country" as well as
+target="continent_filter", lookup="continent". This could be multiple set_control calls
+or better a single set_controls. In this case what would API be? How do you assign the
+right values to the right targets? Maybe a dictionary like {"country_filter": "country",
+"continent_filter": "continent"}?
```
petar-qb: @antonymilne 
I left many comments as I’m not sure what comments or what parts of them you want to leave in the code.

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -43,7 +43,8 @@ class Button(VizroBaseModel):
     actions: ActionsType = []
     variant: Literal["plain", "filled", "outlined"] = Field(
         default="filled",
-        description="Predefined styles to choose from. Options are `plain`, `filled` or `outlined`."
+        description="Predefined styles to choose from. Options are `"
+        "plain`, `filled` or `outlined`."
```
petar-qb: TODO PP:
```suggestion
        description="Predefined styles to choose from. Options are `plain`, `filled` or `outlined`."
```

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -95,6 +96,14 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
         }
 
+    # Not urgent to implement this but just for illustration.
+    # This is how we can return a static value - just ignore the trigger.
+    # so when do vm.Button(actions=set_control(value=3) it would send the value 3.
+    # We could also implement this on selector components.
+    # Then we can relax the restriction that target is a categorical control.
+    # def _get_value_from_trigger(self, action, trigger):
+    #     return action.value
+
```
petar-qb: This is one of the comments I'm not sure should we leave or remove.

vizro-core/tests/unit/vizro/models/_components/test_ag_grid.py
```diff
@@ -135,7 +135,7 @@ class MyAgGrid(vm.AgGrid):
     def test_ag_grid_trigger(self, ag_grid_with_id, identity_action_function):
         ag_grid = vm.AgGrid(figure=ag_grid_with_id, actions=[Action(function=identity_action_function())])
         [action] = ag_grid.actions
-        assert action._trigger == "underlying_ag_grid_id.cellClicked"
+        assert action._trigger == "underlying_ag_grid_id.selectedRows"
```
petar-qb, line 140: TODO PP: Add tests for `_action_triggers` in `TestModelInstantiation` across all models.

Currently, we only test `action._trigger` action's trigger attached on the AgGrid component. However, there are no equivalent tests for `Model._action_triggers` in the `TestModelInstantiation` classes (with the exception of Filter and Parameter). In contrast, both `_action_inputs` and `_action_outputs` are already covered consistently for all models.

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -0,0 +1,159 @@
...
+
+    type: Literal["set_control"] = "set_control"
+    target: ModelID = Field(
+        description="Filter or Parameter component id to be affected by the trigger."
+        "If the target is on a different page to the trigger then it must have `show_in_url=True`."
+    )
+    value: str
+
+    @_log_call
+    def pre_build(self):
```
petar-qb, line 94: All validation steps depend on either another model being instantiated or a pre-built Filter. So, none of these validations can be performed during the initialisation phase.

vizro-core/mkdocs.yml
```diff
@@ -212,9 +212,10 @@ plugins:
               warn_unknown_params: false
               returns_named_value: false
             extensions:
-              - docs/griffe_extensions.py:
+              - docs/griffe_extensions/dynamically_inspect.py:
                   paths:
                     [vizro.figures.kpi_card, vizro.figures.kpi_card_reference]
+              - docs/griffe_extensions/experimental_decorator/extension.py
```
antonymilne: FYI reviewers: this extension adds a nice note in our API docs for things decorated with `experimental`:

image

vizro-core/src/vizro/_vizro_utils.py
antonymilne: This is ironic given the code originally came from `typing_extensions` 😅  I've added a `@typing.no_type_check` to ignore just the `experimental` class in this file and mypy seems ok now.

vizro-core/tests/integration/test_examples.py
```diff
@@ -42,6 +42,7 @@ def dashboard(request, monkeypatch):
 # Ignore as it doesn't affect the test run
 @pytest.mark.filterwarnings("ignore::pytest.PytestUnhandledThreadExceptionWarning")
 @pytest.mark.filterwarnings("ignore:unclosed file:ResourceWarning")
+@pytest.mark.filterwarnings("ignore:.*experimental.*:FutureWarning")
```
antonymilne: We'll presumably need to ignore this globally rather than just in the integration tests?

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -122,7 +122,7 @@ def make_actions_chain(self):
 
         # Temporary hack to help with lookups in filter_interaction. Should not be required in future with reworking of
```
antonymilne:
```suggestion
        # Temporary hack to help with lookups in filter_interaction and set_control. Should not be required in future with reworking of
```

vizro-core/src/vizro/actions/_set_control.py
antonymilne: Cool, thanks for leaving them! I've sorted through them all and moved to other tickets or notes.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -60,7 +60,7 @@ class _BaseAction(VizroBaseModel):
 
     # Temporary hack to help with lookups in filter_interaction. Should not be required in future with reworking of
     # model manager and removal of filter_interaction.
-    _parent_model_id: ModelID = PrivateAttr()
+    _parent_model: ModelID = PrivateAttr()
```
```suggestion
    _parent_model: VizroBaseModel = PrivateAttr()
```

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -126,10 +126,10 @@ def _get_filter_interaction_states(self) -> list[dict[str, State]]:
         page = model_manager._get_model_page(self)
 
         # States are stored in the parent model (e.g. AgGrid) whose actions contains the filter_interaction rather than
-        # the filter_interaction model itself, hence needing to lookup action._parent_model_id.
+        # the filter_interaction model itself, hence needing to lookup action._parent_model.
```
antonymilne, line 129:
```suggestion
        # the filter_interaction model itself, hence needing to lookup action._parent_model.
        # This is also needed to trigger the parent model's_get_value_from_trigger method in set_control.
        # After work on the model_manager we should be able to tidy this to directly get the parent model
        # from inside the action.
```

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -0,0 +1,159 @@
...
+    "but until then it may change or be removed without warning. If you have feedback on the feature then "
+    "[let us know](https://github.com/mckinsey/vizro/issues)."
+)
+class set_control(_AbstractAction):
+    """Sets the value of a control based on data from the trigger.
+
+    Args:
+        target (ModelID): Filter or Parameter component id to be affected by the trigger. If the target is on a
+            different page to the trigger then it must have `show_in_url=True`.
+        value (str): # TODO AM+PP: ADD DESCRIPTION
```
antonymilne: I'll do this in my docs PR so don't worry about it here.

antonymilne (commented):
Partial comments so far.

vizro-core/changelog.d/20250901_210343_antony.milne_experimental_update_controls.md
antonymilne: I'll do this in my docs PR, no need to worry about it here.

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -0,0 +1,112 @@
...
+from vizro.models.types import ControlType, ModelID
+
+
+# This defines what a model needs to implement for it to be capable of acting as the trigger of set_control.
+@runtime_checkable
+class _SupportsSetControl(Protocol):
+    def _get_value_from_trigger(self, value: str, trigger: JsonValue) -> JsonValue: ...
+
+
+@runtime_checkable
```
antonymilne: I don't think this will work since `isinstance` won't check the type hints themselves, just the existence of the attributes.

It doesn't seem to be used anywhere so I guess that's maybe what you found and it should not be deleted?

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -0,0 +1,112 @@
...
+    target: ModelID = Field(
+        description="Filter or Parameter component id to be affected by the trigger."
+        "If the target is on a different page to the trigger then it must have `show_in_url=True`."
+    )
+    value: str = Field(description="TODO AM: ADD DESCRIPTION")
+
+    @_log_call
+    def pre_build(self):
+        # Validate that action's parent model supports `set_control` action.
+        if not isinstance(model_manager[self._parent_model], _SupportsSetControl):
```
antonymilne: FYI @maxschulz-COL this is going to be a quite common pattern for actions (the child) to know what triggers it (the parent). It will be much nicer once we've gone back to the model manager and then we can get rid of this temporary `_parent_model` hack.

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -0,0 +1,112 @@
...
+            raise ValueError(
+                f"`set_control` action was added to the model with ID `{self._parent_model}`, but this action can only "
+                f"be used with models that support it (e.g. Graph, AgGrid)."
+            )
+
+        # Validate that action's target control exists in the dashboard.
+        target_model = cast(ControlType, model_manager[self.target]) if self.target in model_manager else None
+        target_model_page = model_manager._get_model_page(target_model) if target_model else None
```
antonymilne: FYI @maxschulz-COL this check of `target_model_page` is only here because of the current disparity between pages that are in model manager vs. pages that are actually used in the dashboard, so once that no longer exists we can remove this check.

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -0,0 +1,112 @@
...
+        if target_model is None or target_model_page is None:
+            raise ValueError(
+                f"Model with ID `{self.target}` used as a `target` in `set_control` action not found in the dashboard. "
+                f"Please provide a valid control ID that exists in the dashboard."
+            )
+
+        # Validate that target is control that has a categorical selector.
+        if not isinstance(target_model, (vm.Filter, vm.Parameter)) or not isinstance(
```
antonymilne: This is fine in practice but doesn't feel quite right from an extensibility point of view. What if I add a new custom component to `controls` that doesn't subclass `Filter` or `Parameter`? Unlikely in practice, and would probably cause other problems elsewhere, but if it's easy to do then it would be nice to write this so that it could work in theory anyway.

This would mean relaxing the check to something like this:
```
if not hasattr(target_model, "selector") or not isinstance(...)
```

In a perfect world really the check would be even more generic than this but I think this is good for now.

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -0,0 +1,159 @@
...
+            # Validate that target control on different page has `show_in_url=True`.
+            if not target_model.show_in_url:
+                raise ValueError(
+                    f"Model with ID `{self.target}` used as a `target` in `set_control` action is on a different page "
+                    f"than the trigger and must have `show_in_url=True`."
+                )
+            self._same_page = False
+
+    def function(self, _trigger):
+        value = model_manager[self._parent_model]._get_value_from_trigger(  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
antonymilne: Would `cast(_SupportsSetControl)` fix this `ignore`?

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -112,6 +112,16 @@ def _action_inputs(self) -> dict[str, _IdProperty]:
             **{ag_grid_prop: f"{self._inner_component_id}.{ag_grid_prop}" for ag_grid_prop in DAG_AG_GRID_PROPERTIES},
         }
 
+    def _get_value_from_trigger(self, value: str, trigger: list[dict[str, str]]) -> JsonValue:
+        """Value is the name of the column. There is only one point selected, so we just look at trigger[0]."""
```
antonymilne:
```suggestion
        """Value is the name of the column. There is only one row selected, so we just look at trigger[0]."""
```

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -119,6 +120,34 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
+        # value="someKey" and value="some_key" will work. This is enabled by camel_killer_box=True.
+        trigger_box = Box(trigger["points"][0], camel_killer_box=True, box_dots=True)
+        try:
+            # First try to treat value as a column name. Unfortunately the customdata returned in the trigger does
+            # not contain column names (it's just a list) so we must look it up in the called function's `custom_data`
+            # to find its numerical index in this list. This only works if a custom_data was provided in the graph
+            # function call.
+            index = self["custom_data"].index(value)
+            return trigger_box["customdata"][index]
+        except ValueError:
```
antonymilne: This doesn't feel quite right or like it would catch everything here. What are the actual potential sources of error here?
1. `self` doesn't define `custom_data` at all
2. `self["custom_data"]` exists but doesn't contain `value`
3. `index` can't be found in `trigger_box` (Is this possible? Not sure)

Does `ValueError` catch cases 1 and 2? Fine to have the error message for both of these cases but would be good to make sure it is raised at the right time.

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,1347 +1,246 @@
...
-                        ["NavBar", "", "", "", "", ""],
-                        ["NavLink", "", "", "", "", ""],
-                        ["Navigation", "", "", "", "", ""],
+        vm.Container(
+            title="Two filter interactions within Page 1",
+            layout=vm.Grid(grid=[[0, 1]]),
+            variant="outlined",
+            components=[
+                vm.Graph(
+                    id="p1_graph_1",
```
antonymilne: Might be worth removing all these `id`s for Graph and AgGrid just to make absolutely sure it still works that way? And to show reviews it shouldn't be necessary.

vizro-core/tests/integration/test_examples.py
petar-qb: Nice catch! Would adding the following line in the `tests/conftest.py` be a good enough solution or you have something else on your mind? 

```py
@pytest.fixture(autouse=True)
def reset_managers():
    # this ensures that the managers are reset before and after each test
    # the reset BEFORE all tests is important because at pytest test collection, fixtures are evaluated and hence
    # the model_manager may be populated with models from other tests
    Vizro._reset()

    # Some comment
    warnings.filterwarnings("ignore", message=".*experimental.*", category=FutureWarning) # <---LINE ADDED

    yield
    Vizro._reset()
```

vizro-core/src/vizro/actions/_set_control.py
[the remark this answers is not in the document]
petar-qb: It works and looks great! 

`value = cast(_SupportsSetControl, self._parent_model)._get_value_from_trigger(self.value, _trigger)`

vizro-core/src/vizro/actions/_set_control.py
petar-qb: Exactly.. I tried to validate with `_ControlWithCategoricalSelector` and when I figured that it's not working I reimplemented the validation, but forgot to delete this Protocol definition 😄

vizro-core/src/vizro/actions/_set_control.py
petar-qb: Great catch! I did it like:

```py
if not isinstance(getattr(target_model, "selector", None), (vm.Dropdown, vm.Checklist, vm.RadioItems)):
    ...
```
as it's shorter and in my opinion easier to read. If you don't agree I can change it to the `hasattr` syntax.

vizro-core/tests/integration/test_examples.py
antonymilne: No, I mean we should just be able to add it to the ignores in pyproject.toml's `tool.pytest.ini_options` section?

vizro-core/src/vizro/actions/_set_control.py
antonymilne: Yes, perfect! I would also have written it this way too in reality if it hadn't just been a quick PR comment 👍

vizro-core/tests/unit/vizro/models/_components/form/test_checklist.py
```diff
@@ -39,6 +40,7 @@ def test_create_checklist_mandatory_and_optional(self):
         assert checklist.title == "Title"
         assert checklist.actions == []
         assert isinstance(checklist.description, Tooltip)
+        assert checklist._action_triggers == {"__default__": f"{checklist.id}.value"}
```
antonymilne:
```suggestion
        assert checklist._action_triggers == {"__default__": "checklist-id.value"}
```

We can be a bit more explicit in this case since the id was specified. Actually this suggestion applies to all the `_action_outputs/inputs` stuff cross all the unit tests.

vizro-core/tests/unit/vizro/models/_components/test_ag_grid.py
```diff
@@ -378,3 +380,8 @@ def test_aggrid_build_with_description(self, standard_ag_grid):
         )
 
         assert_component_equal(ag_grid, expected_ag_grid, keys_to_strip={"id"})
+
+
+# TODO PP: Add tests
```
antonymilne: This would indeed be nice to do if you have time to do it, which it seems like you do. It should be pretty easy to write I think given that `_get_value_from_trigger` is a pure function and I changed the argument `Action` to `value`.

vizro-core/tests/unit/vizro/models/_components/test_graph.py
```diff
@@ -342,3 +344,8 @@ def test_graph_build_with_extra(self, standard_px_chart):
             overlay_style={"visibility": "visible", "opacity": 0.3},
         )
         assert_component_equal(graph, expected_graph, keys_to_strip={"id"})
+
+
+# TODO PP: Add tests
```
antonymilne: Yes please if you have time!

vizro-core/tests/unit/vizro/actions/test_set_control.py
```diff
@@ -0,0 +1,198 @@
...
+                show_in_url=True,
+            ),
+        ],
+    )
+
+    Vizro._pre_build()
+
+
+@pytest.mark.usefixtures("managers_two_pages_for_set_control")
+@pytest.mark.filterwarnings("ignore:.*experimental.*:FutureWarning")
```
antonymilne: This would be removed here and where else it's done if we move it to a global setting.

vizro-core/tests/unit/vizro/actions/test_set_control.py
```diff
@@ -0,0 +1,198 @@
...
+                id="filter_page_2_show_in_url_true",
+                column="continent",
+                show_in_url=True,
+            ),
+        ],
+    )
+
+    Vizro._pre_build()
+
+
```
antonymilne, line 51: We could have `TestSetControlInstantiation` here like we do for other models. It should be pretty quick and easy to right.

Would be nice if this included a `test_set_controls_experimental` that checks the `FutureWarning` is raised on instantiation - see `test_layout_deprecated` for how we'd do this.

vizro-core/tests/unit/vizro/actions/test_set_control.py
```diff
@@ -0,0 +1,198 @@
...
+        action = set_control(target="filter_page_2_show_in_url_true", value="continent")
+        model_manager["scatter_chart_1"].actions = action
+
+        action.pre_build()
+
+        assert action.outputs == ["vizro_url.pathname", "vizro_url.search"]
+
+
+# TODO PP: Add tests
+class TestSetControlFunction:
```
antonymilne: Yes please if you have time. There's probably two ways of doing this and I'm not sure which is best:
* pure unit test: mock out the `_get_value_from_trigger` line and then test purely the behaviour of `function`
* integration test: more "real world" that uses an actual model manager like you do for the pre-build tests

In an ideal world we'd actually do both, and put the integration tests in our integration tests (which are generally neglected since we tend to just put everything into unit tests - that's not quite the right thing to do but it's ok and I don't mind much). Up to you really depending on how much time you have, how easy the tests are to write and how useful they feel!

antonymilne (commented):
Tests look great so far - it would definitely be nice to have a few more if you have time, which it seems like you do 🙌

vizro-core/schemas/0.1.45.dev0.json
```diff
@@ -135,6 +135,9 @@
               },
               {
                 "$ref": "#/$defs/filter_interaction"
+              },
+              {
+                "$ref": "#/$defs/set_control"
```
maxschulz-COL: Let's discuss

vizro-core/src/vizro/models/types.py
```diff
@@ -715,6 +715,7 @@ class OptionsDictType(TypedDict):
         Annotated["Action", Tag("action")],
         Annotated["export_data", Tag("export_data")],
         Annotated["filter_interaction", Tag("filter_interaction")],
+        Annotated["set_control", Tag("set_control")],
```
maxschulz-COL, line 723: I guess this is where we would fix it - would it be a skip for now?

vizro-core/src/vizro/models/_components/graph.py
petar-qb: Done in -> https://github.com/mckinsey/vizro/pull/1373/commits/7dd31f4ccc762c40c20a0d3bc50f1af3857580be

If you have any further suggestions just let me know. 😃

vizro-core/tests/unit/vizro/models/_components/form/test_checklist.py
petar-qb: This, and a bit more is handled in -> https://github.com/mckinsey/vizro/pull/1373/commits/ea0c45f2b6b2c1683865682408dd7142fe42ee54

vizro-core/tests/unit/vizro/actions/test_set_control.py
petar-qb: I worked on this in -> https://github.com/mckinsey/vizro/pull/1373/files/57ac5019551a5a8c95f1cbd8cc3d0a7daeec8065..a99ca7f21fa54d771ccf7a9836d4259c088d0e24

Looking at the two "unit" tests, I’m not sure they add much value as they largely depend on "managers_two_pages_for_set_control", similar to our pre_build tests. Additionally, I had to monkeypatch `get_relative_path` to get them working.

Do you think we should leave them as is, try to improve them, or remove them entirely and rely on e2e tests instead?

vizro-core/src/vizro/models/_components/graph.py
antonymilne: Looks good! I just left a couple of small comments.

vizro-core/tests/unit/vizro/actions/test_set_control.py
```diff
@@ -190,6 +191,48 @@ def test_pre_build_target_model_on_different_page_show_in_url_false(self):
...
+        # Mock dash.get_relative_path as it's used in set_control.function
+        monkeypatch.setattr(set_control_module, "get_relative_path", lambda path: "/mocked_path")
+
+        # Call function method with a mock trigger value
+        result_relative_path, result_url_query_params = action.function(
+            _trigger={"points": [{"customdata": ["Europe"]}]},
+        )
+        # From mocked get_relative_path
+        expected_relative_path = "/mocked_path"
+        # Value Europe" base64 encoded is b64_IkV1cm9wZSI
```
antonymilne:
```suggestion
        # Value "Europe" base64 encoded is b64_IkV1cm9wZSI
```

vizro-core/tests/unit/vizro/actions/test_set_control.py
```diff
@@ -163,14 +185,55 @@ def test_pre_build_target_model_on_different_page_show_in_url_false(self):
...
+    def test_function_target_model_on_different_page(self, monkeypatch):
+        # Add action to relevant component and set target to a control on different page with show_in_url=True
+        action = set_control(target="filter_page_2_show_in_url_true", value="continent")
+        # Any other model that supports set_control can be used here, but the Graph used for the simplicity.
+        model_manager["scatter_chart_1"].actions = action
+        # Call pre_build to set _same_page attribute
+        action.pre_build()
+
+        # Mock dash.get_relative_path as it's used in set_control.function
+        monkeypatch.setattr(set_control_module, "get_relative_path", lambda path: "/mocked_path")
```
antonymilne: Can we do this with `mocker` instead? That's normally how we do this. There's a `return_value` argument that we can use.

vizro-core/tests/unit/vizro/actions/test_set_control.py
antonymilne: Nice work writing these, I know they're not easy to do! I think they have some value and I don't see an easy way to improve them, so let's keep them as they are. So long as they don't cause problems in future then there's no harm in having these two extra tests in addition to e2e tests even if they're weak.

antonymilne (commented):
I can't approve my own PR, but consider this comment my approval! It would be good to get someone independent to manually test but other than that it's good to go from my side :shipit: 

Amazing work @petar-qb, this is a massive new feature that will change a lot 🤩

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -112,6 +112,20 @@ def _action_inputs(self) -> dict[str, _IdProperty]:
             **{ag_grid_prop: f"{self._inner_component_id}.{ag_grid_prop}" for ag_grid_prop in DAG_AG_GRID_PROPERTIES},
         }
 
+    def _get_value_from_trigger(self, value: str, trigger: list[dict[str, str]]) -> JsonValue:
+        """Value is the name of the column. There is only one row selected, so we just look at trigger[0]."""
+        # In case the selectedRows is empty (eg when the user unselects the row), trigger is an empty list.
```
antonymilne:
```suggestion
        # In case the selectedRows is empty (e.g. when the user unselects the row), trigger is an empty list. 
        # This works for both multi=True and multi=False selectors.
```

vizro-core/changelog.d/20250901_210343_antony.milne_experimental_update_controls.md
petar-qb, line 1: Reminder to write the announcement

petar-qb:
@petar-qb Reminder to:
- [x] - rename "target"->"control"
- [x] - check how it behaves when 4 `set_control` actions are called in a chain. Check whether any `filter_action` is triggered before the last `set_control` finishes. See is there any problem with that. Document it.

vizro-ai/src/vizro_ai/dashboard/_response_models/layout.py
maxschulz-COL, line 1: So there are some core changes to `vizro-ai`. Shall we make a release here, or just leave as is?

@lingyielia FYI

vizro-core/docs/griffe_extensions/experimental_decorator/LICENCE
maxschulz-COL, line 1: Is @Joseph-Perkins aware of this?

vizro-core/src/vizro/_vizro_utils.py
maxschulz-COL, line 1: If we decide against experimental, will we remove this and the griffe extension files?

vizro-core/src/vizro/actions/_actions_utils.py
```diff
@@ -102,7 +102,7 @@ def _apply_filter_interaction(
 
     Returns: filtered DataFrame.
     """
-    # The filter_interaction model actually contains the id we require in its _parent_model_id field.
+    # The filter_interaction model actually contains the id we require in its _parent_model field.
```
maxschulz-COL, line 105: Do we even still have a `filter_interaction` model?

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -0,0 +1,102 @@
...
+    "but until then it may change or be removed without warning. If you have feedback on the feature then "
+    "[let us know](https://github.com/mckinsey/vizro/issues)."
+)
+class set_control(_AbstractAction):
+    """Sets the value of a control based on data from the trigger.
+
+    Args:
+        control (ModelID): Filter or Parameter component id to be affected by the trigger. If the control is on a
+            different page to the trigger then it must have `show_in_url=True`.
+        value (str): TODO AM: ADD DESCRIPTION
```
maxschulz-COL: Just so its not missed

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -0,0 +1,102 @@
...
+    control: ModelID = Field(
+        description="Filter or Parameter component id to be affected by the trigger."
+        "If the control is on a different page to the trigger then it must have `show_in_url=True`."
+    )
+    value: str = Field(description="TODO AM: ADD DESCRIPTION")
+
+    @_log_call
+    def pre_build(self):
+        # Validate that action's parent model supports `set_control` action.
+        if not isinstance(self._parent_model, _SupportsSetControl):
```
maxschulz-COL, line 96: In `_action.py` I found the comment:

>     # Temporary hack to help with lookups in filter_interaction and set_control. Should not be required in future with reworking of model manager and removal of filter_interaction.

Since we are remiving the `filter_interaction`, do we still need the hack, or is the MM work also required?

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -59,9 +58,9 @@ class _BaseAction(VizroBaseModel):
     _first_in_chain_trigger: _IdProperty = PrivateAttr()
     _prevent_initial_call_of_guard: bool = PrivateAttr()
 
-    # Temporary hack to help with lookups in filter_interaction. Should not be required in future with reworking of
-    # model manager and removal of filter_interaction.
-    _parent_model_id: ModelID = PrivateAttr()
+    # Temporary hack to help with lookups in filter_interaction and set_control. Should not be required in future with
```
maxschulz-COL: Ok arrived here now: why do we still mention `filter_interaction`? Is it because it has not yet been removed right?

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -119,6 +120,34 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
             **({"description": f"{self.description.id}-text.children"} if self.description else {}),
         }
 
+    def _get_value_from_trigger(self, value: str, trigger: dict[str, list[JsonValue]]) -> JsonValue:
```
maxschulz-COL, line 126: Would you mind expanding on this black magic. Just a simple example would be good, like what woult `trigger` be, and what becomes possible with `Box`

vizro-core/src/vizro/tables/_dash_ag_grid.py
```diff
@@ -91,6 +91,11 @@ def dash_ag_grid(data_frame: pd.DataFrame, **kwargs: Any) -> dag.AgGrid:
             "domLayout": "autoHeight",
             "pagination": True,
             "paginationPageSize": 20,
+            "rowSelection": {
```
maxschulz-COL, line 94: Should we mentioned this in changelog?

vizro-core/tests/unit/vizro/actions/test_set_control.py
```diff
@@ -0,0 +1,256 @@
...
+        with pytest.raises(
+            ValueError,
+            match=re.escape(
+                "Model with ID `invalid_id` used as a `control` in `set_control` action not found in the dashboard. "
+                "Please provide a valid control ID that exists in the dashboard."
+            ),
+        ):
+            action.pre_build()
+
+    def test_pre_build_control_model_exists_in_model_manager_but_not_in_any_page(self):
```
maxschulz-COL, line 114: FYI, this should not happen after MM rework, but test is good for now I guess

vizro-core/tests/unit/vizro/models/_components/test_graph.py
```diff
@@ -110,6 +112,75 @@ class MyGraph(vm.Graph):
         assert my_graph.actions == []
 
 
+class TestGraphGetValueFromTrigger:
+    """Tests _get_value_from_trigger models method."""
+
+    @pytest.mark.parametrize(
```
maxschulz-COL, line 118: I guess this is the example I was asking about ealier - maybe just comment here what the `Box` allows us to do

maxschulz-COL (approved):
I have tried to understand, and overall I think it is good. I couldn't find anything that was obviously wrong. Most of my comments are just questions.

That being said maybe an overall question: I tried to follow the execution flow, and I don't quite understand anymore how the overall system works...

How does the the `set_control` action get the `_trigger` value, (similar to how other actions get the `_controls` state)? I just wasn't able to piece it together quickly anymore... That made it quite hard to really understand if everything is correct as I was not sure how click events flow through the app - maybe someone can enlighten me how things fit together.

vizro-ai/src/vizro_ai/dashboard/_response_models/layout.py
antonymilne: This is unrelated to the current PR, not sure whether it was meant to be here - @petar-qb? 

I believe @lingyielia had as a ticket a while ago to deprecate `Layout` on vizro-ai (which is what Petar did here) but I guess never got around to doing it as the whole package was deprioritised.

vizro-core/src/vizro/tables/_dash_ag_grid.py
antonymilne: I don't think so, because it's basically an invisible/unnoticeable change.

vizro-ai/src/vizro_ai/dashboard/_response_models/layout.py
petar-qb: After the [actions docs PR](https://github.com/mckinsey/vizro/pull/1357) was merged, I started seeing failing `vizro-ai` unit tests due to the deprecated Layout being used in `vizro-ai/`. This was later addressed in [Max's PR](https://github.com/mckinsey/vizro/pull/1374).

What’s unclear to me is why the vizro-ai CI unit tests passed when the actions docs PR was merged, even though the issue existed at that point. For now, I reverted all the `vizro-ai` changes, and everything works again.

vizro-core/src/vizro/_vizro_utils.py
petar-qb: We decided to remove experimental from `set_control`, but I’ll keep this and the griffe extension as is for now since it doesn’t cause any issues and might be useful in the near future. In the meantime, I hope we will align on whether or not to include the experimental features in the scheme 😄.

vizro-core/src/vizro/actions/_actions_utils.py
petar-qb: Yep, search for `class filter_interaction(_AbstractAction)`. It's inherited from the `_AbstractAction` and `_BaseAction`, so we can access its ` _parent_model` from the `_BaseAction`

vizro-core/src/vizro/actions/_set_control.py
petar-qb: The `_parent_model` will still be useful for `set_control`, so let me adjust the comment. 

I'm now fully aware of the new MM improvements, but I guess that if the `parent_model` will be accessible through the MM by default for any model that we can remove it from the actions.

vizro-core/src/vizro/models/_action/_action.py
petar-qb: Correct. I also clarified the following terms in my head in last couple of weeks/months. So:
- `Deprecated` means a feature is still supported but no longer under active development and is planned for removal. By raising a warning we're giving users time to transition to an alternative.
- `Removed` means the feature is gone from Vizro entirely. 

So, `filter_interaction` along with `vm.Layout`, `vm.Action.inputs` and other are deprecated, and will be removed in the `0.2.0`.

vizro-core/src/vizro/models/_components/graph.py
petar-qb: Some of it you can find [here](https://github.com/mckinsey/vizro/pull/1379/files#diff-fb0a8d4a026fcb413d6db1fd87fc365d966643f3804b244445e5eb2b3cec5bde). But here's an example too:

When you click on the Graph's cell, Plotly.js sends the json to the server with the following format (values below are arbitrary):
```py
{
  "points": [
    {
      "curveNumber": 1,
      "pointNumber": 0,
      "pointIndex": 0,
      "x": 1,
      "y": 3,
      "bbox": {
        "x0": 189.35,
        "x1": 209.35,
        "y0": 1058.07,
        "y1": 1078.07
      },
      "customdata": [
        "setosa", 3, 4.4
      ],
      "this_is_made_up": [0 ,1 , {"petar": "qb"}]
    }
  ]
}
```

This json could vary, so we enabled our users to "walk" through the json and fetch the data they are interested in. So, they are able to extract any data by propagating the Box syntax in the `set_control.value`. For example:
- value="customdata[0]" -> returns `"setosa"`,
- value="x" -> returns `1`
- value="this_is_made_up[2].petar" -> returns `"qb"`

In the link I sent at the beginning of this comment you can find another and simpler way to access the the custom data "setosa" for the Graph (`value="species"`). 

For the AgGrid, the format is `[{"column_1": "abc", "column_2": "123", ...}]`. We don't use Box there as the structure is shallow and known.

vizro-core/src/vizro/_vizro_utils.py
maxschulz-COL: I'd prefer not to keep code just in case, is everyone of a different opinion?

petar-qb: 🤔 I don't have the strong opinion here. What do you think @antonymilne??

There are several upcoming features like: chatbot, cascading filters, dynamic parameters, sync filter on the same page, drill-through with no url, sync filter across the pages, global controls and so on, and I see some of them released as experimental at first. If you guys think that experimental won't be used, I'll remove it.

maxschulz-COL: I am just not sure we should do experimental like that, so unless we decide that I would not merge it

vizro-core/docs/pages/cheatsheet/cheatsheet.html
petar-qb, line 1: Currently the added `set_control` cheatsheet explanation looks like this:
image

I will improve it when the `cross-highlight` is describe in the docs. In that case I suggest one of the two options:
- 2 cells: (cross-filter form table + tip about graphs, cross-highlight from graph + tip about table) 
- 4 cells: (cross-filter from table, cross-filter from graph, cross-highlight from table, cross-highlight from graph) or

When working on this, consider how it would look like when we support cross-filter/highlight from button too.

vizro-core/src/vizro/_vizro_utils.py
antonymilne: Yeah, I think question of whether to use `@experimental` or `experimental` namespace is still open. I'm happy to remove the experimental decorator + docs extension from this PR so long as we can easily find the relevant code again in future if we do want to use it.

petar-qb: Here's the removal commit -> [Remove experimental decorator and experimental docs extension](https://github.com/mckinsey/vizro/pull/1373/files/562e074ab5e7c97eae9bc7d9e0b569c24dbdbf3c..f657e03b3506761fa81ce0fdeb389ac086a636a3)

petar-qb:
Added `experimental` code here for visibility:
- [Adding experimental decorator and experimental docs extension](https://github.com/mckinsey/vizro/pull/1373/files/5d962aa15223f8a76518103024d22794d2b0e3d7..f0f8bf5084925528923f5aa6e96471959c2d6e88)
- [Removing experimental decorator and experimental docs extension](https://github.com/mckinsey/vizro/pull/1373/files/562e074ab5e7c97eae9bc7d9e0b569c24dbdbf3c..f657e03b3506761fa81ce0fdeb389ac086a636a3)
- [Remove experimental from pytest ignores](https://github.com/mckinsey/vizro/pull/1373/commits/4e732a58fd2a0cdc15e882440afcc377bad778b0)

Link to discussion thread https://github.com/mckinsey/vizro/pull/1373#discussion_r2348334543

PR #1374, opened 2025-09-02
[Tidy] Move argument descriptions to Field from docstring

maxschulz-COL:
## Description
Closes https://github.com/McK-Internal/vizro-internal/issues/1929

## Notice

### DIFF
--- a/vizro-core/tests/e2e/vizro/test_dom_elements/test_dynamic_data.py
+++ b/vizro-core/tests/e2e/vizro/test_dom_elements/test_dynamic_data.py
@@ -20,7 +20,6 @@ from e2e.vizro.navigation import (
     select_slider_handler,
 )
 from e2e.vizro.paths import (
-    actions_progress_indicator_path,
     categorical_components_value_path,
     dropdown_arrow_path,
     graph_axis_value_path,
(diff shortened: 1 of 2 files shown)


vizro-mcp/src/vizro_mcp/server.py
```diff
@@ -70,28 +70,28 @@ class ModelJsonSchemaResults:
...
 
 @mcp.tool()
 def get_vizro_chart_or_dashboard_plan(
-    user_plan: Literal["chart", "dashboard"],
-    user_host: Literal["generic_host", "ide"],
-    advanced_mode: bool = False,
+    user_plan: Literal["chart", "dashboard"] = Field(description="The type of Vizro thing the user wants to create"),
+    user_host: Literal["generic_host", "ide"] = Field(
+        description="The host the user is using, if 'ide' you can use the IDE/editor to run python code"
+    ),
```
huong-li-nguyen, line 82: Wondering if these are better as bools if they have two options only? Will we ever do something more than just chart and dashboard for example? If not, we could just refactor this to `is_dashboard_plan` or something similar no?

huong-li-nguyen (approved):
LGTM, just left one question/comment. But I don't feel strongly about it, so if you disagree, just feel free to merge :)

vizro-mcp/src/vizro_mcp/server.py
maxschulz-COL: Hm good question! I would tend to leave it mainly to force the model to match the word `chart` with the user prompt. So I could imagine (although in no way tested) that it helps if the model does not need to make the connection from the user prompt ("make me a chart") to the tool call (`is_dashboard_plan == False`), but rather then just translate this into `user_plan=="chart"`.

It is speculation though, so in the interest of time just leave it?

vizro-ai/pyproject.toml
```diff
@@ -67,9 +67,10 @@ filterwarnings = [
   "ignore:CapturedCallable function is excluded from the schema",
   # Ignore `Layout` FutureWarning until vizro 0.2.0 released:
   "ignore:The `Layout` model has been renamed `Grid`:FutureWarning",
```
petar-qb, line 69: I guess you can remove this line then

maxschulz-COL: Sure but now the tests pass so I will ifnore haha

PR #1379, opened 2025-09-09
[Docs] `set_control action`

antonymilne:
## Description

**Main change**: add documentation on the new `set_control` feature introduced in #1373 and deprecate old `filter_interaction`
* [API reference](https://vizro--1379.org.readthedocs.build/en/1379/pages/API-reference/actions/#vizro.actions.set_control)
* [User guide](https://vizro--1379.org.readthedocs.build/en/1379/pages/user-guides/graph-table-actions/)
* [Deprecation note](https://vizro--1379.org.readthedocs.build/en/1379/pages/API-reference/deprecations/#filter_interaction)

**Other changes**: 
* improve cross-links and narrative for controls, filter, parameters. These were not great before (e.g. we didn't mention on the controls page that they could be used inside containers).
* move the section on `show_in_url` to respond to some old feedback from @maxschulz-COL - see below.

**Note for reviewers**: possibly the user guide on `set_control` contains a bit too much explanation. I intend to write a whole separate page on this to explain the whole paradigm of how controls update components and will move some content there then, but I'm reaching my limit of docs writing so will save that for another PR. https://github.com/McK-Internal/vizro-internal/issues/2170

Terminology used: a _source_ component (graph etc.) updates a _target_ component (graph etc.) via an intermediate control. The `set_control` action _sets_ a control, which then _updates_ the target component.

---

Old feedback on our current `show_in_url` docs from @maxschulz-COL which I agree with:
>  Hm I am not sure I like this [`show_in_url` docs in Parameter/Filter sections] all that much here. I think in the past we have slowly deviated from our north star of the diataxis framework, where "How-to" guide are concise guides in how to get something done, to a more API like description of our models even in the guides. This is by no means about this PR, but for some reason it triggered me more just now. (Similar comments would apply to the extra argument description)
> 
> I think overall we should rather have a guide on URL parameters and drill throughs, link to that guide from here very shortly, and not duplicate twice what is essentially the same description.
> ...
> In general I think if we have a central how to guide, then we would also have more space to show some of the more intricate amazing things that are possible now, such as drill through. This really deserves a place in the docs, but ideally as part of a how to guide that is dedicated to this

## Notice

### DIFF
--- a/vizro-core/docs/pages/user-guides/actions.md
+++ b/vizro-core/docs/pages/user-guides/actions.md
@@ -94,74 +94,74 @@ When you click the "Export data" button, the data for all graphs, tables and fig
 
 Here is an example that [performs a cross-filter](graph-table-actions.md#cross-filter-from-graph) between a graph and a table. When you click on a box in the graph, the table is cross-filtered to show data for only one sex.
 
-````
-=== "app.py"
+!!! example "Cross-filter from graph to table"
 
-    ```{.python pycafe-link hl_lines="15"}
-    import vizro.actions as va
-    import vizro.models as vm
-    import vizro.plotly.express as px
-    from vizro import Vizro
-    from vizro.tables import dash_ag_grid
+    === "app.py"
 
-    tips = px.data.tips()
+        ```{.python pycafe-link hl_lines="15"}
+        import vizro.actions as va
+        import vizro.models as vm
+        import vizro.plotly.express as px
+        from vizro import Vizro
+        from vizro.tables import dash_ag_grid
 
-    page = vm.Page(
-        title="Cross-filter from graph to table",
-        components=[
-            vm.Graph(
-                title="Click on a box to use that box's sex to filter table",
-                figure=px.box(tips, x="tip", y="sex"),
-                actions=va.set_control(control="sex_filter", value="y"),
-            ),
-            vm.AgGrid(id="tips_table", figure=dash_ag_grid(tips)),  # (1)!
-        ],
-        controls=[vm.Filter(id="sex_filter", column="sex", targets=["tips_table"])],  # (2)!
-    )
+        tips = px.data.tips()
 
-    dashboard = vm.Dashboard(pages=[page])
-    Vizro().build(dashboard).run()
-    ```
+        page = vm.Page(
+            title="Cross-filter from graph to table",
+            components=[
+                vm.Graph(
+                    title="Click on a box to use that box's sex to filter table",
+                    figure=px.box(tips, x="tip", y="sex"),
+                    actions=va.set_control(control="sex_filter", value="y"),
+                ),
+                vm.AgGrid(id="tips_table", figure=dash_ag_grid(tips)),  # (1)!
+            ],
+            controls=[vm.Filter(id="sex_filter", column="sex", targets=["tips_table"])],  # (2)!
+        )
 
-    1. We give the `vm.AgGrid` an `id` so that it can be targeted explicitly by `vm.Filter(id="sex_filter")`.
-    1. We give the `vm.Filter` an `id` so that it can be set explicitly by `va.set_control`.
+        dashboard = vm.Dashboard(pages=[page])
+        Vizro().build(dashboard).run()
+        ```
 
-=== "app.yaml"
+        1. We give the `vm.AgGrid` an `id` so that it can be targeted explicitly by `vm.Filter(id="sex_filter")`.
+        1. We give the `vm.Filter` an `id` so that it can be set explicitly by `va.set_control`.
 
-    ```yaml
-    # Still requires a .py to add data to the data manager and parse YAML configuration
-    # See yaml_version example
-    pages:
-      - components:
-          - actions:
-              - control: sex_filter
-                type: set_control
-                value: y
-            figure:
-              _target_: box
-              data_frame: tips
-              x: tip
-              y: sex
-            title: Click on a box to use that box's sex to filter table
-            type: graph
-          - figure:
-              _target_: dash_ag_grid
-              data_frame: tips
-            id: tips_table
-            type: ag_grid
-        controls:
-          - column: sex
-            id: sex_filter
-            targets:
-              - tips_table
-            type: filter
-        title: Cross-filter from graph to table
-    ```
+    === "app.yaml"
 
-=== "Result"
+        ```yaml
+        # Still requires a .py to add data to the data manager and parse YAML configuration
+        # See yaml_version example
+        pages:
+          - components:
+              - actions:
+                  - control: sex_filter
+                    type: set_control
+                    value: y
+                figure:
+                  _target_: box
+                  data_frame: tips
+                  x: tip
+                  y: sex
+                title: Click on a box to use that box's sex to filter table
+                type: graph
+              - figure:
+                  _target_: dash_ag_grid
+                  data_frame: tips
+                id: tips_table
+                type: ag_grid
+            controls:
+              - column: sex
+                id: sex_filter
+                targets:
+                  - tips_table
+                type: filter
+            title: Cross-filter from graph to table
+        ```
 
-    ![](../../assets/user_guides/graph_table_actions/cross_filter_from_graph_2.gif)
-````
+    === "Result"
+
+        ![](../../assets/user_guides/graph_table_actions/cross_filter_from_graph_2.gif)
 
 ## Multiple actions
 
--- a/vizro-core/docs/pages/user-guides/graph-table-actions.md
+++ b/vizro-core/docs/pages/user-guides/graph-table-actions.md
@@ -6,7 +6,7 @@ In this guide we show you how to configure interactions between graphs and table
 - [Cross-parameter](#cross-parameter): a _source_ graph or table updates any argument other than `data_frame` of a _target_ graph or table. The source components sets a [parameter](parameters.md), which in turn updates the target component.
 - [Cross-highlight](#cross-highlight): a _source_ graph or table highlights data in a _target_ graph or table. This is an example of a [cross-parameter](#cross-parameter).
 
-All these interactions use the [`set_control`][vizro.actions.set_control] action. This gives very generic and powerful functionality thanks to the functionality of the intermediate control:
+All these interactions use the [`set_control` action][vizro.actions.set_control]. This gives very generic and powerful functionality thanks to the functionality of the intermediate control:
 
 - The target components can be anything that reacts to a control: [built-in graphs](graph.md), [custom graphs](custom-charts.md), [built-in tables](table.md), [custom tables](custom-tables.md), [built-in figures](figure.md) and [custom figures](custom-figures.md).
 - A single control can update any number of these target components, and a single source component can set any number of controls. Hence a single source component can interact with any number of target components.
@@ -31,7 +31,7 @@ A cross-filter is when the user clicks on one _source_ graph or table to filter 
 1. Call `set_control` in the `actions` argument of the source [`Graph`][vizro.models.Graph] or [`AgGrid`][vizro.models.AgGrid] component that triggers the cross-filter.
 
     1. Set `control` to the ID of the filter.
-    1. Set `value`. The format of this depends on the source model and is given in the [API reference][vizro.actions.set_control], but it is often `column` of the filter.
+    1. Set `value`. The format of this depends on the source model and is given in the [API reference][vizro.actions.set_control], but it is often `column` of the filter. Think of it as an instruction for what to lookup in the source data: whatever value is fetched from this lookup is used to set `control`.
 
     ```python
     import vizro.actions as va
--- a/vizro-core/mkdocs.yml
+++ b/vizro-core/mkdocs.yml
@@ -124,7 +124,7 @@ markdown_extensions:
             kwds:
               type: vizro
               requirements: |
-                vizro == 0.1.45
+                vizro==0.1.45
   - pymdownx.tabbed:
       alternate_style: true
   - pymdownx.mark
--- a/vizro-core/src/vizro/actions/_set_control.py
+++ b/vizro-core/src/vizro/actions/_set_control.py
@@ -34,7 +34,7 @@ class set_control(_AbstractAction):
 
     The following Vizro models can be a source of `set_control`:
 
-    * [`AgGrid`][vizro.models.AgGrid]: triggers `set_control` when user clicks on a row in the table.`value` specifies
+    * [`AgGrid`][vizro.models.AgGrid]: triggers `set_control` when user clicks on a row in the table. `value` specifies
     which column in the clicked row is used to set `control`.
     * [`Graph`][vizro.models.Graph]: triggers `set_control` when user clicks on data in the graph. `value` can be used
     in two ways to specify how to set `control`:
@@ -64,7 +64,7 @@ class set_control(_AbstractAction):
         import vizro.actions as va
 
         vm.Graph(
-            figure=px.scatter(iris, x="sepal_width", y="sepal_length", custom_data=["species"]),
+            figure=px.scatter(iris, x="sepal_width", y="sepal_length", custom_data="species"),
             actions=va.set_control(control="target_control", value="species"),
         )
         ```


stichbury (approved):
This looks great, thanks! I can't find anything to comment on from the docs point of view at a high level (language, readability, etc). Those that know the topic more may have something but it's great and ready to go from my perspective.

With all the docs and changes coming in this area, I wonder how we best communicate the delta / improvements to Vizro of late. I'll continue this discussion away from public view until we have consensus.

vizro-core/docs/pages/user-guides/controls.md
petar-qb, line 1: 10/10🏅

vizro-core/tests/unit/vizro/actions/test_legacy_filter_interaction.py
petar-qb, line 1: Should we delete the `test_filter_interaction.py`?

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -27,20 +27,68 @@ def _encode_to_base64(value):
...
+    The following Vizro models support `set_control`:
+
+    * [`AgGrid`][vizro.models.AgGrid]: triggers `set_control` when user clicks on a row in the table. The `value`
+    argument specifies the column of the value that is sent to the `target` control.
+    * [`Graph`][vizro.models.Graph]: triggers `set_control` when user clicks on data in the graph. The `value` argument
+    can be used in two ways to specify what value to send to the `target` control:
+
+        * String to [traverse a Box](https://github.com/cdgriffith/Box/wiki/Types-of-Boxes#box-dots) that contains the
+        trigger data [`clickData["points"][0]`](https://dash.plotly.com/interactive-graphing), for example `"x"`.
+        * Column from which to take the value. This requires you to set `custom_data` in the graph's `figure` function.
```
petar-qb: I would replace the order of those two as it's much more native to understand the second one (especially as the same analogy is used for `AgGrid.value`).

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -27,20 +27,68 @@ def _encode_to_base64(value):
...
+        ```
+
+    Example: `Graph` as trigger using `custom_data`
+        ```python
+        import vizro.actions as va
+
+        vm.Graph(
+            figure=px.scatter(iris, x="sepal_width", y="sepal_length", custom_data=["species"]),
+            actions=va.set_control(control="target_control", value="species"),
+        )
```
petar-qb: I would replace their place here as well.

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
-<!--
+# Graph and table interactions
 
-When I rewrite this page, make it clear what is difference between cross-filtering, parameter, drilldown, drillthrough. Refer to other BI tools. Make clear paradigm of control affecting components. Reference show_in_url controls. Make sure referred to from pages on graph/table and custom versions of those.
+In this guide we show how to configure interactions between graphs and tables, as is commonly done in business intelligence (BI) tools. In Vizro, all such interactions are enabled by an intermediate [control](controls.md) that you must explicitly define. For example:
 
--->
+- [Cross-filter](#cross-filter) is used to filter a _target_ graph or table from another _source_ graph or table. It is achieved by the source component setting a [filter](filters.md), which in turn updates the target component.
+- [Cross-highlight](#cross-highlight) is used to highlight data in a _target_ graph or table from another _source_ graph or table. It is achieved by the source component setting a [parameter](parameters.md), which in turn updates the target component.
```
petar-qb: Is “cross-highlight” standard BI terminology, and how widely recognised is it within the Vizro community?

What do you think about using “cross-parameter” instead? 

To me, the “cross-parameter” name itself suggests that users need to manually introduce a `vm.Parameter` as an intermediate control. It also feels more intuitive that with a `cross-parameter`, you could leverage all the capabilities of the `parameter` component. For example, parametrizing the x-axis, a title, or even a dynamic data function argument.

By contrast, the explanation like “cross-highlight is used to highlight data in a target graph or table” doesn’t quite convey that broader flexibility.

```suggestion
- [Cross-parameter](#cross-parameter) is used to set any argument other than `data_frame` in the `figure` function of a graph or table from another _source_ graph or table. It is achieved by the source component setting a [parameter](parameters.md), which in turn updates the target component.
```

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
 
-# How to interact with graphs and tables
+All these interactions use the [`set_control`][vizro.actions.set_control] action. This gives very generic and powerful functionality thanks to the functionality of the intermediate control:
 
-## Cross-filtering
+- The target components can be anything that reacts to a control: [built-in graphs](graph.md), [custom graphs](custom-charts.md), [built-in tables](table.md), [custom tables](custom-tables.md), [built-in figures](figure.md) and [custom figures](custom-figures.md).
+- A single control can update any number of these target components, and a single source component can set any number of controls. Hence a single source component can interact with any number of target components.
+- A target component can be on the same page as the source or on a different page (so long as the control has [`show_in_url=True`](run-deploy.md#shareable-url)).
```
petar-qb: Maybe
```suggestion
- A target component can be on the same page as the source or on a different page (so long as the intermediate control has [`show_in_url=True`](run-deploy.md#shareable-url)).
```

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
+                id: sex_filter
+                targets:
+                  - tips_graph
+                type: filter
+            title: Cross-filter from table to graph
+        ```
+
+    === "Result"
+
+        TODO NOW: screenshot
```
petar-qb: I guess/hope that `set_control` will be extensively used feature, so it would be awesome if `.gif` is set here and in the other examples below.

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
-    - You can create a "self-interaction" by providing the source chart id as its own `target`
+!!! tip
 
-Here is an example of how to configure a chart interaction when the source is a [`Graph`][vizro.models.Graph] component.
+    Often the `value` of `set_control` is the same as the `column` of the filter, but this does not need to be the case. You can perform a cross-filter where the source component's column name given by `value` is different from the target component's column name, which is given by the filter's `column`.
 
-!!! example "Graph `filter_interaction`"
+### Cross-filter from table
+
+The trigger for a cross-filter from an [AG Grid](table.md#ag-grid) is clicking on a row in the table or pressing ++space++ while focused on a row. The `value` argument in `va.set_control` specifies the column of the value that is sent to the specified `control`.
```
petar-qb: I would delete the "pressing ++space++" here and in other places. IMO, no one does that, and can only confuse our users 😄. Maybe I'm missing the case where it's really useful

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
+
+!!! example "Cross-filter from table to graph"
```
petar-qb, line 52: I like the example. It would be nice if all examples and code sections on the page would deal with the same data_frame. However, not critical at all 😄

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
+!!! tip
+
+    You can emphasize that a row is selectable by including checkboxes in your AG Grid with `figure=dash_ag_grid(..., dashGridOptions={"rowSelection": {"checkboxes": True}})`. The Dash AG Grid offers many [options to configure row selection](https://dash.plotly.com/dash-ag-grid/single-row-selection). These can be [passed directly](table.md#basic-usage) into `dash_ag_grid` as keyword arguments or set for multiple tables by creating a [custom table function](custom-tables.md).
+
+### Cross-filter from graph
+
+The trigger for a cross-filter from a [graph](graph.md) is clicking on data in the graph. The `value` argument in `va.set_control` can be used in two ways to specify what value is sent to the specified `control`:
+
+- Value to extract from [Plotly's `clickData`](https://dash.plotly.com/interactive-graphing), for example `"x"`.
+- Column from which to take the value. This requires you to set `custom_data` in the graph's `figure` function.
```
petar-qb: Similarly like I commented below: I would replace the order of those two as it's much more native to understand the second one (especially as the same analogy is used for AgGrid.value). The same refers to the example just below these two bullet points.

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
+
+    When performing multiple filters with [dynamic data](data.md#dynamic-data), you should consider [configuring a cache](data.md#configure-cache) so that steps 3 and 4 above do not repeatedly perform a slow data load.
+
+Multiple cross-filters are similarly possible [from a table](#cross-filter-from-table):
+
+```python
+vm.AgGrid(
+    ..., actions=[va.set_control(control="day_filter", value="day"), va.set_control(control="sex_filter", value="sex")]
+)
+```
```
petar-qb, line 620: I would delete the example and move this explanation sentence somewhere else or even delete this one.

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
+
+However, it is not yet possible to cross-filter from a pivot table according to the row and column of the clicked cell.
+
+## Cross-highlight
 
-[graphinteraction]: ../../assets/user_guides/graph_tables_actions/actions_filter_interaction.png
-[tableinteraction]: ../../assets/user_guides/graph_tables_actions/actions_table_filter_interaction.png
+This is already possible, and documentation is coming soon!
```
petar-qb: Is it the plan to have this section completed before the next release, or it will come later (which is totally fine by me)?

petar-qb (approved):
This is definitely one of the clearest docs PRs I've reviewed so far!! 🚀 

I left several comments but none of them are major. Also, just ignore without an explanation these you see irrelevant.

vizro-core/tests/unit/vizro/actions/test_legacy_filter_interaction.py
antonymilne: It's not very important to keep but has a little bit of value since it tests the new `actions=filter_interaction()` syntax with the deprecated `filter_interaction`. That functionality should continue to work until 0.2.0.

vizro-core/src/vizro/actions/_set_control.py
antonymilne: Hmm yes, I wasn't sure of the right order to do it. On the one hand the first example is simpler because it doesn't require `custom_data`, but on the other hand it doesn't much the AgGrid example.

Curious what another reviewer thinks here.

vizro-core/docs/pages/user-guides/graph-table-actions.md
antonymilne: I believe that cross-highlight is standard BI terminology (see e.g. [PowerBI](https://learn.microsoft.com/en-us/power-bi/create-reports/power-bi-reports-filters-and-highlighting)) while cross-parameter isn't. So to a BI user, cross-highlight is likely more familiar than cross-parameter whereas in pure Vizro terms, like you say, cross-parameter makes much more sense.

My intention with this page is to follow diataxis to write about "how to achieve BI functionality". Hence even though cross-parameter is much more general and powerful than just a cross-highlight, a BI user is more likely to search for "how do I do cross-highlight" rather than "how do I do cross-parameter". There will be another page explaining more about the paradigm of how controls update components which will show more generally the power of a cross-parameter.

I agree this is very important though, so I tried to hint on this page that the functionality went well beyond just cross-highlight in the list "All these interactions use the [`set_control`][vizro.actions.set_control] action. This gives very generic and powerful functionality thanks to the functionality of the intermediate control" but didn't explicitly say "you can do anything with parameter". Let me add that to the list.

@maxschulz-COL curious what you think of this organisation of docs - do you think it's the right thing to do here? I believe it's the sort of thing you were thinking of following diataxis given the quote from you in my PR description.

vizro-core/docs/pages/user-guides/graph-table-actions.md
antonymilne: Yes, definitely I will try to do a nice gif of these!

vizro-core/docs/pages/user-guides/graph-table-actions.md
antonymilne: Yeah, I actually didn't realise you could do it by pressing space until I saw it on the AgGrid docs last week. It feels kind of useful to me but agree maybe it's not worth mentioning. In part I added it just because it's fun doing `++` and using the keyboard extension in markdown tbh 😀 

Curious of opinion of second reviewer here.

vizro-core/docs/pages/user-guides/graph-table-actions.md
antonymilne: Not sure. There's a ticket for it this sprint ([#2171](https://github.com/McK-Internal/vizro-internal/issues/2171)) currently assigned to @huong-li-nguyen but that could change. I don't mind whether we do it before or after a release.

vizro-core/docs/pages/user-guides/graph-table-actions.md
stichbury: My 2p is that we should write the docs to be findable, thus answer the queries the readers (humans or AI) are looking to answer. If they're more likely to say cross-highlight, then use that terminology. But you could have a section that is titled "Beyond cross highlight: Vizro's capability for cross parameter" or similar. And set out the extra functionality there.

maxschulz-COL:
Will take a look tomorrow morning

vizro-core/src/vizro/actions/_set_control.py
petar-qb: Also curious what other reviewers think about this. My opinion is that users will easier digest setting the `column_name` setting to the `action.value` and `graph.figure.custom_data` than to dive into `clickData["points"][0]` itself and its structure.

vizro-core/docs/pages/user-guides/graph-table-actions.md
petar-qb: I like this way of thinking! Personally, I’d prefer to officially call it "cross-parameter" for the reasons I mentioned in the comment above, while still mentioning "cross-highlight" in the title and maybe once or twice in the text for easier search. My sense is that only a small portion (`< 20%`) of our users are familiar with the term "cross-highlight". Really curious to hear what others think.

vizro-core/docs/pages/user-guides/actions.md
```diff
@@ -92,7 +92,9 @@ When you click the "Export data" button, the data for all graphs, tables and fig
 
 ## Trigger an action with a graph
 
-This is already possible, and documentation is coming soon!
+Here is an example that [performs a cross-filter](graph-table-actions.md#cross-filter-from-graph) between a graph and a table. When you click on a box in the graph, the table is cross-filtered to show data for only one sex.
+
+TODO NOW: copy and paste example in here once done screenshot
```
maxschulz-COL: TODO

vizro-core/docs/pages/user-guides/controls.md
```diff
@@ -1,9 +1,26 @@
 # Controls
 
-Vizro has a range of controls to make it simple to add common functionality to your app:
+Vizro supports _controls_ to perform common business intelligence (BI) operations. The following [components](components.md) are reactive to controls:
```
maxschulz-COL:
```suggestion
Vizro supports _controls_ to perform common business intelligence (BI) operations. This guide is an overview of the various ways you (the app developer) can configure _controls_ that the app consumer can use to _control_ the components he sees on the screen.

The following [components](components.md) are reactive to controls:
```

vizro-core/docs/pages/user-guides/filters.md
```diff
@@ -1,8 +1,17 @@
...
 
-The [`Page`][vizro.models.Page] model accepts the `controls` argument, where you can enter a [`Filter`][vizro.models.Filter] model. This model enables the automatic creation of [selectors](selectors.md) (for example, `Dropdown` or `RangeSlider`) that operate on the charts/components on the screen.
+- [built-in graphs](graph.md) and [custom graphs](custom-charts.md)
+- [built-in tables](table.md) and [custom tables](custom-tables.md)
+- [built-in figures](figure.md) and [custom figures](custom-figures.md)
+
+It is possible to add filters to a [page](pages.md) or [container](container.md#add-controls-to-container). Both the [`Page` model][vizro.models.Page] and the [`Container` model][vizro.models.Container] have an optional `controls` argument where you can give any number of controls including filters. A filter uses the [`Filter` model][vizro.models.Filter] to filter the `data_frame` of the `figure` function of a target component model such as [`Graph`][vizro.models.Graph].
```
maxschulz-COL:
```suggestion
It is possible to add filters to a [page](pages.md) or [container](container.md#add-controls-to-container). Both the [`Page` model][vizro.models.Page] and the [`Container` model][vizro.models.Container] have an optional `controls` argument where you can give any number of controls - including filters. A filter uses the [`Filter` model][vizro.models.Filter] to filter the `data_frame` of the `figure` function of a target component model such as [`Graph`][vizro.models.Graph].
```
or comma? Not sure :)

vizro-core/docs/pages/user-guides/graph-table-actions.md
maxschulz-COL: I am not strongly opinionated on the main terminology, I think it is done well. I think Jo has good points, so would lean towards that slightly.

@antonymilne To your question: exactly, I think this is the sort of thing I think we were missing or deviating too much from. Very good to see this happen. Eventually we could link to multiple user guides from the API docs (maybe we do already) that show you the various things you can do with model ABCD.

Overall this fits super well in the category Graph and Table interactions and as long as we have a good way to make this findable people will quickly see that they are in the right place!

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
+    1. Remember that if `targets` is not explicitly specified, a filter [targets all components on the page whose data source includes `column`](filters.md#basic-filters).
 
-```py
-Graph(figure=px.scatter(..., custom_data=["continent"]))
-```
+1. Call `set_control` in the `actions` argument of the source [`Graph`][vizro.models.Graph] or [`AgGrid`][vizro.models.AgGrid] component that triggers the cross-filter.
 
-Selecting a data point with a corresponding value of "Africa" in the continent column will result in filtering the data of target charts to show only entries with "Africa" in the continent column. The same applies when providing multiple columns in `custom_data`.
+    1. Set `control` to the ID of the filter.
+    1. Set `value`. The format of this depends on the source model and is given in the [API reference][vizro.actions.set_control], but it is often `column` of the filter.
```
maxschulz-COL: This is by far the hardest concept in my eyes, maybe we can expand on it:
Something along the lines of:

Think of it like an instruction: From which column of the source data do I take my **value** that I set the controls selector to.

I think I would have like a sentence like that or similar.

Maybe then further:

In this example that means: When the app user clicks a point, Vizro checks what row in the source data the clicked point point corresponds to and take the **value** of the `species` column of that row to update the control.

vizro-core/docs/pages/user-guides/graph-table-actions.md
maxschulz-COL: I would have skimmed over it (so no strong opinion) but I think @petar-qb is right. These are developer docs, and really its not so important. Maybe a small hint once?

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
+            title: Cross-filter from graph to table
+        ```
+
+    === "Result"
+
+        TODO NOW: screenshot
+
+When you click on a box in the graph, the table is cross-filtered to show data for only one sex, which is the `y` variable for the plot. This works because `"y"` exists in [Plotly's `clickData`](https://dash.plotly.com/interactive-graphing) for this chart.
+
+??? details "Behind the scenes mechanism"
```
maxschulz-COL, line 220: I think this is great - but maybe some of this is already warranted in a less technical form in the concept section above.

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
     def my_custom_chart(data_frame, custom_data, **kwargs):
         return px.scatter(data_grame, custom_data=custom_data, **kwargs)
+    ```
 
-    ...
+In general, the data contained in [Plotly's `clickData`](https://dash.plotly.com/interactive-graphing) depends on the type of chart used. The rules for how `value` is interpreted by `set_action` are:
 
-    vm.Graph(figure=my_custom_chart(df, custom_data=['continent'], actions=[...]))
+1. If the graph includes `custom_data` then interpret the `value` as a column name and attempt to find it in `custom_data`.
+1. If the graph does not include `custom_data` or the `value` is not specified in `custom_data` then Vizro [traverses a Box](https://github.com/cdgriffith/Box/wiki/Types-of-Boxes#box-dots) formed from the data [`clickData["points"][0]`](https://dash.plotly.com/interactive-graphing). For example:
```
maxschulz-COL:
```suggestion
1. If the graph does not include `value` as a column in `custom_data` then Vizro [traverses a Box](https://github.com/cdgriffith/Box/wiki/Types-of-Boxes#box-dots) formed from the data [`clickData["points"][0]`](https://dash.plotly.com/interactive-graphing). For example:
```

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
+    - `value="x"` is equivalent to looking at `clickData["points"][0]["x"]`.
+    - `value="key.subkey[1]"` is equivalent to looking at `clickData["points"][0]["key"]["subkey"][1]`.
```
maxschulz-COL, line 326:
```suggestion
    - `value="key.subkey[1]"` is equivalent to looking at `clickData["points"][0]["key"]["subkey"][1]`.
    
Based on the chosen source graph, the developer can thus decide which available attribute to set as `value`.
```

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
 
-    ```
+### Cross-filter between containers
 
-Here is an example of how to configure a chart interaction when the source is an [`AgGrid`][vizro.models.AgGrid] component.
+A cross-filter often work best when used [inside a container](container.md#add-controls-to-container). This typically makes it clearer which components the filter applies to, especially when the [container is styled](container.md#styled-containers).
```
maxschulz-COL:
```suggestion
A cross-filter often works best when used [inside a container](container.md#add-controls-to-container). This typically makes it clearer which components the filter applies to, especially when the [container is styled](container.md#styled-containers).
```

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
 
-!!! example "AgGrid `filter_interaction`"
+For example, let us rearrange the above example of a [cross-filter from a table](#cross-filter-from-table) into containers. The rearrangement here is purely visual to give a better user experience; the `va.set_control` is configured exactly the same way and behaves identically while the dashboard is running.
```
maxschulz-COL: This is so so so cool!

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
+            title: Cross-filter between containers
+        ```
+
+    === "Result"
+
+        TODO NOW: screenshot
+
+### Cross-filter between pages
+
+You can perform a cross-filter where the target components are on a different page from the source. The use of `va.set_control` is identical, but the intermediate filter must have [`show_in_url=True`](run-deploy.md#shareable-url).
```
maxschulz-COL: Sorry to write again, but also so so cool!

vizro-core/docs/pages/user-guides/graph-table-actions.md
maxschulz-COL: I think it is important, I think the below might be what people want, so worth mentioning limitations

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1,199 +1,617 @@
...
+
+Multiple cross-filters are similarly possible [from a table](#cross-filter-from-table):
+
+```python
+vm.AgGrid(
+    ..., actions=[va.set_control(control="day_filter", value="day"), va.set_control(control="sex_filter", value="sex")]
+)
+```
+
+However, it is not yet possible to cross-filter from a pivot table according to the row and column of the clicked cell.
```
maxschulz-COL, line 622: That would be to take the row and column headers right?

vizro-core/docs/pages/user-guides/run-deploy.md
```diff
@@ -345,3 +345,70 @@ WASM is a low-level binary instruction format that enables high-performance exec
 [Pyodide is a Python distribution compiled to WebAssembly, enabling Python execution in a browser’s JavaScript runtime](https://hacks.mozilla.org/2019/04/pyodide-bringing-the-scientific-python-stack-to-the-browser/). It includes a standard Python interpreter, common scientific libraries (NumPy, Pandas, Matplotlib, etc.), and interoperability with JavaScript within the browser environment. In the context of Vizro, Pyodide enables the execution of Dash callbacks, component updates, and external Python libraries directly in the client’s browser, without a persistent backend.
 
 While potential advantages include serverless execution, reduced latency and easier "deployment", the disadvantages include limited performance compared to a native server, limited library support, and memory constraints.
+
+### Shareable URL
```
maxschulz-COL, line 349: Yes! Do we need to remove something somewhere else?

vizro-core/mkdocs.yml
```diff
@@ -124,7 +124,8 @@ markdown_extensions:
             kwds:
               type: vizro
               requirements: |
-                vizro==0.1.45
+                # TODO NOW: revert this
```
maxschulz-COL: TODO

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -27,20 +27,68 @@ def _encode_to_base64(value):
 
 
 class set_control(_AbstractAction):
-    """Sets the value of a control based on data from the trigger.
+    """Sets the value of a control.
+
+    Abstract: Usage documentation
+        [Graph and table interactions](../user-guides/graph-table-actions.md)
```
maxschulz-COL, line 33: Maybe the new show_url? Maybe not, but generally I think it might be useful to do generous cross linking

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -27,20 +27,68 @@ def _encode_to_base64(value):
...
+
+    The following Vizro models support `set_control`:
+
+    * [`AgGrid`][vizro.models.AgGrid]: triggers `set_control` when user clicks on a row in the table. The `value`
+    argument specifies the column of the value that is sent to the `target` control.
```
maxschulz-COL:
```suggestion
    argument specifies the column of the value that is sent to the `target` control (while the row is determined by the user's click).
```
Maybe to make it super clear? Ignore if you disagree

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -27,20 +27,68 @@ def _encode_to_base64(value):
...
+    * [`Graph`][vizro.models.Graph]: triggers `set_control` when user clicks on data in the graph. The `value` argument
+    can be used in two ways to specify what value to send to the `target` control:
+
+        * String to [traverse a Box](https://github.com/cdgriffith/Box/wiki/Types-of-Boxes#box-dots) that contains the
```
maxschulz-COL, line 43: I think this is impossible to understand without the user guide, but maybe ok for an API documentation?

vizro-core/src/vizro/actions/_set_control.py
maxschulz-COL: I would swap, as it's impossible to understand the Box comment anyway I think

maxschulz-COL:
```suggestion
        * Column from which to take the value  (while the row is determined by the user's click). This requires you to set `custom_data` in the graph's `figure` function.
```

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -27,20 +27,68 @@ def _encode_to_base64(value):
...
+        trigger data [`clickData["points"][0]`](https://dash.plotly.com/interactive-graphing), for example `"x"`.
+        * Column from which to take the value. This requires you to set `custom_data` in the graph's `figure` function.
 
     Args:
-        control (ModelID): Filter or Parameter component id to be affected by the trigger. If the control is on a
-            different page to the trigger then it must have `show_in_url=True`.
-        value (str): TODO AM: ADD DESCRIPTION
+        target (ModelID): Control whose value is set. If this is on a different page from the trigger then it must have
```
maxschulz-COL:
```suggestion
        control (ModelID): Control whose value is set. If this is on a different page from the trigger then it must have
```

vizro-core/docs/pages/user-guides/graph-table-actions.md
antonymilne: Yes exactly! I think this will be possible in future but not quite yet.

vizro-core/src/vizro/actions/_set_control.py
antonymilne: Yes, agree this is impossible to understand without the user guide but I think it's ok to be very technical and concise in API docs. Otherwise I think they will just get too lengthy.

vizro-core/docs/pages/user-guides/graph-table-actions.md
antonymilne: I have switched this order everywhere.

petar-qb: Great thanks!

vizro-core/docs/pages/user-guides/graph-table-actions.md
antonymilne: I have changed it so it's just mentioned in passing once instead of being repeated lots.

vizro-core/src/vizro/actions/_set_control.py
antonymilne: I have swapped this everywhere.

PR #1381, opened 2025-09-09
[Docs] Fix API docs issues from Griffe

stichbury:
## Notice

### DIFF
--- a/vizro-core/src/vizro/_vizro.py
+++ b/vizro-core/src/vizro/_vizro.py
@@ -31,13 +31,9 @@ class Vizro:
     """The main class of the `vizro` package."""
 
     def __init__(self, **kwargs):
-        """Initializes Dash app, stored in `self.dash`.
-
-        Args:
-            **kwargs : Passed through to `Dash.__init__`, e.g. `assets_folder`, `url_base_pathname`. See
-                [Dash documentation](https://dash.plotly.com/reference#dash.dash) for possible arguments.
-
-        """
+        """Initializes Dash app, stored in `self.dash`."""
+        # **kwargs is passed through to `Dash.__init__`, e.g. `assets_folder`, `url_base_pathname`.
+        # See [Dash documentation](https://dash.plotly.com/reference#dash.dash) for possible arguments.
         # Set suppress_callback_exceptions=True for the following reasons:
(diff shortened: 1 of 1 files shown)


petar-qb (approved):
Thanks @stichbury for the fix 🚀 🚀 

I like the solution as it introduces minimal changes in our code. If the lack of args/kwargs support in docstrings becomes an issue, we can raise it with griffe or explore alternatives. For now, this works great IMO.

vizro-core/src/vizro/_vizro.py
```diff
@@ -31,13 +31,9 @@ class Vizro:
...
     def __init__(self, **kwargs):
-        """Initializes Dash app, stored in `self.dash`.
-
-        Args:
-            **kwargs : Passed through to `Dash.__init__`, e.g. `assets_folder`, `url_base_pathname`. See
-                [Dash documentation](https://dash.plotly.com/reference#dash.dash) for possible arguments.
-
-        """
+        """Initializes Dash app, stored in `self.dash`."""
+        # **kwargs is passed through to `Dash.__init__`, e.g. `assets_folder`, `url_base_pathname`.
```
maxschulz-COL: Could we also add this to the explanation, but just not as "Args"

antonymilne: You mean include it in the `"""` docstring? If so then I agree 👍 That way it will still show in our API docs which would be useful. I would put `**kwargs` in backticks then also since it will appear in the docs.

vizro-core/src/vizro/_vizro.py
```diff
@@ -123,13 +119,7 @@ def build(self, dashboard: Dashboard):
...
 
     def run(self, *args, **kwargs):  # if type annotated, mkdocstring stops seeing the class
-        """Runs the dashboard.
-
-        Args:
-            *args : Passed through to `dash.run`.
-            **kwargs : Passed through to `dash.run`.
-
-        """
+        """Runs the dashboard and passes args and kwargs through to `dash.run`."""
```
antonymilne:
```suggestion
        """Runs the dashboard and passes `args` and `kwargs` through to `Dash.run`."""
```

stichbury: Done!

PR #1392, opened 2025-09-12
[Feat] Officially enable Figure model with vizro-mcp

maxschulz-COL:
Closes https://github.com/McK-Internal/vizro-internal/issues/1928

## Description
See tickets

### How to review
- checkout the branch
- follow this guide: https://vizro.readthedocs.io/projects/vizro-mcp/en/latest/pages/explanation/contribute/?h=developer#how-to-use-a-developer-build-of-vizro-mcp
- ask for a dashboard that involves KPI cards

## Notice


vizro-mcp/src/vizro_mcp/server.py
```diff
@@ -70,7 +72,7 @@ class ModelJsonSchemaResults:
 
 # TODO: check on https://github.com/modelcontextprotocol/python-sdk what new things are possible to do here
 mcp = FastMCP(
-    name="MCP server to help create Vizro dashboards and charts",
+    name=f"MCP server to help create Vizro dashboards and charts. Based on vizro=={vizro.__version__}",
```
lingyielia: How could the MCP know which feature/syntax belongs to which Vizro version? Is this version info been picked up by MCP or it's only user facing info?

vizro-mcp/src/vizro_mcp/_schemas/schemas.py
```diff
@@ -58,6 +59,21 @@ class AgGridEnhanced(vm.AgGrid):
...
+class FigureEnhanced(vm.Figure):
+    """Figure model that allows to use dynamic figure functions."""
+
+    figure: dict[str, Any] = Field(
+        description=f"""This is the figure function to be displayed.
+
+Only use arguments from the below mapping of _target_ to figure function documentation:
+
+{FIGURE_NAMESPACE_FUNCTION_DOCS}"""
+    )
```
lingyielia, line 74: Oh, this is very neat! I thought we need bit of more prompt engineering but you are pulling from the functions directly.

I assume the only trade off is that MCP doesn't know kpi_card and kpi_card_reference exist upfront. I only tested one full round and the steps I took are as follow:
- prompt "create more kpi cards on each page." MCP created custom graphs as KPI cards like:
```
@capture("graph")
def overview_revenue_kpi(data_frame):
    df = data_frame.dropna(subset=["Total"]).copy()
    total_revenue = df["Total"].sum()
    
    fig = go.Figure(go.Indicator(
        mode="number",
        value=total_revenue,
        number={"prefix": "$", "valueformat": ",.0f", "font": {"size": 40}},
        title={"text": "Total Revenue", "font": {"size": 20}},
        domain={"x": [0, 1], "y": [0, 1]}
    ))
    
    fig.update_layout(
        height=200,
        margin=dict(l=20, r=20, t=40, b=20)
    )
    
    return fig
```

- a followup prompt "can you try use Vizro built-in KPI cards whenever possible", then it started with search model "KPI" then `web search` to confirm the correct functions, and MCP call the `Figure` model correctly.

Screenshot 2025-09-16 at 12 07 44 PM

lingyielia (approved):
The KPI cards works well. One remaining question is that whether MCP could be instructed to look at Figure model more directly

vizro-mcp/src/vizro_mcp/_schemas/schemas.py
maxschulz-COL: Yes very good point, I have experienced that as well... I guess we could add them to the overview as well, let me try that! Crazy that it went to a web-search for you haha

maxschulz-COL: @antonymilne This is also to show that the actions being Pydantic models is _really_ good, and the `Figure` function not being it makes it more annoying. Maybe this is something we could consider in the future for `vizro.figures`

vizro-mcp/src/vizro_mcp/server.py
maxschulz-COL: I probably should delete this, I was experimenting with how to best communicate the Vizro version to the user actually. Currently I struggle to for the latest install of Vizro without pinning Vizro. It seems that there is some caching when you run MCP servers such that it doesn't upgrade.

Example:
In our `pyproject.toml` when we say `vizro>=0.1.42`, it may pick `vizro==0.1.42` even if newer versions are out because it has picked it in the past.

Any ideas on how to run an MCP server that picks always the latest available dependency?

vizro-mcp/src/vizro_mcp/_schemas/schemas.py
maxschulz-COL: Ok I added them to the overview, and I think it works well!

vizro-mcp/src/vizro_mcp/server.py
maxschulz-COL: I left it in for now so that we have visibility on the server installed vizro version

vizro-mcp/src/vizro_mcp/_utils/prompts.py
```diff
@@ -50,11 +51,28 @@
...
     "main": [vm.Dashboard, vm.Page],
-    "components": [vm.Card, vm.Button, vm.Text, vm.Container, vm.Tabs, vm.Graph, vm.AgGrid],  #'Figure', 'Table'
+    "static components": [
+        vm.Card,
+        vm.Button,
+        vm.Text,
+        vm.Container,
+        vm.Tabs,
+    ],
+    "dynamic components": [vm.Figure, vm.Graph, vm.AgGrid],
```
huong-li-nguyen, line 75: Do we not support the vm.Table anymore?

vizro-mcp/src/vizro_mcp/_utils/prompts.py
```diff
@@ -64,10 +82,10 @@
         vm.DatePicker,
         vm.Slider,
         vm.RangeSlider,
-        vm.DatePicker,
```
huong-li-nguyen, line 67: Shall we add the vm.Switch here too, since it's also a selector?

huong-li-nguyen (approved):
I've only looked at code, since @lingyielia already tested :) 

Had a few questions, but nothing major

vizro-mcp/src/vizro_mcp/_utils/prompts.py
maxschulz-COL: Trying to get away from Table. It would still pass validation, but actively not steering model towards it. I think it's better to have one source of truth. Would you disagree?

vizro-mcp/src/vizro_mcp/_utils/prompts.py
maxschulz-COL: Yes! Good spot

maxschulz-COL: Will add in next PR, is easier :)

PR #1405, opened 2025-09-17
[Feat] Improve Layout prompting for better OTS results

maxschulz-COL:
## Description
Closes https://github.com/McK-Internal/vizro-internal/issues/2202

This PR is pretty much a prompt only change.

### How to review
- checkout the branch
- follow this guide: https://vizro.readthedocs.io/projects/vizro-mcp/en/latest/pages/explanation/contribute/?h=developer#how-to-use-a-developer-build-of-vizro-mcp
- play a bit with requesting dashboard without mentioning layout, and with mentioning layout
- see if you think the results are more or less acceptable, but certainly better than in the currently released version

## Notice


vizro-mcp/src/vizro_mcp/_utils/prompts.py
```diff
@@ -82,12 +83,39 @@ class HasNameAndDoc(Protocol):
...
         vm.Slider,
         vm.RangeSlider,
+        vm.Switch,
     ],
     "navigation": [vm.Navigation, vm.NavBar, vm.NavLink],
     "additional_info": [vm.Tooltip],
     "functions available for vm.Figure(...,figure=...) model": [vf.__dict__[func] for func in vf.__all__],
 }
 
+LAYOUT_INSTRUCTIONS = """Some extra information about Layout in Vizro:
```
lingyielia, line 94: haven't tried it hands-on. One qq, how did you decide which piece of instruction should we include here?

Some prompts are generic enough:
"Use when you need precise alignment across both rows and columns." or "Use Flex for most dashboards, use Grid only for strict row x column alignment or complex layouts"

Some seems too strict: "Cards: Horizontal row, wrap naturally. Maintain consistent spacing (gap)" -> which might not be always the case.

maxschulz-COL: Hm you may be right, let me remove some of that actually and see if it still works the same

vizro-mcp/src/vizro_mcp/_utils/prompts.py
```diff
@@ -82,12 +83,39 @@ class HasNameAndDoc(Protocol):
...
+Overall:
+- Use Flex for most dashboards, use Grid only for strict row x column alignment or complex layouts
+- Often good strategy: group elements in a `Container` and either use `Flex` inside container
```
lingyielia: Do you mean
“A good strategy is to group elements inside a Container, and then use Flex for arranging elements inside the container, while using Grid to organize the overall page — or do the reverse.”
or
“It is often a good strategy to group elements in a Container, and then you can either:
(a) use Flex inside the container and Grid for the page, or
(b) use Grid inside the container and Flex for the page.”?

vizro-mcp/src/vizro_mcp/_utils/prompts.py
```diff
@@ -82,12 +83,39 @@ class HasNameAndDoc(Protocol):
...
+        vm.Switch,
     ],
     "navigation": [vm.Navigation, vm.NavBar, vm.NavLink],
     "additional_info": [vm.Tooltip],
     "functions available for vm.Figure(...,figure=...) model": [vf.__dict__[func] for func in vf.__all__],
 }
 
+LAYOUT_INSTRUCTIONS = """Some extra information about Layout in Vizro:
+Overall:
+- Use Flex for most dashboards, use Grid only for strict row x column alignment or complex layouts
```
lingyielia: I started with a request without mention layout at all ("create a dashboard for this data"). And it returns code without layout (defaults to grid, stacking all components). This is annoying, as although there is "Use Flex for most dashboards", it still disrespect the overall rule.

In this case, do you think it makes sense to further simplify the layout prompt?

like "always specify layout=vm.Flex(...) unless there is a good reason not to" -> so it will not fallback to Grid as default

maxschulz-COL: Hm yes I have had this too... I am a bit worried of always using `Flex` as it is not always the best option... In some sense it would be better if our default was `Flex` but alas it's not.. But maybe I should try that

maxschulz-COL: Done

PR #1409, opened 2025-09-19
[Fix] Pin `click` version in CI and `hatch.toml`

huong-li-nguyen:
## Description
- Pin `click` temporarily in CI due to https://github.com/pypa/hatch/issues/2050
- Pin `click` temporarily in `hatch.toml`

## Notice

### DIFF
--- a/vizro-core/.readthedocs.yaml
+++ b/vizro-core/.readthedocs.yaml
@@ -10,7 +10,9 @@ build:
   tools:
     python: "3.12"
   commands:
-    - pip install hatch
+    # Remove the `click` installation constraint once the
+    # https://github.com/pypa/hatch/issues/2050 is resolved.
+    - pip install hatch "click<8.3.0"
     - cd vizro-core/ && hatch run docs:pip tree
     - cd vizro-core/ && hatch run docs:build
     - cd vizro-core/ && hatch run docs:logo-check


petar-qb:
I guess that we should add it in the `pyproject.toml` dependencies (or maybe `hatch.toml`)?

maxschulz-COL:
> I guess that we should add it in the `pyproject.toml` dependencies (or maybe `hatch.toml`)?

Just wanted to comment the same :)

huong-li-nguyen:
> * docs/readthedocs.org:vizro

Updated all workflows, but I couldn't find the workflow file to update for the docs build? https://app.readthedocs.org/projects/vizro/builds/29635024/

One of the vizroAI tests is failing, but it has nothing to do with this PR, but some pydantic issue. Can you check @maxschulz-COL ?

petar-qb:
@huong-li-nguyen Could you try adding the pinned click in the `vizro-core/hatch.toml` under the `[envs.docs]` dependecies section too? 

Also, I guess that you can remove all the changes from the `.yml` files and leave only those you made in the `hatch.toml` files.

huong-li-nguyen:
> @huong-li-nguyen Could you try adding the pinned click in the `vizro-core/hatch.toml` under the `[envs.docs]` dependecies section too?
> 
> Also, I guess that you can remove all the changes from the `.yml` files and leave only those you made in the `hatch.toml` files.

Ah, i thought we need both! Let me check 👍

huong-li-nguyen:
> * e2e vizro dom elements tests for Vizro / test-e2e-vizro-dom-elements on Py3.12 upper bounds-1 (pull_request)

@petar-qb - I checked removing from one yml file, but we do really need both! So adding the change in CI on every yaml file and the hatch.toml: https://github.com/mckinsey/vizro/pull/1409/commits/0e0849abbd84edb43fab430d25a904dc2b4fbf9c

Also this still doesn't fix the docs 🤔 Where can I update that workflow file? :D

huong-li-nguyen:
> I just checked my assumptions in the #1410 and I was wrong 😄, you can't only change it in the `hatch.toml` files. I'm pretty sure that the `click` version used by `hatch` can be set somewhere in a single line, but we just don't now exactly how 😄
> 
> However, pinning it all over the place is good enough solution for now. 💹 Thanks for handling it.

If you feel better about it, I think that's the way to go actually. Because I saw other repositories do the same. So update hatch.toml or pyproject.toml and then all the single yml files as well 😄

petar-qb:
> > I just checked my assumptions in the #1410 and I was wrong 😄, you can't only change it in the `hatch.toml` files. I'm pretty sure that the `click` version used by `hatch` can be set somewhere in a single line, but we just don't now exactly how 😄
> > However, pinning it all over the place is good enough solution for now. 💹 Thanks for handling it.
> 
> If you feel better about it, I think that's the way to go actually. Because I saw other repositories do the same. So update hatch.toml or pyproject.toml and then all the single yml files as well 😄

`hatch` library has `"click>=8.0.6"` as its dependency in its `pyproject.toml` ([here](https://github.com/pypa/hatch/blob/master/pyproject.toml#L42)). I thought that setting the pinned click version in our `pyproject.toml` will force the `hatch` to use that `click` version, but unfortunately that's not true..

I'm wondering if there is any way to specify the `hatch` dependencies through `vizro` configuration files instead of explicitly setting it as the `hatch run` argument. I know the problem is solved, but I'm still wondering if that's possible 🤓 @antonymilne do you know some hack?

antonymilne:
Thanks for sorting this out @huong-li-nguyen and everyone else!

Just to comment on a few things here since I think there's a bit of confusion. There are three different sorts of dependencies in question. Going from top level to lowest level, these are:
1. dependencies of `vizro` i.e. what gets installed when a user does `pip install vizro`. These are specified in `pyproject.toml`.
1. development dependencies of `vizro` i.e. what gets installed automatically when we do `hatch run example` etc. These are not user-facing. These are specified in `hatch.toml`
1. dependencies of `hatch` i.e. what gets installed when you `pip install hatch`. They are also not user-facing. These are manually specified when we do `pip install hatch` e.g. in github workflows, or whenever you first installed `hatch` on your computer.

`click` is a dependency of `hatch` and hence in category 3 here. So technically I think we should only need to change all `pip install hatch` to `pip install hatch click!=...`. If for some reason that isn't sufficient then it's also fine to add `click!=...` to our development environments in `hatch.toml`. But we should never write it in `pyproject.toml` like in https://github.com/mckinsey/vizro/pull/1410 because then all users who do `pip install vizro` would then get `click` as a dependency, which is not right.

As for @petar-qb's question of whether there's a better way to specify `click!=...` everywhere rather than manually copying and pasting into every `pip install hatch`, the answer is no. At least not without doing some other copy-paste solution e.g. to change some `PIP` environment variable everywhere to specify it.

PR #1428, opened 2025-09-30
Feat/make param filter hidden

huong-li-nguyen:
## Notice


vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -303,7 +308,7 @@ def build(self):
             selector_build_obj.children.append(dcc.Store(id=f"{selector.id}_guard_actions_chain", data=False))
 
         if not self._dynamic:
-            return html.Div(id=self.id, children=selector_build_obj)
+            return html.Div(id=self.id, children=selector_build_obj, hidden=self.hidden)
```
huong-li-nguyen: I initially added it to the dcc.Loading as well, but the dcc.Loading component doesn't actually have a hidden attribute. I think for dynamic filters we probably also don't need them, as we want these visible? Otherwise, I could also hide it via this argument from the dcc.Loading spinner:

`display (a value equal to: 'auto', 'show' or 'hide'; default 'auto'): Setting display to "show" or "hide" will override the loading state coming from dash-renderer.`

and then we would have to extend our `_all_hidden` logic, but at the moment I don't think it's worth doing?

petar-qb: >  I think for dynamic filters we probably also don't need them, as we want these visible?

It's not the clearest to me why we wouldn't want this?

Could you try to wrap `dcc.Loading` into empty `html.Div` and set the hidden there, or just by setting `className=”d-none”` to `dcc.Loading` directly. I haven’t tried it and not sure would it help or hinder the current setup, but I guess it’s worth trying.

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -12,6 +13,18 @@
 logger = logging.getLogger(__name__)
 
 
+def _all_hidden(components: Union[Component, list[Component]]):
```
petar-qb, line 16: I pushed the scratch PR that has three pages:
- page with controls shown
- page with no controls
- page with controls hidden

Ideally, the second and the third page should look the same, but even when I comment out the dynamic controls (that's currently not supported) on the third page, I see the horizontal line in the container above the chart. Should we utilise this `_all_hidden` function when creating the container too?

vizro-core/src/vizro/models/_controls/filter.py
antonymilne: Agree with @petar-qb - _in principle_ I don't see any reason why we wouldn't want to hide a dynamic filter, just like any other control.

In practice, if it's hard to do then I'm not worried about implementing it, but if it's reasonably easy then I think we should do it.

antonymilne (commented):
Generally looks good, I'm happy to go along with something like this 👍 

What do you think of `show_on_page: bool = True`? It would match `show_in_url: bool = False`. If you or @petar-qb don't like that then I would prefer `visible: bool = True` rather than `hidden: bool = False` to avoid the double negative which I think is confusing (I know it matches HTML elements, but still it's confusing I think).

FYI I linked this to ticket https://github.com/McK-Internal/vizro-internal/issues/2191. It won't complete that ticket but it's part of this. I think next step would be to work out how/whether we should have some logic that does "if any controls are hidden then include reset button". There's a question there of _which_ controls it would reset - all controls or just the hidden ones?

vizro-core/src/vizro/models/_models_utils.py
huong-li-nguyen: Yeeees, let me properly fix all of these cases if we decide to go for it 👍

vizro-core/src/vizro/models/_controls/filter.py
huong-li-nguyen: It works if we add a parent div around it, I think that's fine, even though semantically not optimal 😬  "d-none" could also work, because I think we've wrote the _all_hidden function to capture that as well. Let me check.

huong-li-nguyen:
> What do you think of `show_on_page: bool = True`? It would match `show_in_url: bool = False`. If you or @petar-qb don't like that then I would prefer `visible: bool = True` rather than `hidden: bool = False` to avoid the double negative which I think is confusing (I know it matches HTML elements, but still it's confusing I think).
> 

I don't feel strongly either way. I lean slightly toward `visible` since controls could be in modals or containers - in the case of modals they are outside the page, and in the case of containers `show_in_container` might make more sense there. But honestly, that's probably edge cases. So I'm good with whatever - I'll keep it as is for now though until we decide 👍

petar-qb:
> > What do you think of `show_on_page: bool = True`? It would match `show_in_url: bool = False`. If you or @petar-qb don't like that then I would prefer `visible: bool = True` rather than `hidden: bool = False` to avoid the double negative which I think is confusing (I know it matches HTML elements, but still it's confusing I think).
> 
> I don't feel strongly either way. I lean slightly toward `visible` since controls could be in modals or containers - in the case of modals they are outside the page, and in the case of containers `show_in_container` might make more sense there. But honestly, that's probably edge cases. So I'm good with whatever - I'll keep it as is for now though until we decide 👍

I agree with the "double negating", and "show_in_container" arguments, so let's proceed with `"visible: bool = True"`.

vizro-core/docs/pages/user-guides/parameters.md
```diff
@@ -198,5 +198,12 @@ Note that in the above example, one parameter affects multiple targets.
 
 If you use [dynamic data](data.md/#dynamic-data) that can be updated while the dashboard is running then you can pass parameters to the dynamic data function to alter the data loaded into your dashboard. For detailed instructions, refer to the section on [parametrized data loading](data.md/#parametrize-data-loading).
 
+## Further customization
```
huong-li-nguyen, line 201: I was thinking about adding a separate section with code examples, but that doesn't really make sense right now since none of them would actually be suitable cases where I'd suggest hiding the control. For now, I think we should mention that it's possible. But once we create proper code examples where hiding the controls is recommended (cross-highlighting and so on), we can add those and link back here. But I wouldn't just create a random code example to hide the control.

vizro-core/docs/pages/user-guides/parameters.md
```diff
@@ -198,5 +198,12 @@ Note that in the above example, one parameter affects multiple targets.
...
+
+For further customizations, you can always refer to the [`Parameter`][vizro.models.Parameter] model reference and the [guide to selectors](selectors.md). Some popular choices are:
+
+- specify configuration of the `selector`, for example `multi` to switch between a multi-option and single-option selector, `options` for a categorical filter or `min` and `max` for a numerical filter
+- hiding the parameter from the dashboard interface by setting `visible=False`
+
```
petar-qb: I would replace this entire section with something simple like:

```
For further customizations, refer to the [guide to selectors](selectors.md), and the [`Parameter`][vizro.models.Parameter] model reference.
```

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
```diff
@@ -1080,9 +1092,29 @@ def test_dynamic_filter_build(self, test_column, test_selector, gapminder_dynami
...
             color="grey",
             overlay_style={"visibility": "visible"},
+            className="",
+        )
+
+        assert_component_equal(result, expected, keys_to_strip={"children"})
+
+    @pytest.mark.usefixtures("managers_one_page_two_graphs_with_dynamic_data")
+    @pytest.mark.parametrize("visible", [True, False])
+    def test_dynamic_filter_build_visible(self, gapminder_dynamic_first_n_last_n_function, visible):
```
petar-qb, line 1102: There's already a test for the dynamic filer when the filter is visible (test_dynamic_filter_build), so I guess that here we can test just what happens if it's not visible. 

```suggestion
    def test_dynamic_filter_build_not_visible(self,gapminder_dynamic_first_n_last_n_function):
```

It looks like GitHub's multi-line suggestions aren’t working properly, but you can see what I was aiming for.

vizro-core/tests/unit/vizro/models/_controls/test_parameter.py
```diff
@@ -273,13 +273,26 @@ class TestParameterBuild:
...
+    @pytest.mark.usefixtures("managers_one_page_two_graphs")
+    @pytest.mark.parametrize("visible", [True, False])
+    def test_parameter_build_visible(self, visible):
+        test_input = vm.Checklist(options=["lifeExp", "gdpPercap", "pop"])
+        parameter = Parameter(id="parameter-id", targets=["scatter_chart.x"], selector=test_input, visible=visible)
+        page = model_manager["test_page"]
+        page.controls = [parameter]
+        parameter.pre_build()
+        result = parameter.build()
+        expected = html.Div(id="parameter-id", children=html.Div(children=[test_input.build()]), hidden=not visible)
```
petar-qb, line 295: Same here

```suggestion
    @pytest.mark.usefixtures("managers_one_page_two_graphs")
    def test_parameter_build_visible(self):
        test_input = vm.Checklist(options=["lifeExp", "gdpPercap", "pop"])
        parameter = Parameter(id="parameter-id", targets=["scatter_chart.x"], selector=test_input, visible=False)
        page = model_manager["test_page"]
        page.controls = [parameter]
        parameter.pre_build()
        result = parameter.build()
        expected = html.Div(id="parameter-id", children=html.Div(children=[test_input.build()]), hidden=True)
```

vizro-core/changelog.d/20250930_132715_90609403+huong-li-nguyen_make_param_filter_hidden.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Enable hiding the `Filter` or `Parameter` from the dashboard view by setting `visible=False`. ([#1428](https://github.com/mckinsey/vizro/pull/1428))
```
antonymilne:
```suggestion
- `Filter` and `Parameter` can be hidden by setting `visible=False`. ([#1428](https://github.com/mckinsey/vizro/pull/1428))
```

vizro-core/docs/pages/user-guides/controls.md
```diff
@@ -18,6 +18,8 @@ Vizro has two types of control:
 
 All controls have an [argument `selector`](selectors.md) that configures the visual interface for the control, for example a checklist or a range slider. The same selectors are available for all controls.
 
+You can control the visibility of controls using the `visible` argument. By default, controls are visible (`visible=True`), but you can hide them by setting `visible=False`. Hidden controls remain functional and can still be controlled through actions.
```
antonymilne: This page is getting a bit long without subsections - I'll write a new suggested version and push it here.

antonymilne (approved):
Tests and functionality looks great! 💯 I'll push some suggested docs changes in a minute.

vizro-core/docs/pages/user-guides/parameters.md
antonymilne: Definitely agree. We don't need an example for every type of argument in our narrative docs. Better instead to put the example on the cross-highlighting page and then link to it from here.

vizro-core/docs/pages/user-guides/parameters.md
antonymilne: This was copy and pasted from `filters.md`. I've made some changes to both pages now.

vizro-core/tests/unit/vizro/models/_controls/test_filter.py
huong-li-nguyen: In the previous test, the visible argument isn't explicitly set though. it’s just checking the default behavior when the argument isn't provided. I actually prefer keeping it the current way, since it makes the intent more explicit 🙂

Issue #1426, opened 2025-09-29
[Docs] llms.txt does not respect version number in provided links


maxschulz-COL:
### Have you already looked into this bug?

- [x] I've reviewed the [release notes](https://github.com/mckinsey/vizro/releases) for relevant migration steps
- [x] I've searched through [existing issues](https://github.com/mckinsey/vizro/issues) for similar reports
- [x] I've already searched online (e.g. Dash documentation) but couldn't find anything relevant

### Which package?

vizro

### Package version

0.1.46

### Description

Links within `llms.txt` are incoherent with the overall docs version of the URL

### How to Reproduce

Acces the docs at a specific version, e.g. https://vizro.readthedocs.io/en/0.1.45 then try to get the llms.txt: https://vizro.readthedocs.io/en/0.1.45/llms.txt

The links within the llms.txt will still carry `stable`.

### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/mckinsey/vizro/blob/main/CODE_OF_CONDUCT.md).

huong-li-nguyen:
@stichbury - could you take a look? :)

stichbury:
A possible solution for exposing versioned `llms.txt` is to add a link in a visible place for users/llm indexing. We could also add versioned links explicitly to a detailed sitemap (we don't build one at present but rely on the automated version from readthedocs) but this is probably not useful since crawlers reading sitemaps are not llms.

stichbury:
Made a PR to the plugin developer https://github.com/pawamoy/mkdocs-llmstxt/pull/25 to add a configuration option that would enable us to version the links inside `llms.txt`.

PR #1433, opened 2025-10-01
[Docs] Update `mkdocs.yml` to enable versioned links in `llms.txt`

stichbury:
## Description
Closes https://github.com/mckinsey/vizro/issues/1426
## Notice

### DIFF
--- a/vizro-core/docs/index.md
+++ b/vizro-core/docs/index.md
@@ -40,8 +40,6 @@ Vizro is an open-source Python-based toolkit. You can use it to build data visua
 
     [Vizro-AI](https://vizro.readthedocs.io/projects/vizro-ai/)
 
-    [`llms.txt`](llms.txt)
-
 - :fontawesome-solid-book-open:{ .lg .middle } __Find out more__
 
     ---


stichbury:
This PR adds links to `llms.txt` and `llms-full.txt` to the front page of the docs. See https://vizro--1433.org.readthedocs.build/en/1433/ for illustration.

Note that this points to the content built for the version the user is viewing, so makes it accessible for anyone unaware it's at the root of the versioned build. For crawlers that seek out `llms.txt` at the root of the docs project, this will still be pointed through redirect to `stable`, but making this addition enables an explicit level of versioning.

HOWEVER, there is a problem, which is that the `llms.txt` files are built by a plugin, which runs after the mkdocs check that each file is present. So it returns a warning for each file to say it's not present. When you suppress the `strict` flag, the build runs to completion and the warning is redundant because the file is built by the plugin eventually. I don't want us to take off the `strict` flag, obviously, but am unsure how else to suppress the warning so the build runs to completion. Any ideas @maxschulz-COL ?

maxschulz-COL:
I am not sure this solves the issue, nor whether this is what we should do? As previously mentioned, I am not too sure about `llms.txt`, but adding it to the front-page is even less what it's intended for.

The issue I was describing is that the `llms.txt` for a specific version does not show the correct links to a version, hence might be wrong or even outdated. It always points to stable

stichbury:
> The issue I was describing is that the `llms.txt` for a specific version does not show the correct links to a version, hence might be wrong or even outdated. It always points to stable

Understood. But we cannot version in the root of the Vizro docs ([as advised by RTD](https://github.com/readthedocs/readthedocs.org/issues/12321), we are using a redirect from the root to stable). The best way to flag that there actually _is_ a versioned llms.txt is to link it (making it visible to humans and bots alike).

maxschulz-COL:
> > The issue I was describing is that the `llms.txt` for a specific version does not show the correct links to a version, hence might be wrong or even outdated. It always points to stable
> 
> Understood. But we cannot version in the root of the Vizro docs ([as advised by RTD](https://github.com/readthedocs/readthedocs.org/issues/12321), we are using a redirect from the root to stable). The best way to flag that there actually _is_ a versioned llms.txt is to link it (making it visible to humans and bots alike).

Maybe I don't understand the matter, but to me there is no versioned `llms.txt` in a sense that the links within it are pointing to stable. It may be that the text in there is versioned, but that is only half the truth no?

stichbury:
> Maybe I don't understand the matter, but to me there is no versioned `llms.txt` in a sense that the links within it are pointing to stable. It may be that the text in there is versioned, but that is only half the truth no?

It was me that misunderstood. I thought because we were using relative links in the `mkdocs.yml` we would automatically get the version picked up, but it uses the `stable` for every build because it takes the `site_url`. One solution is to specify a particular version URL field to the plugin, so when it builds `llms.txt` it points each file to that version and not stable.

@maxschulz-COL I've made a PR over on the repo to see if that is an acceptable solution for them: https://github.com/pawamoy/mkdocs-llmstxt/pull/25

stichbury:
Now ready for review. The plugin we use to generate `llms.txt` has been updated to use a `base_url` setting, which we specify, and will need to keep updated in alignment with what we do currently in the yaml for versioning the pycafe plugin requirements.

I have updated the release build checklist to indicate this, and updated the `mkdocs.yml` setting for the current build (will need to be incremented to 0.1.47 before release) so it will point to the existing stable build by version number. 

Note, I have removed the code for the original purpose of this PR, which was to link the `llms.txt` on the front page of the docs. I cannot work out how to do that with a relative link, since strict build converts this warning to an error:

`Doc file 'index.md' contains a link to 'llms.txt', but the target is not found among documentation files.`

I can't turn off strict building (don't want to) and there isn't a validation setting to remove this warning for the individual page AFAICT.

This PR is still valid though, since it sets up the versioning for `llms.txt`.

vizro-core/mkdocs.yml
```diff
@@ -152,7 +152,7 @@ plugins:
         Configuration can be written in multiple formats including Pydantic models, JSON, YAML or Python dictionaries for added flexibility of implementation.
 
         Optional high-code extensions enable almost infinite customization in a modular way, combining the best of low-code and high-code - including bespoke visual formatting and custom components.
-      full_output: llms-full.txt
+      base_url: https://vizro.readthedocs.io/en/0.1.46/
```
maxschulz-COL, line 155: And just so that I understand, we would manually change that on release right?

maxschulz-COL (approved):
Fine with me as long as we change the release guide :)

vizro-core/mkdocs.yml
stichbury: Correct! I've updated the wiki page to add this info.

Issue #726, opened 2024-09-20
Create a Vizro app where the pydantic configuration is translated to yaml


huong-li-nguyen:
**Requirements:**
- Simple Vizro app
- Text input field (accepts yaml or pydantic)
- Text output field (Returns yaml or pydantic)
- Action that translated from one to the other
- Host it on huggingface or similar
- Add your app to the community demos

antonymilne:
LOVE this idea! FYI @maxschulz-COL we discussed similar before.

Note the to YAML version won't be possible yet until we implement the full `to_dict` or `to_yaml` functionality, which we haven't prioritised yet and maybe best to wait for our move to pydantic v2.

huong-li-nguyen:
I'd love to have this because I think it would also speed up our dev workflow 😄. Creating the YAML examples for our docs takes a lot of time and isn't much fun, so automating it would be awesome.

The standard Pydantic to YAML converters are okay and give a good starting point, but they don't handle all the Vizro-specific syntax like `__target__` and stuff.

maxschulz-COL:
> No, not yet. It's something that would still be nice to have.

Do you mean because of the MM? Because one way its possible now right? yaml -> pydantic

antonymilne:
Yes, that way is possible. It's the opposite direction that doesn't exist. This isn't a limitation of the model manager, it's just because if we do `dashboard.model_dump(_json)` then it doesn't yet know how to handle `CapturedCallable` and probably some other things.

huong-li-nguyen:
Yes, I was mostly interested in the opposite direction — converting **Pydantic → YAML**. This is mainly out of personal interest, to be honest. I’ve been writing a lot of documentation, tutorials, and dashboard examples lately, and creating and testing the YAML examples is always the most time-consuming (and least enjoyable 😄) part of the process.  

I agree it’s not a high priority and more of a “nice to have”.

maxschulz-COL:
> Yes, that way is possible. It's the opposite direction that doesn't exist. This isn't a limitation of the model manager, it's just because if we do `dashboard.model_dump(_json)` then it doesn't yet know how to handle `CapturedCallable` and probably some other things.

Yes that is also a limitation, but I was thinking if like Li describes at the top, we have a simple Vizro app that does this, once we instantiate the model and thus the MM in order to do `dashboard.model_dump(_json)`, then we might get conflicts no?

antonymilne:
[...]

PR #1435, opened 2025-10-02
[Docs] Add cross-highlight examples

huong-li-nguyen:
## Description
Add cross-highlighting examples
- Cross-highlight from graph to graph
- Cross-highlight from table to graph
- Cross-highlight from graph to graph (self-highlighting)

## Notice

### DIFF
--- a/vizro-core/docs/pages/user-guides/graph-table-actions.md
+++ b/vizro-core/docs/pages/user-guides/graph-table-actions.md
@@ -1166,8 +1166,6 @@ A self-highlight is often part of an [actions chain](actions.md#multiple-actions
         ```
         # Still requires a .py to add data to the data manager and parse YAML configuration
         # See yaml_version example
-        # Still requires a .py to add data to the data manager and parse YAML configuration
-        # See yaml_version example
         pages:
           - components:
               - type: graph


vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -628,8 +628,384 @@ However, it is not yet possible to cross-filter from a pivot table according to
...
+
+    ```python
+    import vizro.actions as va
+
+    components = [
+        vm.Graph(
+            id="source_graph",  # (1)!
+            actions=va.set_control(control="highlight_parameter", value="y"),  # (2)!
+        )
+    ]
```
huong-li-nguyen, line 908: @antonymilne @petar-qb - I was initially very confused by this and configured this wrong. I had `value="highlight_country`, instead of `value="y"`, so basically thought that we want to affect the "highlight_parameter" and the value we want to change is the target graphs argument so "highlight_country". 

It didn't work, and I was a bit confused. The error message then says something with adding custom_data, which I think is even more confusing. When I added custom_data, it didn't actually fix anything. So I am sure it's probably mentioned somewhere in the docs here, but I don't know where? 😬 But this threw me off at the beginning. I do think the explanation of this in the cross-filter makes sense (like "it's often the column of the Filter"), but here I wasn't sure how to provide an equally sensitive explanation of how they could know that it's "y". 


```suggestion
        vm.Graph(
            id="source_graph",  # (1)!
            actions=va.set_control(control="highlight_parameter", value="y"),  # (2)!
        )
    ]
```

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -628,8 +628,554 @@ However, it is not yet possible to cross-filter from a pivot table according to
...
+
+    ```python
+    import vizro.actions as va
+
+    components = [vm.Graph(..., actions=va.set_control(control="my_parameter", value="y"))]
+    ```
+
+## Cross-highlight
+
+A cross-highlight is when the user clicks on one _source_ graph or table to highlight corresponding data, typically in a graph. The highlighting can occur in two ways:
```
huong-li-nguyen:
```suggestion
A cross-highlight is when the user clicks on one _source_ graph or table to highlight corresponding data, typically in a graph. The highlighting can occur in two ways:
```

I've specifically mentioned that cross-highlighting is typically done in a graph and the current examples I've added also only show these examples. Just to repeat my reasoning: cross-highlighting is most useful for charts because it lets you focus on one trace while keeping the others visible for context, which is especially helpful in crowded views. For tables, highlighting doesn't add much beyond what filtering already provides - for example, a highlighted row on another table page wouldn't even be seen, and filtering is usually the more intuitive way to narrow down data. For figures, there might be potential, but I can't think of a strong use case at the moment. So I think the three examples where a graph is always highlighted, probably covers the main scenarios, at least for now :)

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -628,8 +628,564 @@ However, it is not yet possible to cross-filter from a pivot table according to
...
+            targets=["target_component.my_argument"],
+            selector=vm.RadioItems(options=["option1", "option2", "option3"]),
+        )
+    ]
+    ```
+
+1. Call `set_control` in the `actions` argument of the source [`Graph`][vizro.models.Graph] or [`AgGrid`][vizro.models.AgGrid] component that triggers the cross-parameter.
+
+    1. Set `control` to the ID of the parameter.
+    1. Set `value` to specify which data from the source component to pass to the target component's parameter. The format depends on the source component type (see the [API reference][vizro.actions.set_control]). For tables, use the column name. For graphs, use the argument name where that data is plotted (e.g., `"y"` for data on the y-axis).
```
huong-li-nguyen: @antonymilne @petar-qb  - Could you take a look at all my explanations about setting the `value` inside `set_control`? That part actually confused me the most when I was configuring it, sometimes it was `y`, other times `country` 😄. I think I've figured it out now and have basically codified my understanding.

However, if you could review the explanations here, as well as in the other code examples, that would be great! I've slightly rephrased them in the other examples to make them more relevant to each specific case.

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -628,8 +628,564 @@ However, it is not yet possible to cross-filter from a pivot table according to
...
+        fig = px.box(data_frame, x="lifeExp", y="country", color="country")  # (2)!
+        fig.update_traces(marker=dict(color="#00b4ff"))  # (3)!
+        fig.update_layout(showlegend=False)
+
+        if highlight_country:  # (4)!
+            for trace in fig.data:
+                if trace.name == highlight_country:
+                    trace.marker.color = "#ff9222"
+
+        return fig
```
huong-li-nguyen: @antonymilne - I rewrote the highlighting logic of your code example here. Your version definitely works, but I thought the overlay approach might be a bit confusing. Using the trace name approach consistently seemed cleaner and easier to explain, but let me know what you think 👍

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -628,8 +628,564 @@ However, it is not yet possible to cross-filter from a pivot table according to
...
+    ```
+
+    1. The `highlight_country` argument receives the selected country name from the cross-parameter interaction and is used to identify which trace to highlight.
+    1. Using `color="country"` creates one trace per country and sets each trace's name to the country name. This makes it easy to identify traces by name in the highlighting logic. Note that we later overwrite the automatic color assignment to make all boxes uniformly colored.
```
huong-li-nguyen: I was initially really confused about why the `trace.name` approach didn’t always work until I realized that trace names are only populated if you specify the `color` argument or use `facet_row` / `facet_col`. In this case, we artificially set the `color` argument, even though we want everything to be uniformly encoded, so that we can consistently identify the trace that needs to be highlighted. I don't know if there is an easier / better way, but this was the easiest I could find.

vizro-core/docs/pages/user-guides/graph-table-actions.md
huong-li-nguyen: My understanding codified here: https://github.com/mckinsey/vizro/pull/1435#discussion_r2405894863

vizro-core/docs/pages/user-guides/parameters.md
```diff
@@ -203,7 +203,7 @@ If you use [dynamic data](data.md/#dynamic-data) that can be updated while the d
 For further customizations, refer to the [guide to selectors](selectors.md) and the [`Parameter` model][vizro.models.Parameter]. Some popular choices are:
 
 - Customize the `selector`, for example `multi` to switch between a multi-option and single-option selector, `options` for a categorical parameter or `min` and `max` for a numerical parameter.
```
stichbury:
```suggestion
- Customize the `selector`. For example: `multi`, to switch between a multi-option and single-option selector; `options` for a categorical parameter; or `min` and `max` for a numerical parameter.
```

vizro-core/docs/pages/user-guides/parameters.md
```diff
@@ -203,7 +203,7 @@ If you use [dynamic data](data.md/#dynamic-data) that can be updated while the d
...
-- Make the parameter's selector invisible by setting `visible=False`.
+- Make the parameter's selector invisible by setting `visible=False`. This is particularly useful for graph interactions where you want to hide the selector from the user interface while keeping the functionality active. Cross-highlighting is a common example of this pattern. For a complete code example, see the [cross-highlighting section](graph-table-actions.md#cross-highlight) in the graph and table interactions guide.
```
stichbury:
```suggestion
- Make the parameter's selector invisible by setting `visible=False`. This is particularly useful for graph interactions to hide the selector from the user interface while keeping the functionality active. Cross-highlighting is a common example of this pattern. For a complete code example, see the [cross-highlighting section](graph-table-actions.md#cross-highlight) in the graph and table interactions guide.
```

stichbury (approved):
LGTM 🏆 just a couple of minor tweaks, and one was to a bullet that you didn't actually change.

vizro-core/docs/pages/user-guides/graph-table-actions.md
antonymilne: Agreed! I think this a very sensible subset of examples. I changed the box chart one to a bar chart since I think it's more common, and now we have examples of how to do highlighting in what are maybe the three most common sorts of chart to highlight (hopefully?): scatter, bar and line.

vizro-core/docs/pages/user-guides/graph-table-actions.md
antonymilne: I've simplified this a bit so now we don't need to mess around with `trace.name` unless it makes sense to do so

antonymilne:
Great PR, thank you for writing all this! I know how much work it is 😅 I really like all the examples 👍 

I've made a few edits but the general gist of it is exactly as before. The major points are:
* simplified all the custom chart code as much as I could
* changed the self-highlighted box plot to a bar chart since it's more common
* added docs for #1437 since we now have it and it's kind of essential in these examples where all controls are invisible
* how you can work out what format `value` takes:
  * your understanding of thiswas pretty much right, but what you said on slack about positional vs. visual encodings made me wonder whether there's a much better way of explaining it. In short: it's not quite true to say that positional encoding is always in `clickData` while visual encoding isn't (there's many special cases and inconsistencies in plotly here that make it hard to say anything is always true) BUT we can say that in general positional encoding is there. Hence the logic should be "is the variable positional? If yes, specify it directly like `value="y"`; if no, use `custom_data` and column name". This applies for both filters and parameters.
  * so I've explained this again in the section on cross-filters and referred back to it from the section on cross-parameters. Please do see if it makes sense.
* the "Behind the scenes mechanism" for the self-highlighting example wasn't quite right in terms of the order of operations, so I corrected that
* it actually _is_ now possible to do custom charts through yaml configuration, just it's not widely known... And yes, this means that luckily we can now convert the Python examples to yaml, which I know is your favourite activity. But don't worry, as per https://github.com/mckinsey/vizro/issues/726 I [wrote a script](https://py.cafe/snippet/vizro/v1?#c=H4sIAApq6WgEA61Y23Ljthl-FYx8sVJGog4-yPaM2m6dTbK92N1J3ObC8jAQCYlcUwADgJK5Hr9EX6DTN-wj9PsBnqTYSWYSeDwSgf_4_SdQT71IxaJ33Tv5U9ZSnrD30ghtWakKzWJukpXiOmakhqWSJUKLa5D9KWsp022uoGyXftEqyDNlszIQj7kWxjBuWP54RLKFGZk72m2PjnhkUyX9GV_KtVbbA6bAlrkwrOKJeG4LLbp09dG_6KF7EFi-ylpWAiXkm3Cj03gpl3LD820qY6HZAgYHMbc8aPb6g-DnQuiyv0SkpE2lkJYtFuzNu0KrXLxhXMasFBzMCzabTObL3oCE_q0yEHwbzfPEbcdizUzErRU63Kc2CZN0k2T4t33SGq4134oha3bDSBXS6nLxQUkxuGZsKRlWtRumphUgYljfCrkjcx3VsndPpjWEtcxa2DrdeL8rw_p-m1YrbthuPi7gUZx_EhoeLnudkxInWboW7x4P9036ReAoV7_cDrf8cXEx6eyqnEepLReT4LSzG6lM6cXLfnfJuBUbBQqlETqzeCIUwLnsXbO7b3hmAO6tLsT987B23wWLvqTrFzBKDZPKMoL_umKgBcyCIgc8IrSaR8L0jchEZGHjdMi2XD8ITcozpAsCHVuE_5rNhmzZq_yj5-nzgKzwYjsiM47atX2TqH0mNkLGC2c6LCVCLZBWkuiJdSlzvhEI4G4bfMK3Kno2tRlhfqOVMaPGL-ZqwlVDEwqvbQH-b1EPfSqKxd3dZMim9_dDAj5ELUDU5QTF3Bs0wEUK1UTVYBZ3focWxLzdOEHtHq1EcEQEYv73n3__l91kafTAlGScabVnVrXQM5twVHeNv8SzqGuGUYNpDK8XkECdLbpV3ammRVPJgyO-qt8sdjwwgkIOjSrrV58wtbEpzDmJggVQznY8KwjbysYuJrS6ahyoKP4jMIDwslf3gSjhmpzC_iFV5VdNd9QvXnSr_n5ffVa-HEfoU-3OS3Y1Kg69PrbOcr0RLvaHngQtfwMQ8uiYvakXmPM9j1P13oqt6avcxwRSP3z88I4Uf9V4etDT7gn1I6G71KRIbV8uh8ddZFzJt5PR1c7X9WOf6gkG0Mc9KN2c-8OLNFbTp-TbrHn4bFR7cjCrOKagff1ofXT08uil68DbLEMNoZGtUsv2OkWoJFuV7AZ19u2nW7YqLKIB8KkM90o_sL3IsiAIgBGNK6FTnqFV99Xq8wA3B0IRvRJQS2O5jNwBqmIbuPkc3Pi5F9_wLCN7ax5aJ3X34ixOI4snMhc9hFPYWzqJvENcIDgI14V0hRqEIW2HYUsGM6g7J9wgAXU_fxw6zoGbyd2TnX31ZF2ddKykdcK-42hRrhlxY4otcK8mOu5SlWs0HmpEq07F8zzIS9QunQIl-8ZPEHRKptaO5FjRqkgzS_y1pwbQu7NqkX1AY73shZiXqQzD4In2npe9lrACFkMn9KUZ0oghMpTQVw5JbMMPNOxnz6Z5agS7RdTeaa10HwrogT1RIF24n53tdQqQy-4a4zMD8cc9M3TEK04Fp3ZCo_s2YB6miaehuPswdI5qxuq4E4sT9k2qDcKnRVRomJtKpCnwZw-iNHhiJLalRzoVQHPBnionaa2VZg9DtqvJg9Q1m64eWp737gH3JVC2zu1azwIA24eop-cBcHBcWCfsNhFyWDFhyBPbsuenbc3KgEaV7QxJ5UoyEzuBZuCEYIGi4YWpNefLZtaUZG1N2W62LFVieC6_LXBL68LvI5Ol5tci4447plAt-5CglVNajfaUToQ1kRrWr-tmy22UpHIDd-XGJqaDXGXcnUeuyiXAPXBy6BsB8SXNjzOschCvBV14KnFESwZmgu_c5SHj2rAxklmOXN_xX8lMFPcoNZTSJ-xHwfacouPKNC62OWodCg1qkpqjeIyyAoEsYKxFiYssRldF-0ylP_iJ7P8poGzQEKCEkW98cyX2lQC4THBTQk1JO7HyaUDKabyzvIxhQBqx_paXoM9xe6OJxgpEZINSgMqRS5pGp01UsUnQxKhQ0Z3-OgjYD2jkgn0u4N6WywLVUjLcI4XG7ZKtFBRRlOGkdKotfyA5-If5VFcMF0JWdSsjcImIGfUvlxo-xV0V0vGaqhP96oT989PXb2_fXeOIEoOyvPEGhFqgEVYAhsg-DCjuIWT7JI0STJ4C33OtVmgzZQMNpUGBFFIANEI8JYYc2g-COEXeN0PcvyKGFLN-rcWFaXGL6z4Sjjhmr3Hg3JcHCLqZ6NT4njQDDXn5fk3Q-pbtDHRIuDdOcpOGADYhwndMaucMF3nyDfGsWz1YjsckRoUR2RpNmdCCGOSKFCKmuIOcdBfIasxwAEzXhoBMp1FBD6Gx9Pba7ve9QzBerDm-LGqTCAwS9hGJTW0U3gDrBhYgDVjY6C_Oa_okmQziyZDONl1jqu3AWddFplbm_Xf-tnDXnrZw4VYHfFy-AqTY5Zr5NZu6ykm3LziOfyQ30EN9GVdgKVIR96JXXOyIYX309qBJ2REdDR1BwOM41PVFRegBfERu14lKdYdbE2qKDMnREAuHFIFRBR4pZfH6g8oSOfvHDx8_jJ0fcJF-osH9GR-Rvx7ogl4VD5LAl743NRdRuqaCElEiU_wuQfHPYb_tO1MJ3r5Lgkzx2LivlBq4K6eY2dIuZjS3ltIlLH7aoHDR21IFzSDQhewPesOeFj8XAM9dF_DrVGJtbq7H47wMIr4W400y5tqma7xDjbfRAxVNOXY30XH1WjWGJDOeXp5N5tP5xeXp6ThPc08ymgTT4GwexGI3GeXl6QjNWIy4LIN9gmmYl-QLbKAahG7Hg8e8RK-T2MhLFWMGjHaTYHYRTHHk32F710-9ql33rmfwQSn7vdrTdiUKL5ugxjDKYi1AdNec4GqNboFD97beu56eT4Y9vHb86B9P_dN3gl5s_GMag43ueX-HVCT8jZIWlzOhX9FApKOVpwVJjhBjt_eM3yIamtaKStHVPJhMrs4vrk7n86urydnpy7LB14r0V9Deodj2uBvZwD7aQws8QrX66WQWOLWTixl-27o6n7-q_9Dui3lwPj-9uLqYnp1PZ2fz6e8BswLnt4CsMTyw-xcWTE9nwdns4nRyenl-djk_n8H0pDo7vzz7Pfag2LHNs98yqKYji-jveeiCjbq5u3_-P63Ngc7hFQAA) that actually makes this pretty much painless! All you need to do is copy and paste the Python into the top and then label the `data_frame` sources. I've done one example for you in [c6dee94](https://github.com/mckinsey/vizro/pull/1435/commits/c6dee9490926d55442cc7d10fd75b1ed1f1e8913) and it really took a couple of minutes.

There's not much to do now - let's aim to merge next week 🚢 

TODO @antonymilne:
- [ ] check through everything again

TODO @huong-li-nguyen:
- [x] skim through and see what you think. Hopefully the `value` explanation makes more sense now
- [x] add yaml examples for last 2 cross-highlight examples. This would be a good way to check my script works and to understand what the yaml configuration looks like now. It really shouldn't be painful, I promise! Let me know if you have any problems.

TODO by whoever gets to it:
- [x] new gifs including reset controls once #1437 has merged or is close to doing so

huong-li-nguyen:
> * it actually _is_ now possible to do custom charts through yaml configuration, just it's not widely known... And yes, this means that luckily we can now convert the Python examples to yaml, which I know is your favourite activity. But don't worry, as per Create a Vizro app where the pydantic configuration is translated to yaml #726

Ahh, wow! Is this now only possible for custom charts, or also custom actions etc? Because then we should probably update our yaml examples for our docs at some point. At the moment, they all say that this is not possible in yaml 😄 I've added the yaml examples, but I haven't tested them out yet 👍

huong-li-nguyen:
@antonymilne - the self-highlighting example is still a bit off (the color switches when you click the first bar), but in terms of text changes, it looks good to me 👍 


Screenshot 2025-10-13 at 12 15 23

antonymilne:
> > * it actually _is_ now possible to do custom charts through yaml configuration, just it's not widely known... And yes, this means that luckily we can now convert the Python examples to yaml, which I know is your favourite activity. But don't worry, as per Create a Vizro app where the pydantic configuration is translated to yaml [Create a Vizro app where the pydantic configuration is translated to yaml #726](https://github.com/mckinsey/vizro/issues/726)
> 
> Ahh, wow! Is this now only possible for custom charts, or also custom actions etc? Because then we should probably update our yaml examples for our docs at some point. At the moment, they all say that this is not possible in yaml 😄 I've added the yaml examples, but I haven't tested them out yet 👍

It's possible for custom charts, tables, figures and actions since [0.1.42](https://github.com/mckinsey/vizro/releases/tag/0.1.42):
> yaml/json configurations can specify _target_ functions with any import path rather than being restricted to built-in Vizro functions. (https://github.com/mckinsey/vizro/pull/1255)

The only thing it's not possible for is custom component swhich uses a different mechanism. Technically that is actually possible already but we might need to make it impossible to get something working easily on the model manager... So I don't intend to change the current documentation there from saying that it's not possible in case we make it impossible 😅

vizro-core/docs/pages/user-guides/graph-table-actions.md
```diff
@@ -1166,26 +1166,28 @@ A self-highlight is often part of an [actions chain](actions.md#multiple-actions
         ```
         # Still requires a .py to add data to the data manager and parse YAML configuration
         # See yaml_version example
```
antonymilne:
```suggestion
```

PR #1437, opened 2025-10-06
[Feat] Add page-level reset controls button

petar-qb:
Closes https://github.com/McK-Internal/vizro-internal/issues/2191

## Description
Draft PR opened without fixing/writing any tests or docs. 

## TODO
- [x] - PoC
- [x] fix existing unit tests
- [x] fix screenshot tests
- [x] fix other tests and merge
- [x] write new JS unit tests -> TBD in https://github.com/McK-Internal/vizro-internal/issues/2256
- [x] add new e2e dom and http test -> TBD in https://github.com/McK-Internal/vizro-internal/issues/2255
- [x] Should we explain this feature in docs? TBD in #1435 
- [x] Should I adjust all docs screenshots so they contain reset button? -> ❌ 

## Screenshot

1. text-button

https://github.com/user-attachments/assets/cf735eb1-714e-4d1f-847f-8bbc3c8a731d

2. icon-button with tooltip

https://github.com/user-attachments/assets/9e9144c2-fcc2-448c-a303-b107d9612db3

3. text-icon-button in control_panel or icon-with-tooltip in header (the current PR implementation)

https://github.com/user-attachments/assets/de4795d8-ebec-4bae-8698-74e847b6d506

## Notice


vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -188,16 +189,18 @@ def __call__(self, target_to_data_frame: dict[ModelID, pd.DataFrame], current_va
...
+                selector_call_obj,
+                dcc.Store(id=f"{selector.id}_guard_actions_chain", data=True),
+            ]
+        )
 
         return selector_call_obj
 
     @_log_call
-    def pre_build(self):
+    def pre_build(self):  # noqa: PLR0912
```
petar-qb: Filter's `pre_build` is too complex for a while, so it's worth breaking it into separate helper methods similarly like it's done for the `Parameter` model (but not in this PR).

vizro-core/src/vizro/models/_controls/parameter.py
```diff
@@ -167,6 +167,21 @@ def _check_categorical_selectors_options(self):
...
+            # RangeSlider and DatePicker(range=True)
+            if isinstance(self.selector, RangeSlider) or getattr(self.selector, "range", None):
+                self.selector.value = [self.selector.min, self.selector.max]
+            # Slider and DatePicker(range=False)
+            else:
+                self.selector.value = self.selector.min
+        elif isinstance(self.selector, (Checklist, Dropdown, RadioItems)):
+            multi = isinstance(self.selector, Checklist) or getattr(self.selector, "multi", False)
+            self.selector.value = get_dict_options_and_default(options=self.selector.options, multi=multi)[1]
+
```
petar-qb: Try to extract to separate controls util and reuse in two places here and in `filter.py`.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -175,6 +177,20 @@ def build(self):
...
+                    "originalValue": control.selector.value,
+                }
+                for control in cast(
+                    Iterable[ControlType],
+                    [*model_manager._get_models(Parameter, page), *model_manager._get_models(Filter, page)],
+                )
+            }
+            for page in self.pages
+        }
+
```
petar-qb: @antonymilne

I was able skip the `page.id` level of hierarchy and to move this to `page.py` (similarly how the `page_id_store` is added/handled), but I left it here on the dashboard level as this will be reused when someone starts working on the cross-page actions.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -314,6 +339,10 @@ def _inner_page(self, page: Page) -> _InnerPageContentType:
         else:
             header_right_content.append(header_controls)
 
+        # Show reset button in the control panel when page controls and control panel exist
+        if has_page_controls and not _all_hidden(control_panel):
+            control_panel.children.append(reset_page_controls_btn)
+
```
petar-qb, line 366: This code covers all the cases when:
 - controls exist in the controls panel => reset button is in the control panel
 - controls exist but hidden => reset button is in the header (page/dashboard depending on whether the dashboard header exist)
 - there's no controls on the page => reset button is not added

vizro-core/src/vizro/models/_page.py
```diff
@@ -151,26 +151,35 @@ def pre_build(self):
...
-            for control in cast(
-                Iterable[ControlType],
-                [*model_manager._get_models(Parameter, self), *model_manager._get_models(Filter, self)],
+        page_controls = cast(
+            Iterable[ControlType],
+            [*model_manager._get_models(Parameter, self), *model_manager._get_models(Filter, self)],
+        )
+
+        if page_controls:
+            # TODO-AV2 D: Think about merging this with the URL callback when start working on cross-page actions.
```
petar-qb, line 160: The comment added so i can be cross linked.

petar-qb: linked here -> https://github.com/McK-Internal/vizro-internal/issues/2054

vizro-core/src/vizro/models/_dashboard.py
petar-qb: The `vizro_controls_store` structure will likely be adjusted later as we add more reset buttons (container-level, dashboard-level, figure-level) or cross-page actions. A flat structure is a strong candidate to avoid deep hierarchies.

vizro-core/src/vizro/static/js/models/page.js
```diff
@@ -180,11 +180,35 @@ Received input: ${JSON.stringify(values_ids)}`,
...
+    const originalValue = controlInfo["originalValue"];
+
+    dash_clientside.set_props(`${selectorId}_guard_actions_chain`, {
+      data: true,
+    });
+    dash_clientside.set_props(selectorId, { value: originalValue });
+  }
+
+  // Trigger the OPL after resetting all controls.
+  return null;
```
antonymilne: Would it work if we set all of these properties using `Output` at the same time instead of doing some of them with `set_props`?

vizro-core/src/vizro/models/_dashboard.py
antonymilne: Is the current structure where `page.id` is a top-level key done like this just so that `reset_page_controls` can easily look up all the controls on that page?

petar-qb: Yes exactly. However I think that the structure we be adjusted later to something like:
```json
{
  "control_1_id": {
    "selector_id": "selector_control_1_id",
    "page_id": "my_homepage",
    "parent_container_id": "container_id"
    "original_value": ["1", "2"],
    "current_value": ["1," "2", "3"],
    ...
  },
  "control_2_id": ...
}
```

petar-qb:
> I'm not a big fan of the button moving next to the theme switcher but if others prefer it then it's ok. The alternative would be to keep the button in the same place and just have a big empty left sidebar for it which is also not great I know. If we do move the button next to the theme switcher then maybe in that case I would make it just a circular button with icon so it's more in scale with the theme switcher. wdyt now we have a real live example to look at @huong-li-nguyen? Happy to go with whatever you prefer here.

I just pushed some changes and you can see how it looks like now in the third screen-recording in the PR description. So, the current implementation adds:
- (if controls and control_panel) an icon-text-button without a tooltip (as it's obvious from the button text what happens) to the control_panel
- (if controls and not control_panel) clickable icon with a tooltip to the header

I like the current solution. What do you think @huong-li-nguyen?

petar-qb:
The only thing I’m currently uncertain about is whether we should commit to adding the reset button by default without an easy way of removing it. We plan (not immediately but it should be an ultimate goal) to expand the reset-controls functionality beyond the page level by adding it to the dashboard, container, control, and source figure levels.

However, making it a default behaviour (“set in stone”) might feel too intrusive for Vizro users, even now when it’s only applied at the page level.

I’d recommend an opt-in / opt-out approach instead, so users can easily enable or disable the button at different levels.

Here’s my suggested API design:

```py
vm.Dashboard(show_reset_controls_button: bool = False)
vm.Page(show_reset_controls_button: bool = True)  # Maybe even set it to False by default.. or not
vm.Container(show_reset_controls_button: bool = False)
vm.Graph(show_reset_controls_button: bool = False)
vm.Filter(show_reset_control_button: bool = False)
```

In that case I would expose additional `vm.Page` argument and default it to False/True depending on what we agree.

I'd love to get you opinions on this: @antonymilne @huong-li-nguyen @Joseph-Perkins 🙏

vizro-core/src/vizro/static/js/models/page.js
petar-qb: Very good point! 👍 

I needed to use the `set_props` instead in the `sync_url_query_params_and_controls` because output values there are also the inputs of the same function. Here in the `reset_page_controls`, I can use proper dash Outputs so I pushed the changes.

vizro-core/src/vizro/models/_dashboard.py
petar-qb: I just refactored the code and switched to the format suggested in the previous comment. 😃

vizro-core/src/vizro/models/_controls/_controls_utils.py
```diff
@@ -52,3 +53,17 @@ def warn_missing_id_for_url_control(control: ControlType) -> None:
             "If you want to ensure that links continue working, set a fixed `id`.",
             UserWarning,
         )
+
+
+def set_selector_default_value(control_selector: SelectorType) -> None:
```
petar-qb: Another approach to calculating selectors' default values:
1. During the creation of `vizro_controls_store` (`dashboard.build`), calculate the default value for all selectors using `get_dict_options_and_default` and other relevant methods for min/max.
2. Implement a getter property (e.g `get_default_value`) for all selectors to access their current value.

NB: Default selector values cannot be calculated during the initialisation phase because options/min/max could be determined in the `Filter.pre_build` phase.

I found the current implementation good enough but we should rethink the solution when we deal with the https://github.com/McK-Internal/vizro-internal/issues/1554

antonymilne:
> I just pushed some changes and you can see how it looks like now in the third screen-recording in the PR description. So, the current implementation adds:
> - (if controls and control_panel) an icon-text-button without a tooltip (as it's obvious from the button text what happens) to the control_panel
> - (if controls and not control_panel) clickable icon with a tooltip to the header
> 
> I like the current solution. What do you think @huong-li-nguyen?

I'm still not 100% sure I like the button moving around depending on whether there's a control panel, but I also don't much like the alternative solution. So overall this is fine by me. 👍 

> The only thing I’m currently uncertain about is whether we should commit to adding the reset button by default without an easy way of removing it. We plan (not immediately but it should be an ultimate goal) to expand the reset-controls functionality beyond the page level by adding it to the dashboard, container, control, and source figure levels.
> 
> However, making it a default behaviour (“set in stone”) might feel too intrusive for Vizro users, even now when it’s only applied at the page level.
> 
> I’d recommend an opt-in / opt-out approach instead, so users can easily enable or disable the button at different levels.
> 
> Here’s my suggested API design:
> 
> ```python
> vm.Dashboard(show_reset_controls_button: bool = False)
> vm.Page(show_reset_controls_button: bool = True)  # Maybe even set it to False by default.. or not
> vm.Container(show_reset_controls_button: bool = False)
> vm.Graph(show_reset_controls_button: bool = False)
> vm.Filter(show_reset_control_button: bool = False)
> ```
> 
> In that case I would expose additional `vm.Page` argument and default it to False/True depending on what we agree.
> 
> I'd love to get you opinions on this: @antonymilne @huong-li-nguyen @Joseph-Perkins 🙏

We should not do this yet. Maybe in the future, but it's premature optimisation/functionality right now. I think the current page-level reset button is not too intrusive to appear on every page by default. If users really want to hide it then it's easy to do so through CSS, and if it's a popular request then we can _at that point_ add some option to the API to do it.

In general the problem with adding something like `show_reset_controls_button` as a top-level field to `vm.Page` is:
* it places this "fine tuning" on a similar level of important to other more fundamental things like the page title, what components are on the page, etc. This seems out of place since it's a small stylistic tweak rather than a fundamental property of the page
* hence if we do add it then it would probably make more sense inside a field like `vm.Page(style=vm.PageStyle(show_reset_controls_button=...))` or somehow inside `vm.Page(controls=...)` itself. This would give space for other similar kinds of settings. But we don't yet know what those other settings would be, so let's not make a space for it already since it's hard to put in the right place until we have more such settings (e.g. move whole theme switcher/settings panel to left hand side; do groups of controls; alter collapsible properties of sidebar etc.).

vizro-core/src/vizro/models/_controls/_controls_utils.py
antonymilne: I think ultimately the right solution here would be that the selector itself sets its own `self.value`. As you say, that's hard to sequence correctly at the moment due to `Filter.pre_build`, but hopefully in future after MM work is complete that will be easier.

The whole `value` setting and validation for selectors is indeed a bit of a mess at the moment, but for now it works ok anyway.

vizro-core/examples/scratch_dev/app.py
antonymilne, line 1: Great app to test this all out, as ever! 🙏 💯 

The only case I think is a bit debatable and you didn't mention here is what happens when here's navigation and hidden controls. In this case the side bar already exists so we could potentially put the reset button there since it wouldn't take up any more space.

I see advantages and disadvantages both ways. So don't change it from how you have now unless you think it's much better that way.

Image

vizro-core/src/vizro/models/_page.py
```diff
@@ -151,46 +151,46 @@ def pre_build(self):
...
             clientside_callback(
-                ClientsideFunction(namespace="page", function_name="reset_page_controls"),
+                ClientsideFunction(namespace="page", function_name="reset_controls"),
                 Output(f"{ON_PAGE_LOAD_ACTION_PREFIX}_trigger_{self.id}", "data", allow_duplicate=True),
                 *selector_outputs,
                 *selector_guard_outputs,
                 Input(f"{self.id}_reset_button", "n_clicks"),
                 State("vizro_controls_store", "data"),
-                State("page_id_store", "data"),
+                State(self.id, "id"),  # Assigned to outermost Div in Dashboard._make_page_layout.
```
petar-qb, line 177: Great spot! 🥇

petar-qb:
@antonymilne thanks for extracting a few util functions to `_controls_utils.py` and the rest of the refactoring. The PR changes look much better now! 🎉

vizro-core/examples/scratch_dev/app.py
petar-qb: Great catch! Let's leave it as it is for now because we'll be rewriting the system anyway when we introduce control reset buttons on different levels. In that moment we will exactly know what do we want and how the things should look like from the UI/UX perspective.

PR #1441, opened 2025-10-14
[Feat] Enable that `vm.Button` can trigger a `set_control` action

petar-qb:
Closes https://github.com/McK-Internal/vizro-internal/issues/2148

## Description
Enabled that `vm.Button` can be the source of a `set_control` action, where `set_control.value` is propagated to the targeted control.

## Notice

### DIFF
--- a/vizro-ai/tests/unit/vizro-ai/conftest.py
+++ b/vizro-ai/tests/unit/vizro-ai/conftest.py
@@ -4,8 +4,8 @@ from typing import Any
 
 import pytest
 import vizro.plotly.express as px
-from langchain_core.output_parsers.pydantic import PydanticOutputParser
 from langchain_community.llms.fake import FakeListLLM
+from langchain_core.output_parsers.pydantic import PydanticOutputParser
 
 
 @pytest.fixture


vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,165 +1,68 @@
...
         ),
-        vm.Text(id="text-2", text="Click the button to see the action output here."),
-        vm.Graph(id="graph-2", figure=px.scatter(df, x="sepal_width", y="sepal_length", color="species")),
     ],
-    controls=[vm.Filter(id="filter-id-2", column="species", targets=["graph-2", "figure-2"])],
+    controls=[
+        vm.Filter(id="filter-id-2", column="species", selector=vm.RadioItems()),
+        vm.Filter(id="filter-id-3", column="sepal_length"),
+    ],
 )
```
petar-qb, line 69: Hi @lingyielia 👋

Here’s an example showing how to reset a specific control when a button is clicked.
Is that what you were looking for in your example?

If you need to reset multiple controls (say 3 out of 5), you can chain several `set_control` actions and it’ll work fine, though each one triggers a separate HTTP request. That’s okay for now, but we plan to optimise this soon. 

FYI: @antonymilne as `set_control.value` is `str`, it's impossible to reset anything but single-choice categorical selector. That's another reason why I would change it to JsonValue. I know finding the ideal API isn’t simple. I’m sharing this here for visibility, but let’s continue the discussion online.

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -112,7 +124,7 @@ def pre_build(self):
         if not isinstance(self._parent_model, _SupportsSetControl):
             raise ValueError(
                 f"`set_control` action was added to the model with ID `{self._parent_model.id}`, but this action "
-                f"can only be used with models that support it (e.g. Graph, AgGrid, Figure)."
+                f"can only be used with models that support it (e.g. Graph, AgGrid, Figure, Button)."
```
maxschulz-COL: Starts to be annoying :P ?

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -74,12 +74,19 @@ class Button(VizroBaseModel):
...
 
     @model_validator(mode="after")
-    def validate_text(self):
+    def validate_text_and_icon(self):
         if not self.text and not self.icon:
             raise ValueError("You must provide either the `text` or `icon` argument.")
 
         return self
 
+    @model_validator(mode="after")
```
maxschulz-COL, line 83: Can this no be a field validator on the later field? In general I think it may be wise not to rely to heavily on model validators. You would have access to the other field via [data](https://docs.pydantic.dev/latest/concepts/validators/#validation-data)

vizro-core/tests/unit/vizro/actions/test_set_control.py
```diff
@@ -86,13 +86,13 @@ def test_pre_build_parent_model_does_not_support_set_control(self):
...
         # Add action to the component that does not support set_control
-        model_manager["button_1"].actions = action
+        model_manager["table_1"].actions = action
 
         with pytest.raises(
             ValueError,
             match=re.escape(
-                "`set_control` action was added to the model with ID `button_1`, but this action can only be used with"
-                " models that support it (e.g. Graph, AgGrid, Figure)."
+                "`set_control` action was added to the model with ID `table_1`, but this action can only be used with"
```
maxschulz-COL: Or rather annoyed here?

vizro-core/src/vizro/actions/_set_control.py
petar-qb: Yep, I kinda agree 😅 

Currently there's no docs except [set_control API docs](https://vizro.readthedocs.io/en/stable/pages/API-reference/actions/#vizro.actions.set_control) where every model that can source a `set_control` is described. What do you think, would the following change be more concise and informative?

```py
                "can only be used with models that support it (e.g. Graph, AgGrid, Figure etc). "
                "See all models that can source a `set_control` at https://vizro.readthedocs.io/en/stable/pages/"
                "API-reference/actions/#vizro.actions.set_control"
```

vizro-core/src/vizro/models/_components/button.py
petar-qb: Could be done that way too. After thinking about it more, I’d say I prefer the following approach order:
1. **Field validator on both arguments:** The most robust solution, resistant to argument reordering and to post-assignment of arguments in any order.
2. **Model validator:** A semi-robust solution, resistant to argument reordering but not to post-assignment in any order.
3. F**ield validator only on the second argument:** Not robust against argument reordering or post-assignment in any order.

In other parts of our codebase, we typically use solution `#2`, for example in `validate_text`, and I haven’t run into issues there.
However, we also use solution `#`3 in places like `validate_max` for numerical selectors, and I’ve encountered problems there several times (e.g. when `max` is set during initialisation and `min` is added later from `Filter.pre_build`).

I actually started adding a `validate_min` recently but got interrupted before finishing.

Thanks for bringing this up!! I suggest creating a ticket to standardise and align all multi-argument validations across the codebase in one go by using the  **Field validator on both arguments:**. What do you think?

vizro-core/tests/unit/vizro/actions/test_set_control.py
petar-qb: This annoyed me too. We’re running out of models that don’t support `set_control`. Even if we implement it for every Vizro model, we’ll still need this check for cases where the user has created a custom model.

Since `set_control` is under active development, let’s leave it as is for now. We’ll likely define default behaviour for every component before we run into a situation with Vizro models that don’t support `set_control`.

vizro-core/src/vizro/actions/_set_control.py
antonymilne: Yes, I like it!

vizro-core/src/vizro/models/_components/button.py
antonymilne: For me this is a classic case of a model after validator - very similar to the [example given in the pydantic docs](https://docs.pydantic.dev/latest/concepts/validators/#model-after-validator) where the validator could be attached to the second field but is nicer to do as a model validator.

I agree that we shouldn't overuse model validators, but in this sort of scenario it feels like the most natural approach to me. Putting a field validator on both arguments seems like overkill. Doing it on the second field is ok but just feels a bit wrong. I don't understand what is the problem mentioned in "resistant to argument reordering but not to post-assignment in any order"?

Anyhow I leave it up to you guys to decide since maybe I missed some context here on the problems associated with a model validator, but just from looking at the code here it makes sense as a model validator to me.

vizro-core/examples/scratch_dev/app.py
antonymilne: Yes, I had forgotten that... And actually it also doesn't make much sense to keep it restricted to `str` because even in a single-choice categorical selector you could have `RadioItems(options=[1, 2, 3])` in which case you want to send `int`.

So you have my permission to change to `JsonValue` wherever seems suitable 😀 I guess that's just Button and Card for now since they should be able to pass any kind of literal value. Leave the Graph/AgGrid value as just a string since that's interpreted in a special way rather than directly passed on, and we'll work out another time how to handle non-categorical selectors there in https://github.com/McK-Internal/vizro-internal/issues/2266.

In theory this means that Button and Card would be able to set non-categorical selectors already, but let's keep that restriction there for now until we handle https://github.com/McK-Internal/vizro-internal/issues/2266.

antonymilne (approved):
Perfect! ⭐ 🌟 ⭐ 🌟 

We should do a release soon with all this good stuff going in.

vizro-core/src/vizro/models/_components/button.py
petar-qb: I left the model after validator for now and we can refactor all validations for Vizro models if we decide to do that and when we find some capacity. 

> I don't understand what is the problem mentioned in "resistant to argument reordering but not to post-assignment in any order"

The **"Model validator"** (second approach) is more robust than the **"Field validator on the second argument"** (third approach) regarding the fields reordering. 
If we use the third approach and later reorder the model fields, for example moving `actions` before `href` in `Button` class definition, the validator will not catch conflicts correctly. A similar issue appears in the `Slider` model where the third approach is implemented: swapping `min` and `max` breaks most of the tests in `test_slider.py`, showing the approach is brittle.

The **"Field validator on both arguments"** (first approach) is more robust than the **"Model validator"** (second approach) regarding the post field assignment. 
Compared to the **Field validator on both arguments** (first approach), the **Model validator** is less strict when fields are updated after initialisation. With the first approach, setting `href` first and assigning `actions` later (e.g during `pre_build`) correctly raises a validation error. The model after validator, on the other hand, will not catch this since it only validates during initialisation, showing the approach is brittle.

vizro-ai/tests/unit/vizro-ai/conftest.py
```diff
@@ -4,8 +4,8 @@
 
 import pytest
 import vizro.plotly.express as px
-from langchain.output_parsers import PydanticOutputParser
 from langchain_community.llms.fake import FakeListLLM
+from langchain_core.output_parsers.pydantic import PydanticOutputParser
```
petar-qb, line 8: FYI: @lingyielia 

The `langchain` released a breaking `1.0.0` version, and some unit tests failed so I fixed that by changing this line.

lingyielia: Thanks @petar-qb for the quick action! 🥇

PR #1443, opened 2025-10-16
[Feat] Add additional arguments to `vm.Card`

nadijagraca:
## Description
closes: https://github.com/McK-Internal/vizro-internal/issues/2265
closes: https://github.com/McK-Internal/vizro-internal/issues/1643

This PR enabled:
-  adding  `header`, `footer` and `title` arguments to `vm.Card` model for easier formatting
- adding `actions` argument to enable clickable `vm.Card` 

## Notice


vizro-core/changelog.d/20251016_110924_nadija_ratkusic_graca_update_card_model.md
```diff
@@ -0,0 +1,48 @@
...
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
```
petar-qb:
```suggestion

### Added

- Enable `header`, `title` and `footer` arguments inside `vm.Card`. ([#1443](https://github.com/mckinsey/vizro/pull/1443))
 
- Actions can be triggered when clicking a `vm.Card`. ([#1443](https://github.com/mckinsey/vizro/pull/1443))

- `vm.Card` can be the source of a `set_control` action. ([#1443](https://github.com/mckinsey/vizro/pull/1443))

```

vizro-core/docs/pages/user-guides/card.md
```diff
@@ -197,6 +197,52 @@ The [`Card`][vizro.models.Card] uses the `dcc.Markdown` component from Dash as i
 
         [![CardText]][cardtext]
 
+## Add header and footer
+
+You can add header and footer to your [`Card`][vizro.models.Card] model by specifying `header` and `footer` arguments.
+
```
petar-qb, line 203: Should it be explained that the header and the footer can be any Markdown object?

vizro-core/docs/pages/user-guides/card.md
```diff
@@ -569,6 +615,58 @@ To create a KPI card, use the existing KPI card functions from [`vizro.figures`]
 
 For detailed examples on how to create a KPI card, refer to the [figure user guide on KPI cards](figure.md#key-performance-indicator-kpi-cards).
 
+## Add a tooltip
+
+The `description` argument enables you to add helpful context to your button by displaying an info icon next to its text. Hovering over the icon shows a tooltip with your chosen text.
```
petar-qb:
```suggestion
The `description` argument enables you to add helpful context to your card by displaying an info icon next to its title. Hovering over the icon shows a tooltip with your chosen text.
```

As `Card.title` is not explained anywhere (as far as I can see), maybe this is the best place for this. How about renaming this section to `"Add title with a tooltip "` and then add a sentence that title is plain string (not a markdown).

vizro-core/examples/scratch_dev/app.py
petar-qb, line 1: I really appreciate well-made `app.py` examples like this one. 🥇

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -92,6 +94,17 @@ class set_control(_AbstractAction):
             actions=va.set_control(control="target_control", value="A"),
         )
         ```
+
+    Example: `Card` as trigger
+        ```python
+        import vizro.actions as va
```
petar-qb: You can skip this line as it's skipped in other example from above as well.
```suggestion
```

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -55,13 +98,25 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
             id=f"{self.id}-text", children=self.text, dangerously_allow_html=False, className="card-text"
         )
 
-        card_content = (
+        description = self.description.build().children if self.description is not None else [None]
+
+        title = html.Div(
+            children=[html.H4(self.title, className="card-title") if self.title else None, *description],
+            className="card-title-outer",
+        )
```
petar-qb: [no action required - just a personal expression that no one has asked for 😄]

I tend to prefer fewer helper variables and building content directly inline for readability.
Totally get that you followed the existing pattern from other files though, so no action needed here.

```suggestion
        title = html.Div(
            children=[
                html.H4(self.title, className="card-title") if self.title else None,
                self.description.build().children if self.description is not None else [None]
             ],
            className="card-title-outer",
        )
```

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -55,13 +98,25 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
+        return value
+
     @_log_call
     def build(self):
         text = dcc.Markdown(
             id=f"{self.id}-text", children=self.text, dangerously_allow_html=False, className="card-text"
         )
 
-        card_content = (
+        description = self.description.build().children if self.description is not None else [None]
```
petar-qb: I guess this would also work just as well. If you agree, let's change it here and in the `container.py` too.
```suggestion
        description = self.description.build().children if self.description else [None]
```

vizro-core/src/vizro/models/_components/card.py
petar-qb, line 1: Could you add IDs to header, title, footer, so the  `_action_outputs` mappings could be enhanced too? 
So, similarly like it's done in the Graph:

```py
    @property
        return {
            "__default__": f"{self.id}.figure",
            **({"title": f"{self.id}_title.children"} if self.title else {}),
            **({"header": f"{self.id}_header.children"} if self.header else {}),
            **({"footer": f"{self.id}_footer.children"} if self.footer else {}),
        }
```

P.S. initialisation unit tests in that case should be improved (again similarly like it's done for the graph)

vizro-core/tests/unit/vizro/models/_components/test_card.py
```diff
@@ -45,53 +52,209 @@ def test_none_as_text(self):
...
+                class_name="card-nav",
             ),
-            class_name="card-nav",
+            className="card-wrapper",
         )
 
         assert_component_equal(card, expected_card)
 
     def test_card_build_wo_href(self):
         card = vm.Card(id="card_id", text="Hello")
```
petar-qb: Let's move this one as the first test in the `TestBuildMethod` class and rename it to `test_card_build_mandatory`.

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -16,24 +17,51 @@ class Card(VizroBaseModel):
...
+        footer (str): Markdown text positioned below the `Card`. Follows the CommonMark specification.
+            Ideal for providing further details such as sources, disclaimers, or additional notes. Defaults to `""`.
         href (str): URL (relative or absolute) to navigate to. If not provided the Card serves as a text card
             only. Defaults to `""`.
         extra (Optional[dict[str, Any]]): Extra keyword arguments that are passed to `dbc.Card` and overwrite any
             defaults chosen by the Vizro team. This may have unexpected behavior.
             Visit the [dbc documentation](https://www.dash-bootstrap-components.com/docs/components/card/)
             to see all available arguments. [Not part of the official Vizro schema](../explanation/schema.md) and the
             underlying component may change in the future. Defaults to `{}`.
+        actions (ActionsType): See [`ActionsType`][vizro.models.types.ActionsType].
```
antonymilne, line 33: You need `description` in this docstring too.

huong-li-nguyen:
> With hesitation 1 + 2 combined, my feeling is that at the moment there's maybe too many fields (4) compared to the number of distinct sections I can perceive on the screen (3, divided by the horizontal lines). So I think we need to figure out which fields are genuinely useful and/or how to distinguish them a bit better.
> 
> wdyt @huong-li-nguyen? 🙏
> 
> Whatever you guys come up with, feel free to merge while I'm away. I'm confident the result will be good 🚀

Hmm… I think the `header` and `footer` should be markdown fields so people can customize the text content more easily.  
For example, I would have expected the text below to be rendered as titles rather than displaying the hash symbol.  

Also, while we usually place the info icon next to the title text, I think for the Card it should always appear in the **top-right corner**. Otherwise, if multiple cards are displayed side by side, the icons won’t align properly.  

If the info icon is consistently placed in the top-right corner, I don't really see the need for a separate `title` argument - that could be handled through the `header` instead. Removing the `header` would actually feel more restrictive.  

We could also wait until Steph joins next week to revisit the Card design, but for now, I'd lean toward keeping both `header` and `footer`, and fixing the info icon to the top-right corner.

Screenshot 2025-10-20 at 10 13 11

huong-li-nguyen:
These card designs were initially the reason why we wanted to create more sophisticated Cards, but I assume that even with markdown fields, we can't render Buttons inside for example. Images and links should be possible via the markdown fields. Is that correct?

Screenshot 2025-10-20 at 10 21 43

vizro-core/docs/pages/user-guides/card.md
huong-li-nguyen: Yeees, and I think they should actually be Markdown objects! I think at the moment, they are simple strings?

nadijagraca:
> These card designs were initially the reason why we wanted to create more sophisticated Cards, but I assume that even with markdown fields, we can't render Buttons inside for example. Images and links should be possible via the markdown fields. Is that correct?
> 
> Screenshot 2025-10-20 at 10 21 43

`dbc.CardBody` can receive any dash component, so in theory we could allow users to add buttons or any other component to vm.Card.

antonymilne:
> `dbc.CardBody` can receive any dash component, so in theory we could allow users to add buttons or any other component to vm.Card.

The problem though is that we don't have a good way for a user to specify that they want a button _from inside Markdown text_ - we'd have to add a new `button` or whatever field, which I don't think we should do. There actually are ways we could extend Markdown to enable `[button]` or something like that in future but let's not do that now anyway. Especially given we now have `actions` as well as `href`, so clicking the whole card acts similar to clicking a button, I think enabling buttons here is not important.

So based on what @huong-li-nguyen says and those designs, I would probably do this:
* `header` as markdown placed where you have it now
* `footer` as markdown placed where you have it now
* `description` but move the icon into the top right of the header
* remove `title` (it can just go inside `text`)
* forget about adding buttons - just use `href`/`actions` which effectively makes the whole card a button

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -55,13 +98,25 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
             "text": f"{self.id}-text.children",
         }
 
+    @staticmethod
+    def _get_value_from_trigger(value: str, trigger: int) -> JsonValue:
```
petar-qb: Here's a context about type hints of `_get_value_from_trigger` -> https://github.com/mckinsey/vizro/pull/1441#discussion_r2431413894

So, let's change it here like:

```suggestion
    def _get_value_from_trigger(value: JsonValue, trigger: int) -> JsonValue:
```

---

Would you mind also changing the signature of the `Button._get_value_from_trigger`
 from:
`def _get_value_from_trigger(value: JsonValue, *args) -> Any:`
to:
`def _get_value_from_trigger(value: JsonValue, trigger: int) -> JsonValue:`

I forgot to do that in already merged [PR](https://github.com/mckinsey/vizro/pull/1441).

vizro-core/docs/pages/user-guides/card.md
```diff
@@ -197,6 +197,52 @@ The [`Card`][vizro.models.Card] uses the `dcc.Markdown` component from Dash as i
 
         [![CardText]][cardtext]
 
+## Add header and footer
+
+You can add header and footer to your [`Card`][vizro.models.Card] model by specifying `header` and `footer` arguments. Each supports any Markdown object, providing flexibility for text, links, or inline formatting, as the underlying components are dbc.CardHeader and dbc.CardFooter.
```
petar-qb:
```suggestion
You can add header and footer to your [`Card`][vizro.models.Card] model by specifying `header` and `footer` arguments. Both accept Markdown content, allowing you to add text, links, and inline formatting with ease.
```

vizro-core/docs/pages/user-guides/card.md
```diff
@@ -569,6 +615,60 @@ To create a KPI card, use the existing KPI card functions from [`vizro.figures`]
 
 For detailed examples on how to create a KPI card, refer to the [figure user guide on KPI cards](figure.md#key-performance-indicator-kpi-cards).
 
+## Add title with a tooltip
```
petar-qb: I guess the decision was to remove the `title`, which means that this part should be updated and only describe the `description` parameter.

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -108,7 +108,7 @@ def _action_inputs(self) -> dict[str, _IdProperty]:
         return {"__default__": f"{self.id}.n_clicks"}
 
     @staticmethod
-    def _get_value_from_trigger(value: JsonValue, *args) -> Any:
+    def _get_value_from_trigger(value: JsonValue, *args) -> JsonValue:
```
petar-qb:
```suggestion
    def _get_value_from_trigger(value: JsonValue, trigger: int) -> JsonValue:
```

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -47,21 +90,46 @@ class Card(VizroBaseModel):
             ),
         ]
     ]
+    actions: ActionsType = []
+
+    @model_validator(mode="after")
+    def _validate_href_if_actions_set(self):
+        if self.href and self.actions:
+            raise ValueError("You cannot define both `href` and `actions` in `Card` model.")
+        return self
```
petar-qb, line 86: Let's align with the button's implementation (method name and exception message). Change here or there whatever you think it's better.

petar-qb (approved):
Really amazing work @nadijagraca! 🥇 ✅

vizro-core/docs/pages/user-guides/card.md
nadijagraca: Nice catch. 🙌

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -1,13 +1,29 @@
...
-from vizro.models import VizroBaseModel
-from vizro.models._models_utils import _log_call
-from vizro.models.types import _IdProperty
+from vizro.models import Tooltip, VizroBaseModel
+from vizro.models._models_utils import _log_call, make_actions_chain
+from vizro.models._tooltip import coerce_str_to_tooltip
+from vizro.models.types import ActionsType, _IdProperty
+
+
+def warn_description_without_header(description, info: ValidationInfo):
```
petar-qb: Why do we raise a warning when a `description` is defined without a `header`?

In most cases, users will likely want to include a `description` without necessarily having a `header`, so it might make sense to support that scenario instead of warning.

From an implementation perspective, we could render the `description` inside the `dbc.CardBody` next to the `card_text` when no `header` is defined. To achieve the desired layout, we can apply the following CSS to the CardBody:

```css
display: flex;
flex-direction: row;
```

vizro-core/changelog.d/20251016_110924_nadija_ratkusic_graca_update_card_model.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Added
+
+- Enable `header`, `description` and `footer` arguments inside `vm.Card`. ([#1443](https://github.com/mckinsey/vizro/pull/1443))
+
+- Actions can be triggered when clicking a `vm.Card`. ([#1443](https://github.com/mckinsey/vizro/pull/1443))
+
+- `vm.Card` can be the source of a `set_control` action. ([#1443](https://github.com/mckinsey/vizro/pull/1443))
```
huong-li-nguyen: Remember this should be easy to understand for the user :) You could also link to the relevant docs examples below then. 

```suggestion
- Add support for `header`, `description`, and `footer` in `Card` for more flexible styling. See our user guide on [XXX]() for an example. ([#1443](https://github.com/mckinsey/vizro/pull/1443))

- Make `Card` clickable to trigger actions. ([#1443](https://github.com/mckinsey/vizro/pull/1443))

- Allow `Card` to trigger `set_control` actions (e.g., to filter another chart).  See our user guide on [XXX]()  for an example. ([#1443](https://github.com/mckinsey/vizro/pull/1443))
```

vizro-core/docs/pages/user-guides/card.md
```diff
@@ -197,6 +197,52 @@ The [`Card`][vizro.models.Card] uses the `dcc.Markdown` component from Dash as i
 
         [![CardText]][cardtext]
 
+## Add header and footer
+
+YouYou can add header and footer to your [`Card`][vizro.models.Card] model by specifying `header` and `footer` arguments. Both accept Markdown content, allowing you to add text, links, and inline formatting with ease.
```
huong-li-nguyen:
```suggestion
You can now add a header and footer to your [`Card`][vizro.models.Card] by using the `header` and `footer` arguments. Both fields support [Markdown text](https://markdown-guide.readthedocs.io/), so you can easily include text, links, and inline formatting.
```

vizro-core/docs/pages/user-guides/card.md
```diff
@@ -569,6 +615,60 @@ To create a KPI card, use the existing KPI card functions from [`vizro.figures`]
 
 For detailed examples on how to create a KPI card, refer to the [figure user guide on KPI cards](figure.md#key-performance-indicator-kpi-cards).
 
+## Add a tooltip
+
+Add a title to your card by passing a string to the `title` argument.
+
+The `description` argument enables you to add helpful context to your button by displaying an info icon in the top right corner of the card header. Hovering over the icon shows a tooltip with your chosen text.
+
+You can provide [Markdown text](https://markdown-guide.readthedocs.io/) as a string to use the default info icon or a [`Tooltip`][vizro.models.Tooltip] model to use any icon from the [Google Material Icons library](https://fonts.google.com/icons).
```
huong-li-nguyen: I agree with @petar-qb by the way that the info-icon should always appear in the top-right corner. So if the `header` is not provided, it should be in the top-right corner of the card-text :) 

```suggestion
Use the `description` argument to add helpful context to your `Card`. This displays an info icon in the top-right corner, and hovering over it reveals a tooltip with your chosen text.

You can provide [Markdown text](https://markdown-guide.readthedocs.io/) to `Card.description` to use the default info icon, or pass a [`Tooltip`][vizro.models.Tooltip] model to customize the icon using any symbol from the [Google Material Icons library](https://fonts.google.com/icons).
```

vizro-core/docs/pages/user-guides/card.md
```diff
@@ -569,6 +615,60 @@ To create a KPI card, use the existing KPI card functions from [`vizro.figures`]
...
+                  Lorem Ipsum is simply dummy text of the printing and typesetting industry.
+
+                  Lorem Ipsum has been the industry's standard dummy text ever since the 1500s
+                type: card
+            title: Card with a tooltip
+        ```
+
+    === "Result"
+
+        [![CardInfoIcon]][cardinfoicon]
```
huong-li-nguyen, line 668: Shall we add another example where the Card is filtering another chart? That would be great to showcase

vizro-core/src/vizro/models/_components/card.py
huong-li-nguyen: Agree! It should always be in the top-right corner, but it doesn't have to be the header. Can also be the card-text if no header is defined.

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -16,24 +32,51 @@ class Card(VizroBaseModel):
     Args:
         type (Literal["card"]): Defaults to `"card"`.
         text (str): Markdown string to create card title/text that should adhere to the CommonMark Spec.
+        header (str): Markdown text positioned above the `Card.title`. Follows the CommonMark specification.
+            Ideal for adding supplementary information. Defaults to `""`.
+        footer (str): Markdown text positioned below the `Card`. Follows the CommonMark specification.
+            Ideal for providing further details such as sources, disclaimers, or additional notes. Defaults to `""`.
```
huong-li-nguyen: Thee is no Card.title anymore, please update the descriptions here :)

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -16,24 +32,51 @@ class Card(VizroBaseModel):
     Args:
         type (Literal["card"]): Defaults to `"card"`.
         text (str): Markdown string to create card title/text that should adhere to the CommonMark Spec.
+        header (str): Markdown text positioned above the `Card.title`. Follows the CommonMark specification.
+            Ideal for adding supplementary information. Defaults to `""`.
+        footer (str): Markdown text positioned below the `Card`. Follows the CommonMark specification.
```
huong-li-nguyen:
```suggestion
        footer (str): Markdown text positioned at the bottom of the `Card`. Follows the CommonMark specification.
```

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -16,24 +32,51 @@ class Card(VizroBaseModel):
...
         "",
         description="URL (relative or absolute) to navigate to. If not provided the Card serves as a text card only.",
     )
+    description: Annotated[
+        Optional[Tooltip],
+        BeforeValidator(coerce_str_to_tooltip),
+        AfterValidator(warn_description_without_header),
+        Field(
+            default=None,
+            description="""Optional markdown string that adds an icon next to the card header.
```
huong-li-nguyen:
```suggestion
            description="""Optional markdown string that adds an icon in the top-right corner of the Card.
```

huong-li-nguyen:
I think some of the docs + descriptions are now outdated, because we removed the `title` argument. I fixed a few occasions, but please double-check again :) I agree with @petar-qb , that the info-icon can appear in the card-text if no header is provided. It should just appear in the top-right corner - doesn't have to be in the header only. 

Also I think it would be nice to actually show a code example where the card is filtering a chart!

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -16,24 +32,51 @@ class Card(VizroBaseModel):
...
+        description="Markdown text positioned below the `Card`. Follows the CommonMark specification. Ideal for "
+        "providing further details such as sources, disclaimers, or additional notes.",
+    )
     href: str = Field(
         "",
         description="URL (relative or absolute) to navigate to. If not provided the Card serves as a text card only.",
     )
+    description: Annotated[
+        Optional[Tooltip],
+        BeforeValidator(coerce_str_to_tooltip),
```
maxschulz-COL, line 59: Do we not need the comment here that we have for the other models:
```
```

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -16,24 +32,51 @@ class Card(VizroBaseModel):
...
 
     """
 
     type: Literal["card"] = "card"
     text: str = Field(
         description="Markdown string to create card title/text that should adhere to the CommonMark Spec."
     )
+    header: str = Field(
+        default="",
+        description="Markdown text positioned above the `Card.title`. Follows the CommonMark specification. Ideal for "
```
maxschulz-COL: I prefer triple quotes :P

vizro-core/src/vizro/models/_components/card.py
```diff
@@ -47,21 +90,47 @@ class Card(VizroBaseModel):
             ),
         ]
     ]
+    actions: ActionsType = []
+
+    @model_validator(mode="after")
```
maxschulz-COL, line 81: Can this not be a field validator with access to the other field?

https://docs.pydantic.dev/latest/concepts/validators/#validation-data

petar-qb: In [this conversation](https://github.com/mckinsey/vizro/pull/1441#discussion_r2438654625) we decided to handle this kind of situations using a model validator. Ideally, situations like this should be solved with two field validators (not a single field validator on the latest class field) for the reason explained in the linked conversation.

Here's a separate ticket for solving it -> https://github.com/McK-Internal/vizro-internal/issues/1554

maxschulz-COL: Ah I never looked at that thread again for some reason - not sure why I missed it in my notifications.

Happy to follow your guidance there, the field re-ordering although unlikely is a valid point! 

Without much of a very deep argument I think `model_validators` should be used as last resort become you are changing things post instantiation and acting on the instance directly, and that may lead potentially (again more gut feeling than anything else) to other complication. My main current point would be the order of model validators potentially starting to be difficult to understand in case of inheritance etc

But since I am clearly not finding any striking cases, let's use the model validator.

vizro-core/changelog.d/20251016_110924_nadija_ratkusic_graca_update_card_model.md
```diff
@@ -0,0 +1,48 @@
...
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Add support for `header`, `description`, and `footer` in `Card` for more flexible styling. See our user guide on [Cards](https://vizro.readthedocs.io/en/stable/pages/user-guides/card/#add-header-and-footer) for an example. ([#1443](https://github.com/mckinsey/vizro/pull/1443))
+- Make `Card` clickable to trigger actions. ([#1443](https://github.com/mckinsey/vizro/pull/1443))
+- Allow `Card` to trigger `set_control` actions (e.g., to filter another chart).  See our user guide on [Actions]((https://vizro.readthedocs.io/en/stable/pages/user-guides/actions/#trigger-an-action-with-a-card))  for an example. ([#1443](https://github.com/mckinsey/vizro/pull/1443))
+<!--
```
huong-li-nguyen:
```suggestion
- Make `Card` clickable to trigger actions (e.g., to filter another chart).  See our user guide on [Actions]((https://vizro.readthedocs.io/en/stable/pages/user-guides/actions/#trigger-an-action-with-a-card))  for an example. ([#1443](https://github.com/mckinsey/vizro/pull/1443))
<!--
```

vizro-core/docs/pages/user-guides/actions.md
```diff
@@ -22,7 +22,7 @@ To use an action:
 
 You can also execute [multiple actions with a single trigger](#multiple-actions).
 
-Some actions are usually triggered by certain components, for example [`export_data`](data-actions.md#export-data) is usually triggered by clicking a [button](button.md). However, many actions can be triggered by any component, for example you could also trigger `export_data` by clicking a point on a graph.
+Some actions are usually triggered by certain components, for example [`export_data`](data-actions.md#export-data) is usually triggered by clicking a [button](button.md). However, many actions can be triggered by any component, for example you could also trigger `export_data` by clicking a point on a graph, or you can trigger `set_control` action by clicking on a [`card`](card.md#trigger-an-action-with-a-card).
```
huong-li-nguyen:
```suggestion
Some actions are usually triggered by certain components, for example [`export_data`](data-actions.md#export-data) is usually triggered by clicking a [button](button.md). However, many actions can be triggered by any component, for example you could also trigger `export_data` by clicking a point on a graph, or you can trigger `set_control` action by clicking on a [`card`](card.md#trigger-an-action-with-a-card).
```

@stichbury - do you have a guidance on how to phrase this here? "trigger `set_control` is technically correct, but overall I find this hard to understand as some users might not even know what `set_control` is (distinct to `export_data`). `set_control` in the end means that you can filter a Graph or Table, by clicking on another Graph or Table, so it's a chart interaction / filter interaction. So my question is whether to be technically correct as it is, or to rephrase this in non-technical terms.

vizro-core/docs/pages/user-guides/card.md
```diff
@@ -569,6 +615,136 @@ To create a KPI card, use the existing KPI card functions from [`vizro.figures`]
 
 For detailed examples on how to create a KPI card, refer to the [figure user guide on KPI cards](figure.md#key-performance-indicator-kpi-cards).
 
+## Add a tooltip
+
+Use the `description` argument to add helpful context to your `Card`. This displays an info icon in the top-right corner, and hovering over it reveals a tooltip with your chosen text.
+
+You can provide [Markdown text](https://markdown-guide.readthedocs.io/) to `description` argument to use the default info icon, or pass a [`Tooltip`][vizro.models.Tooltip] model to customize the icon using any symbol from the [Google Material Icons library](https://fonts.google.com/icons).
```
huong-li-nguyen:
```suggestion
You can provide [Markdown text](https://markdown-guide.readthedocs.io/) to the `description` argument to use the default info icon, or pass a [`Tooltip`][vizro.models.Tooltip] model to customize the icon using any symbol from the [Google Material Icons library](https://fonts.google.com/icons).
```

vizro-core/docs/pages/user-guides/card.md
```diff
@@ -569,6 +615,136 @@ To create a KPI card, use the existing KPI card functions from [`vizro.figures`]
...
+            title: Card with a tooltip
+        ```
+
+    === "Result"
+
+        [![CardInfoIcon]][cardinfoicon]
+
+## Trigger an action with a card
+
+Here is an example action that uses the [`set_control` action][vizro.actions.set_control] when a [card](card.md) is clicked.
```
huong-li-nguyen: In the below, I would reference the Card API reference, because this paragraph is already written in the `card.md` file. So if you were to click on this, this just jumps to the beginning of the same page :)

```suggestion
Below is an example showing how to use the [set_control action][vizro.actions.set_control] to filter another Graph or Table when a [`Card`](API REFERENCE) is clicked.
```

vizro-core/docs/pages/user-guides/actions.md
stichbury: Great question! I think there's space in this page to go from simple statement to example, and from what I can see the section "Trigger an action with a graph" does illustrate this. So how about we revise. Let me have a go in a suggestion...

stichbury:
```suggestion
Some actions are typically triggered by certain components, for example [`export_data`](data-actions.md#export-data) is usually triggered by clicking a [button](button.md). However, many actions can be triggered by any component, for example:

 - you could also trigger `export_data` by clicking a point on a graph. There's an illustration of this in the <NAME SECTION HERE AND INCLUDE LINK>
- you can trigger a chart and filter interaction by clicking on a [`card`](card.md#trigger-an-action-with-a-card). There's an illustration of this in the <NAME SECTION HERE AND INCLUDE LINK>
```

vizro-core/docs/pages/user-guides/card.md
stichbury:
```suggestion
The example below shows how to use the [set_control action][vizro.actions.set_control] to filter another Graph or Table when a [`Card`](API REFERENCE) is clicked.
```

stichbury: Blending both together...Where possible, we avoid the -ing from of verbs and instead keep it really straightforward (so use "shows" rather than "showing")

vizro-core/docs/pages/user-guides/actions.md
nadijagraca: I don't think we have an example of triggering `export_data` by clicking a point on a graph. Do you want me to add an example, and if yes, where would that example live? @petar-qb do you have any suggestions?

huong-li-nguyen: No, we don't need an example for the export_data.  I don't think it makes much sense to trigger it from a graph. So just remove that bullet point :)

petar-qb: I agree with Li

PR #1449, opened 2025-10-21
[Feat] Simplify outputs for custom download actions

nadijagraca:
## Description
closes https://github.com/McK-Internal/vizro-internal/issues/2051

This PR simplifies custom download data actions. 

Instead of writing `vm.Action(function=download_stuff(), outputs=["vizro_download.data"])` 

users can write  `vm.Action(function=download_stuff(), outputs=["vizro_download"])`

## Notice


vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -188,6 +194,10 @@ def _transform_dependency(dependency: _IdOrIdProperty, type: Literal["output", "
         try:
             return getattr(model_manager[component_id], attribute_type)[component_property]
         except (KeyError, AttributeError) as exc:
+            # Fallback: try builtin_components
+            with suppress(KeyError, AttributeError):
+                return getattr(builtin_components[component_id], attribute_type)[component_property]
```
nadijagraca: @antonymilne Is the placement of the builtin_components okay? Should we handle it in actions or in model manager?

petar-qb: After thinking more about the overall structure, I believe the best placement for this logic is in `model_manager.py`. The `builtin_components` concept aligns more naturally with the responsibilities of the model manager rather than the actions module.

Placing it there would also improve discoverability, so if we ever need to access it elsewhere, it feels more intuitive to import it as: `from vizro.model_manager import BUILTIN_COMPONENTS`


So I’d suggest removing any changes in the `_action.py` and refactoring the `model_manager.py` as follows:
```py
# imports
...

Model = TypeVar("Model", bound="VizroBaseModel")

class _VizroDownload:
    _action_outputs = {"__default__": "vizro_download.data"}

BUILTIN_COMPONENTS = {"vizro_download": _VizroDownload()}

class FIGURE_MODELS:
    pass
...

class ModelManager:
    ...
    def __getitem__(self, model_id: ModelID) -> VizroBaseModel:
        # new comment
        with suppress(KeyError):
            return BUILTIN_COMPONENTS[model_id]
        return self.__models[model_id]
    ...
...
```

Also, I suggest skipping the explicit _VizroDownload class instantiation and simplifying it by doing this:
```py
BUILTIN_COMPONENTS = {
    "vizro_download": type("", (), {"_action_outputs": {"__default__": "vizro_download.data"}})()
}
```
or
```py
from types import SimpleNamespace

BUILTIN_COMPONENTS = {
    "vizro_download": SimpleNamespace(_action_outputs={"__default__": "vizro_download.data"})
}
```

petar-qb:
Another open question from the ticket is what should be done for the `vizro_url`.
Is there the most common property we can set as a default? I can't tell from top of my head is it the `href` or the `pathname`. And, if it's the `pathname` should we somehow internally handle the `dash.get_relative_path(input)` or not.

antonymilne:
First of all, a thousand apologies it has taken me so long to get around to this 🙈 I hope it's a case of better late than never, anyway... 🙏 

I took a look at the code myself this morning and found what we have already pretty confusing so I started by refactoring that: [49ca1ba](https://github.com/mckinsey/vizro/pull/1449/commits/49ca1baf13c9740119b406eb92e6f4f3162d6203).

Then I've suggested an approach that's somewhere between @nadijagraca's original one and @petar-qb suggestion: [76e25bc](https://github.com/mckinsey/vizro/pull/1449/commits/76e25bce9edeb5056c6fe0f9b0c3e801f1f82e42)
* at least for now I'd like to keep `model_manager` strictly for Vizro models and not mix in other things that aren't quite full `VizroBaseModel` like we have here
* using `SimpleNamespace` is a great idea

See what you both think of it!

I'm not sure what to do about `vizro_url` so let's just not make that change for now. I think it might become clearer when we do more cross-page actions. I will note this on the ticket.

So I think hopefully @nadijagraca can now write some tests for this and we're good to go! No need to document or add to the changelog - I'll add it to my next batch of docs later.

petar-qb:
> I took a look at the code myself this morning and found what we have already pretty confusing so I started by refactoring that: [49ca1ba](https://github.com/mckinsey/vizro/pull/1449/commits/49ca1baf13c9740119b406eb92e6f4f3162d6203).

Thanks for that! 
 
> Then I've suggested an approach that's somewhere between @nadijagraca's original one and @petar-qb suggestion: [76e25bc](https://github.com/mckinsey/vizro/pull/1449/commits/76e25bce9edeb5056c6fe0f9b0c3e801f1f82e42)
> 
> * at least for now I'd like to keep `model_manager` strictly for Vizro models and not mix in other things that aren't quite full `VizroBaseModel` like we have here
> * using `SimpleNamespace` is a great idea
> 
> See what you both think of it!

I think that this shows the best of both! 👍

> I'm not sure what to do about `vizro_url` so let's just not make that change for now. I think it might become clearer when we do more cross-page actions. I will note this on the ticket.

I agree 👍 

> So I think hopefully @nadijagraca can now write some tests for this and we're good to go! No need to document or add to the changelog - I'll add it to my next batch of docs later.

I'll properly review the PR when it's fully completed with tests 😃

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -171,10 +172,14 @@ def _transform_dependency(dependency: _IdOrIdProperty, type: Literal["output", "
                 f"'<model_id>.<argument_name>'."
             )
 
+        # Combine builtin components with real models in model_manager (which take priority).
+        builtin_components = {"vizro_download": SimpleNamespace(_action_outputs={"__default__": "vizro_download.data"})}
+        models = builtin_components | {key: model_manager[key] for key in model_manager}
```
petar-qb:
```suggestion
        models = builtin_components | model_manager
```

or 

I found that [ChainMap](https://docs.python.org/3/library/collections.html#collections.ChainMap) incorporates the underlying mappings by reference. So, there's no unnecessary dictionary copying every time any user triggers any action.
```suggestion
        from collections import ChainMap
        models = ChainMap(model_manager, builtin_components)
```

antonymilne: This would be great if it worked but I suspect it doesn't, because `model_manager` doesn't actually implement `Mapping`. I considered adding methods to it so we could do this here, but decided not to - better to leave it until after the model manager work is done.

If it does work already then feel free to do it like this though!

petar-qb: It works with the `ChainMap`. I just pushed the changes

PR #984, opened 2025-02-03
[Bug] Fix flickering scroll bar and increase lower bound for `plotly`

huong-li-nguyen:
## Description
- Upgrade `plotly` version to `6.0.0 `in examples env and visual-vocabulary demo
- Increase lower bound of `plotly==5.24.0`, see discussion: https://github.com/mckinsey/vizro/pull/984#discussion_r1940809090
- Set default for `map_style` given increase in lower bound for `plotly`
- Fix flickering scroll bug when using new Plotly maps inside `vizro` with `plotly==6.0.0`.

## Screenshot

Before: https://py.cafe/huong-li-nguyen/vizro-us-maps-car-share

## Notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

    - I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
    - I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
    - I certify that the use of this contribution as authorized by the Apache 2.0 license does not violate the intellectual property rights of anyone else.
    - I have not referenced individuals, products or companies in any commits, directly or indirectly.
    - I have not added data or restricted code in any commits, directly or indirectly.

PR #1458, opened 2025-10-24
[Bug] Fix graph rendering when theme changes in `Flex` layout

petar-qb:
closes https://github.com/McK-Internal/vizro-internal/issues/2243

## Description
Issue: When using the `Flex` layout, switching the theme selector does not properly update the graph’s theme.

## Screen recording

https://github.com/user-attachments/assets/279be46e-c53f-4be8-b82c-23d8d7c224c6

## Notice

### DIFF
--- a/vizro-core/src/vizro/static/css/tiles.css
+++ b/vizro-core/src/vizro/static/css/tiles.css
@@ -1,7 +1,6 @@
 .table-container,
 .dash-graph {
   flex: 1 1 auto;
-  overflow: auto;
   width: 100%;
 }
 
--- a/vizro-core/examples/scratch_dev/app.py
+++ b/vizro-core/examples/scratch_dev/app.py
@@ -7,7 +7,10 @@ df = px.data.iris()
 page_1 = vm.Page(
     title="BUG theme switch doesn't work with Flex layout",
     layout=vm.Flex(),
-    components=[vm.Graph(figure=px.scatter(data_frame=df, x="sepal_width", y="sepal_length"))],
+    components=[
+        vm.Graph(figure=px.scatter(data_frame=df, x="sepal_width", y="sepal_length")),
+        vm.Graph(figure=px.scatter(data_frame=df.head(10), x="sepal_width", y="sepal_length")),
+    ],
 )
 
 dashboard = vm.Dashboard(pages=[page_1])


vizro-core/src/vizro/static/css/tiles.css
```diff
@@ -32,8 +32,3 @@
 .maplibregl-ctrl {
   display: flex;
 }
-
-/* Fixes flickering scroll-bar when using new plotly maps: https://github.com/mckinsey/vizro/pull/984 */
-.js-plotly-plot {
-  overflow: hidden;
-}
```
petar-qb, line 39: Hey @huong-li-nguyen 👋 You added this CSS change in your [PR](https://github.com/mckinsey/vizro/pull/984) and I'm wondering what impact it would have on Vizro if it's removed. Currently, this CSS is preventing charts from re-rendering when the theme changes in a Flex layout.

petar-qb:
hi @l0uden 👋 would you mind creating a screenshot test for this bug?

petar-qb:
Converting to draft PR as I figured out that just removing the css below doesn't solve the problem with multiple graphs on the page.
```css
.js-plotly-plot {
  overflow: hidden;
}
```

To enable multiple graphs rendering when theme is changed, I removed also `overflow: auto;` from `tiles.css .dash-graph` selector. It now renders multiple graphs but it's flickering a lot.

petar-qb:
Any thoughts @huong-li-nguyen @nadijagraca?

vizro-core/src/vizro/static/css/tiles.css
huong-li-nguyen: There is a PR comment with a PR and a code example :) If that works well, then it should be fine to remove!

vizro-core/src/vizro/static/css/tiles.css
```diff
@@ -1,7 +1,6 @@
 .table-container,
 .dash-graph {
   flex: 1 1 auto;
-  overflow: auto;
```
huong-li-nguyen, line 4: Here I am actually not sure - I think Vlad added it back then to fix other bugs. If you could just double-check all the examples here with that new change: https://py.cafe/huong-li-nguyen/vizro-flex-grid-layouts

If the charts look fine, then I guess it should be okay 👍

vizro-core/src/vizro/static/css/tiles.css
huong-li-nguyen: /* Fixes flickering scroll-bar when using new plotly maps: https://github.com/mckinsey/vizro/pull/984 */

huong-li-nguyen:
> closes [McK-Internal/vizro-internal#2243](https://github.com/McK-Internal/vizro-internal/issues/2243)

I don't really understand though, how the overflow behaviour could affect the chart theme to be properly updated? 🤔

petar-qb:
Here's another bug I found on main, so linking it here in case we solve even that one with a single PR -> https://github.com/McK-Internal/vizro-internal/issues/2276

l0uden:
Here's the current behaviour from the branch. Can be reproduced with any number of components more than 1.
Looks pretty the same as in @petar-qb bug from the precious comment.

Original:
image

After theme switch:
image

petar-qb:
@l0uden 
> Here's the current behaviour from the branch.

Does the same thing happen on the `main` branch too?

l0uden:
> @l0uden
> 
> > Here's the current behaviour from the branch.
> 
> Does the same thing happen on the `main` branch too?

No, the page becomes empty after switching theme with this config in main

maxschulz-COL:
Just became aware of this PR - please please fix it. It really often occurs in the MCP stuff, because it uses Flex so much now!

petar-qb:
Closing as totally different change has to be done (if even possible 😄)

PR #1462, opened 2025-10-28
[Feat] Dash v4 support

petar-qb:
closes https://github.com/McK-Internal/vizro-internal/issues/2295
closes https://github.com/McK-Internal/vizro-internal/issues/2200

## Description
This PR adds relevant backend and UI support for all `dcc` components used inside Vizro for `dash>4`.

### Key Improvements

**Dropdown Component**
- **Removed custom `select_all` implementation**: Dash v4 now provides native "Select All"/"Clear All" functionality, eliminating our custom JavaScript implementation
- **Simplified options height handling**: Previously, we manually calculated dropdown options' height on every build. Now, Dash v4 handles this automatically based on CSS max-width
- **Improved maintainability**: The simpler implementation makes `vm.Dropdown` easier to inherit from and extend

**Slider and RangeSlider Components**
- **Removed custom input fields**: Dash v4 now natively supports editable input fields for sliders, so we've removed our custom implementation and related JavaScript code
- **Deleted obsolete tests**: Removed unit tests for the now-unnecessary custom input field logic

**Styling Updates**
- Mapped Bootstrap CSS variables to Dash CSS variables for consistent theming
- Added new custom CSS for Dropdown, Slider, and RangeSlider components
- Removed legacy styling logic

## Known Issues

@petar-qb 
The only "creepy" thing that stays the same is the weird placeholder for dynamic dropdowns that can’t be replaced with standard one (just by calling the `__call__` like other models). (find the reason and the dash PR link in the code comments).

I found a few more minor inconsistencies and posted it in the [comment](https://github.com/plotly/dash/issues/3449#issuecomment-3451573683) for the dash issue.

## Screenshot

Run `hatch run example themes`

## TODO:
- [x] @nadijagraca Dropdown variant feature
- [x] @l0uden rewrite e2e tests
- [x] @petar-qb Update https://github.com/mckinsey/vizro/wiki/Developer-docs#page-loading

## Notice

### DIFF
--- a/vizro-core/examples/scratch_dev/app.py
+++ b/vizro-core/examples/scratch_dev/app.py
@@ -14,7 +14,7 @@ data_manager["dynamic_df"] = lambda number_of_points=10: df.head(number_of_point
 
 
 SPECIES_COLORS = {"setosa": "#00b4ff", "versicolor": "#ff9222", "virginica": "#3949ab"}
-"""
+
 page_1 = vm.Page(
     title="Single/Multi static DD",
     components=[
@@ -136,7 +136,7 @@ page_4 = vm.Page(
         vm.Filter(column="date_column", selector=vm.DatePicker(range=True, title="Range DatePicker")),
     ],
 )
-"""
+
 page_5 = vm.Page(
     title="Sliders stress-test",
     components=[
@@ -282,7 +282,7 @@ page_5 = vm.Page(
     ],
 )
 
-"""
+
 page_6 = vm.Page(
     title="Dropdown variants",
     components=[
@@ -343,13 +343,7 @@ page_6 = vm.Page(
         ),
     ],
 )
-"""
-dashboard = vm.Dashboard(
-    pages=[
-        # page_1, page_2, page_3, page_4,
-        page_5,
-        # page_6,
-    ]
-)
+
+dashboard = vm.Dashboard(pages=[page_1, page_2, page_3, page_4, page_5, page_6])
 if __name__ == "__main__":
     Vizro().build(dashboard).run()


vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -33,24 +31,6 @@ def validate_multi(multi, info: ValidationInfo):
...
-
-
-def _calculate_option_height(full_options: OptionsType, char_count: int) -> int:
-    """Calculates the height of the dropdown options based on the longest option."""
-    # We look at the longest option to find number_of_lines it requires. Option height is the same for all options
-    # and needs 24px for each line + 8px padding.
-    list_of_labels = _get_list_of_labels(full_options)
-    max_length = max(len(str(option)) for option in list_of_labels)
-    number_of_lines = math.ceil(max_length / char_count)
-    return 8 + 24 * number_of_lines
```
huong-li-nguyen, line 51: I haven't checked it, but before merge it's probably good to double-check all these examples we had on the optionHeight @nadijagraca. I don't know if it's still relevant with the new Dropdown, but the examples should show :)

petar-qb: I tested it for a bit and it looks like we don't need this mechanism anymore as `dash>=4` handles it automatically. See the screen recording from the PR description (from the 45th second onwards). However, I guess it should be tested more thoroughly.

huong-li-nguyen: Ah, I didn't see that! Yes, if we think it's save to remove, then we can also remove the unit tests for that 👍

huong-li-nguyen:
Some visual change requests for Dash tracked here:
- https://github.com/plotly/dash/issues/3510#issuecomment-3542468296
- https://github.com/plotly/dash/issues/3512

It might already be possible, but eventually I just didn't know how. Anyway, I am waiting for their response :)

petar-qb:
~~TODO: @petar-qb - Update to the new rc5, clean the code, check if anything is broken.~~

petar-qb:
Hey @huong-li-nguyen @nadijagraca 👋
I’ve updated to the latest `dash==4.0.0rc6`, and it looks like the UI is broken again 😕

You can see the issues by running `hatch run example` and `hatch run example themes`. The dropdown UI problems that this PR fixes with CSS seem to have started in `dash==4.0.0rc5` again.

huong-li-nguyen:
> Hey @huong-li-nguyen @nadijagraca 👋 I’ve updated to the latest `dash==4.0.0rc6`, and it looks like the UI is broken again 😕
> 
> You can see the issues by running `hatch run example` and `hatch run example themes`. The dropdown UI problems that this PR fixes with CSS seem to have started in `dash==4.0.0rc5` again.

Fixed for now :)

Screenshot 2026-01-20 at 13 42 19

vizro-core/changelog.d/20251117_125834_90609403+huong-li-nguyen_update_css_styling.md
petar-qb, line 1: Should we announce this change under the "Highlights" or "Changed"?

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,57 +1,122 @@
...
+    title="Long values in Dropdown",
+    components=[vm.Graph(figure=px.scatter(long_df, x="x", y="y"))],
+    controls=[vm.Filter(column="long_string")],
+)
+
+page_4 = vm.Page(
+    title="All selectors",
+    components=[
+        vm.Container(
+            controls=[
```
petar-qb, line 98: On the last `scratch_dev` page, filters look cramped. When they wrap to multiple rows above the graph, there’s no spacing from the container.

Image

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -132,41 +112,17 @@ def _action_inputs(self) -> dict[str, _IdProperty]:
...
+        if self.multi and not isinstance(value, list):
+            value = cast(MultiValueType, [value])
 
         description = self.description.build().children if self.description else [None]
         defaults = {
             "id": self.id,
             "options": dict_options,
             "value": value,
             "multi": self.multi,
-            "optionHeight": option_height,
```
petar-qb, line 169: In Dash v4, the dropdown width automatically adjusts to match the longest option. However, if an option exceeds the screen width, it overlaps with adjacent options in the list. Should we ignore this "edge" case and leave this behaviour as is? P.S. This scenario is practically possible only when app is opened on the cell phone.

Image

vizro-core/src/vizro/models/_components/form/dropdown.py
petar-qb, line 145: image


Multi-select dropdown looks much better now, but still not perfect in my opinion as only one or two selected values are visible. A lot of space is reserved for label `N selected  ✖️ 🔻`. Here are a few possible solutions from top of my head:
1. Leaving 3 rows space for multi dropdowns similarly like we do on the main branch?
2. Shortening `N selected  ✖️ 🔻` -> `N ✖️ 🔻` or `xN ✖️ 🔻`.
3. Remove ✖️ since the "Deselect All" option is already available in the dropdown selection menu.
4. Remove 🔻 because clicking the dropdown already opens the selection menu.
5. Decreasing the gap between `value` <gap> `N selected` <gap> `✖️` <gap> `🔻`.


Here's how it looks like when the 2. rule is applied:
image


In case we decide to shorten `N selected  ✖️ 🔻` -> `N ✖️ 🔻`, we can apply something similar to this to highlight the number `N`.

Image

Image

Do you think some of these solutions are suitable @stephkaiser, @huong-li-nguyen, @nadijagraca?

vizro-core/src/vizro/static/css/dcc.css
```diff
@@ -0,0 +1,151 @@
...
+}
+
+/* SLIDER / RANGE SLIDER  ------------------------------------------------------------ */
+.dash-slider-track {
+  height: 2px;
+}
+
+.dash-slider-container{
+  gap: 4px !important;
+  min-width: 200px;
```
petar-qb: I set min-width for sliders as they look too small when rendered within a container. It happened as their width was defined by the width of their title. Somewhere between `200px`-`250px` looks the best in the opinion of my backend-oriented eye 😄

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -132,41 +112,17 @@ def _action_inputs(self) -> dict[str, _IdProperty]:
...
-                        persistence_type="session",
-                        className="dropdown-select-all",
-                    ),
-                    # Special sentinel value used in update_dropdown_select_all.
-                    # This never gets sent to the server.
-                    "value": "__SELECT_ALL",
-                },
-                *dict_options,
-            ]
+        if self.multi and not isinstance(value, list):
```
antonymilne, line 122: Please can you remind me why we have this conversion here? When could `value` be something other than list?

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -198,51 +154,24 @@ def _build_dynamic_placeholder(self):
...
-            self._update_dropdown_select_all()
-            # hidden_select_all_dropdown is needed to ensure that clientside callback doesn't raise the no output error.
-            hidden_select_all_dropdown = [dcc.Dropdown(id=f"{self.id}_select_all", style={"display": "none"})]
-            placeholder_model = dcc.Checklist
-            placeholder_options = self.value
-        else:
-            hidden_select_all_dropdown = [None]
-            placeholder_model = dbc.RadioItems
-            placeholder_options = [self.value]  # type: ignore[assignment]
+        plh_model, plh_options = (dcc.Checklist, self.value) if self.multi else (dbc.RadioItems, [self.value])
```
antonymilne: Please can we name these `placeholder_` again? As soon as I saw `plh` I was confused and only understood what it was for because we previously called it `placeholder`. In future when that context has completely gone this will be very mysterious. 😅 

If it means breaking it back into a multi-line `if/else` like before then that's fine.

vizro-core/src/vizro/models/_components/form/range_slider.py
```diff
@@ -101,84 +101,31 @@ def _action_inputs(self) -> dict[str, _IdProperty]:
...
-        )
-
         current_value = self.value or [min, max]
 
         defaults = {
             "id": self.id,
             "min": min,
             "max": max,
-            "step": self.step,
+            # Only include `step` when defined. Passing None prevents dcc.RangeSlider from displaying input values.
```
antonymilne, line 110: Should we just change our `step` field default from `None` to `{}` to avoid this `if`?

antonymilne: And similarly for `Slider`, obviously.

vizro-core/pyproject.toml
```diff
@@ -15,7 +15,7 @@ classifiers = [
   "Programming Language :: Python :: 3.14"
 ]
 dependencies = [
-  "dash>=3.4.0,<4",  # temporary upper bound to avoid breaking changes in Dash v4
+  "dash>=4",
```
antonymilne:
```suggestion
  "dash>=4,<5",
```

Unlikely to be relevant for a while I guess but this is a good one to keep pinned.

vizro-core/src/vizro/models/_components/form/dropdown.py
huong-li-nguyen: I definitely see your point, but I'm not really a fan of options 2, 3, or 4.

- For 2, the abbreviations aren't very obvious unless you already know how the dropdown works, so I'd avoid those. And for 3 and 4, I think the visual indicators are actually useful, so I wouldn't remove them.
- Even if we reduce the gap as suggested in 5, the content would still be hidden, so it doesn't really fix the issue.

Honestly, I don't think the current behavior is that bad. It seems intentional by their designer, so the dropdown doesn't grow endlessly, and "3 selected" makes it clear not everything is visible. So I don't think the current behaviour is unexpected.

However, since @stephkaiser  is on PTO for the next two weeks, I'd just leave it as the default Dash behavior for now and maybe create a ticket for her to look at when she's back :)

vizro-core/src/vizro/models/_components/form/dropdown.py
huong-li-nguyen: Yes, I think it's fine to ignore. In any case, I think if such a use case does become a realistic issue, we could raise it on the Dash GitHub repo now :) I think most people would just shorten their labels when they see this.

vizro-core/changelog.d/20251117_125834_90609403+huong-li-nguyen_update_css_styling.md
huong-li-nguyen: Not sure 🤔 Functionally, nothing has changed for the user, right? Because most of that newly added functionality was provided by us in a custom way before, so now Dash is just taking care of it. Visually, it looks a bit different, but for the visual changes alone, I don't think it's worth mentioning. I'll leave that up to you guys!

vizro-core/src/vizro/models/_components/form/dropdown.py
antonymilne, line 1: One thing that doesn't feel great to me is that the "n selected" text almost always seems to obliterate the selected options since it takes up so much space on the dropdown.
Image

I wonder whether we should make one or both of these changes:
* use a [custom label](https://dash.plotly.com/dash-core-components/dropdown#custom-labels) of just `{num_selected}`? Not sure if this is too unclear to users though what the number means (though IIRC I think the old QBDS did this). Maybe worth asking a designer (Agata?) what they think
* remove some padding from the "n selected" box?

I don't have strong feelings though - if others think it's good as it is then it's ok for me.

vizro-core/src/vizro/models/_components/form/dropdown.py
antonymilne: Yeah, I think it's ok. Maybe worth reporting as a dash bug, but I don't think we should try and fix it on Vizro.

vizro-core/src/vizro/models/_components/form/slider.py
antonymilne, line 1: Image

The version of this with `step` defined doesn't look right to me:
* vertical alignment of input field is off
* `dots` don't seem aligned on bar either?
* do we even want `dots` turned on? They seem quite "cluttered" at least in this example

More generally, it looks like the logic that governs `steps/marks` might have changed a bit from looking at https://dash.plotly.com/dash-core-components/slider and https://plotly.com/blog/dash-core-components-gets-a-design-driven-refresh-with-dash-4/ 😬 I know we spent quite a while playing around with this before, but does the new "Intelligent tick mark generation prevents visual clutter" work well for us? Is our `set_default_marks` still relevant? That was there in part as a workaround for a bug which maybe no longer exists.

vizro-core/src/vizro/models/_components/form/range_slider.py
antonymilne: See also https://github.com/mckinsey/vizro/pull/1462#discussion_r2811781318. I'm not sure exactly what the different behaviours of `marks/steps` should be now...

vizro-core/examples/scratch_dev/app.py
antonymilne: Your screenshot didn't come out, but which bit do you think looks cramped exactly? It actually seems ok to me.

vizro-core/changelog.d/20251117_125834_90609403+huong-li-nguyen_update_css_styling.md
antonymilne: Yeah, I think just put in Changed since to the Vizro user the functionality is basically the same as we had before.

vizro-core/examples/scratch_dev/app.py
antonymilne: Ah, I see the screenshot now. The only bit that maybe feels a bit too cramped to me is the gap on either side of the two dropdowns but still I think that's ok

vizro-core/src/vizro/models/_components/form/dropdown.py
antonymilne: Ahhh I see there's already a discussion on this that I missed, sorry! Let me continue there... https://github.com/mckinsey/vizro/pull/1462#discussion_r2804971092

antonymilne: I commented on this separately so let me put my thoughts here so it's all in one place. I tend to agree with @petar-qb here since I had thought the same thing completely independently. All of the suggested solutions sound ok to me actually though we would have to see how they "feel" in reality. I think the old QBDS did something very similar to what you suggest here with changing the text to just `n` in a circle. 

I see @huong-li-nguyen already removed the padding in https://github.com/mckinsey/vizro/pull/1462/changes/f431ec5cbf79933ac7c3537c422682f90f22e1fd - great minds think alike since that was one of my ideas too 😀 

Overall, I also don't love it but it's ok. So I agree with @huong-li-nguyen here to just leave as we have it for now and then ask Steph to take a look when she's back.

vizro-core/src/vizro/models/_components/form/slider.py
huong-li-nguyen: Screenshot 2026-02-16 at 12 04 20

I just pushed some visual fixes, but in general above example is not our default, it's artificially created by us because we configure a Slider from 10 to 150 with a step size of 10 in the `app.py` :D So no surprise, it looks cramped as above 😄  By default, it doesn't do that. It will go without the marks. However, I think something is off on the dash side, because i can't color the active track anymore

huong-li-nguyen:
For me this is good to go visually except for the sliders, the active trace can't be colored. I'll raise an issue on dash shortly. In general, all visuals here are okay for me, because @stephkaiser and I will redo all of them anyway, so for now I don't think it's worth changing any of the default behavior :) 

EDIT: Seems like this actually works fine visually in a pure dash app. So have to double-check later again why it doesn't work on vizro.

vizro-core/src/vizro/models/_components/form/slider.py
antonymilne: Ah ok, fair enough. This looks a lot better already with your fixes, thank you! Do you think there's anything we can do to align the input field (90 in the above) with the bar?

Agree that something seems wrong that the active track isn't highlighted. Should it also be highlighted for the one without steps also?
image

huong-li-nguyen: Yes, it should also be highlighted. Seems to work in a pure dash app though, so I am not sure what the issue is at the moment

vizro-core/src/vizro/models/_components/form/dropdown.py
petar-qb: Just double checked it and it's actually the bug on our side. Here's how it looks like in pure Dash:
image

I'll try to address these several UI issues, and will reach out to someone more experienced if I stuck

vizro-core/src/vizro/models/_components/form/dropdown.py
petar-qb: Agree it's not critical to handle it immediately. Here's the separate ticket for that -> https://github.com/McK-Internal/vizro-internal/issues/2488

vizro-core/changelog.d/20251117_125834_90609403+huong-li-nguyen_update_css_styling.md
petar-qb: Added it as:

```md
### Changed

- Add support for `dash>=4.0.0`, including the redesigned `Dropdown`, `Slider`, and `RangeSlider` components. ([#1462](https://github.com/mckinsey/vizro/pull/1462))

```

vizro-core/examples/scratch_dev/app.py
petar-qb: Added padding between controls and the container to resolve a layout issue that originates from the `main` branch. This only appeared when date picker selectors were positioned in the last controls row.

vizro-core/src/vizro/models/_components/form/dropdown.py
petar-qb: Fixed in https://github.com/mckinsey/vizro/pull/1462/commits/0b7a95cf20a3e86545dddcd07bd6bae2b2d6879f

vizro-core/src/vizro/models/_components/form/slider.py
huong-li-nguyen: Fixed

huong-li-nguyen:
I can't seem to be able to approve the PR 😄 But for me it's good to go now ✅

vizro-core/src/vizro/models/_components/form/dropdown.py
petar-qb: There's no way I can remember even though git blames me for this change 😄
What I can tell is that this enables `value="setosa"` input configuration even if it's multi-select Dropdown. 

Should I delete these two lines and validate (raise an exception in) this case (`value='setosa', multi=True`) in `validate_multi` too?

vizro-core/src/vizro/models/_components/form/range_slider.py
petar-qb: > Should we just change our step field default from None to {} to avoid this if?

I don't think we should do it. It could become quite hacky to maintain or to handle in our schema (or maybe not). I applied this if-logic here similarly like we do in many `_action_outputs` implementations. For example `_slider.py::_action_outputs`
```py
    @property
        return {
            "__default__": f"{self.id}.value",
            **({"title": f"{self.id}_title.children"} if self.title else {}),
        }
```

vizro-core/src/vizro/models/_components/form/slider.py
petar-qb: Thank you both!! Let me "stress-test" our sliders with different configurations and will post all inconveniences here (if I find any) 🔍 👀

petar-qb: Posting this for visibility after testing multiple Slider (find in control panel) and RangeSlider (find in container controls) configurations. I’ve identified a few layout inconsistencies:
1. Numeric inputs becomes misaligned vertically with the slider when marks are displayed (maybe this is intended and expected).
2. Numeric inputs can be partially cut off when min/max are floats but step is an integer. (See fourth RangeSlider in the screenshot below)
3. The gap between the slider and numeric inputs is sometimes excessively large, likely due to Dash’s internal width calculation: `width: calc(Xch + calc(var(--Dash-Spacing) * 2))` (`X` is somehow automatically calculated)

Here's the part of the screen from the scratch-dev example:
image

Personally, I wouldn't make any changes in Vizro as all three listed inconsistencies are minor and something only ~5% of users will encounter. However, I'll open a Dash issue explaining the `2.` and the `3.` inconsistencies as both are caused due to incorrect `X` calculation of this `width: calc(Xch + calc(var(--Dash-Spacing) * 2))` `innerHTML` prop.

vizro-core/src/vizro/static/js/models/dropdown.js
antonymilne, line 1: @huong-li-nguyen What is this vibe coded madness?! There's an argument to do this already: https://dash.plotly.com/dash-core-components/dropdown#custom-labels :p

huong-li-nguyen: Upsi, yes, feel free to kill that 😄

vizro-core/changelog.d/20251117_125834_90609403+huong-li-nguyen_update_css_styling.md
antonymilne: Looks good 👍

vizro-core/src/vizro/models/_components/form/dropdown.py
antonymilne: Hmmm I think it's worth keeping, the code just seems to be in a bit of a weird place (`__call__` rather than a validator). But I do see why we might want to do this in `__call__`. So all good I think 😀

vizro-core/src/vizro/models/_components/form/slider.py
antonymilne: Thanks for checking all this! I find the vertical misalignment very annoying but @huong-li-nguyen fixed in [a40e79f](https://github.com/mckinsey/vizro/pull/1462/commits/a40e79f238643fbf781a41b51bd6ffc3d2ab8547)😀 🙏

The other two issues I agree are minor and ok by me to just raise on the Dash side and wait for a fix there.

petar-qb: Awesome work @huong-li-nguyen 🚀 

I posted a few issues regarding new Dash slider [here](https://github.com/plotly/dash/issues/3611#issuecomment-3915304661) as a comment in the issue Li opened. Now, they have a single ticket that describes many sliders inconsistencies in one place 😃.

petar-qb:
> Just one small thing: do we still need the workaround in set_default_marks? Not sure whether that's still a Dash bug with this new version.

Great catch! 🚀 

From my testing, the only observable difference when removing `AfterValidator(set_default_marks)` is that `marks` are applied by default (ironically 😄).

Current version:
image

By removing `AfterValidator(set_default_marks)`:

image

I actually find the default marks behaviour preferable, particularly for `vm.Slider` (see in control panel) (imagine `value=min` or `value= max`). Without marks, both min and max are not visible, which makes the slider less intuitive. For `RangeSlider`, this is less of an issue when `value=[min, max]` is shown via numeric inputs, but for `Slider` there’s no clear indication of both bounds without marks.

Given that marks can still be explicitly disabled via `marks=None`, keeping them enabled by default seems like a reasonable (follows Dash implementation) and more user-friendly default behaviour.

vizro-core/tests/tests_utils/e2e/vizro/checkers.py
```diff
@@ -152,18 +153,29 @@ def check_selected_categorical_component(
...
+    if driver.find_elements(f"{dropdown_id_path(dropdown_id)}[aria-expanded='false']"):
+        driver.multiple_click(dropdown_id_path(dropdown_id), 1)
+    if multi is False:
+        selected_options = driver.find_elements(f"{dropdown_id_path(dropdown_id)} .dash-dropdown-value-item span")
+    if multi is True:
+        selected_options = driver.find_elements(
+            f"{dropdown_id_path(dropdown_id)} + div "
+            f"label[class='dash-options-list-option selected dash-dropdown-option'] "
+            f".dash-options-list-option-text span"
+        )
```
petar-qb: Here's a minor suggestion:

```suggestion
    if multi:
        selected_options = driver.find_elements(
            f"{dropdown_id_path(dropdown_id)} + div "
            "label[class='dash-options-list-option selected dash-dropdown-option'] "
            ".dash-options-list-option-text span"
        )
    else:
        selected_options = driver.find_elements(f"{dropdown_id_path(dropdown_id)} .dash-dropdown-value-item span")
```

vizro-core/tests/tests_utils/e2e/vizro/checkers.py
```diff
@@ -152,18 +153,29 @@ def check_selected_categorical_component(
...
     selected_options_list = ["".join(option.text.split()) for option in selected_options]
```
petar-qb, line 185: What this line does?

vizro-core/tests/tests_utils/e2e/vizro/checkers.py
```diff
@@ -152,18 +153,29 @@ def check_selected_categorical_component(
...
     assert_that(selected_options_list, equal_to(expected_selected_options))
     if expected_unselected_options:
```
petar-qb, line 189: Very minor suggesting (ignore if irrelevant)
```suggestion

assert_that(selected_options_list, equal_to(expected_selected_options))

if expected_unselected_options:
```

vizro-core/tests/tests_utils/e2e/vizro/checkers.py
```diff
@@ -152,18 +153,29 @@ def check_selected_categorical_component(
 
 
 def check_selected_dropdown(
-    driver, dropdown_id, expected_selected_options, expected_unselected_options=False, all_value=False
+    driver, dropdown_id, expected_selected_options, expected_unselected_options=False, multi=True
```
petar-qb: Very minor: It's a kind of Python convenience to use `None` as default value for something that's not defined. 
```suggestion
    driver, dropdown_id, expected_selected_options, expected_unselected_options=None, multi=True
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_filters.py
```diff
@@ -67,24 +65,21 @@ def test_dropdown_options_value(dash_br):
     4. Check how options in dropdown looks like.
     """
     page_select(dash_br, page_path=cnst.FILTERS_PAGE_PATH, page_name=cnst.FILTERS_PAGE)
-    # delete last option 'virginica'
-    dash_br.clear_input(f"div[id='{cnst.DROPDOWN_FILTER_FILTERS_PAGE}']")
+    # uncheck 'virginica'
```
petar-qb:
```suggestion
    # unselect 'virginica'
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_filters.py
petar-qb:
```suggestion
    1. Unselect 'virginica' option.
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_filters.py
```diff
@@ -67,24 +65,21 @@ def test_dropdown_options_value(dash_br):
...
     check_selected_dropdown(
         dash_br,
         dropdown_id=cnst.DROPDOWN_FILTER_FILTERS_PAGE,
-        all_value=False,
         expected_selected_options=["setosa", "versicolor"],
-        expected_unselected_options=["SelectAll", "virginica"],
+        expected_unselected_options=["virginica"],
     )
-    # choose one option 'virginica'
+    # choose 'virginica' again
```
petar-qb: I’ve added a few simple suggestions in this file to help align the naming conventions. To keep things consistent and clear, I recommend using “select” and “unselect” instead of variations like delete, uncheck, select, or choose. Would be great if you clean other tests similarly (but don't do that right now if there's a lot of repetitions)
```suggestion
    # select 'virginica'
```

antonymilne:
> > Just one small thing: do we still need the workaround in set_default_marks? Not sure whether that's still a Dash bug with this new version.
> 
> Great catch! 🚀
> 
> From my testing, the only observable difference when removing `AfterValidator(set_default_marks)` is that `marks` are applied by default (ironically 😄).
> 
> Current version: image
> 
> By removing `AfterValidator(set_default_marks)`:
> 
> image
> I actually find the default marks behaviour preferable, particularly for `vm.Slider` (see in control panel) (imagine `value=min` or `value= max`). Without marks, both min and max are not visible, which makes the slider less intuitive. For `RangeSlider`, this is less of an issue when `value=[min, max]` is shown via numeric inputs, but for `Slider` there’s no clear indication of both bounds without marks.
> 
> Given that marks can still be explicitly disabled via `marks=None`, keeping them enabled by default seems like a reasonable (follows Dash implementation) and more user-friendly default behaviour.

Yes agreed, let's keep the Dash implementation then and remove our custom `set_default_marks` stuff.

petar-qb:
> Yes agreed, let's keep the Dash implementation then and remove our custom set_default_marks stuff.

[commit](https://github.com/mckinsey/vizro/pull/1462/commits/24376829924c946397321f794a6af4828a74f7dc) 

Util `set_default_marks` can be removed completely as:
1. This is no more an issue in the `dash==4.0.0` -> https://github.com/plotly/dash-core-components/issues/159#issuecomment-380581043, 
2. We don't want to remove marks by default (setting `marks=None`)  when neither `marks` nor `step` is defined.

P.S. @l0uden this change will affect some screenshot tests, but it shouldn't cause any DOM or HTTP test to fail.

vizro-core/tests/tests_utils/e2e/vizro/checkers.py
l0uden: gathering values from dropdown into a list to compare it later to the expected ones

l0uden: added a comment

vizro-core/tests/e2e/vizro/test_dom_elements/test_filters.py
l0uden: Hopefully changed everywhere)

huong-li-nguyen:
> 2. We don't want to remove marks by default (setting `marks=None`)  when neither `marks` nor `step` is defined.
> 
@antonymilne @petar-qb I don't have a strong preference whether we show marks or not, but the current implementation isn't quite right for me. If we show marks, we need to show the reference dots (so have steps configured), otherwise the numbers are just free-floating there 😄
> 
So implement either of these:

* No marks , no steps (then we need to bring back the tooltip that's visible when using the handle)
* Marks + steps (show only min / max)
* Marks + steps (show min/max and in between)

Right now, we show marks **without steps/dots** for min/max and everything in between.

QBDS designs for reference: 

A tooltip with the number will appear when moving the handle

Screenshot 2026-02-19 at 19 45 45

EDIT: Actually, after looking at it longer, I slightly prefer how we had it before 😄  For example, in the below example, I feel like the little numbers are just cluttering up the view. The default inside the QBDS is to show a mark + dot only for the min / max value, and nothing in between. I've looked at other frameworks too, and the most common implementation is to not display the marks.

Screenshot 2026-02-19 at 19 53 24


Default QBDS:

Screenshot 2026-02-19 at 20 03 45


Screenshot 2026-02-19 at 20 06 49

vizro-core/tests/tests_utils/e2e/vizro/paths.py
```diff
@@ -78,9 +86,20 @@ def table_ag_grid_checkbox_path_by_row(table_id, row_index):
...
+        f"g:nth-of-type({axis_value_number}) text[data-unformatted='{axis_value}'"
+    )
+
+
+def graph_x_axis_value_path(graph_id, axis_value_number, axis_value):
+    """Path or x axis values of the graph."""
+    return (
+        f"div[id='{graph_id}'] .xaxislayer-below "
+        f"g:nth-of-type({axis_value_number}) text[data-unformatted='{axis_value}'"
+    )
```
petar-qb: I would merge these two functions into a single one and refactor it in the following way (I didn't check whether it works):

```suggestion
def graph_axis_value_path(graph_id, axis, tick_index, value):
    axis = axis.lower()
    if axis not in {"x", "y"}:
        raise ValueError("axis must be either 'x' or 'y'")

    return (
        f"div[id='{graph_id}'] "
        f".{axis}axislayer-below "
        f"g:nth-of-type({tick_index}) "
        f"text[data-unformatted='{value}']"
    )
```

vizro-core/tests/tests_utils/e2e/vizro/navigation.py
```diff
@@ -64,18 +76,36 @@ def page_select_selenium(driver, page_path, page_name, timeout=cnst.SELENIUM_WAI
...
 
 
-def select_slider_handler(driver, elem_id, value, handler_class="rc-slider-handle"):
-    driver.multiple_click(slider_value_path(elem_id=elem_id, value=value), 1)
-    driver.multiple_click(slider_handler_path(elem_id=elem_id, handler_class=handler_class), 1)
+def select_slider_value(driver, elem_id, value):
+    slider_values = driver.find_elements(f"div[id='{elem_id}'] div div")
+    for slider_value in slider_values:
+        if slider_value.text == value:
+            driver.click_at_coord_fractions(slider_value, fx=0, fy=0)
```
petar-qb: @l0uden, I’ll push an update today implementing the changes outlined in [this comment](https://github.com/mckinsey/vizro/pull/1462#issuecomment-3929266680).  

After that, I’ll review the impact and reach you out with a summary of any additional refactoring needed to ensure this function works correctly again.

vizro-core/tests/tests_utils/e2e/vizro/paths.py
l0uden: thanks!

petar-qb:
@huong-li-nguyen thanks for the input! I agree that showing marks for min/max and everything in between without steps/dots is not ideal.

> So implement either of these:
> 1. No marks , no steps (then we need to bring back the tooltip that's visible when using the handle)
> 2. Marks + steps (show only min / max)
> 3. Marks + steps (show min/max and in between)

I went with **option 2** because:

- **Option 1** has UX issues (described here: https://github.com/mckinsey/vizro/pull/1462#issuecomment-3919156286).
- **Option 3** is how Dash behaves when marks are not defined. That’s how it worked before [my changes](https://github.com/mckinsey/vizro/pull/1462/commits/aa3857c29aad68d0e9797e854bba6e707e7c4daa). But since `step` and `marks` are often not aligned, it just adds visual noise.

In theory, we could improve this (Option 3) compared to Dash by auto-generating `marks` aligned with `step` when marks are not provided. But to do it properly, we would need to handle:
- float vs int `min/max`
- whether `step` fits exactly into the range
- round marks when step is float and min/max are int
- too many marks on screen  
Covering all edge cases would make the implementation quite complex (the complex implementation is probably the reason why aligning marks and step doesn't properly work in Dash).

### How `marks` are treated in Vizro now?
- If marks are set (e.g. `{1: "1", 2: "2"}`), those marks are shown.
- If marks are explicitly set to `None`, no marks are shown.
- If marks not set at all, default **min/max marks** are applied.

I tried skipping default min/max marks when `self.step` is defined, but then Dash adds implicit marks that often don’t align with `step`, which looks inconsistent (option 3).

---

### Implementation details

I didn’t use `AfterValidator(set_default_marks)` because dynamic sliders require recalculating marks every time the filter is recreated. Instead, marks are recalculated inside `vm.Slider.__call__` (same for `vm.RangeSlider`) based on the current `min/max`.

---

### Before
image

### After
image

huong-li-nguyen:
Hi @petar-qb,

it definitely looks a lot better! And is working nicer than our current default. However, there are still two points I am a bit confused by:

1) I've put the same configuration into pyCafe, because I wanted to see how it works at the moment, and noticed that some of the max. values displayed in the Sliders now differ 🤔 I would have expected all values being the same, but just the visual display differing a bit. See this pyCafe example: https://py.cafe/app/huong-li-nguyen/vizro-iris-sliders
One example:


Currently in this branch:
Screenshot 2026-02-23 at 10 00 10

Same config in main:
Screenshot 2026-02-23 at 10 00 21

Why is the maximum 8 in main and 5 in this current branch? Did we change the calculation of the min / max values? 🤔 I copy/pasted only the relevant page, so not sure if there are maybe parts I didn't copy/paste correctly. 

2) It definitely works better, but the case that throws me off is when someone configures steps and therefore the handle does stop at these points, but there are no marks / dots displayed. I don't think it's right, and it's a bit confusing. I think if someone configures steps (marks+dots should be displayed too no?). The way how we had it before in main (see screenshot example above actually). In some of the other sliders, you have some numbers in between but there is not dot (so reference point to stop), I think that's also not correct. Anytime we show any numbers in between, there should be a dot that users can go to.


I just went through some of these. Before you implement anything, maybe let's discuss with @antonymilne 🤔 . I also don't want to overcomplicate things, so if you think the above is the best we can do without complicating the logic too much, then maybe we have to accept the drawbacks in terms of UX?

 ```
       # OK
        vm.Filter(column="sepal_length", selector=vm.Slider(title="No Config")),
        # OK
        vm.Filter(column="sepal_length", selector=vm.Slider(min=0, max=10, title="Min/Max")),
        # OK
        vm.Filter(column="sepal_length", selector=vm.Slider(min=0.13, max=10.13, title="Min-flot/Max-float")),
        # Doesn't work right, because mark+dot are not shown, but handle stops
        vm.Filter(column="sepal_length", selector=vm.Slider(step=1, title="Step")),
        # Doesn't work right, because mark+dot are not shown, but handle stops
        vm.Filter(column="sepal_length", selector=vm.Slider(step=0.5, title="Step-float")),
```

**Comment by petar-qb** [comment]
> I've put the same configuration into pyCafe, because I wanted to see how it works at the moment, and noticed that some of the max. values displayed in the Sliders now differ 🤔 I would have expected all values being the same, but just the visual display differing a bit. See this pyCafe example: https://py.cafe/app/huong-li-nguyen/vizro-iris-sliders

Sliders `min` and `max` in the `scratch_dev` example depend on the data-frame parameter value (the first slider in the control panel). Could it be that the data-frame parameter value in one example is 10 (resulting with `max=5.4`) and in another 150 (resulting with `max=7.9`)?

>  I copy/pasted only the relevant page, so not sure if there are maybe parts I didn't copy/paste correctly.

I think that the default value here has changed: `data_manager["dynamic_df"] = lambda number_of_points=10: df.head(number_of_points)`  # 150 -> 10

**Comment by petar-qb** [comment]
> It definitely works better, but the case that throws me off is when someone configures steps and therefore the handle does stop at these points, but there are no marks / dots displayed. I don't think it's right, and it's a bit confusing. I think if someone configures steps (marks+dots should be displayed too no?). The way how we had it before in main (see screenshot example above actually).

In the previous comment I explained why I decided not to let Dash to internally handles the marks. The reason is that `step` and `marks` are often not aligned, and it just adds visual noise then. Now I tested it more thoroughly and found when this "often" situation happens. In the latest dash version, dash handles step/marks incorrectly when `step=integer` and `min=float`. I fixed that in the latest version (will push a few code comments after we confirm on a call that the current version is correct)

> In some of the other sliders, you have some numbers in between but there is not dot (so reference point to stop), I think that's also not correct. Anytime we show any numbers in between, there should be a dot that users can go to.

According to the dcc documentation, dots will show on marks only if step is greater than 1:
```
    - dots (boolean; optional):
        When the step value is greater than 1, you can set the dots to
        True if you want to render the slider with dots.
```
That's why our data-frame parameter is the only slider on the screen with dots visible (the only slider on the page that has the step greater than one: `step=10`).

As we can't force dots to show in every case (there's no dots when `step<=1`), should we set `dots=False` by default?

---

EDIT: Reverting changes introduced in https://github.com/mckinsey/vizro/pull/1462/commits/d0a8b8ece35d0ae8f107df92456aaee36b06d564. Now, only min/max boundary marks are visible by default.

vizro-core/tests/tests_utils/e2e/vizro/navigation.py
petar-qb: Thanks for addressing the tests once again. Let me review the latest changes and then mergethe PR 😃

PR #1477, opened 2025-11-04
[Tidy] Refactor color system and template architecture

huong-li-nguyen:
## Description
Closes https://github.com/McK-Internal/vizro-internal/issues/2297
Discussed with @antonymilne that we'll bring over the theme improvements we've already implemented internally.

**1. Color System (`_colors.py`)**
- Migrated from function-based dictionary to `SimpleNamespace` structure
- Added named colors organized by category (qualitative, sequential, special)
- Colors now accessible via dot notation: `colors.cyan_500`, `colors.orange_300`

**2. Palettes Module (`_palettes.py`)**
- Created dedicated module for color sequences
- Defined reusable palettes: `qualitative_10`, `sequential_*`, `diverging_*`
- Replaced hardcoded hex values with color references where applicable

**3. Template Generation (`generate_plotly_templates.py`)**
- Refactored to use `pio.templates.merge_templates()` pattern
- Separated common properties from theme-specific overrides

## Notice


vizro-core/changelog.d/20251104_141009_90609403+huong-li-nguyen_refactor_themes_colors.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
```
huong-li-nguyen, line 22: @antonymilne - shall we make the `themes` directory public? I don't see a reason why not actually. I think the current structure will stay the same. The only reason why we might want to wait is that we are going to redo all the colors very soon and eventually have dark/light specific palettes, so might be worth waiting until then. WDYT?

vizro-core/src/vizro/_themes/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""Themes that can be used in and outside Vizro framework.
+
+vizro._themes.colors: Vizro colors.
+vizro._themes.palettes: Vizro palettes, divided into light and dark themes.
```
huong-li-nguyen: Wondering whether we should just make this public

huong-li-nguyen:
I didn't rename the files yet, as it's easier to review then. But I would probably change the naming of these files to:

Suggested:
├── _colors.py
├── _palettes.py
├── _template_common.py            ← rename from _common_templates.py
├── _template_generator.py          ← rename from generate_plotly_templates.py

vizro-core/changelog.d/20251104_141009_90609403+huong-li-nguyen_refactor_themes_colors.md
maxschulz-COL: What would people import from it?

huong-li-nguyen: The colors and palettes. Our Plotly charts currently always default to the sequential blue palette for example. If people want to use a diverging one, they have to go into our source code at the moment instead of just importing the options, similarly to https://plotly.com/python/builtin-colorscales/

maxschulz-COL: Ok dokes, then I am all for making it public

maxschulz-COL: What would be the suggested public API?

huong-li-nguyen: Same as we have it for the internal one I assume :) `from vizro.themes import colors, palettes` and then just use `palettes.diverging_red_blue`, unless we care about having vizro as a prefix / suffix, but I don't think it's required.

maxschulz-COL (approved):

vizro-core/src/vizro/_themes/_colors.py
```diff
@@ -1,169 +1,84 @@
...
+    cyan_01="#00B4FF",
+    orange_02="#FF9222",
+    purple_03="#3949AB",
+    red_04="#FF5267",
+    teal_05="#08BDBA",
+    amber_06="#FDC935",
+    green_07="#689F38",
+    purple_08="#976FD1",
+    pink_09="#F781BF",
+    dark_green_10="#52733E",
```
antonymilne: The naming here is confusing I think. In other cases the number 100, 200, etc. are the darkness of the shade, but here it's just a sequential number.

I feel like we should relabel in one of two ways:
1. `cyan`, `orange` etc. without any numbers
2. `cyan_350` or whatever it would be if we were to do different darknesses of each shade

Option 2 seems like a lot of work so I'd just do option 1 unless you think there's good reason not to?

vizro-core/src/vizro/_themes/vizro_dark.json
```diff
@@ -33,60 +82,60 @@
...
-        [0.5, "#E6E8EA"],
-        [0.5555555555555556, "#afe7f9"],
-        [0.6111111111111112, "#5bd6fe"],
-        [0.6666666666666666, "#3bbef1"],
-        [0.7222222222222222, "#24a6e1"],
-        [0.7777777777777778, "#0d8ed1"],
-        [0.8333333333333334, "#0077bd"],
-        [0.8888888888888888, "#0061a4"],
-        [0.9444444444444444, "#004c8c"],
+        [0.0, "#7D000F"],
```
antonymilne, line 85: This doing more than just changing from lowercase to uppercase since actual hex values have changed. Is that intentional?

vizro-core/src/vizro/_themes/_palettes.py
```diff
@@ -0,0 +1,183 @@
+"""Color palettes by Vizro."""
+
+from types import SimpleNamespace
+
+from ._colors import colors
+
+qualitative_10 = [
```
antonymilne: I'd call this just `qualitative` unless you think in future we're likely to have different palettes for `qualitative_2`, `qualitative_3` etc. that aren't just the first 2/3 entries of this palette.

Even if we do have those, probably it's useful to have a "standard default" `qualitiative` palette anyway, so I'd still just call this `qualitative`.

vizro-core/src/vizro/_themes/_palettes.py
```diff
@@ -0,0 +1,183 @@
...
+    colors.cyan_300,
+    colors.cyan_400,
+    colors.cyan_500,
+    colors.cyan_600,
+    colors.cyan_700,
+    colors.cyan_800,
+    colors.cyan_900,
+]
+
+# Are the same for dark and light theme at the moment, but could be different in the future.
```
antonymilne: I really hope they're not different in future because this would cause big issues and probably not even be possible at all 😀 The fact that they are the same at the moment is actually vital to achieve consistent colors after switching themes (we spent aaages working this out maybe 18 months ago or so).

```suggestion
# These are necessarily the same for dark and light themes. 
# For plotly express plots, some colors are taken from the template's palettes and stored in
# fig.data rather than fig.layout. This means they cannot be changed consistently by post-fig 
# updates to fig.layout.template (e.g. in the clientside callback we currently use).
```

vizro-core/src/vizro/_themes/generate_plotly_templates.py
```diff
@@ -46,66 +46,69 @@ def extract_bs_variables_from_css(variables: list[str], css_content: str) -> tup
...
-        xaxis_linecolor=AXIS_COLOR,
-        xaxis_tickcolor=AXIS_COLOR,
-        xaxis_tickfont_color=FONT_COLOR_SECONDARY,
-        xaxis_title_font_color=FONT_COLOR_PRIMARY,
-        yaxis_gridcolor=GRID_COLOR,
-        yaxis_linecolor=AXIS_COLOR,
-        yaxis_tickcolor=AXIS_COLOR,
-        yaxis_tickfont_color=FONT_COLOR_SECONDARY,
-        yaxis_title_font_color=FONT_COLOR_PRIMARY,
+    theme_overrides = go.layout.Template(
```
antonymilne:
```suggestion
    return go.layout.Template(
```

vizro-core/src/vizro/_themes/_static_template_generator.py
```diff
@@ -46,66 +46,69 @@ def extract_bs_variables_from_css(variables: list[str], css_content: str) -> tup
...
-    ]
+    return theme_overrides
+
+
+def generate_json_template(extracted_values: dict[str, str]) -> go.layout.Template:
+    """Generates the Plotly JSON chart template by merging common and theme-specific parts."""
+    common = create_template_common()
+    theme_overrides = create_theme_overrides(extracted_values)
+    # Merge common template with theme-specific overrides
+    template = pio.templates.merge_templates(common, theme_overrides)
```
antonymilne, line 110: Love this!

vizro-core/src/vizro/_themes/_common_template.py
```diff
@@ -2,98 +2,135 @@
...
-        yaxis_tickfont_size=14,
-        yaxis_ticklabelposition="outside",
-        yaxis_ticklen=8,
-        yaxis_ticks="outside",
-        yaxis_tickwidth=1,
-        yaxis_title_font_size=16,
-        yaxis_title_standoff=8,
-        yaxis_visible=True,
-        yaxis_zeroline=False,
+    common = go.layout.Template(
```
antonymilne:
```suggestion
    return go.layout.Template(
```

vizro-core/src/vizro/_themes/_templates.py
```diff
@@ -2,98 +2,135 @@
 
 from plotly import graph_objects as go
 
-from vizro._themes._colors import get_colors
+from vizro._themes import colors, palettes
 
 
 def create_template_common() -> go.layout.Template:
     """Creates template with common values for dark and light theme.
```
antonymilne, line 9:
```suggestion
    """Creates template with common values for dark and light theme.
    
    This is a function rather than just a variable to avoid running unnecessary code on import.
    It's only used to generate static vizro_dark/light.json files and not consumed directly at runtime.
```

vizro-core/src/vizro/_themes/_templates.py
antonymilne, line 1: Suggested name: `_templates.py`.

vizro-core/src/vizro/_themes/_static_template_generator.py
antonymilne, line 1: Suggested name: `_static_template_generator.py `.

vizro-core/changelog.d/20251104_141009_90609403+huong-li-nguyen_refactor_themes_colors.md
antonymilne: Will post more in my overall comment but basically yes, I'm happy to make this public! 🙌 I've wanted a few times to use Vizro colors in an app and was annoyed needing find a hardcoded hex value.

antonymilne (approved):
Very beautiful work! I wasn't planning on actually making these changes but it does look much nicer now, so thank you for doing it 🙌 

Basically all good, but just some notes on making it public:
* the proposed API of `from vizro.themes import colors, palettes` sounds perfect to me
* proposed filenames good to me other than a couple of suggestions I made
* double check that there's no Python files without an initial `_` in the directory (the json ones are fine without underscore)
* minimal requirement for being public would be to have a changelog entry, but in an ideal world we'd also have some docs on https://vizro.readthedocs.io/en/stable/pages/user-guides/themes/ and https://vizro.readthedocs.io/en/stable/pages/API-reference/ (could maybe even include a nice graphic like I did on the Brix post?). If you can't be bothered to do it now then feel free to make a separate ticket for it to do another time or I can help write it.

vizro-core/src/vizro/_themes/vizro_dark.json
huong-li-nguyen: Yeeees, the new values are taken from the recent Figma files. We haven't been updating them for a while, hence why this has changed

vizro-core/src/vizro/_themes/_palettes.py
huong-li-nguyen: I do think that we will have qualitative_2, qualitative_3 that aren't just the first 2/3 entries of this palette in the future. Discussed this already with J., but I can still just rename it to qualitative 👍

vizro-core/src/vizro/_themes/_palettes.py
huong-li-nguyen: Noted 😆

vizro-core/src/vizro/_themes/_colors.py
huong-li-nguyen: I initially had just `cyan`, `orange` etc., but I thought that this was confusing, because I was then wondering how they are different to `cyan_100`, `cyan_200` etc. At the moment, they are completely different values 😅 The index suffix is actually coming from the QB design library, they are called `cyan_01` there. 

Let me rename then, but note that one of the purple values I would have to rename to blue or dark purple, otherwise it doesn't work as we have two purple values in the qualitative palette: purple_03 and purple_08

huong-li-nguyen:
> Very beautiful work! I wasn't planning on actually making these changes but it does look much nicer now, so thank you for doing it 🙌
> 
> Basically all good, but just some notes on making it public:
> 
> * minimal requirement for being public would be to have a changelog entry, but in an ideal world we'd also have some docs on https://vizro.readthedocs.io/en/stable/pages/user-guides/themes/ and https://vizro.readthedocs.io/en/stable/pages/API-reference/ (could maybe even include a nice graphic like I did on the Brix post?). If you can't be bothered to do it now then feel free to make a separate ticket for it to do another time or I can help write it.

I've thought about it and decided to keep it private for now since we’ll be redoing the palettes this sprint. I’ll continue working on the documentation in parallel, but I’d make it public once the new palettes and sequences are ready. For now, I'd merge the PR as is and keep it private since everything will be refreshed soon anyway. 🥳 

And I'll prep a separate PR in the meantime

vizro-core/src/vizro/_themes/_colors.py
antonymilne: Hmm yes, this is all quite confusing and I don't think there's a perfect solution here 😅 

Overall I think best to go for `cyan`, `orange` etc. even though they don't relate to the `cyan_100` etc. values. And rename one of the `purple` like you say.

PR #1483, opened 2025-11-11
[Bug] Fix alignment of button icon and text

huong-li-nguyen:
## Description
Closes https://github.com/McK-Internal/vizro-internal/issues/2262

- Have `icon` and `text` in separate sub-containers such that styling doesn't conflict

## Notice

### DIFF
--- a/vizro-core/src/vizro/static/css/tooltip.css
+++ b/vizro-core/src/vizro/static/css/tooltip.css
@@ -35,6 +35,10 @@
   font-size: 1rem;
 }
 
+.form-label .tooltip-icon:hover {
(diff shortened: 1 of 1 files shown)


huong-li-nguyen:
> This looks good. 👍 The only thing that is bugging me is the misalignment between icons in buttons with and without text, but that is not introduced in this PR.

Yes, I checked that before, but it has to be set on the flex-box on a case-by-case basis. It doesn't look center aligned because of the different height, but better to do this via custom CSS. If you set align-self or align-items to center, it won't look good on all flex box cases :)

PR #813, opened 2024-10-15
[Feat] Drop Python 3.8 and add Python 3.13 support to Vizro

antonymilne:
## Description

1. Drop Python 3.8 and add Python 3.13 for Vizro.
2. Vizro-AI already had Python >= 3.9 so no need to drop anything there. It's not possible to add support for Python 3.13 yet due to https://github.com/huggingface/tokenizers/issues/1639 and possibly others. @lingyielia please could you make a ticket to check this again in a few weeks/months and try to add Python 3.13 support?
3. Add various new tools to `ruff`, especially `pyupgrade` which has helped to update our codebase to Python 3.9+. 
    * The biggest difference this made is that [PEP 585](https://stackoverflow.com/questions/65120501/typing-any-in-python-3-9-and-pep-585-type-hinting-generics-in-standard-collect) means you no longer need to do the ugly `from typing import List, Dict, ...` - you can just use the builtin `list` etc. 🎉 
    * Unfortunately we can't yet implement PEP 604 even with `from __future__ import annotations` because of pydantic. So we still need to do the ugly `Union` and `Optional`. When we drop Python 3.9 in a year's time this will go
4. Bump lower bound of pydantic to 1.10.16 and changed lower bounds environment to use Python 3.9
5. Bump our "standard" version for CI from Python 3.11 to 3.12
6. There's no discrepancy for `hatch run schema` depending on Python version now so we can finally tidy up that job to not depend on a specific version.
7. Speed up the vizro-core checks CI by not running pre-commit hooks after `hatch run schema-check` any more

TODO:
- [ ] Update Github required jobs to add Python 3.9 lower bounds and 3.13
- [x] Update vizro-qa

## Notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

  - I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
  - I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
  - I certify that the use of this contribution as authorized by the Apache 2.0 license does not violate the intellectual property rights of anyone else.
  - I have not referenced individuals, products or companies in any commits, directly or indirectly.
  - I have not added data or restricted code in any commits, directly or indirectly.

PR #1484, opened 2025-11-11
[Feat] Drop Python 3.9 and add Python 3.14 support to Vizro 

antonymilne:
## Description

Same strategy as done a year ago for dropping Python 3.8 and adding 3.13 in #813.

The biggest changes introduced by dropping Python 3.9 are:
* [PEP 604](https://peps.python.org/pep-0604/) means you no longer need to write `Union[X, Y]`, instead you write `X | Y`. Similarly, `Optional[X]` becomes `X | None`. This should be fixed automatically by ruff's pyupgrade rules 🎉 
* [structural pattern matching](https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching) is now possible! This is the only big new feature of Python 3.10 that's likely to be relevant to us. I haven't implemented it anywhere yet

Vizro: dropped Python 3.9 support, added tests for 3.14
Vizro-MCP: was already on Python>=3.10, added tests for 3.14
Vizro-AI: dropped Python 3.9 support, added tests for 3.13 and 3.14

Other small changes:
* introduced new [flake8-bandit linter](https://docs.astral.sh/ruff/rules/#flake8-bandit-s)
* changed how our Python badge works to be based on `requires-python` rather than manually entered
* removed old temporary pin `click<8.3.0`

TODO:
- [x] Complete changes as in #813
- [x] @antonymilne fix vizro-core tests that fail on Python 3.14 due to kedro
- [x] @antonymilne check docs
- [x] @antonymilne check changelogs
- [ ] @l0uden - check through all the github workflows to see I updated it all correctly and after merging:
  - [ ] update Github required jobs to include new 3.14 ones
  - [ ] remove old Python 3.9 Github jobs

## Notice

### DIFF
--- a/vizro-core/src/vizro/models/types.py
+++ b/vizro-core/src/vizro/models/types.py
@@ -725,7 +725,7 @@ OutputsType = Annotated[list[str] | dict[str, str], BeforeValidator(_coerce_to_l
 a list of strings, or a dictionary mapping strings to strings. Each output can be specified as
 `<model_id>` or `<model_id>.<argument_name>` or `<component_id>.<property>`. Defaults to `[]`."""
 
-# Extra type groups usd only for static type checking, not at runtime.
+# Extra type groups used only for static type checking, not at runtime.
 FigureWithFilterInteractionType: TypeAlias = "Graph | Table | AgGrid"
 FigureType: TypeAlias = "Graph | Table | AgGrid | Figure"
 


.github/workflows/test-integration-vizro-core.yml
```diff
@@ -43,21 +40,24 @@ jobs:
           - python-version: "3.13"
             hatch-env: all.py3.13
             os: ubuntu-latest
-          - python-version: "3.9"
+          - python-version: "3.14"
```
l0uden, line 43: Looks like we need to skip kedro tests here and for unit tests as it is not compatible with python 3.14

antonymilne: Yeah, I'll figure this out when I do the kedro ticket later this week.

vizro-core/src/vizro/models/types.py
```diff
@@ -704,37 +699,35 @@ class _OptionsDictType(TypedDict):
...
-OutputsType = Annotated[Union[list[str], dict[str, str]], BeforeValidator(_coerce_to_list), Field(default=[])]
+OutputsType = Annotated[list[str] | dict[str, str], BeforeValidator(_coerce_to_list), Field(default=[])]
 """List or dictionary of outputs modified by the action function. Accepts either a single string,
 a list of strings, or a dictionary mapping strings to strings. Each output can be specified as
 `<model_id>` or `<model_id>.<argument_name>` or `<component_id>.<property>`. Defaults to `[]`."""
 
-# Extra type groups used for mypy casting
-FigureWithFilterInteractionType = Union["Graph", "Table", "AgGrid"]
-FigureType = Union["Graph", "Table", "AgGrid", "Figure"]
+# Extra type groups usd only for static type checking, not at runtime.
```
```suggestion
# Extra type groups used only for static type checking, not at runtime.
```

PR #1491, opened 2025-11-13
[Docs] API docs cross-links and tidy

antonymilne:
## Description

Closes [#2181](https://github.com/McK-Internal/vizro-internal/issues/2181) and this phase of tidying API docs:
* all API docs are now cross-linked apart from the Kedro ones that I'll handle in a separate PR
* various things that should never have appeared in the `vizro.models.types` API docs have been made private
* `Dashboard.custom_header` is no longer a `staticmethod` just to make it simpler Python for a user to implement (non-breaking change, `staticmethod` still works ok, no need to put it in changelog)
* remove `type` from all our models docstrings - these were never needed and just added clutter
* ensure we use actions as `import vizro.actions as va`

## Notice


vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -75,16 +75,12 @@ def _filter_isin(series: pd.Series, value: MultiValueType) -> pd.Series:
...
 class Filter(VizroBaseModel):
-    """Filter the data supplied to `targets` on the [`Page`][vizro.models.Page].
+    """Filter the data supplied to `targets`.
 
     Abstract: Usage documentation
         [How to use filters](../user-guides/filters.md)
 
-    Examples:
-        >>> print(repr(Filter(column="species")))
-
```
huong-li-nguyen, line 85: You removed the examples here but kept it in the Parameter - any reason for that? Shall we just add:

```
Example:
        ```python

        import vizro.models as vm
        vm.Filter(column="species")
        ```
```

antonymilne: Yes good idea, thanks!

PR #1494, opened 2025-11-18
[Tidy] Add selected defaults for mantine theme

huong-li-nguyen:
## Description
Closes https://github.com/McK-Internal/vizro-internal/issues/1951
Closes https://github.com/McK-Internal/vizro-internal/issues/2153

This PR only adds a selected set of properties/CSS variables. We can always extend this later if it makes sense. Fully mapping all Mantine variables to our Bootstrap variables would take significantly more time, and there isn't a clean 1:1 mapping anyway.

Since our main components are Bootstrap-based, I'd recommend adding mappings between Bootstrap and Dash Mantine variables only where it's clearly useful (e.g., body-color, text-color, card-background are obvious candidates). There are two places where these mappings need to be added, as shown in this PR.

## Notice


vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,16 +1,109 @@
...
+            card_children.append(
+                dmc.Text(
+                    self.description,
+                    size="sm",
+                )
+            )
+
+        # Add button
+        card_children.append(
+            dmc.Button(
```
lingyielia, line 75: where does this button pick bg color from? is it looking inconsistent with the dark theme?

huong-li-nguyen: Yes, that's the default theme for the button using the grey `primaryColor` we defined. We *could* update the button as well (using the same logic), but since our main components are `dbc`, our assumption has always been that `dmc` components should be styled by the user.

We can add more logic on our side with the current system, but doing this for every `dmc` component would take a lot of time and effort - we already went through this whole exercise for `dbc`, and you can’t just change a single background color; you also need to handle hover, focus, and all the other states. That’s why I’m not too keen on extending this to `dmc` components right now :)

huong-li-nguyen:
> Thanks for taking this ticket! So next we just keep accumulating this mapping right?

Yes, exactly :)

PR #1495, opened 2025-11-18
[Bug] Fix flex-row graph rendering

nadijagraca:
## Description
closes https://github.com/McK-Internal/vizro-internal/issues/2353

This PR sets min-width of dash-graph to 450px if within flex-item.

```
.flex-item .dash-graph {
  min-width: 450px;
}
```

The css snippet above fixes the issue where the graph width is affected by the length of vm.Graph.title (header & footer as well - depending that is longest), but there are limitations with this approach:

- By default the graph will always have a width of 450px in a flex-row layout, even when more space is available.
- Users can set width inside of px chart to be wider than 450px, but not narrower. (width=300 will not be applied) unless user defines height as well. This is due to how plotly controls sizing. Specifying both width and height forces plotly to fully control sizing, so CSS min-width is effectively ignored.

In short, with this fix graphs within flex-row layout will have `450px` width by default. If users want to increase the width they can provide `width` argument within `px` chart. If they want to decrease the width, they need to provide both `height` and `width` arguments. 

## Notice

### DIFF
--- a/vizro-core/examples/scratch_dev/app.py
+++ b/vizro-core/examples/scratch_dev/app.py
@@ -1,55 +1,97 @@
+from typing import Literal
+
+from dash import html
+
+import vizro.models as vm
 import vizro.plotly.express as px
 from vizro import Vizro
-import vizro.models as vm
-from vizro.tables import dash_ag_grid
+from vizro.models.types import ControlType
+
+df_gapminder = px.data.gapminder()
+
+
+class ControlGroup(vm.VizroBaseModel):
+    """Container to group controls."""
 
     type: Literal["control_group"] = "control_group"
     title: str
     controls: list[ControlType] = []
 
+    def build(self):
+        return html.Div(
+            [html.H4(self.title), html.Hr()] + [control.build() for control in self.controls],
+            className="control_group_container",
+        )
+
+
+vm.Page.add_type("controls", ControlGroup)
+vm.Page.add_type("components", ControlGroup)
+vm.Page.add_type("components", vm.Filter)
+
+vm.Container.add_type("controls", ControlGroup)
+vm.Container.add_type("components", ControlGroup)
+vm.Container.add_type("components", vm.Filter)
+
 page = vm.Page(
-    title="Vizro on PyCafe",
-    layout=vm.Flex(direction="row", wrap=False),
+    layout=vm.Flex(),
+    title="Page:",
     components=[
-        vm.Graph(
-            id="scatter_chart",
-            title="title one",
-            figure=px.scatter(df, x="sepal_length", y="petal_width", color="species", width=300, height=300),
+        # vm.Filter(id="page_components", column="country"),  # -> EXCEPTION
+        # ControlGroup(  # -> EXCEPTION
+        #     title="page_components_group",
+        #     controls=[vm.Filter(id="page_components_group", column="continent")],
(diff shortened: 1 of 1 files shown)


huong-li-nguyen:
> Apply min-width to dash-graph only

@nadijagraca - I’ve pushed this  [commit](https://github.com/mckinsey/vizro/pull/1495/commits/882b229046794e774c831df1ee3c2016e76234b4). For now, it’s probably safest to apply the change only to the `dash-graph` container, since the Cards and Tables seem to behave correctly. Could you double-check all of our examples again?

As you mentioned, the drawback is that if a user sets a width below the `min-width`, it won’t have any effect. For example, something like the config below would be ignored. I think that’s fine for now, especially since `width` is also ignored inside the `Grid`. But it would be great if you could add this to the PR description and briefly explain the behavior so we have it documented for later.

Also, in @l0uden’s screenshots, it’d be good to update the config to use a width > 450px then :) 

```
vm.Graph(
            id="hist_chart",
            title="title onetwothree",
            figure=px.histogram(df, x="sepal_width", color="species", width=300),
        ),
```

vizro-core/changelog.d/20251119_100434_90609403+huong-li-nguyen_flex_row_graph_width.md
```diff
@@ -0,0 +1,48 @@
...
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
```
petar-qb: Should we mention something here?

petar-qb:
@l0uden could you consider adding a e2e test for this case? In my opinion this bug/solution is quite brittle and would be awesome if we have a e2e test that prevents us to change the important css in a way to introduce a bug again.

l0uden:
> @l0uden could you consider adding a e2e test for this case? In my opinion this bug/solution is quite brittle and would be awesome if we have a e2e test that prevents us to change the important css in a way to introduce a bug again.

Sure, changed the config to test this fix

vizro-core/changelog.d/20251119_100434_90609403+huong-li-nguyen_flex_row_graph_width.md
```diff
@@ -0,0 +1,48 @@
...
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
```
antonymilne:
```suggestion
### Fixed

- Set minimum width for graphs inside a flex layout. ([#1495](https://github.com/mckinsey/vizro/pull/1495))
```

antonymilne (approved):

PR #1496, opened 2025-11-18
[Feat] Enable configuration of notifications

huong-li-nguyen:
## Description

Closes https://github.com/McK-Internal/vizro-internal/issues/2052

Implements new `show_notification` and `update_notification` actions using dmc.NotificationContainer to enable configurable alerts in Vizro dashboards. For reference, see: https://www.dash-mantine-components.com/components/notification

## How it works

The `dmc.NotificationContainer` runs a client-side JavaScript notification manager in the browser:

- Our Python action outputs a notification config dict to `notification-container.sendNotifications`
- DMC's JavaScript manager receives the config and handles everything client-side:
  - Renders notification UI with animations
  - Manages queue (limit: 5, first in first out) when multiple notifications are triggered
  - Runs independent `setTimeout()` timers for auto-close
  - Handles `action="update"` by finding and transitioning existing notifications by ID
  - Cleans up DOM and timers when notifications close

### New Actions (`_notifications.py`)

**`show_notification`** - Display a new notification with support for:
- Five semantic variants: `info`, `success`, `warning`, `error`, `progress`
- Customizable Google Material icons (with variant-based defaults)
- Markdown text support

**`update_notification`** - Update an existing notification:
- References original notification via `notification` parameter matching `show_notification` ID
- Useful for progress → completion flows

### Dashboard Integration

- Added `dmc.NotificationContainer` to Dashboard model as direct child of MantineProvider
- Configured with top-right positioning and a notification limit of 5 (default)

### Styling (`notifications.css`)

- Custom CSS classes for notification variants (`alert-info`, `alert-success`, `alert-warning`, `alert-error`)

### Basic notification
```python
import vizro.actions as va
import vizro.models as vm

vm.Button(
    text="Save",
    actions=va.show_notification(
        text="Operation completed!",      # Markdown notification message (required)
        variant="success",                # info, success, warning, error, progress (default: info)
        title="Success",                  # Title (default: capitalized variant name)
        icon="celebration",               # Google Material icon (default: variant-specific)
        auto_close=5000,                  # Auto-close in ms, or False to persist (default: 4000)
    ),
)
```

### Update existing notification (progress → success flow)
```python
import vizro.actions as va
import vizro.models as vm

vm.Button(
    text="Save",
    actions=[
        va.show_notification(
            id="save_notification",       # ID to reference for updates
            text="Saving data...",
            variant="progress",           # Shows loading spinner
        ),
        va.export_data(),
        va.update_notification(
            notification="save_notification",  # Must match show_notification ID
            text="Data saved successfully!",
            variant="success",
        ),
    ],
)
```

**TODO**

- [x] Align on arguments for `show_notification`
- [x] @antonymilne and @petar-qb to check whether we can just use `id` instead of ` notification_id`
- [x] Ignore notifications on page load
- [x] Write tests and docs https://github.com/mckinsey/vizro/pull/1498

## Screenshot

Run `hatch run example`

## Notice

### DIFF
--- a/vizro-core/hatch.toml
+++ b/vizro-core/hatch.toml
@@ -172,7 +172,7 @@ extra-dependencies = [
   "dash-ag-grid==31.3.1",
   "dash-mantine-components==2.0.0",
   "plotly==5.24.0",
-  "pydantic==2.7.0",  # TODO: Update to 2.10.0 when pyCafe is made compatible
+  "pydantic==2.7.0",
   "pandas==2.0.0",
   "numpy==1.23.0",  # Need numpy<2 to work with pandas==2.0.0. See https://stackoverflow.com/questions/78634235/.
   "kedro==0.19.9"
--- a/vizro-core/schemas/0.1.50.dev0.json
+++ b/vizro-core/schemas/0.1.50.dev0.json
@@ -2425,7 +2425,7 @@
     },
     "show_notification": {
       "additionalProperties": false,
-      "description": "Shows a notification message.\n\nAbstract: Usage documentation\n    [Notifications](../user-guides/notification-actions.md)\n\nArgs:\n    text (str): Markdown text for the main notification message. Follows the CommonMark specification.\n    variant (Literal[\"info\", \"success\", \"warning\", \"error\", \"progress\"]): Variant that determines color and\n        default icon. If `progress`, the notification will show a loading spinner instead of an icon.\n        Defaults to \"info\".\n    title (str): Notification title. Defaults to the capitalized variant name, for example `\"Info\"` for\n        `variant=\"info\"`. Set to `\"\"` to hide the title.\n    icon (str): Icon name from the [Google Material Icon Library](https://fonts.google.com/icons). Defaults\n        to the variant-specific icon, for example 'info' for 'info' variant. Ignored if `variant=\"progress\"`.\n    auto_close (bool | int): Auto-close duration in milliseconds. Set to `False` to keep the notification\n        open until the user closes it manually. Default value depends on variant: `4000` for\n        info/success/warning/error, `False` for progress.\n\nExample:\n    ```python\n    import vizro.actions as va\n    import vizro.models as vm\n\n    vm.Button(\n        text=\"Save\",\n        actions=va.show_notification(\n            title=\"Useful information\",\n            text=\"This is some useful information that you should know.\",\n        ),\n    )\n    ```",
+      "description": "Shows a notification message.\n\nAbstract: Usage documentation\n    [Notifications](../user-guides/notification-actions.md)\n\nArgs:\n    text (str): Markdown text for the main notification message. Follows the CommonMark specification.\n    variant (Literal[\"info\", \"success\", \"warning\", \"error\", \"progress\"]): Variant that determines color and\n        default icon. If `progress`, the notification will show a loading spinner instead of an icon.\n        Defaults to \"info\".\n    title (str): Notification title. Set to `\"\"` to hide the title. Defaults to the capitalized variant name,\n        for example `\"Info\"` for `variant=\"info\"`.\n    icon (str): Icon name from the [Google Material Icon Library](https://fonts.google.com/icons). Ignored if\n        `variant=\"progress\"`. Defaults to the variant-specific icon, for example 'info' for 'info' variant.\n    auto_close (bool | int): Auto-close duration in milliseconds. Set to `False` to keep the notification\n        open until the user closes it manually. Default value depends on variant: `4000` for\n        info/success/warning/error, `False` for progress.\n\nExample:\n    ```python\n    import vizro.actions as va\n    import vizro.models as vm\n\n    vm.Button(\n        text=\"Save\",\n        actions=va.show_notification(\n            title=\"Useful information\",\n            text=\"This is some useful information that you should know.\",\n        ),\n    )\n    ```",
       "properties": {
         "id": {
           "description": "ID to identify model. Must be unique throughout the whole dashboard. When no ID is chosen, ID will be automatically generated.",
@@ -2452,13 +2452,13 @@
         },
         "title": {
           "default": "",
-          "description": "Notification title. Defaults to the capitalized variant name, for example\n            `\"Info\"` for `variant=\"info\"`. Set to `\"\"` to hide the title.",
+          "description": "Notification title. Set to `\"\"` to hide the title. Defaults to the capitalized\n            variant name, for example `\"Info\"` for `variant=\"info\"`",
           "title": "Title",
           "type": "string"
         },
         "icon": {
           "default": "",
-          "description": "Icon name from Google Material icons library. Defaults to variant-specific icon.\n            Ignored if `variant=\"progress\"`.",
+          "description": "Icon name from the [Google Material Icon Library](https://fonts.google.com/icons).\n            Ignored if `variant=\"progress\"`. Defaults to the variant-specific icon, for example\n            'info' for 'info' variant.",
           "title": "Icon",
           "type": "string"
         },
@@ -2482,7 +2482,7 @@
     },
     "update_notification": {
       "additionalProperties": false,
-      "description": "Updates an existing notification message.\n\nThis action updates notifications that were previously created with\n[`show_notification`][vizro.actions.show_notification]. `notification` must match the `id` of the original\n`show_notification` action.\n\nAbstract: Usage documentation\n    [Update notification](../user-guides/notification-actions.md#update-existing-notification)\n\nArgs:\n    notification (ModelID): Notification to update. Must match the id of the original `show_notification` action.\n    text (str): Markdown text for the main notification message. Follows the CommonMark specification.\n    variant (Literal[\"info\", \"success\", \"warning\", \"error\", \"progress\"]): Variant that determines color and\n        default icon. If `progress`, the notification will show a loading spinner instead of an icon.\n        Defaults to \"info\".\n    title (str): Notification title. Defaults to capitalized variant name if not provided, for example\n        'Info' for 'info' variant.\n    icon (str): Icon name from the [Google Material Icon Library](https://fonts.google.com/icons). Defaults\n        to the variant-specific icon, for example 'info' for 'info' variant. Ignored if `variant=\"progress\"`.\n    auto_close (bool | int): Auto-close duration in milliseconds. Set to `False` to keep the notification\n        open until the user closes it manually. Default value depends on variant: `4000` for\n        info/success/warning/error, `False` for progress.\n\nExample:\n    ```python\n    import vizro.actions as va\n    import vizro.models as vm\n\n    vm.Button(\n        text=\"Save\",\n        actions=[\n            va.show_notification(id=\"save_notification\", text=\"Saving data...\", variant=\"progress\"),\n            va.export_data(),\n            va.update_notification(\n                notification=\"save_notification\", text=\"Data saved successfully!\", variant=\"success\"\n            ),\n        ],\n    )\n```",
+      "description": "Updates an existing notification message.\n\nThis action updates notifications that were previously created with\n[`show_notification`][vizro.actions.show_notification]. `notification` must match the `id` of the original\n`show_notification` action.\n\nAbstract: Usage documentation\n    [Update notification](../user-guides/notification-actions.md#update-existing-notification)\n\nArgs:\n    notification (ModelID): Notification to update. Must match the id of the original `show_notification` action.\n    text (str): Markdown text for the main notification message. Follows the CommonMark specification.\n    variant (Literal[\"info\", \"success\", \"warning\", \"error\", \"progress\"]): Variant that determines color and\n        default icon. If `progress`, the notification will show a loading spinner instead of an icon.\n        Defaults to \"info\".\n    title (str): Notification title. Set to `\"\"` to hide the title. Defaults to the capitalized variant name,\n        for example `\"Info\"` for `variant=\"info\"`.\n    icon (str): Icon name from the [Google Material Icon Library](https://fonts.google.com/icons). Ignored if\n        `variant=\"progress\"`. Defaults to the variant-specific icon, for example 'info' for 'info' variant.\n    auto_close (bool | int): Auto-close duration in milliseconds. Set to `False` to keep the notification\n        open until the user closes it manually. Default value depends on variant: `4000` for\n        info/success/warning/error, `False` for progress.\n\nExample:\n    ```python\n    import vizro.actions as va\n    import vizro.models as vm\n\n    vm.Button(\n        text=\"Save\",\n        actions=[\n            va.show_notification(id=\"save_notification\", text=\"Saving data...\", variant=\"progress\"),\n            va.export_data(),\n            va.update_notification(\n                notification=\"save_notification\", text=\"Data saved successfully!\", variant=\"success\"\n            ),\n        ],\n    )\n```",
       "properties": {
         "id": {
           "description": "ID to identify model. Must be unique throughout the whole dashboard. When no ID is chosen, ID will be automatically generated.",
@@ -2509,13 +2509,13 @@
         },
         "title": {
           "default": "",
-          "description": "Notification title. Defaults to the capitalized variant name, for example\n            `\"Info\"` for `variant=\"info\"`. Set to `\"\"` to hide the title.",
+          "description": "Notification title. Set to `\"\"` to hide the title. Defaults to the capitalized\n            variant name, for example `\"Info\"` for `variant=\"info\"`",
           "title": "Title",
           "type": "string"
         },
         "icon": {
           "default": "",
-          "description": "Icon name from Google Material icons library. Defaults to variant-specific icon.\n            Ignored if `variant=\"progress\"`.",
+          "description": "Icon name from the [Google Material Icon Library](https://fonts.google.com/icons).\n            Ignored if `variant=\"progress\"`. Defaults to the variant-specific icon, for example\n            'info' for 'info' variant.",
           "title": "Icon",
           "type": "string"
         },
--- a/vizro-core/src/vizro/actions/_notifications.py
+++ b/vizro-core/src/vizro/actions/_notifications.py
@@ -38,10 +38,10 @@ class show_notification(_AbstractAction):
         variant (Literal["info", "success", "warning", "error", "progress"]): Variant that determines color and
             default icon. If `progress`, the notification will show a loading spinner instead of an icon.
             Defaults to "info".
-        title (str): Notification title. Defaults to the capitalized variant name, for example `"Info"` for
-            `variant="info"`. Set to `""` to hide the title.
-        icon (str): Icon name from the [Google Material Icon Library](https://fonts.google.com/icons). Defaults
-            to the variant-specific icon, for example 'info' for 'info' variant. Ignored if `variant="progress"`.
+        title (str): Notification title. Set to `""` to hide the title. Defaults to the capitalized variant name,
+            for example `"Info"` for `variant="info"`.
+        icon (str): Icon name from the [Google Material Icon Library](https://fonts.google.com/icons). Ignored if
+            `variant="progress"`. Defaults to the variant-specific icon, for example 'info' for 'info' variant.
         auto_close (bool | int): Auto-close duration in milliseconds. Set to `False` to keep the notification
             open until the user closes it manually. Default value depends on variant: `4000` for
             info/success/warning/error, `False` for progress.
@@ -62,7 +62,6 @@ class show_notification(_AbstractAction):
     """
 
     type: Literal["show_notification"] = "show_notification"
-
     text: str = Field(
         description="Markdown text for the main notification message. Follows the CommonMark specification.",
     )
@@ -77,13 +76,14 @@ class show_notification(_AbstractAction):
     # Placeholder defaults are used to satisfy type checking; actual defaults are set in the validator.
     title: str = Field(
         default="",
-        description="""Notification title. Defaults to the capitalized variant name, for example
-            `"Info"` for `variant="info"`. Set to `""` to hide the title.""",
+        description="""Notification title. Set to `""` to hide the title. Defaults to the capitalized
+            variant name, for example `"Info"` for `variant="info"`""",
     )
     icon: Annotated[str, AfterValidator(validate_icon)] = Field(
         default="",
-        description="""Icon name from Google Material icons library. Defaults to variant-specific icon.
-            Ignored if `variant="progress"`.""",
+        description="""Icon name from the [Google Material Icon Library](https://fonts.google.com/icons).
+            Ignored if `variant="progress"`. Defaults to the variant-specific icon, for example
+            'info' for 'info' variant.""",
     )
     auto_close: bool | int = Field(
         default=4000,
@@ -136,10 +136,10 @@ class update_notification(show_notification):
         variant (Literal["info", "success", "warning", "error", "progress"]): Variant that determines color and
             default icon. If `progress`, the notification will show a loading spinner instead of an icon.
             Defaults to "info".
-        title (str): Notification title. Defaults to capitalized variant name if not provided, for example
-            'Info' for 'info' variant.
-        icon (str): Icon name from the [Google Material Icon Library](https://fonts.google.com/icons). Defaults
-            to the variant-specific icon, for example 'info' for 'info' variant. Ignored if `variant="progress"`.
+        title (str): Notification title. Set to `""` to hide the title. Defaults to the capitalized variant name,
+            for example `"Info"` for `variant="info"`.
+        icon (str): Icon name from the [Google Material Icon Library](https://fonts.google.com/icons). Ignored if
+            `variant="progress"`. Defaults to the variant-specific icon, for example 'info' for 'info' variant.
         auto_close (bool | int): Auto-close duration in milliseconds. Set to `False` to keep the notification
             open until the user closes it manually. Default value depends on variant: `4000` for
             info/success/warning/error, `False` for progress.
@@ -163,7 +163,6 @@ class update_notification(show_notification):
     """
 
     type: Literal["update_notification"] = "update_notification"  # type: ignore[assignment]
-
     notification: ModelID = Field(
         description="Notification to update. Must match the id of the original `show_notification` action.",
     )


vizro-core/src/vizro/actions/_show_notification.py
```diff
@@ -0,0 +1,111 @@
...
+        description="Duration in milliseconds before auto-closing. Set to False to disable auto-close.",
+    )
+    action: Literal["show", "update"] = Field(
+        default="show",
+        description="Action to perform: 'show' adds new notification, 'update' updates existing one.",
+    )
+    loading: bool = Field(
+        default=False,
+        description="Show loading spinner instead of icon.",
+    )
```
huong-li-nguyen: title, message and variant are no-brainers for me. The rest we don't need to expose now if you think these are too many arguments, but all of them I do find useful.

vizro-core/src/vizro/actions/_show_notification.py
```diff
@@ -0,0 +1,156 @@
...
+        """Creates and returns a notification configuration for DMC NotificationContainer."""
+        # L: Is there a better to do this? Essentially, what I want is to cover two use cases:
+        # 1. Show a notification on page load (should be shown on page load without having to click on anything)
+        # 2. Show a notification when the action is triggered by a button or other
+        # interactive component (should not be shown on page load)
+        from vizro.models import Page
+
+        is_page_action = isinstance(self._parent_model, Page)
+        if not is_page_action and (_trigger is None or _trigger == 0):
+            return no_update
```
huong-li-nguyen: Is there a better way of doing this? This also doesn't always work 😬  
Do we need to expose the `on_page_load` action and attach it there? What's your recommended approach for displaying notifications on page load? How should this be configured. At the moment, I thought this should be triggered as long as the `show_notification` action is attached to `Page.actions`

petar-qb: There's another PR I'm preparing that will enable `Page.actions` to be properly customised. This PR is temporarily put on hold until I finish the `clickmode` improvement). Let's discuss it online

huong-li-nguyen: I found a quick workaround here: https://github.com/mckinsey/vizro/pull/1496/commits/3f15d4530c13fd82bc37e2f929669780dcdeb295

But I am not sure how compatible it is with your solution. Yes, let's discuss 👍

huong-li-nguyen: Not relevant anymore, as we won't enable on page notifications

vizro-core/src/vizro/actions/_show_notification.py
```diff
@@ -0,0 +1,124 @@
...
+from vizro.models._models_utils import _log_call, validate_icon
+from vizro.models.types import _IdOrIdProperty
+
+# Mapping of notification types to their corresponding colors and default icons (Google Material Icons)
+VARIANT_CONFIG = {
+    "info": {"className": "alert-info", "icon": "info"},
+    "success": {"className": "alert-success", "icon": "check_circle"},
+    "warning": {"className": "alert-warning", "icon": "warning"},
+    "error": {"className": "alert-error", "icon": "error"},
+    "progress": {"className": "alert-info", "icon": "info"},
```
petar-qb: Should we apply `auto_close=False` by default when the "progress" variant is used? 
(edit: It would probably be over engineered and confusing for users..)

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -0,0 +1,124 @@
...
+        description="""Notification identifier for updates. Multiple actions can share the same `notification_id`
+            to update a single notification.""",
+    )
+    # L: We do need this argument and can't make it depend on notification_id because both actions will have
+    # notification id provided. But one needs to have action 'update' and the other 'show'.
+    action: Literal["show", "update"] = Field(
+        default="show",
+        description="""Action type. Use `"show"` to display a new notification or `"update"` to modify an existing
+            notification with matching `notification_id`.""",
+    )
```
petar-qb: The perfect solution here would be to build a solution only based on the `self.id` and to remove both `notification_id` and `action` from the API. However, we can’t rely solely on `self.id` because `dmc` notifications must distinguish between the original notification (`action="show"`) and its replacement (`action="update"`). By only looking at multiple notifications with the same`self.id`, we can’t tell which is which.

Instead of keeping both `notification_id` and `action`, we could use a single argument, e.g. `update_notification_id` (please help me with the naming here 😄). Then we can derive the `"action"` dynamically:

`"action": "update" if self.update_notification_id else "show"`

P.S. we should still keep: `"id": self.update_notification_id or self.id`

It means that users don't have to provide `self.id` neither the `update_notification_id`. In that case it will be a standard `"show"` notification. In case users want to replace one notification with another, they have to do something like this:

```py
        vm.Button(
            actions=[
                va.show_notification(
                    id="my-notification",
                    message="Processing...",
                    variant="progress",
                )
            ],
        ),
        vm.Button(
            actions=[
                va.show_notification(
                    update_notification_id="my-notification",  # The arg name is terrible I know 😅 
                    message="Done!",
                )
            ],
        ),
```

huong-li-nguyen:
> Minor naming suggestion: “show_notification” feels a bit long. So, "alert" (with `"alert_id"` instead of `"notification_id"`) would be a cleaner.

I actually went back and forth on this quite a bit! There's a great article that helped clarify the difference: https://lunarlab.io/blog/whats-the-difference-between-alerts-and-notifications/

The key distinction is: **"Alerts require immediate action, usually during app use, while notifications don't need immediate action and can occur at any time"**

In the end, I went for `notification` (even though it's longer) for a few reasons:

- **How people will actually use it:** In Vizro, you're way more likely to show status updates like "Export completed!" than urgent "Drop everything and fix this!" messages. That's also why the default is `variant="info"` and `auto_close=4000s`, so our defaults are self-dismissable notification messages.

- **Staying consistent with DMC:** We're using Dash Mantine Components under the hood, which calls it `NotificationContainer`, so it felt weird to call our wrapper something different.

- **Alert is only a subset of notification**: https://www.smashingmagazine.com/2025/07/design-guidelines-better-notifications-ux/

Screenshot 2025-11-24 at 10 51 08


We definitely do cover both scenarios (notification and alerts), hence why I do understand that it's a bit ambiguous,  but I think the most common use case is more in the "notification" territory, rather than the "alert" one 🤔 But happy to have a vote here!

vizro-core/src/vizro/actions/_notifications.py
huong-li-nguyen: Hmm, I think keeping `notification_id` + `action` is actually clearer than having two different ID fields.

**The two IDs feel more confusing to me**, as I also mentioned during the call. If we have both `id` and `update_notification_id`, I feel like I'd constantly be going "wait which one was it again?" 😄 
With `notification_id` and `action`, you just... say what you mean:

```python
# This reads pretty clear to me
va.show_notification(notification_id="x", action="update", ...)

# This is less obvious? The field name has to basically encode both intents (action and id)
va.show_notification(update_notification_id="x", ...)
```

Also the fact that we can't figure out what to call `update_notification_id` doesn't feel like a good sign haha We're trying to put two ideas into one field name, which is why the naming gets weird. Also I think it makes sense to keep the dash mantine syntax instead of introducing a new one just to get rid of one argument 😬

huong-li-nguyen: The only thing that would be nice is if we don't have to use `notification_id`, but could just use the existing `id` argument 😄  But there I get the duplicate errors at the moment

vizro-core/src/vizro/actions/_show_notification.py
huong-li-nguyen: I actually like that! I do suggest that in the docs I wrote in the other PR, so why not just have it default 🤔 Let's see what @antonymilne thinks

vizro-core/src/vizro/actions/_notifications.py
petar-qb: how about `update_alert_id`? 😄 
you can handle duplication id by overwriting the base `id` from the VizroBaseModel. In case it already exists in the MM, rewrite it with the  `str(uuid.UUID(int=rd.getrandbits(128)))`.  
(edit: I'm not sure if this implementation would cause headaches later on for some reason 😄)

huong-li-nguyen:
Hi @l0uden 

> 1. Right now if we don't have full screen and have any number of notifications in the queue that don't fit the screen, the last ones can't be seen. There's no scroll to get to them.
>    Maybe we should limit number of the queue to 5 so maximum number of notifications could be seen on any screen size?

Yes, agree! I removed the `limit` config, so it will default to 5.

> 2. For the mobile devices in vertical mode notifications covers all of the page width, so dashboard can't be used until you close the notification messages.
>    @stephkaiser, what do you think, can we do smth here more user friendly?

I've added the same zoom on the `notification-container` as we have for the `right-side` now. So the notifications won't look disproportionately large anymore. When it comes to mobile optimization in general, I think we'll do that as a separate ticket. The zoom of 80% seems to be a bit big for me for this example app, but I think we have to properly test it with charts etc. So for now, just work with the current version :) Aligned that with @stephkaiser as well.

Screenshot 2025-11-24 at 11 27 54

vizro-core/src/vizro/actions/_notifications.py
huong-li-nguyen: I think I don't like the general concept of having two different `id` variables. So having to use `id` on one notification and then `XXX_id` on the other, just to get rid of `action`. I think I'd prefer to keep the `action` argument + `id` and be more explicit. But I'll leave that for you guys to decide, as @antonymilne wanted to take a proper look and might not be able to do that before my PTO. So feel free to merge whatever you think is best in the end 😄

vizro-core/src/vizro/actions/_show_notification.py
huong-li-nguyen: Pushed here: https://github.com/mckinsey/vizro/pull/1496/commits/e58c3ab6eb59de59f69a2f1b4c10c3aa688dd07b

antonymilne: I like it! 👍

vizro-core/src/vizro/actions/_notifications.py
antonymilne: After much consideration I decided to go for a whole new action `update_notification`. See my overall comment for more explanation.

vizro-core/changelog.d/20251118_135209_90609403+huong-li-nguyen_enable_alerts.md
```diff
@@ -0,0 +1,47 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+
+### Highlights ✨
+
+- Add `show_notification` action to display notification messages to users. Notifications can be customized with different variants, icons, and timing, and can be triggered by e.g. buttons, graphs, or cards. See the [notification actions user guide](../user-guides/notification-actions.md) for details. ([#1496](https://github.com/mckinsey/vizro/pull/1496))
```
antonymilne:
```suggestion
- Add `show_notification` and `update_notification` actions to display notification messages to users. Notifications can be customized with different variants, icons, and timing, and can be triggered by, for example, buttons, graphs, or cards. See the [notification actions user guide](../user-guides/notification-actions.md) for details. ([#1496](https://github.com/mckinsey/vizro/pull/1496))
```

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -73,6 +73,7 @@
   display: flex;
   gap: 0.5rem;
   height: 100%;
+  z-index: 1000;
```
petar-qb: Here's a UI bug I found. The theme selector goes over the notification. I guess we should remove the `z-index` here but also increase the `margin-top` of the notification container so it's possible to change the theme while the notification is on the screen.

Image

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -0,0 +1,173 @@
...
+        description="""Variant that determines color and default icon.
+        If `progress`, the notification will show a loading spinner instead of an icon.""",
+    )
+
+    title: str = Field(
+        default_factory=lambda data: VARIANT_DEFAULTS[data["variant"]].title,  # type: ignore[arg-type, misc]
+        description="""Notification title. Defaults to capitalized variant name if not provided,
+        for example 'Info' for 'info' variant.""",
+        validate_default=True,
+    )
```
petar-qb, line 87: I like the current defaulting behaviour for "title". Notifications generally look cleaner with it, so keeping the default absolutely makes sense to me.

However, I think that not all Vizro users will want a "title". That said, I also think we should allow them to disable it via `title=""` or `title=None`. This already works (see the second page of the scratch example or the screen recording below), but the CSS needs a small tweak so that notifications without a "title" render completely correctly.

I’d rather not spend time guessing why a particular dashboard designer might want things styled one way or another. My general view is that if flexibility comes at essentially no cost (for coding/testing/documenting/maintaining), we should provide it and avoid forcing a single aesthetic on everyone. However, I can think of a few use cases for this case where hiding the "title" seems like a legitimate requirement:
- The dashboard designer doesn't have a meaningful title to use, and the default "Info" feels inappropriate.
- A user triggers multiple parallel actions, producing several notifications. Removing the repeated title makes the stack of notifications more compact and readable.
- The designer wants to use the `Markdown` by utilising the `text` field to style a custom title look.

Here's an example that shows how it looks now when `title=""`.
https://github.com/user-attachments/assets/018faae1-380b-4773-816c-89706f8a2c33

vizro-core/src/vizro/actions/_notifications.py
petar-qb: I spent some time playing with the notifications, and I have to admit I’m still not convinced that introducing two separate actions for what is essentially the same operation.

The proposal is very close to what I originally suggested: a single additional argument like `notification`/`replace_notification` to control the behaviour. In this version, though, the logic gets split across separate models, 
hich feels harder for maintainers to digest before fixing/improving and adds unnecessary overhead in terms of code, tests, documentation, and long-term maintenance. For example, if we decide to reuse the `show_notification` for solving the dynamic notification system (creating notification based on the previous action), we will need to think or validate how to handle if `update_notification` is chained instead.

I still think we can simplify the implementation by moving the logic into a single argument on the existing `show_notification` action. I say "simplify" because this is essentially just a couple of lines of code:

```py
replace_notification: Optional[ModelID] = ...
...
def pre_build():
   ...
...
"id": self.replace_notification or self.id
"action": "update" if self.replace_notification else "show"
...
```

If we go down that path, I would also suggest renaming the action to `set_notification` to align with `set_control` and better reflect its role, since it would now cover both creating and updating notifications. 🤔 

Curious to hear what others think.

vizro-core/changelog.d/20251118_135209_90609403+huong-li-nguyen_enable_alerts.md
```diff
@@ -0,0 +1,47 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+
+### Highlights ✨
+
+- Add `show_notification` and `update_notification` actions to display notification messages to users. Notifications can be customized with different variants, icons, and timing, and can be triggered by, for example, buttons, graphs, or cards. See the [notification actions user guide](../user-guides/notification-actions.md) for details. ([#1496](https://github.com/mckinsey/vizro/pull/1496))
```
maxschulz-COL, line 10:
```suggestion
- Add `show_notification` and `update_notification` actions to display notification messages to users. See the [notification actions user guide](../user-guides/notification-actions.md) for details. ([#1496](https://github.com/mckinsey/vizro/pull/1496))
```

maxschulz-COL: To make it a bit more concise in the changelog?

vizro-core/examples/scratch_dev/app.py
maxschulz-COL, line 1: I get this on PyCafe, did not run locally yet.

vizro-core/schemas/0.1.50.dev0.json
maxschulz-COL, line 1: This makes me so happy ❤️

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -0,0 +1,173 @@
...
+        default="info",
+        description="""Variant that determines color and default icon.
+        If `progress`, the notification will show a loading spinner instead of an icon.""",
+    )
+
+    title: str = Field(
+        default_factory=lambda data: VARIANT_DEFAULTS[data["variant"]].title,  # type: ignore[arg-type, misc]
+        description="""Notification title. Defaults to capitalized variant name if not provided,
+        for example 'Info' for 'info' variant.""",
+        validate_default=True,
```
maxschulz-COL: Do we need to validate the default? Is that to do with the `default_factory`?

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -0,0 +1,173 @@
...
+                Ignored if `variant="progress"`""",
+            validate_default=True,
+        ),
+    ]
+    auto_close: bool | int = Field(
+        default_factory=lambda data: VARIANT_DEFAULTS[data["variant"]].auto_close,  # type: ignore[arg-type, misc]
+        description="""Auto-close duration in milliseconds. Set to `False` to keep the notification
+            open until the user closes it manually. Default value depends on variant: `4000` for
+            info/success/warning/error, `False` for progress.""",
+        validate_default=True,
```
maxschulz-COL: Similar here

vizro-core/src/vizro/static/css/notifications.css
maxschulz-COL, line 1: Are we happy with the layout here? I was immediately taken aback with the notification being overlayed by the theme change. 
image

vizro-core/src/vizro/static/css/layout.css
stephkaiser: just a quick comment that i dont think users need to be able to change the theme while the notification is visible - they can dismiss/close the notification and then toggle the theme if they wish :)

vizro-core/src/vizro/actions/_notifications.py
stephkaiser: i agree with your points. If we allow removing of the title, then the body text will need to be aligned/centred in the middle of the notification container - currently the icon is centred but not the text.

vizro-core/src/vizro/static/css/layout.css
petar-qb: Thanks for the input. In that case I think the theme-selector should be placed behind the notification.

vizro-core/src/vizro/actions/_notifications.py
maxschulz-COL: Ok I thought about this for a while as well now, and I think I agree with Antony. I say think, because I am not 100% sure myself - and possibly I overlooked something because I wasn't in the calls before.

I think introducing the concept of a secondary ID system is bad. It was hard enough to remove that from the pages and navigation (although in slightly different form), and I would prefer not to do that again.

If I understand correctly, that leaves us with A and P's options. Why do I prefer A's solution then:
- I think the overhead is actually minimal, in this case it is a very slim subclass
- _future_ subclassing is actually easier as we can base ourselves on some some sort of common denominator (show) which does **not** have a reference ID argument (such as `notification` or `update_notification_id`)

That indeed leaves the question of what is the best name for `show_notification` because this is the base that really _creates_ (or sets?) the notification. So I agree with P there, although I am not entirely sure if _set_ is such a good term... `create_notification`? I think I might like that one best though, it's explicit, it serves as the base, and it distinguishes it from other modification actions. Wdyt?

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -0,0 +1,173 @@
...
+        if type(notification) is not show_notification:
+            raise ValueError(
+                f"`notification={self.notification}` in `update_notification` action must refer to the ID of a "
+                f"`show_notification` action."
+            )
+
+    @_log_call
+    def function(self):
+        [notification] = super().function()
+        notification["id"] = self.notification
```
maxschulz-COL, line 184: Does a second update then have to refer to the _updated_ ID? Is that maybe confusing?

vizro-core/examples/scratch_dev/app.py
petar-qb: I can't see it
image

maxschulz-COL: image

vizro-core/src/vizro/static/css/layout.css
stephkaiser: yes the notification should be on top

vizro-core/src/vizro/actions/_notifications.py
petar-qb: I agree with your points, Max. let me edit-strikethrough parts of my earlier comment accordingly. Ultimately, the choice between option A and option P comes down to which is simpler for users to understand:
- introducing a new `update_notification` action, or
- adding a new `replace_notification` argument to the existing action.

Neither approach is terrible, so I’m okay with you decide team.

vizro-core/src/vizro/actions/_notifications.py
antonymilne: You're right, we don't need to! I've changed this in [30105f7](https://github.com/mckinsey/vizro/pull/1496/commits/30105f7cb57afc7ab98725faade89014f500e0e3).

vizro-core/src/vizro/actions/_notifications.py
antonymilne: No, a second update refers to the same original `id`. Think of it this way:
* `show_notification` creates the notification in the first place, and it has a single fixed `id`
* any subsequent `update_notification` actions refer to that original notification
* this would work the same way if we wanted to perform some other operation on the the original notification, e.g. delete it, hide it, make it dance round the screen, etc.

vizro-core/examples/scratch_dev/app.py
antonymilne: Ahh good catch, this is because `default_factory` can only take arguments from pydantic>=2.10.0 (which is more than a year old now). Currently our pycafe runtime only has `pydantic==2.7.0`.

Looking at the different versions supported on pyodide it looks like we'd need to go to the latest pyodide 0.27.2 in order to have pydantic>=2.10.0 but that [doesn't instantly work](https://py.cafe/snippet/vizro/v1?#c=H4sIAECzOmkEA81ae3PbNhL_Kjj1j5MyFvVwXDm60d0lbtJ2Jkkzda6djuVhIRIScaFIFgRtK5l89_stHhJlSraTU5MiD5PAcnexbyz8oRXlsWiNW3JZ5EqzK_le5UGR5jpdBeKmUKIsGS9ZcTPNtkCW-Cw1S1fLW0s80jLP7BqfZnOVL-0Kc2C_0Et9wWEL9KoQpYeKeKErJRyclkvhV8pUiGKaTbN4ziZgLYi55oFUsmx3aLrgCxEOsHS1DN7guT3NGIaWOhWTaeutKDWLlxHLci3nMuLELitXpRbLaevIAqd8lVd6AgwvUnHT7rjpKAcLmch0ObmwMzQA9azSOs8cJT9klGcgGCUiehdGUkWpWBPwQ4sbDZjzKopI1q9rPDVgnWTrpP244kGZ5NdhfU-3uPHDUfypEMpunTaVCi1iVlou5lWarv7WIO_HFVeSZ4TDwe-EhCrMTzcuayBemjTukd01V5nMFg0KbhO_2uUvKLY3qeClYEpcSXHNdCJL5nhkS4gC5hY0GPBjI7h926LxfwrO8flcqVzdLRcnYkGQjcWDCu1pxgwVlkdRpRQsLa4UiUwnguXeEB8iuN3M0jiM2H7M5vmDpCYB2Fg7qNB-EEr8vWRljshXlQJOyYioWppvGZ4YYtRDhLaTVxqHkdnLnMf3OqETW5JXapHysgzFstCrBthBJfhG5RSeiDdISsFp_6gQ_IPgIUIrVL6g9LcT9DCCe52zp5XOu2dpXjYzg5NYtHPxoHJ6S1GsDs6uZZqyUvMVqzItUxIfM4wwqdmSZxVHhtgvRZ9q3whVSuTVTO8F3Qh8r5XS4JBTaDiYvOBpKXaAHUYnZ1WpUW_8COE3mPEaEamY2ZDVADmoXs6ocGCoQ2yeyZBw5oJTVbQ_OXvRvwbwCwu8F3Yj-7sSOY0H7JzGYVTwiqt3cX5NVqgT9lJm7xrEHD_prrWDquAXWcLiKU1dmLKVxXlULWHQ5tvLdqJ1UY57PVvHKsFjwAKmDGTeE1kPPjRLRa9jwvUyV4LFQnOZlvdr8KVAkcBe4Zu9oF_MKwYBO4fgmAv1DYacPr5MdJcxSFUFqn7RjcVyf9BwzH9GGvA62Hy6F3TjRF8gYwwD9h-zcaZzdubq9gY9p40YJ5XG2t0qsFJ9iBLqIKD1Cer4raYDluCUOBMiYxAeSfrBhxCvob1C8GOjn7uC3GHUY3zkNb-SCyOXB1VEHBXtdYjwgDNB3AC6W10P9RjH3lkqkU8uEhSWl-0eHZO7-jrvkC0tcvr_0SM6L7Pho0eMZzFDAyClkGXilss8DwhcPyMKrgid-_5fe7_Y6OYLJH-aA9i6Q4CtN3oEJrR4waz58RszwnlbW4AGtcpTqAdoXshUC9WO8rRakmLLQkRSwNw6npu93YPvFS-S9lwuIOAJGhrI9hqxhC_b8fyI3RAyUfA0TEW20Akw-j3RwPc7bdLZ11ymIqR0miJ2rzn3w1nG8xvTWYEP8y3PbsAfxh6NmGEdoNk14A06fuzgDylVaUTkh8TvvQAby7s7au-Yg7yfGim051VmHiauV9WGN5sJaKid8uUs5mPbrWoPg5NOp72lNz8gNyuKkLa3bjXVB0A-OzJ_rpj_pK7QrTnvGjT82tpRY14msxxh0frpd_61TR4KGzRuPDhi3p0vnXubSq3dCWaVTOP2GksnUFVGLcLWUYvyj1SCKrkS7U9fxhWrIOJz0VskPQ4jm0OdvWX0TmalWPVMjddzHtADrrI37A-Oh8Mn345OH_cKWViQbj8YBCf9IBZX_W6xOobsM9Hl2Sq4TlLQpjYnaBpYvBYrneQZJopVHstYdK_6wXAUDLFku5Ct8YfWFZ2kCGoI3vNc_5xf07RDhQwC6CjBdpUA0MV6BeVnKTQWryUq09Z4cNI_ai1l9qt9PbZvPwi5SECHXmWMzyhoPAPWUqgzxDguM6H2UCDQ7szCAqTghLfV-nj58WgNs-HCETodBd-Onjwenpwej46H_ZPT3bjx3QYlL4qgWLW20W6W6xoN9I3e5sBKyJMfDIYBkR09Hnw7NIzspW_5vl9iTgL3ScsLaou5NZnE8zc6eQhNpBtM8_Q-oh6OqNKfj0dGazB8wGR8STA9VO1Cl72oxD9zBg7wCISVSrHsHUTx62CBY1k1Q0tKUf6DvAPEittuctwf9E_mo2h2MuLzaMRnT05Hw5M557E4fdIfPJ6f9PuiL46dy0SoE3rihlPMKXtlhENmlITwoL185Uou_ObBA23hEXOn97Pzc8Ye9aZZMEf3vivR2WdwbBV33xWSfaAwY0U9ZnCI4uYfNGM8ZMyO-3biI0WSYKazbqHkkquV_a4JZSzSSzFKEDdKstOvL7k6L01pTc2fszxNhYloLJ8zK2DmPrQQJAZ38VJwCskMf4t4M2kvi1CtFEmYz_4LbAZk8QlXPPTn35skanAhhyIDCM9UWCKjUX1FySmcozYSY7CBpKD5C3pDnYRkq9URW5mfnTGpizG7C2cYDolher0_glICtDMwjTKOqjBDZnKB93NH92ayoXxxcwky9YnVZYeyz5Yt0HKw4kuK-V_bFuq8NG3hG3aW5CivcRTT1ziNjdmwP3zc7Q_wlwp5-3bSHT6hpIwyBOHEwxxjwc9yqMDPDvq3xEHUQ5fHei6Yf33B7OaqKSJrLM8tCpIJ1bqolHyFAl5N_W7KLmZq_7lE8UT-wn57-uolo5C7tjnnG5S7UjnzDvEGr7TmXg3vd1213nlRSztzZIhNT4OeQ2Qc3EVuFkO-CBdKxh7o6eJ7vDkAQ8uv7Li85RnqL7X2ajI1mIKZ3Aa8OwRs3eYueAEbi-F3nQCdArVCWFihIcYmE1hYf2SjQ43SBcKG_ya0EJfAF88J9VZ0sXv30WW5Cl2AMfPtKIEfZKE9xZVjluIodoFwAoffuPvkNSq6nQGGalT8p_lbwrZWN8E5lGDq4sPUeMW0NWbyCAgganr8aLqEkkncyW6xcWkxgF9epQivE5uL7Ji2Sr3CGY-XITEb0sUkoXurqvpJ2YPRLmgZPMxyRdKa5Tg94up7DJhBcYMbpxSWgHq-3e3Oyq6FggvhLN-ZtohfmzztF4-RCKdw9CalBH0AmIChtV6k8TmUu1rxrEyrCO5IXOzBqPPCoiMN7QCrc348JM5rIH4TSOr0w-WEjcOYZGMUa9PDxiACNChiGWmYlxIINjEd-4-8yifu5xEaLF6HML9bdulccIdl8gUtHMQ0rWNv2aXn6LZVRdTIfe3MlNjr6gTldfePCtXB-26MLj1bTxoPn9aFif6jRYwS4zuBo-cOM0DUku9JnrR6y15pTFtlrpy_7gFAVEV2vmf5DYdooJIGCzRgO6Z9YtYRRpAEUmpdYwH8CRhL_ZDqh7uY-yl7asF30vcG5Qf4QT1KwIPGCkKXKfdpddSvLXsk21ZpFVlrAVizgKRxMIZvI_mkJqpg3sYVPOyILEcMJxKy6_uteWO8lqozYfIPf0qnnBOUOEGH1G1qUz4DkjWEqUCAy1yUhNR0aIsMv4EkswX6BpWed0-xSvbvvwDOrZM_ePDPjrycszCkaBqGlBymrRApQWZhiM1bPu9sB8RiVi0mpDvg21-vOE-sl9R_qcplB3_7ahgXDiyosQy4MKPSx64bodrs_FUL_hnSwMYm7yn2Wfefm-J9V_QDNrvl9S4JZH_Z_wzUP73k39LJ2tJ84f2XspgGd017-Yad4_I_hZRMawVaZjAtKn55jF_hgSjomazHPLtKDNlIzNGNYGdWv_EZflmAYjjyH-wJRqVw0jAl8VbRPM2-Mbed5toAqEwhjYBl8OdRyfD6-zoS_O5uKMAqbQenVdPVL43uu8AMizOtdmcMNGkb7OsJGkjsaiF0OHZWu11RB7CaOvRG-2N2q9qsQcE0SZsQAX7voTYPUy1QntRmEIRztROa3AY0jF-YSdPRHrNzeBLOH7BOHNk4XMqdSMgOLz_-D5Wu6NfPKQAA).

@maxschulz-COL please would you be able to play around with dependencies a bit to see if you can find something that works here? Regardless of this PR it's not great to be stuck on such an old version of pydantic anyway. If we can't then we should change the `default_factory` to something equivalent like this:

```
    @field_validator("title", "icon", "auto_close")
    @classmethod
    def set_variant_defaults(cls, value, info: ValidationInfo):
        # Remember to check for "is None" rather than Falsey values because auto_close=False is valid.
        if value is None:
            return getattr(VARIANT_DEFAULTS[info.data["variant"]], info.field_name)
        return value
```

vizro-core/src/vizro/actions/_notifications.py
antonymilne: Also agree - if this is easy to add then please can you make `title=None` or `title=""` (whichever, I don't mind) render nicely @huong-li-nguyen? 🙏

vizro-core/examples/scratch_dev/app.py
antonymilne: Also, should our CI check have failed the pycafe job here? 🤔 I can't remember what it's acceptance criteria are, but it would be great if in cases like this it detected the failure for us rather than relying on manual testing.

antonymilne:
To answer the questions about whether this should be two models and what it should be called (sorry, github makes it impossible to put the comments in the conversation itself 😬): let's stick to how it is now.

@petar-qb's comments are very valid. I did already consider several alternatives here including what he suggests, and I still think the current API is best. Since independent 3rd party @maxschulz-COL tends to prefer it too then I'm 100% happy with it.

* I am not worried about two classes adding extra maintenance burden/overhead. One of them is very small, and I don't think it makes any difference to the complexity of documentation whether this is two separate models or one.
* One class just doesn't feel natural to me here - if we aim for actions to be small, modular operations then "show" and "update" are distinct operations in my mind, even if the underlying implementation using DMC is the same.
* "Show" is the right terminology to use here. I did toy with different names before, but really for ephemeral UI elements like notifications, "show" is the right term rather than "create" or "set".

Overall what we should aim for here is something that feels natural to a human/AI user. And I really think this flow is the most intuitive one for Vizro: "show a notification, label it X" and then "update notification that has label X".

vizro-core/src/vizro/actions/_notifications.py
antonymilne: Ah, found this thread! Here's my response: https://github.com/mckinsey/vizro/pull/1496#issuecomment-3641755040.

Sorry @petar-qb, I removed your strikethroughs because I assumed they were a mistake and I couldn't read the comment with them there 🤦

petar-qb: Hahah legit move 🤣 
I'm resolving this conversation as we're all on the same page now 😃

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -0,0 +1,173 @@
...
+            actions=[
+                va.show_notification(id="save_notification", text="Saving data...", variant="progress"),
+                va.export_data(),
+                va.update_notification(
+                    notification="save_notification", text="Data saved successfully!", variant="success"
+                ),
+            ],
+        ```
+        )
+    """
```
antonymilne:
```suggestion
        )
    ```
    """
```

vizro-core/docs/pages/user-guides/notification-actions.md
```diff
@@ -0,0 +1,369 @@
...
+        from vizro import Vizro
+
+        page = vm.Page(
+            title="Update notification example",
+            layout=vm.Flex(direction="row"),
+            components=[
+                vm.Button(
+                    text="Start process",
+                    actions=va.show_notification(
+                        id="process_status",
```
petar-qb: Should we highlight this line and add a note here that explains why this ID is added here and how it's reused below? If so, then the similar should be done for other examples in the docs here.

vizro-core/tests/unit/vizro/actions/test_notifications.py
```diff
@@ -0,0 +1,139 @@
...
+            # Progress variant has auto_close=False by default
+            ("progress", "alert-info", "Progress", "info", True, False),
+        ],
+    )
+    def test_function_variants(
+        self, variant, expected_class, expected_title, expected_icon, expected_loading, expected_auto_close
+    ):
+        model_manager["button_one"].actions = [
+            show_notification(id="test_notification", text=f"{variant} message", variant=variant)
+        ]
```
petar-qb:
```suggestion
        model_manager["button_one"].actions = [
            show_notification(id="test_notification", text="test", variant=variant)
        ]
```

vizro-core/src/vizro/static/css/layout.css
huong-li-nguyen: Pushed here: https://github.com/mckinsey/vizro/pull/1496/commits/8452e3ff5f208382e53199b6303ed7e29f57f944 :) 

@l0uden - important for you to know that this is intended to work like this now :)

vizro-core/src/vizro/static/css/notifications.css
huong-li-nguyen: @petar-qb mentioned it in this thread as well. It's solved :) https://github.com/mckinsey/vizro/pull/1496#discussion_r2601589634

vizro-core/changelog.d/20251118_135209_90609403+huong-li-nguyen_enable_alerts.md
huong-li-nguyen: I would leave it in. It's not like our changelog files have lots of text anyway :)

vizro-core/examples/scratch_dev/app.py
huong-li-nguyen: @maxschulz-COL - did you have a chance to take a look at @antonymilne comment? Seems to be unrelated to this PR then, it does work locally! :) So I think, this shouldn't prevent the merge.

huong-li-nguyen: I've changed the lower pydantic version in the meantime here: https://github.com/mckinsey/vizro/pull/1496/commits/49873cb0e4e26822662ed7a60a4e5e669533d93c

So I think we just need to get pyCafe working with pydantic>=2.10.0, can you help here @maxschulz-COL ?
If that's not possible, I'd do the suggested change from @antonymilne above 👍

vizro-core/src/vizro/actions/_notifications.py
huong-li-nguyen: Fixed here https://github.com/mckinsey/vizro/pull/1496/commits/5d1ee45bd4c7f05bbc77604191424e35133076be :) 

So bascially:
- Without title: Everything is center aligned
- With title: Everything is aligned on top

Is that what you wanted @stephkaiser ? I could also make it center-aligned always (see second screenshot)

A (current):
Screenshot 2025-12-17 at 12 20 34

B (always center aligned):
Screenshot 2025-12-17 at 12 25 15

huong-li-nguyen: @stephkaiser prefers A

vizro-core/docs/pages/user-guides/notification-actions.md
huong-li-nguyen: Done here: https://github.com/mckinsey/vizro/pull/1496/commits/96e66f37f794d4cd2c0d7acebbaba2f008d8b062 :)

vizro-core/examples/scratch_dev/app.py
maxschulz-COL: I would use a version of the code that does not require `>=2.10.0` if you want to merge this soon. I have started a discussion with PyCafe to get ourselves upgraded to a new version of Pyodide, but this is not happening ASAP.

Here is the [thread](https://mckinsey-hub.slack.com/archives/C09L95GNXRS/p1765803801702579) in which we discuss this.

This is is the same type of issue as we have for[ `json_schema_input_type`](https://docs.pydantic.dev/latest/concepts/validators/#json-schema-and-field-validators), where we left a note that it should be changed once we upgrade.

So best course of action:
- use a version that works with the current lower bound of pydantic
- put a comment in the code just as we do for  `json_schema_input_type`
- refer to that comment, just as for the above issue, in the pydantic dependency (like that we won't forget)
- when Mario ressolves the Pyodide issue, we can upgrade it all together

huong-li-nguyen: Pushed here: https://github.com/mckinsey/vizro/pull/1496/commits/9e56232f3af2757f9af0af5fe5093bc6656f2575

vizro-ai/pyproject.toml
```diff
@@ -22,7 +22,7 @@ dependencies = [
   "langgraph>=0.1.2",
   "python-dotenv>=1.0.0",  # TODO decide env var management to see if we need this
   "vizro>=0.1.32",
-  "pydantic>=2.7.0",
+  "pydantic>=2.7.0",  # TODO: Update to 2.10.0 when pyCafe is made compatible
```
maxschulz-COL: Why are we updating this? Not relevant no?

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -72,26 +80,38 @@ class show_notification(_AbstractAction):
...
         Field(
-            default_factory=lambda data: VARIANT_DEFAULTS[data["variant"]].icon,
+            default=None,
             description="""Icon name from Google Material icons library. Defaults to variant-specific icon.
                 Ignored if `variant="progress"`""",
+            validate_default=True,
+        ),
+    ]
+    auto_close: Annotated[
+        bool | int | None,
```
maxschulz-COL: What actually happens if the user uses `True`?

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -72,26 +80,38 @@ class show_notification(_AbstractAction):
         If `progress`, the notification will show a loading spinner instead of an icon.""",
     )
 
-    title: str = Field(
-        default_factory=lambda data: VARIANT_DEFAULTS[data["variant"]].title,  # type: ignore[arg-type, misc]
-        description="""Notification title. Defaults to capitalized variant name if not provided,
-        for example 'Info' for 'info' variant.""",
-    )
+    title: Annotated[
+        str | None,
```
maxschulz-COL: Hm I am not a fan of this. This would be a breaking change if we go back to the default factory (see the above schema) which now explicitly allows `None`. 

But help me out, did we not think about this type of problem before (with things like `title=""` vs `title=None`) - it may be that I am forgetting something...

However as it stands I think this would be a problem because in the future someone can add `title=None` and it would fail while it does not fail now.

vizro-core/pyproject.toml
```diff
@@ -18,13 +18,13 @@ dependencies = [
...
   "dash-ag-grid>=31.3.1",  # 31.3.1 needed to support dash>=3.0.0
-  "dash_mantine_components>=1",  # 1.0.0 needed to support dash>=3.0.0
+  "dash_mantine_components>=2.0.0",  # 2.0.0 needed to support dmc.NotificationContainer
   "pandas>=2",
   "plotly>=5.24.0",
   # Must be <= the version in pre-commit mypy hook, sync manually
   # When updating this to >=2.9, search code for `but this requires pydantic >= 2.9` to find all
   # occurrences of code that is now ready to change
-  "pydantic>=2.7.0",
+  "pydantic>=2.7.0",  # TODO: Update to 2.10.0 when pyCafe is made compatible
```
maxschulz-COL:
```suggestion
  "pydantic>=2.7.0",  # When updating to `>=2.10`, search code for `once we bump to pydantic>=2.10.0` to find all occurrences ready for this change
```

vizro-core/src/vizro/actions/_notifications.py
maxschulz-COL: We should definitely:
- add a test that checks for `title=None` 
- change the schema such that future use of `default_factory` is no breaking change

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -77,7 +77,9 @@ def _build_inner_layout(layout, components):
     return components_container
 
 
-def validate_icon(icon) -> str:
+def validate_icon(icon) -> str | None:
```
maxschulz-COL: Would it not be better to change the order of the validators? Like that one can be sure that the input is not `None`. But in any case I think we should not allow `None` (see above)

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -0,0 +1,199 @@
...
+
+class update_notification(show_notification):
+    """Updates an existing notification message.
+
+    This action updates notifications that were previously created with
+    [`show_notification`][vizro.actions.show_notification]. `notification` must match the `id` of the original
+    `show_notification` action.
+
+    Abstract: Usage documentation
+        [Update notification](../user-guides/notification-actions.md#update-notification)
```
petar-qb:
```suggestion
        [Update notification](../user-guides/notification-actions.md#update-existing-notification)
```

vizro-core/src/vizro/actions/_notifications.py
huong-li-nguyen: It will automatically close with a default of 4s. So basically auto_close=True and auto_close=4000 and auto_close=None are the same: https://www.dash-mantine-components.com/components/notification#autoclose

vizro-core/src/vizro/models/_models_utils.py
huong-li-nguyen: I think we need to allow for `None` though, otherwise it doesn't work with the variant-specific defaults. Or is there another way I might be missing?

- title: Defaults to the capitalized variant name (e.g., "Info", "Success", "Warning"). Using `None `triggers this default; using "" explicitly disables the title. But both need to be enabled "" and None.
- auto_close: Defaults to 4000 for most variants, but False for progress (since progress notifications should stay visible until explicitly updated). Using None allows the validator to apply the correct variant-specific default.
- icon: Defaults to a variant-specific icon (e.g., "info" for info, "check_circle" for success). While we could use "" as the default, the validator would then need to check for both None and "" to determine whether to apply the variant default. Not sure if that's better 🤔

vizro-ai/pyproject.toml
huong-li-nguyen: Upsi, yes, I just did a search and find 😄

vizro-core/src/vizro/models/_models_utils.py
huong-li-nguyen: Will comment on the other thread, but agree 👍  :)

vizro-core/src/vizro/actions/_notifications.py
huong-li-nguyen: @maxschulz-COL @petar-qb 

After our discussion, I changed it to use the `model_validator(mode="after")` with `model_fields_set` (similar to how we handle `path` in the `Page` model): https://github.com/mckinsey/vizro/pull/1496/commits/fc1df3e666135c903604411776b21cd39f7fe9b6

So now this:
- Raises a Pydantic ValidationError if `None` is provided by the user
- Still sets variant-specific defaults when the user doesn't provide a value
- Keeps the type annotations clean (no `None` in the types)

The only downside is that we need placeholder defaults (e.g., `default=""`) for this to work, which are technically incorrect since they get immediately overwritten by the validator. I think that's fine for now, but we should switch to `default_factory` once we bump to `pydantic ≥2.10`.

Added unit tests for `None` as well 👍

petar-qb: This looks pretty good to me! ✅

vizro-core/hatch.toml
```diff
@@ -170,9 +170,9 @@ extra-dependencies = [
   "dash==3.1.1",
   "dash-bootstrap-components==2.0.0",
   "dash-ag-grid==31.3.1",
-  "dash-mantine-components==1.0.0",
+  "dash-mantine-components==2.0.0",
   "plotly==5.24.0",
-  "pydantic==2.7.0",
+  "pydantic==2.7.0",  # TODO: Update to 2.10.0 when pyCafe is made compatible
```
maxschulz-COL:
```suggestion
  "pydantic==2.7.0",
```

maxschulz-COL: We will get the reminder automatically

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -0,0 +1,187 @@
...
+    Abstract: Usage documentation
+        [Notifications](../user-guides/notification-actions.md)
+
+    Args:
+        text (str): Markdown text for the main notification message. Follows the CommonMark specification.
+        variant (Literal["info", "success", "warning", "error", "progress"]): Variant that determines color and
+            default icon. If `progress`, the notification will show a loading spinner instead of an icon.
+            Defaults to "info".
+        title (str): Notification title. Defaults to the capitalized variant name, for example `"Info"` for
+            `variant="info"`. Set to `""` to hide the title.
```
maxschulz-COL:
```suggestion
        title (str): Notification title. Set to `""` to hide the title. Defaults to the capitalized variant name, 
            for example `"Info"` for `variant="info"`. 
```

maxschulz-COL: Would keep it consistent with other docstrings - that is to say have the default description last.

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -0,0 +1,187 @@
...
+        icon (str): Icon name from the [Google Material Icon Library](https://fonts.google.com/icons). Defaults
+            to the variant-specific icon, for example 'info' for 'info' variant. Ignored if `variant="progress"`.
```
maxschulz-COL: Same here

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -0,0 +1,187 @@
...
+    Args:
+        notification (ModelID): Notification to update. Must match the id of the original `show_notification` action.
+        text (str): Markdown text for the main notification message. Follows the CommonMark specification.
+        variant (Literal["info", "success", "warning", "error", "progress"]): Variant that determines color and
+            default icon. If `progress`, the notification will show a loading spinner instead of an icon.
+            Defaults to "info".
+        title (str): Notification title. Defaults to capitalized variant name if not provided, for example
+            'Info' for 'info' variant.
+        icon (str): Icon name from the [Google Material Icon Library](https://fonts.google.com/icons). Defaults
+            to the variant-specific icon, for example 'info' for 'info' variant. Ignored if `variant="progress"`.
```
maxschulz-COL: Same here

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -0,0 +1,187 @@
...
+        [Update notification](../user-guides/notification-actions.md#update-existing-notification)
+
+    Args:
+        notification (ModelID): Notification to update. Must match the id of the original `show_notification` action.
+        text (str): Markdown text for the main notification message. Follows the CommonMark specification.
+        variant (Literal["info", "success", "warning", "error", "progress"]): Variant that determines color and
+            default icon. If `progress`, the notification will show a loading spinner instead of an icon.
+            Defaults to "info".
+        title (str): Notification title. Defaults to capitalized variant name if not provided, for example
+            'Info' for 'info' variant.
```
maxschulz-COL: Is there a reason we are not saying "Set to "" to hide title"here?

PR #1502, opened 2025-11-24
[Feat] Hide all Vizro internal callbacks

antonymilne:
## Description

FYI @noklam 

Dash 3.3.0 and 3.4.0 introduce a `hidden` argument to hide callbacks from the dev tools graph. I've marked all our Vizro callbacks as `hidden` so that only the "core" action callbacks remain.

For some debugging purposes this might be a bit too aggressive and hiding too much, but I think that's only relevant for Vizro developers rather than users. So then we can temporarily remove the `hidden=True` as and when needed during development.

There's a couple of issues with the new Dash features that I've raised on their repo:
[[BUG] hide_all_callbacks doesn't apply to clientside callbacks](https://github.com/plotly/dash/issues/3579) - this means it's not possible for us to use this argument, so we just manually mark all relevant callbacks as `hidden=True` (which is probably the best approach for us anyway)
[[BUG] Dash pages callback not fully hidden](https://github.com/plotly/dash/issues/3578) - just a small problem but would be nice to fix @l0uden 

## Screenshot

The callback graph is still massively overwhelming in a big app, so it's best just to comment out all pages but one if you actually want to use this graph.

### Example 1: one page on export data action in `hatch run example:dev`.

**Before this PR**:
image

**After this PR**
image

### Example 2: chained actions taken from [custom actions tutorial](https://vizro.readthedocs.io/en/stable/pages/tutorials/custom-actions-tutorial/)

**Before this PR**
image

**After this PR**
image

## Notice

### DIFF
--- a/vizro-core/pyproject.toml
+++ b/vizro-core/pyproject.toml
@@ -17,7 +17,7 @@ classifiers = [
 dependencies = [
   "dash>=3.4.0,<4",  # temporary upper bound to avoid breaking changes in Dash v4
   "dash_bootstrap_components>=2",  # 2.0.0 needed to support dash>=3.0.0
-  "dash-ag-grid>=31.3.1",  # 31.3.1 needed to support dash>=3.0.0
+  "dash-ag-grid>=31.3.1,<33",  # 31.3.1 needed to support dash>=3.0.0; temporary upper bound until we fix https://github.com/McK-Internal/vizro-internal/issues/2441
   "dash_mantine_components>=2.0.0",  # 2.0.0 needed to support dmc.NotificationContainer
   "pandas>=2",
   "plotly>=5.24.0",


petar-qb (approved):
Thanks @antonymilne for introducing this improvement and for creating new dash issues! 🙌 

I want to raise further questions/ideas around callback visibility.

Should we consider enabling visibility of all callbacks in Vizro when the dashboard creator explicitly opts in like `Vizro(hide_all_callbacks=False).build(dashboard).run()` (by default it's `True`)? It's maybe an overkill to propagate this flag to all clientside callback you hide like `hidden=hide_all_callbacks`

Ideally, the default behaviour could be to show only the callbacks relevant to the currently opened page. I realise this is non-trivial to implement on the Dash side as an option, but it would be a great UX improvement.

Related to that, it would be useful if certain `DashDependency` objects could be hidden from the callback graph entirely. For example: `Output(f"{id}_finished", "data", hidden=True)`. Other candidates would be "_guarded_trigger" and "action_progress_indication_placeholder".

These changes would allow us to exclude all our internal dependencies and keep the graph focused on what actually matters to the dashboard creator.

antonymilne:
Thanks for the quick review guys!

@petar-qb 

> Should we consider enabling visibility of all callbacks in Vizro when the dashboard creator explicitly opts in like `Vizro(hide_all_callbacks=False).build(dashboard).run()` (by default it's `True`)? It's maybe an overkill to propagate this flag to all clientside callback you hide like `hidden=hide_all_callbacks`

Yes, potentially we could do that in future. It might be useful to make it more fine-grained like a kind of logging level/verbosity:
* level = 0: show only explicitly defined actions (like now)
* level = 1: level 0 + show callbacks used in actions (like guard trigger)
* level = 2: show all callbacks

Or something like that anyway.

@noklam already asked out of curiosity whether there was a way to turn back on the vizro callbacks but didn't have a use for it yet. So let's wait until someone would actually find it useful to understand what callbacks they would like to see.

There's also a question about where this argument would go exactly. Dash's `hide_all_callbacks` is in `run` rather than `Vizro`. We could potentially also do it with an environment variable.

> Ideally, the default behaviour could be to show only the callbacks relevant to the currently opened page. I realise this is non-trivial to implement on the Dash side as an option, but it would be a great UX improvement.

Agreed, this would be a huge improvement. Maybe worth making a feature request on Dash? I have no idea off the top of my head how it would work or if it's even possible. But realistically in any slightly complex multi-page app at the moment the callback graph is pretty unusable.

> Related to that, it would be useful if certain `DashDependency` objects could be hidden from the callback graph entirely. For example: `Output(f"{id}_finished", "data", hidden=True)`. Other candidates would be "_guarded_trigger" and "action_progress_indication_placeholder".
>
> These changes would allow us to exclude all our internal dependencies and keep the graph focused on what actually matters to the dashboard creator.

Yes agreed, I wanted to do exactly this but couldn't. The graph would need to resolve some scenarios for chained callbacks like what happens if `Output(hidden=True)` is the `Input(hidden=False)` to another callback, but I don't think it would too hard to resolve to something sensible. Maybe also worth another feature request on Dash? 🤔 I don't think these would be high priority for them but are potentially something @l0uden could contribute with his new Dash expertise 😀

petar-qb:
Here are the feature requests:
- [[Feature Request] Dev Tools Callback Graph per Page](https://github.com/plotly/dash/issues/3586)
- [[Feature Request] Dev Tools Callback Graph - Allow hiding specific DashDependency objects](https://github.com/plotly/dash/issues/3587)

petar-qb:
Hey @antonymilne, would you mind adding the `black>=26.0.0` constraint? I missed it in https://github.com/mckinsey/vizro/pull/1545/changes. Thanks!

antonymilne:
> Hey @antonymilne, would you mind adding the `black>=26.0.0` constraint? I missed it in https://github.com/mckinsey/vizro/pull/1545/changes. Thanks!

Oops sorry, I missed this... But actually I don't think it's worth doing because it's not really a requirement for users, just for our tests to pass. If our tests break again then we can pin it in the test environment but if not then I think let's just leave it as is.

PR #1503, opened 2025-11-24
[Tidy] Enable sticky headers inside Grid and add docs for Flex

huong-li-nguyen:
## Description
closes https://github.com/McK-Internal/vizro-internal/issues/2223

This PR:
- enables sticky headers in Ag Grid when its used within a Grid layout. 
- Inside the Flex layout it is not enabled by default as we can't predict the number of tables, instead section is added to the docs explaining how the same functionality can be achieved with custom css. 

## Notice


vizro-core/docs/pages/user-guides/table.md
```diff
@@ -420,6 +420,25 @@ As mentioned above, all [parameters of the Dash AG Grid](https://dash.plotly.com
 
 If the available arguments are not sufficient, there is always the option to [create a custom AG Grid callable](custom-tables.md).
 
+#### Add sticky headers inside a Flex layout
+
+AG Grid provides sticky headers by default when used in a Grid layout. To achieve the same behaviour within a Flex layout, you must add the below CSS to your custom CSS file in your local `assets` folder.
```
stichbury:
```suggestion
AG Grid provides sticky headers by default when used in a Grid layout. To achieve the same behavior within a Flex layout, you must add this CSS to your custom CSS file within your local `assets` folder.
```

vizro-core/docs/pages/user-guides/table.md
```diff
@@ -420,6 +420,25 @@ As mentioned above, all [parameters of the Dash AG Grid](https://dash.plotly.com
...
+
+.ag-body {
+    top: 40px;
+}
+```
+
+!!! note
+
+    This approach works best when there is only one `AgGrid` on the screen, or when all `AgGrid` tables are in the same row. Otherwise, all headers will stack on top of one another.
+
```
petar-qb, line 454: It seems that this custom CSS workaround behaves correctly only when the ag-grid elements (or element) are positioned at the top of the page. If that's true, I would rephrase this sentence.

https://github.com/user-attachments/assets/7131e22c-3a03-4ee7-a4de-cfd6bc0f2a99

vizro-core/changelog.d/20251124_165505_90609403+huong-li-nguyen.md
```diff
@@ -0,0 +1,48 @@
...
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
```
petar-qb, line 30: Forgot to comment this section..
Should we mention the change here and link the docs where it's explained on how to turn on the sticky header for ag-grid in Flex?

petar-qb:
@nadijagraca could you add a small PR description here?
@l0uden do you think it's possible and worth to cover this use case with the e2e test?

vizro-core/changelog.d/20251124_165505_90609403+huong-li-nguyen.md
```diff
@@ -0,0 +1,47 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Enable Ag Grid sticky headers when used within Grid layout. See our [user guide](https://vizro.readthedocs.io/en/stable/pages/user-guides/table/#add-sticky-headers-inside-a-flex-layout) for an example how to enable it in the Flex layout. ([#1503](https://github.com/mckinsey/vizro/pull/1503))
```
maxschulz-COL:
```suggestion
- Enable AG Grid sticky headers when used within Grid layout. See our [user guide](https://vizro.readthedocs.io/en/stable/pages/user-guides/table/#add-sticky-headers-inside-a-flex-layout) for an example of how to enable it in the Flex layout. ([#1503](https://github.com/mckinsey/vizro/pull/1503))
```

vizro-core/docs/pages/user-guides/table.md
```diff
@@ -420,6 +420,25 @@ As mentioned above, all [parameters of the Dash AG Grid](https://dash.plotly.com
 
 If the available arguments are not sufficient, there is always the option to [create a custom AG Grid callable](custom-tables.md).
 
+#### Add sticky headers inside a Flex layout
```
maxschulz-COL: Is there a reason we did not make this an argument of the model? Or even better, detect automatically?

antonymilne:
@petar-qb pointed me here and I pretty much agree with what he said: I don't think `position: fixed` can ever consistently solve this when it's built into Vizro. 

Doing `position: sticky` sounds much closer to what we want (sticking relative inside the parent container rather than fixed to viewport) but relies on something like the container component being scrollable e.g. with `overflow: auto`. From what I've read, it seems like sticky headers are turned on with AgGrid automatically. What I'm not 100% of is whether the condition for that is exactly the same as the condition for `position: sticky` to work, but if they are then there's not much hope here because it means that if this were possible then it would just work out of the box already.

So I think where this leaves us is:
* try for a bit to play around with `position: sticky` to see if that works. I remember there's been lots of discussion before about where we can/cannot assign container heights, scrollability, etc. so if this is feasible then we have to not break whatever we decided before 😅 
* if that doesn't work then I don't think we can solve this and it's best to just put in docs, like @petar-qb said

maxschulz-COL:
@petar-qb Good catch - I was not aware of these consequences. I think going for the docs approach seem the most sensible course of action here!

vizro-core/changelog.d/20251124_165505_90609403+huong-li-nguyen.md
```diff
@@ -0,0 +1,47 @@
...
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Changed
+
+- Enable AG Grid sticky headers when used within Grid layout. See our [user guide](https://vizro.readthedocs.io/en/stable/pages/user-guides/table/#add-sticky-headers-inside-a-flex-layout) for an example of how to enable it in the Flex layout. ([#1503](https://github.com/mckinsey/vizro/pull/1503))
+
```
petar-qb, line 29: Let's remove this.

vizro-core/docs/pages/user-guides/table.md
```diff
@@ -420,6 +420,25 @@ As mentioned above, all [parameters of the Dash AG Grid](https://dash.plotly.com
...
+
+```css
+.ag-header {
+    position: fixed !important;
+    z-index: 1;
+}
+
+.ag-body {
+    top: 40px;
+}
```
petar-qb, line 435: When added, this CSS will affect all ag-grids in the dashboard.
If possible, let's add a hint for users explaining how the CSS can be applied per ag-grid model.

vizro-core/docs/pages/user-guides/table.md
```diff
@@ -420,6 +420,25 @@ As mentioned above, all [parameters of the Dash AG Grid](https://dash.plotly.com
...
+```
+
+!!! note
+
+    This approach works reliably only when a single `AgGrid` is positioned on top of a non-scrollable page, or when all `AgGrid` tables are in the same row in Flex layout. Using a fixed header may cause unexpected behavior in more complex layouts.
```
petar-qb: It looks like the issue isn’t tied specifically to scenarios like:
- "a single AgGrid being positioned is positioned on top of" or 
- "all `AgGrid` tables are in the same row in Flex layout"

Rather, the behaviour appears to reproduce reliably on any non-scrollable page, so we should avoid implying those layout conditions are the root cause.

vizro-core/examples/scratch_dev/assets/css/custom.css
```diff
@@ -3,3 +3,14 @@
   height: 150px;
   width: 300px;
 }
+
```
petar-qb, line 6: Let's add the css from the docs here

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,97 +1,118 @@
...
+filtered_gapminder = gapminder[(gapminder["continent"] == "Europe") & (gapminder["year"] == 2007)]
 
-    type: Literal["control_group"] = "control_group"
-    title: str
-    controls: list[ControlType] = []
+page_grid_0 = vm.Page(
+    title="Grid",
+    layout=vm.Grid(
+        grid=[[0, 1], [2, 2]],
+        row_min_height="1000px",
```
petar-qb: Let's remove `row_min_height` here and in other pages to show how the custom css works for non-scrollable pages.

vizro-core/src/vizro/static/css/aggrid.css
```diff
@@ -5,7 +5,7 @@
   --ag-background-color: transparent;
   --ag-odd-row-background-color: transparent;
   --ag-data-color: var(--bs-heading-color);
-  --ag-header-background-color: transparent;
+  --ag-header-background-color: var(--right-side-bg-iterated);
```
petar-qb, line 8: What is this change?

vizro-core/tests/e2e/screenshots/chrome/main_ag_grid_page.png
petar-qb, line 1: Are the screenshot changes still relevant?

vizro-core/src/vizro/static/css/aggrid.css
nadijagraca: The header background was previously transparent. When using sticky headers, scrolling the ag-grid caused row values to show through, so a solid background color was added to prevent this.

vizro-core/docs/pages/user-guides/table.md
```diff
@@ -420,6 +420,38 @@ As mentioned above, all [parameters of the Dash AG Grid](https://dash.plotly.com
 
 If the available arguments are not sufficient, there is always the option to [create a custom AG Grid callable](custom-tables.md).
 
+#### Add sticky headers
+
+To add sticky headers to your AG Grid, you must add this CSS to your custom CSS file within your local `assets` folder.
```
petar-qb:
```suggestion
To add sticky headers to your AG Grid, add the following CSS to your custom CSS file within your local `assets` folder.
```

vizro-core/docs/pages/user-guides/table.md
```diff
@@ -420,6 +420,38 @@ As mentioned above, all [parameters of the Dash AG Grid](https://dash.plotly.com
...
+}
+
+#my-aggrid .ag-body {
+    top: 40px;
+}
+```
+
+!!! note
+
+    This approach works reliably only when a single `AgGrid` is positioned on top of a non-scrollable page.
```
petar-qb:
```suggestion
    This approach works reliably only when the `AgGrid` is positioned in the non-scrollable page.
```

PR #1507, opened 2025-11-26
[Dev] Pin default hatch environment to Python 3.13 to avoid Kedro compatibility issues

antonymilne:
Pins the `default` Hatch environment to Python 3.13 so that `hatch run test` gives consistent results on everyone's machines.

### DIFF
--- a/vizro-core/hatch.toml
+++ b/vizro-core/hatch.toml
@@ -45,7 +45,7 @@ dependencies = [
 ]
 features = ["kedro"]
 installer = "uv"
-# Pin Python version so that `hatch run test` etc. give consistent behaviour.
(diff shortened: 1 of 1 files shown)


vizro-core/hatch.toml
```diff
@@ -45,6 +45,9 @@ dependencies = [
 ]
 features = ["kedro"]
 installer = "uv"
+# Pin to Python 3.13 to avoid Kedro compatibility issues on Python 3.14.
+# Developers can still test with Python 3.14 using: hatch run all.py3.14:test-unit
```
antonymilne:
```suggestion
installer = "uv"
# Pin Python version so that `hatch run test` etc. give consistent behaviour.
```

PR #1508, opened 2025-11-27
[BUG] Fix finding controls when in Page/Container

petar-qb:
closes https://github.com/McK-Internal/vizro-internal/issues/2375

## Description
- Fix that control is properly found in the Page.controls / Container.controls if used nested within containers or nested custom components.
- Fix that `control.selector._in_container = True` is properly set when controls if used nested within containers or nested custom components.
- Write unit tests.
- Forbid placing controls under the Page.components or Container.components

## Notice

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,7 +41,7 @@ repos:
         args: [--autofix]
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.3
+    rev: v0.14.7
     hooks:
       - id: ruff
         args: [--fix]
@@ -49,14 +49,14 @@ repos:
       - id: ruff-format
 
   - repo: https://github.com/PyCQA/bandit
-    rev: 1.8.6
+    rev: 1.9.2
     hooks:
       - id: bandit
         args: [-c, pyproject.toml, -ll]
         additional_dependencies: ["bandit[toml]"]
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.18.2
+    rev: v1.19.0
     hooks:
       - id: mypy
         files: ^vizro-core/src/
@@ -76,7 +76,9 @@ repos:
         exclude: ^vizro-core/src/vizro/static/css/.+\.min.*|^vizro-core/src/vizro/static/css/mantine_dates.css
 
   - repo: https://github.com/executablebooks/mdformat
-    rev: 0.7.22 # mdformat 1.0.0 is incompatible with mdformat-frontmatter (max version 2.0.8 requires mdformat<0.8.0)
+    # mdformat 1.0.0 is incompatible with mdformat-frontmatter (max version 2.0.8 requires
+    # mdformat<0.8.0). No way to skip this automatically in pre-commit autoupdate.
+    rev: 0.7.22
     hooks:
       - id: mdformat
         args:


vizro-core/changelog.d/20251127_101136_petar_pejovic_fix_finding_nested_controls.md
```diff
@@ -0,0 +1,48 @@
...
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
```
petar-qb: Should I mention it here?

antonymilne:
Thank you for dealing with this so promptly! A few initial comments/questions:
1. I see this enables controls to live inside `components`. I think we would ideally not allow that (even if it works now). The previous reason we might have wanted to make it possible was only relevant when we didn't have the ability to do controls inside containers, so that the only way to get a control to appear in the main part of the page was by putting it inside `components`. Now that we controls inside containers, I think we should keep controls strictly in controls and not let them work inside `components`. Otherwise we start mixing up the "control dictates components" paradigm and it will make future changes harder. If the code gets much more complicated just to forbid the "control inside components" case then don't worry about it if the code still allows it, but the changelog entry should then just say we've enabled nesting inside `Page.controls` and `Container.controls` rather than anywhere inside those models.[^1]
1. one things this _should_ ideally enable is to put a control group arbitrarily nested inside multiple containers, e.g. `Container(Container(Container(controls=ControlGroup))`. Does the logic currently work correctly there? If it's a big effort to enable then don't worry though.
1. sorry I accidentally deleted your comment about changelog entry with my clumsy hands 😅 I think we should indeed put this in the changelog, once we've figured out exactly what the change is (see question 1)

[^1]: See also [this conversation](https://github.com/mckinsey/vizro/pull/903#discussion_r1864269529) from nearly a year ago where I said something similar. Good to know I'm consistent on this one 😀

vizro-core/src/vizro/models/_controls/_controls_utils.py
```diff
@@ -54,24 +54,23 @@ def _validate_targets(targets: list[str], root_model: VizroBaseModel) -> None:
...
     if (page := model_manager._get_model_page(model=control)) is None:
         return None
 
-    # Return the Page if the control is its direct child.
-    if control in page.controls:
-        return page
+    # Return the Container if the control is nested inside it.
+    for container in model_manager._get_models(model_type=Container, root_model=page):
+        if control in model_manager._get_models(model_type=type(control), root_model=container):
+            return container
```
antonymilne: Will this work with containers nested inside each other? Not the end of the world if it doesn't, but if it's easy to enable that then it would be nice.

vizro-core/src/vizro/models/_page.py
```diff
@@ -151,10 +151,8 @@ def pre_build(self):
         if targets:
             self.actions = [_on_page_load(id=f"{ON_PAGE_LOAD_ACTION_PREFIX}_{self.id}", targets=targets)]
 
-        controls = cast(
-            Iterable[ControlType],
-            [*model_manager._get_models(Parameter, self), *model_manager._get_models(Filter, self)],
-        )
+        # Convert generator to list as it's going to be iterated multiple times.
```
antonymilne, line 154: This is the bit (and comparable code in container) where I think ideally we'd look at child models of `self.controls` rather than `self` (which includes `self.anything`). I know the `_get_models` stuff is probably not currently well setup to do that though, so if it's way too complicated then don't worry about it. Something like this maybe? With some comment explaining what's going on.
```
controls = []
for control in self.controls:
    if isinstance(control, (Filter, Parameter)):
        controls.append(control)
    else:
        controls.extend(model_manager._get_models((Filter, Parameter), control))
```

vizro-core/src/vizro/models/_controls/_controls_utils.py
petar-qb: Great catch! I didn't work completely correctly with nested containers as it didn't return the nearest ancestor container, but the farthest ancestor container.

vizro-core/src/vizro/models/_page.py
petar-qb: I fixed the `_get_models` stuff so it allows the following input `root_model=self.controls`. So, I left the `self` because here in `_page.py` because controls could be nested deeply in the container's controls under the `Page.components`:
```py
        # Convert generator to list as it's going to be iterated multiple times.
        controls = list(cast(Iterable[ControlType], model_manager._get_models((Filter, Parameter), self)))
```

However, here's what I did in the `Container.pre_build`:
```py
        # Mark controls under this container as `_in_container`. Note this relies on the fact that filters are pre-built
        # upfront in Vizro._pre_build. Otherwise, control.selector might not be set.
        for control in cast(
            Iterable[ControlType], model_manager._get_models(model_type=(Filter, Parameter), root_model=self.controls)
        ):
            control.selector._in_container = True
```
I scoped the search on `self.controls` here to avoid setting the `control.selector._in_container = True` multiple times for nested containers.

petar-qb:
> I see this enables controls to live inside components. I think we would ideally not allow that (even if it works now). The previous reason we might have wanted to make it possible was only relevant when we didn't have the ability to do controls inside containers, so that the only way to get a control to appear in the main part of the page was by putting it inside components. Now that we controls inside containers, I think we should keep controls strictly in controls and not let them work inside components. Otherwise we start mixing up the "control dictates components" paradigm and it will make future changes harder. If the code gets much more complicated just to forbid the "control inside components" case then don't worry about it if the code still allows it, but the changelog entry should then just say we've enabled nesting inside Page.controls and Container.controls rather than anywhere inside those models.

My opinion about this and similar topics doesn't change over time. I think we should enable it. Why? Because our code is easier to maintain (by coding, testing, documenting and maintaining less use cases). Anyway, Vizro users have to enable it explicitly with `.add_type` so it's out of our responsibility to take care about that.

Also, by introducing that rule explicitly (by raising an exception if a control is added within the components) we should take care of what happens if component (e.g. Text/Card/CustomSomething) is added under the "controls". Shouldn't we? It means again more code more tests more documentation and more mental effort for someone to understand it. What do we get by introducing this? To forbid Vizro users to create the dashboard they (or their designers) want. I don't think we get anything else.

This is much bigger question than "Should we allow users to specify controls within the components". This is the question about Vizro vision. I'm not the one whose opinion will decide anything about it, but I just want to share what I think once again.

---

However, I did restrict placing controls inside the deepest Page/Container components argument, simply because it was straightforward to implement given that `MM._get_models` uses a depth-first pre-order search. With any other traversal strategy (or BFS, or something else), the implementation would have become significantly more complex.

vizro-core/changelog.d/20251127_101136_petar_pejovic_fix_finding_nested_controls.md
```diff
@@ -0,0 +1,46 @@
...
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+### Fixed
+
+- Fix that custom controls can be nested arbitrarily deep inside `Page.controls` and `Container.controls`. ([#1508](https://github.com/mckinsey/vizro/pull/1508))
```
antonymilne:
```suggestion
- Fix regression that prevented custom controls being nested arbitrarily deep inside `Page.controls` and `Container.controls`. ([#1508](https://github.com/mckinsey/vizro/pull/1508))
```

vizro-core/src/vizro/models/_controls/_controls_utils.py
```diff
@@ -54,25 +59,26 @@ def _validate_targets(targets: list[str], root_model: VizroBaseModel) -> None:
...
+    # Fallback to the page if not nested inside any container.
+    return nearest_ancestor_container or page
 
 
 def check_control_targets(control: ControlType) -> None:
-    if (root_model := _get_control_parent(control=control)) is None:
-        raise ValueError(f"Control {control.id} should be defined within a Page object.")
+    root_model = get_control_parent(control=control)
+    if root_model is None or control not in model_manager._get_models(root_model=root_model.controls):
+        raise ValueError(f"Control {control.id} should be defined within a Page.controls or Container.controls.")
```
antonymilne:
```suggestion
        raise ValueError(f"Control {control.id} should be defined within Page.controls or Container.controls.")
```

vizro-core/src/vizro/models/_controls/_controls_utils.py
```diff
@@ -54,25 +59,26 @@ def _validate_targets(targets: list[str], root_model: VizroBaseModel) -> None:
...
-    return next(container for container in page_containers if control in container.controls)
+    # Fallback to the page if not nested inside any container.
+    return nearest_ancestor_container or page
 
 
 def check_control_targets(control: ControlType) -> None:
-    if (root_model := _get_control_parent(control=control)) is None:
-        raise ValueError(f"Control {control.id} should be defined within a Page object.")
+    root_model = get_control_parent(control=control)
+    if root_model is None or control not in model_manager._get_models(root_model=root_model.controls):
```
antonymilne, line 81: What's the `control not in model_manager._get_models(root_model=root_model.controls)` for here? I don't see when it would be relevant. 🤔

vizro-core/src/vizro/models/_controls/_controls_utils.py
```diff
@@ -54,25 +59,26 @@ def _validate_targets(targets: list[str], root_model: VizroBaseModel) -> None:
 #  This would make the following renaming logical: model_manager._get_models -> model_manager._get_model_children.
 #  These two new methods could have the same signature.
 #  Consider adding the parent_model_id to the VizroBaseModel and use that to find the parent model more easily.
-def _get_control_parent(control: ControlType) -> VizroBaseModel | None:
-    """Get the parent model of a control."""
+def get_control_parent(control: ControlType) -> Page | Container | None:
```
antonymilne, line 62: This function is so fragile and complicated for something that should really just be `control._parent_model` or similar... I can't wait until we finally get the model manager stuff done to simplify all this stuff.

(Not a criticism of your code at all - I think you did a great job with the tools currently available!)

vizro-core/src/vizro/models/_components/container.py
```diff
@@ -122,9 +124,13 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
     @_log_call
     def pre_build(self):
-        # Note this relies on the fact that filters are pre-built upfront in Vizro._pre_build. Otherwise
-        # control.selector might not be set.
-        for control in self.controls:
+        from vizro.models import Filter, Parameter
+
+        # Mark controls under this container as `_in_container`. Note this relies on the fact that filters are pre-built
+        # upfront in Vizro._pre_build. Otherwise, control.selector might not be set.
+        for control in cast(
```
antonymilne, line 133: Maybe worth adding a comment here explaining why we don't just do `for control in self.controls`. I was just trying to understand that myself when it's obvious because it's the whole point of this PR. So I dread to think what my brain would understand here in a few months time outside the context of this PR 😅 

(Good think you added the tests to prevent someone like me inadvertently rewriting this code and breaking everything in future 💯 )

vizro-core/src/vizro/models/_page.py
antonymilne: Ahh yes, that makes perfect sense, thank you! Please could you add a comment here about why we need to look in `self` and not just `self.controls`.

vizro-core/src/vizro/managers/_model_manager.py
```diff
@@ -63,7 +63,7 @@ def __iter__(self) -> Generator[ModelID, None, None]:
     def _get_models(
         self,
         model_type: type[Model] | tuple[type[Model], ...] | type[FIGURE_MODELS] | None = None,
-        root_model: VizroBaseModel | None = None,
+        root_model: VizroBaseModel | Mapping[str, Model] | Collection[Model] | None = None,
```
antonymilne, line 66: Great spot that this was already possible 👍

vizro-core/tests/unit/vizro/models/conftest.py
```diff
@@ -16,3 +18,22 @@ def _identity_action_function(arg=None):
 @pytest.fixture(params=[vm.Container, vm.Page])
 def model_with_layout(request):
     return request.param
+
+
+@pytest.fixture
+def mock_control_wrapper_class():
+    class MockControlWrapper(vm.VizroBaseModel):
```
antonymilne:
```suggestion
def MockControlWrapper():
    class _MockControlWrapper(vm.VizroBaseModel):
```

Otherwise it just looks weird at point of use.

vizro-core/tests/unit/vizro/models/conftest.py
```diff
@@ -16,3 +18,22 @@ def _identity_action_function(arg=None):
...
+        """Wrapper model around the control."""
+
+        type: Literal["mock_control_wrapper"] = "mock_control_wrapper"
+        control: vm.Filter | vm.Parameter = None
+
+        def __new__(cls, *args, **kwargs):
+            vm.Page.add_type("controls", cls)
+            vm.Container.add_type("controls", cls)
+            return super().__new__(cls)
```
antonymilne, line 37: Where we do `add_type` to real Vizro models (i.e. anything outside `test_base.py`) we should also put in the following comment. Currently we do this once in `test_filter.py`:

```
    # This test does add_type so ideally we would clean up after this to restore vizro.models to its previous state.
    # This is difficult to fix fully by un-importing vizro.models though, since we use `import vizro.models as vm` - see
    # https://stackoverflow.com/questions/437589/how-do-i-unload-reload-a-python-module.
```

vizro-core/src/vizro/models/_controls/_controls_utils.py
petar-qb: The line above calls `root_model = get_control_parent(control=control)`, where `root_model` can be:

1. `None` - the control is defined outside any page context.
2. `vm.Container` - the nearest ancestor container in which the control is defined.
3. `vm.Page` - the page under which the control is defined.

`get_control_parent` simply returns the model (page or container or none) that owns the control, regardless of whether it appears under `components` or `controls`.

To ensure/validate the control actually lives under the `controls` (Page.controls or Container.controls), I added the check:

`control not in model_manager._get_models(root_model=root_model.controls)`

If the control is not found under `Page.controls` or `Container.controls`, an exception is raised.

vizro-core/src/vizro/models/_controls/_controls_utils.py
petar-qb: yepp totally agree! Can't wait to have improved model_manager 🤞 🤞

vizro-core/src/vizro/models/_controls/_controls_utils.py
antonymilne: Ohhh ok, I see. So this checks that it's in `Page.controls` rather than e.g. `Page.components`? Can't we just write it as `control not in root_model.controls` then?

Or just remove that clause completely and silently allow it to work, I don't mind (as per overall comment) 🙂

petar-qb: > Can't we just write it as control not in root_model.controls then?

Nop, in that case we wouldn't cover the case when e.g. a `ControlGroup(controls=[vm.Filter(id="page_controls_group", column="country")])` is placed under the `Page.controls` or `Container.controls` 😄

You can find all cases I found in the scratch example where these that raises an exception are commented out.

petar-qb: > I think we are actually not so different in our opinions here, and I wasn't very clear in my previous message, sorry! In this particular case there's maybe three possible solutions:
>1. allow controls inside components and advertise it as a feature. I don't like this because it means we have to maintain it as a feature, and I don't know if it's something we want to encourage
>2. don't allow controls inside components on purpose (i.e. raise clear exception when someone tries to do it). Agree with you that it's not worth the effort since means raising exceptions, testing for those exceptions, etc.
>3. don't care whether controls inside components works, i.e. it's not advertised as a feature or documented. It might work now but not in future or vice versa. We haven't deliberately enabled it or forbidden it. This is great because it means no need to worry about special cases, testing, documentation, etc. 
>
>So both with your original PR and how it stands now, we're in category 3. It's less about whether it's actually possible to do controls inside components and more a question of whether we officially document/support/test the functionality (which includes mentioning it in changelog).

Actually the `2. don't allow controls inside components on purpose` is now implemented. 😄 

Now, with this PR, controls and nested controls are allowed under `Page.controls` and `Container.controls`. If a control or a nested custom control is placed directly under `Page.components` or `Container.components`, an exception will be raised. Given that the current behaviour is implemented without much of additional logic, and given that tests are written too, I wouldn’t make additional changes at this point. 

I explained that in the comment above:
> However, I did restrict placing controls inside the deepest Page/Container components argument, simply because it was straightforward to implement given that MM._get_models uses a depth-first pre-order search. With any other traversal strategy (or BFS, or something else), the implementation would have become significantly more complex.

vizro-core/src/vizro/managers/_model_manager.py
petar-qb: The bigger discovery here for me was that `_get_models` returns models using a depth-first pre-order traversal. With any other traversal strategy such or BFS, implementing `get_control_parent` to locate the nearest ancestor would become significantly more complex.

vizro-core/tests/unit/vizro/models/conftest.py
petar-qb: Totally forgot about that! Thanks

vizro-core/src/vizro/models/_controls/_controls_utils.py
antonymilne: Ohhh ok I finally understand this bit of code then 😀 Yeah, so I don't mind whether that `model_manager._get_models(root_model=root_model.controls)` clause stays in there or not (i.e. we go for solution 2 as you have now or solution 3). So I'd just stick with what you have now.

PR #1515, opened 2025-12-03
[Feat] Set `clickmode("event+select")` by default for `plotly.express` charts and `multiRow` selection for AgGrid

petar-qb:
Closes https://github.com/McK-Internal/vizro-internal/issues/2239
Closes https://github.com/McK-Internal/vizro-internal/issues/2128

Related to https://github.com/McK-Internal/vizro-internal/issues/2188 (closed as irrelevant after this PR)

## Description
- Multi-select controls can now be updated by the `set_control` action. For example, you can set controls by using lasso select on a graph or selecting multiple rows on a table.
- Set `clickmode="event+select"` as the default for all `plotly.express` charts, enabling point/data highlighting on click
- Set `dashGridOptions.rowSelection.multi = "multiRow"` by default for `dash_ag_grid`. Also, set `dashGridOptions.rowSelection.checkboxes` and `dashGridOptions.rowSelection.headerCheckbox` to `True` if `set_control` is attached to the `AgGrid` model.
- Returning `None` from a `set_control` trigger now resets the targeted control to its original value.

## Why this change?
1. Enabled multi-point selection in Graphs and multi-row selection in AgGrid.
2. Graphs now visually highlight selected data, matching the AgGrid behaviour.
3. Added support for unselecting points/rows.

## Screen Recordings: All Features in scratch/app.py

https://github.com/user-attachments/assets/16062384-fa93-48b8-a3f1-b5b10f983e71

https://github.com/user-attachments/assets/95d09d4a-8d98-496f-b295-b6c72b684373

## TODO
- [x] - @antonymilne - Manual testing and PR review
- [x] - @petar-qb - Address PR comments
- [x] - @petar-qb - Fix bug with refreshed Graph and its selectedData (consider to do no_update if no selection happened)
- [x] - @petar-qb - Clean `graph.js` and answer the last batch of PR comments
- [x] - @petar-qb - Fix failing unit tests
- [x] - @petar-qb - Add new unit tests 
- [x] - @petar-qb - Prepare a e2e-tests ticket with detailed description for @l0uden 
- [ ] - @petar-qb - Document changes and also explain how `clickmode(E+S)` can be disabled.

## Notice

### DIFF
--- a/vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
+++ b/vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
@@ -404,7 +404,7 @@ def test_drill_down_graph(dash_br):
     )
 
     # check that graph title changed to 'versicolor'
-    dash_br.wait_for_text_to_equal(".gtitle", "Graph shows `[versicolor]` species.")
+    dash_br.wait_for_text_to_equal(".gtitle", "Graph shows `['versicolor']` species.")
 
 
 def test_action_properties_shortcut_title_description_header_footer(dash_br):


vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -168,11 +169,26 @@ def pre_build(self):
...
+
+        # Normalize returned value based on target selector type.
+        selector = model_manager[self.control].selector
+        is_multi = getattr(selector, "multi", isinstance(selector, Checklist))
+        if is_multi:
+            value = sorted(set(value)) if isinstance(value, list) else [value]
+        else:
+            if value == []:
+                # Single-value selector cannot be set to empty list.
+                return no_update if self._same_page else (no_update, no_update)
```
petar-qb, line 201: Should we log something here?

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -168,11 +169,26 @@ def pre_build(self):
...
-    def function(self, _trigger):
+    def function(self, _trigger, _vizro_controls_store):
+        from vizro.models import Checklist
         value = cast(_SupportsSetControl, self._parent_model)._get_value_from_trigger(self.value, _trigger)
 
+        # If value is None then reset control to original value.
+        if value is None:
+            value = _vizro_controls_store[self.control]["originalValue"]
+
+        # Normalize returned value based on target selector type.
```
petar-qb, line 183: I handled single-select controls like this:
1. Graph/AgGrid setting `None` or `[]` is treated as None -> reset control
2. Figure/Button/Card setting `None` -> reset control
3. Figure/Button/Card setting `[]`
    3.1. to multi-select control -> sets `[]` 
    3.2. to single-select control -> `no_update`

@antonymilne is this intuitive in your opinion?

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -168,11 +169,26 @@ def pre_build(self):
...
+        selector = model_manager[self.control].selector
+        is_multi = getattr(selector, "multi", isinstance(selector, Checklist))
+        if is_multi:
+            value = sorted(set(value)) if isinstance(value, list) else [value]
+        else:
+            if value == []:
+                # Single-value selector cannot be set to empty list.
+                return no_update if self._same_page else (no_update, no_update)
+            value = value[0] if isinstance(value, list) else value
```
petar-qb: See whether to return `values[-1]` instead of `values[0]` when many points/rows selected but a single-select control is targeted.

The problem here is that nothing guarantees the order of selected points and the result might be change but sometimes may stay the same.

I think returning `value[0]` is fine, but maybe we can do `no_update` here as well.

vizro-core/src/vizro/static/js/models/graph.js
```diff
@@ -0,0 +1,55 @@
...
+  // 2. clickmode is not "event+select" (default value is overwritten) and clickData is triggered.
+  if (!isGraphSelectable || (!isClickmodeEventSelect && isClickDataTriggered)) {
+    return clickData;
+  }
+
+  // Ignore "clickData" event for "event+select" mode for charts that support selectedData.
+  // This is a workaround for a Plotly issue where "event+select" is set but clickData is the only thing that's updated.
+  if (!isSelectedDataTriggered) {
+    return dash_clientside.no_update;
+  }
```
petar-qb: By setting `clickmode=("event+select")` it can happen that graph's `clickData` changes without affecting `selectedData` and graph's highlight -> [see this comment](https://github.com/McK-Internal/vizro-internal/issues/2239#issuecomment-3415123499). This happens when a graph's point is clicked immediately after it's hovered.

vizro-core/src/vizro/static/js/models/graph.js
```diff
@@ -0,0 +1,56 @@
...
+  ) === "event+select";
+
+  // Extract trigger IDs.
+  const triggeredIds = dash_clientside.callback_context.triggered;
+  const isClickDataTriggered = triggeredIds.some(t => t.prop_id.includes("clickData"));
+  const isSelectedDataTriggered = triggeredIds.some(t => t.prop_id.includes("selectedData"));
+
+  // Return clickData if:
+  // 1. graph is not selectable or
+  // 2. clickmode is not "event+select" (default value is overwritten) and clickData is triggered.
```
petar-qb: It was really important to me to enable everything to work fine when `clickmode="event"` is manually set (in that case use `clickData`).

vizro-core/src/vizro/static/js/models/graph.js
```diff
@@ -0,0 +1,66 @@
...
+function update_graph_actions_trigger_prop(
+  clickData,
+  selectedData,
+  figure,
+  figureId,
+) {
+  // Check whether graph supports selection by looking for "Box Select" option in the graph's modebar in chart's header.
+  const isGraphSelectable = !!document
+    .getElementById(figureId)
+    .querySelector('[data-title="Box Select"]');
```
petar-qb: I tried different methods like: `figure.data[0].selectedpoints !== undefined;` or by matching the `figure.data[0].type` and `figure.data[0].mode` with the list of whitelisted types/modes, but this one turned out as the best approach.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -228,6 +228,7 @@ def _transformed_inputs(self) -> list[State] | dict[str, State | ControlsStates]
                 "filter_interaction": self._get_filter_interaction_states(),
             },
             "_trigger": State(*self._first_in_chain_trigger.split(".")),
+            "_vizro_controls_store": State("vizro_controls_store", "data"),
```
```suggestion
            "_controls_store": State("vizro_controls_store", "data"),
```

I think this is probably a good convention to follow:
* we use `vizro_` prefix for the Dash components so that user's don't accidentally clash with them, like `vizro_controls_store`, `vizro_download`, etc.
* when they're built-in arguments to actions, no need for the `vizro_` prefix. If a user makes their own `controls_store` argument then it takes priority already

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -114,15 +114,21 @@ def _action_inputs(self) -> dict[str, _IdProperty]:
...
 
-    def _get_value_from_trigger(self, value: str, trigger: list[dict[str, str]]) -> JsonValue:
-        """Value is the name of the column. There is only one row selected, so we just look at trigger[0]."""
-        # In case the selectedRows is empty (e.g. when the user unselects the row), trigger is an empty list.
-        # This works for both multi=True and multi=False selectors.
+    def _get_value_from_trigger(self, value: str, trigger: list[dict[str, str]]) -> MultiValueType | None:
+        """Extract values from the trigger that represents selected dag.AgGrid rows. Value is the name of the column.
+
+        Returns:
+          - list of values (one per point) or None if no points selected (signals reset).
```
antonymilne:
```suggestion
          - list of values (one per row) or None if no row selected (signals reset).
```

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -114,15 +114,21 @@ def _action_inputs(self) -> dict[str, _IdProperty]:
...
+          - ValueError if `value` column name can't be found.
+        """
+        # Returning None signals a reset of control to its original value.
         if not trigger:
-            return []
+            return None
 
         try:
-            return trigger[0][value]
+            return sorted({row[value] for row in trigger})
```
antonymilne: Do we actually want to do `sorted`? I think probably not. Probably best to just send values in the same order that they are selected. There's a few cases I can think of where we want to preserve the order that multiple things are selected rather than sorting them.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -110,7 +117,7 @@ def _make_actions_chain(self):
 
     @property
     def _action_triggers(self) -> dict[str, _IdProperty]:
-        return {"__default__": f"{self.id}.clickData"}
+        return {"__default__": f"{self.id}_actions_trigger.data"}
```
antonymilne:
```suggestion
        return {"__default__": f"{self.id}_action_trigger.data"}
```
To match the property name.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -280,6 +320,16 @@ def build(self):
             prevent_initial_call=True,
         )
 
+        clientside_callback(
+            ClientsideFunction(namespace="graph", function_name="update_graph_actions_trigger_prop"),
```
antonymilne:
```suggestion
            ClientsideFunction(namespace="graph", function_name="update_graph_action_trigger"),
```

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -122,33 +129,66 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
+    def _get_value_from_trigger(self, value: str, trigger: dict[str, list[JsonValue]]) -> MultiValueType | None:
+        """Extract values from the trigger that represents clicked or selected Plotly graph points.
+
+        Priority:
+          1) If `value` matches a column name in self["custom_data"], take it as from customdata[index].
+          2) Otherwise treat `value` as a Box lookup (e.g. "x", "customdata[0]").
+
+        Notes:
+          - Allows camelCase and snake_case value keys interchangeably (camel_killer_box=True).
+          - Enables dot-style (e.g. value="customdata[0]") access to nested dict values (box_dots=True).
```
antonymilne:
```suggestion
          - Enables dot-style (e.g. value="key.subkey") access to nested dict values (box_dots=True).
```

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -122,33 +129,66 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
+          - Automatically creates missing keys as empty boxes instead of raising errors (default_box=True). This is done
+            to avoid exceptions when the `trigger` has a key with a dot in it.
```
antonymilne, line 145: When might this happen in reality?

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -122,33 +129,66 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
+
+        Returns:
+          - list of values (one per point) or None if no points selected (signals reset).
+
+        Raises:
+          - ValueError if `value` can't be found.
+        """
+        # Returning None signals a reset of control to its original value.
+        if not (trigger or {}).get("points"):
```
antonymilne: What cases is this meant to cover? 
* `trigger = {}`
* `trigger` is dictionary but doesn't have `points`. Is that even possible?

Anything else? Just wondering why `if not trigger` by itself isn't sufficient here.

I was wondering before whether this reset logic should live inside `set_control` itself so we do `raise PreventUpdate` before we call `_get_value_from_trigger`, but I think it's probably best to handle it in each `_get_value_from_trigger` independently. Because e.g. if we put `if not trigger` in `set_control` directly then if you sent the value `[]` or `0` from clicking on a button, it would treat it as a "reset control" operation rather than "set to empty value" or "set to `0`". And since `trigger` can take a different form depending on the trigger component, probably it's best to handle it in the trigger component itself to work out when it means "reset control". So I think it's right how you have it now. Do you agree?

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -122,33 +129,66 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
-                    f"Couldn't find value `{value}` in trigger for `set_control` action. "
-                    f"This action was added to the Graph model with ID `{self.id}`. "
-                    "If you expected the value to come from custom data, add it in the figure's custom_data signature."
-                )
+            lookup_keys.append(f"customdata[{index}]")
+
+        # Treat the value as a box lookup string, as in https://github.com/cdgriffith/Box/wiki/Types-of-Boxes#box-dots
+        # This works for e.g. value="x" or value="customdata[0]".
+        lookup_keys.append(value)
+
```
antonymilne: I don't quite understand the logic here. Why could there be multiple lookup keys? Don't we want to treat it as we either have `value=<column_name>` or `value=<box-lookup-string>`?

I was expecting something a bit more like this:

```suggestion
try:
    # Try value=<column_name>
    index = self["custom_data"].index(value)
    lookup = f"customdata[{index}]"
except (KeyError, ValueError):
    # If that fails, use value=<box-lookup-string>
    lookup = value

try:
    return [point[lookup] for point in trigger_box]
except (KeyError, TypeError):
    raise ValuError("...")
```

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -122,33 +129,66 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
+        for key in lookup_keys:
+            # Safely attempt to extract values by each lookup key from each point. Any lookup-related errors that could
+            # occur by accessing the Box (missing key, wrong type, etc.) are ignored, and we move on to the next key.
+            with suppress(Exception):
+                # Skip Box instances as these represent missing keys when due to the `default_box=True`.
+                if any(isinstance(point[key], Box) for point in trigger_box):
+                    continue
+
+            return sorted(
```
antonymilne: I think we should probably leave unsorted unless there's a good reason to sort.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -122,33 +129,66 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
+                {
+                    # If value is a list, flatten it to single value to skip returning nested object.
+                    val[0] if isinstance((val := point[key]), list) else val
```
antonymilne: Why could it be a list?

vizro-core/src/vizro/static/js/models/graph.js
```diff
@@ -0,0 +1,66 @@
...
+
+  // Extract the "clickmode" setting from the figure layout or default template.
+  // figure.layout.clickmode can be defined if user has explicitly set it with figure.update_layout(clickmode=...).
+  // figure.layout.template.layout.clickmode can be defined if user has overwritten the default template's clickmode.
```
antonymilne: I think we can assume this is not possible since `event+select` is hardcoded into `dashboard_overrides` and not the `vizro_light/dark` themes themselves. So I would only check `figure.layout.clickmode` here.

vizro-core/src/vizro/static/js/models/graph.js
antonymilne: I don't understand the need for `isGraphSelectable` here. Why can't we just do `if (isClickDataTriggered && !isSelectedDataTriggered)`? Does `!isGraphSelectable` imply `!isSelectedDataTriggered`? What am I missing?

vizro-core/src/vizro/static/js/models/graph.js
antonymilne: This is a smart idea. I think there might be some edge cases where it doesn't work e.g. what happens if I just turn off the mode bar in a plot? Maybe it's the best solution overall though compared to something more complicated, and we just ignore the edge case.

With that said I don't completely understand the need for this variable at all (see below) :sweat_smile: You're going to have to explain this function to me slowly...

vizro-core/src/vizro/actions/_set_control.py
antonymilne: Yeah, probably worth a `debug` log. Should we just `raise PreventUpdate` though instead of doing the `no_update`?

vizro-core/examples/scratch_dev/app.py
```diff
@@ -3,59 +3,432 @@
...
+                vm.Button(
+                    text="Set []",
+                    actions=va.set_control(control=f"{pre_target}filter_1", value=[]),
+                ),
+                vm.Card(
+                    text="Set 'versicolor'",
+                    actions=va.set_control(control=f"{pre_target}filter_1", value="versicolor"),
+                ),
+                vm.Card(
+                    text="""Set [['setosa', 'versicolor']]""",
```
antonymilne, line 328:
```suggestion
                    text="""Set \['setosa', 'versicolor'\]""",
```

I guess you were trying to escape the square brackets? The double square brackets doesn't work for that but I think this will.

vizro-core/examples/scratch_dev/app.py
```diff
@@ -3,59 +3,432 @@
...
+                px.data.iris(), x="sepal_width", y="sepal_length", color="species", custom_data=["species"]
+            ),
+            actions=va.filter_interaction(targets=[f"{pre}table"]),
+        ),
+        vm.AgGrid(id=f"{pre}table", figure=dash_ag_grid(px.data.iris())),
+    ],
+)
+
+pre = "p8_"
+page_8 = vm.Page(
```
antonymilne, line 392: This page does not seem to work correctly for me - when I do a selection, the control is updated for a moment but then seems to get reset straight away?

vizro-core/src/vizro/static/js/models/graph.js
antonymilne: Also, can you remind me why someone might want to set `clickmode="event"`? Is it just to avoid the "clicking doesn't work if you don't wait 0.3 seconds" Dash bug?

vizro-core/src/vizro/static/js/models/graph.js
```diff
@@ -0,0 +1,66 @@
...
+
+  // Return clickData if:
+  // 1. Graph is not selectable or
+  // 2. clickmode is not "event+select" (default value is overwritten) and clickData is triggered.
+  if (!isGraphSelectable || (!isClickmodeEventSelect && isClickDataTriggered)) {
+    return clickData;
+  }
+
+  // Ignore "clickData" event for "event+select" mode for charts that support selectedData.
+  // This is a workaround for a Plotly issue where "event+select" is set,
```
antonymilne: Please could you put in a link to that plotly Github issue we found?

vizro-core/examples/scratch_dev/app.py
```diff
@@ -3,59 +3,432 @@
...
+                    ],
+                ),
+            ]
+        ),
+        vm.AgGrid(id=f"{pre}table", figure=dash_ag_grid(px.data.iris())),
+    ],
+    controls=_create_filters(prefix=pre),
+)
+
+pre = "p3_"
```
antonymilne, line 253: I can't select multiple arbitrary rows in AgGrid using Ctrl + click, can you? Doing shift + click works but only for adjacent rows?

Also did you find any way to un-select a row in AgGrid? I guess this is possible with checkboxes enabled at any rate?

antonymilne: I wonder whether we should after all turn checkboxes on by default if `AgGrid` contains one of more `set_control` actions. I don't think would be too hard to do. wdyt? If it seems like a good idea then let's make a separate ticket for it.

vizro-core/src/vizro/actions/_set_control.py
antonymilne: Yes, very intuitive I think. It works well, great job!

vizro-core/src/vizro/actions/_set_control.py
antonymilne: I found this a bit confusing while using the app. I think we should just do `no_update` in the case that you select multiple things and the target is single-select. So maybe change the overall logic to just this:
```
# for single-select
if isinstance(value, list):
    if len(value) != 1:
        # Handles [] and [1, 2, 3] cases
        raise PreventUpdate
    [value] = value
return value
```

vizro-core/changelog.d/20251203_125033_petar_pejovic_use_selectedData_as_graphs_trigger.md
```diff
@@ -0,0 +1,45 @@
...
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+### Added
+
+- Support multi-selection (e.g. Graph points, AgGrid rows) that can now target multi-select controls via `set_control` actions. ([#1515](https://github.com/mckinsey/vizro/pull/1515))
```
antonymilne: Let's make this a highlight!

```suggestion
- Multi-select controls can now be updated by the `set_control` action. For example, you can set controls by using lasso select on a graph or selecting multiple rows on a table. ([#1515](https://github.com/mckinsey/vizro/pull/1515))
```

vizro-core/changelog.d/20251203_125033_petar_pejovic_use_selectedData_as_graphs_trigger.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Changed
+
+- Set `clickmode="event+select"` as the default for all `plotly.express` charts, enabling point/data highlighting on click. ([#1515](https://github.com/mckinsey/vizro/pull/1515))
+- Enable multi-row selection for `dash_ag_grid`. ([#1515](https://github.com/mckinsey/vizro/pull/1515))
```
antonymilne:
```suggestion
- Set `clickmode="event+select"` as the default for all graphs to enable data highlighting on click. ([#1515](https://github.com/mckinsey/vizro/pull/1515))
```

Since the multi-row thing is mentioned in highlights I wouldn't mention it again here.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -122,33 +129,66 @@ def _action_outputs(self) -> dict[str, _IdProperty]:
...
-        # are no realistic examples which need values outside this. We use Box for two reasons:
-        # 1. it makes it possible to address nested values in the trigger["points"][0] dictionary using a simple
-        # string syntax, e.g. value="x" or "value=customdata[0]". This is enabled by box_dots=True.
-        # 2. it converts camelCase to snake_case. e.g. if trigger contains something called "someKey" then both
-        # value="someKey" and value="some_key" will work. This is enabled by camel_killer_box=True.
-        trigger_box = Box(trigger["points"][0], camel_killer_box=True, box_dots=True)
-        try:
-            # First try to treat value as a column name. Unfortunately the customdata returned in the trigger does
+    def _get_value_from_trigger(self, value: str, trigger: dict[str, list[JsonValue]]) -> MultiValueType | None:
+        """Extract values from the trigger that represents clicked or selected Plotly graph points.
```
antonymilne, line 133: Please could you give an example here of what `trigger` might look like? I'm always forgetting...

antonymilne: Same goes for AgGrid as well :pray:

vizro-core/src/vizro/static/js/models/graph.js
```diff
@@ -0,0 +1,66 @@
...
+  // and clickData changes but selectedData and graph's highlight stays unaffected.
+  if (!isSelectedDataTriggered) {
+    return dash_clientside.no_update;
+  }
+
+  // Otherwise, graph is selectable and selectedData is triggered.
+  return selectedData;
```
antonymilne: Is there anything in `selectedData` or `clickData` that anyone might ever want to use that lives outside `points`? If no then I think should just return `selectedData[points]` here and then not need to do the lookup of `points` in `_get_value_from_trigger`.

This way when we eventually make `_trigger` public we have already done one stage of "data wrangling/cleaning" for the user. This is one of the things we wanted to achive in https://github.com/McK-Internal/vizro-internal/issues/2128. If it still seems like a good idea then we could do it here or in a separate PR, I don't mind.

antonymilne (approved):
Wow, amazing work!! I have lots of questions and comments, but basically this is awesome :100:

I found one thing that didn't seem to work correctly (self-filtering) and have a few minor changes I think we should make, but I'm going to approve already. You should feel free to merge this while I'm away once you're happy with it.

On the remaining TODOs:
> See and implement this: McK-Internal/vizro-internal#2128

Fine to implement this here or fine to save for a different PR, I don't mind.

On tests: one random thing I noticed that you should remember to check is that the `_get_value_from_trigger` gets the unique values. This is something that works now but would be easy to break in future without realising.

I have an idea for what some MVP does for this will look like - will comment further next week.

vizro-core/src/vizro/static/js/models/graph.js
petar-qb: >  I think there might be some edge cases where it doesn't work e.g. what happens if I just turn off the mode bar in a plot? Maybe it's the best solution overall though compared to something more complicated, and we just ignore the edge case.

I also noticed that edge case, and figured out that this best solution and that we should ignore it. I just forgot to comment it in the PR 😄 

Let me revisit the function and then will answer the question.

vizro-core/src/vizro/actions/_set_control.py
petar-qb: debug log added :+1: 

I avoid raising the PreventUpdate exception here as it stops other actions in the chain from running. The comment that explains that is added too.

vizro-core/src/vizro/models/_components/ag_grid.py
petar-qb: Great catch, thanks for digging into this!

I tested this behaviour as well, and it feels more intuitive that the most recently selected value is appended to the current selection (the control value). `AgGrid.selectedRows` preserves order as expected, which allows us to safely rely on logic like:

```python
# Use dict.fromkeys to remove duplicates while preserving order.
return list(dict.fromkeys([row[value] for row in trigger]))
```
However, `Graph.selectedData` does not appear to maintain a predictable order. So adding points with Shift+Click adds the most recently selected point non-deterministically somewhere in the list of `selectedData.points`.

Given this, I'd propose the following rule:
- If the source provides a meaningful order (latest selection appears last), we should preserve it. - AgGrid
- If the order is arbitrary or inconsistent, we should apply sorting to ensure determinism. - Graph

Does that approach make sense to you?

antonymilne: Oh, that's annoying that `Graph.selectedData` doesn't order things in a consistent way. In that case, yes I completely with your proposal. Just one small nitpick would be to remove those `[]`:
```
# Use dict.fromkeys to remove duplicates while preserving order.
return list(dict.fromkeys(row[value] for row in trigger))
```

antonymilne:
Here's my idea for how to do docs: https://github.com/McK-Internal/vizro-internal/issues/2392. Let me know what you think! I would probably put this in a different PR (which could go into the same release when this feature comes out or a different one, I don't mind).

One small annoyance I found with AgGrid while working on this: doing a multi-select with checkboxes works great, but if you click on a single row then it resets the selection to just that row. This is quite easy to do if you have clumsy fingers like me... Is there some way to change the behaviour of clicking on a single row to match clicking on a checkbox, i.e. it adds to the selection rather than replacing it? Even if it's possible, maybe that would cause other problems, I'm not sure... Up to you to see whether it's possible and whether it would be an improvement!

vizro-core/src/vizro/models/_components/graph.py
petar-qb: I've covered this case as I encountered this problem. However, I can't reproduce the issue now.. Should I remove handling this edge case?

vizro-core/src/vizro/models/_components/graph.py
petar-qb: >What cases is this meant to cover?
>
> - trigger = {}
> - trigger is dictionary but doesn't have points. Is that even possible?
>
>Anything else? Just wondering why if not trigger by itself isn't sufficient here.

1. When you click/select a graph's point and then unclick/unselect it, it results with `trigger=None`
2. When you use "Select Box" and select nothing it results with `trigger={"points": [], bbox={...}}`

So `if not trigger` would be insufficient for the second case and would set selectors on empty list instead of resetting them. So I did `if not (trigger or {}).get("points")` to handle both cases in one liner.

> I was wondering before whether this reset logic should live inside set_control itself so we do raise PreventUpdate before we call _get_value_from_trigger, but I think it's probably best to handle it in each _get_value_from_trigger independently. Because e.g. if we put if not trigger in set_control directly then if you sent the value [] or 0 from clicking on a button, it would treat it as a "reset control" operation rather than "set to empty value" or "set to 0". And since trigger can take a different form depending on the trigger component, probably it's best to handle it in the trigger component itself to work out when it means "reset control". So I think it's right how you have it now. Do you agree?

Totally agree. As you already said, there are several reasons why it's much better to handle it in the `_get_value_from_trigger`.

petar-qb: I added a comment for that (will push it later with other changes so it's easier to re-review):

```py
        # Returning None signals a reset of control to its original value. No point selected when trigger is None
        # (unclicking a point) or its 'points' dictionary field is empty (selecting no points).
        if not (trigger or {}).get("points"):
            return None
```

vizro-core/src/vizro/models/_components/graph.py
petar-qb: Relevant comment -> https://github.com/mckinsey/vizro/pull/1515#discussion_r2593093656

Decision is to remove sorted from AgGrid and leave it here in Graph.

vizro-core/src/vizro/models/_components/graph.py
petar-qb: I found that it's a list in some cases. 

https://github.com/user-attachments/assets/29201b81-776f-4d77-b306-142ab8c73f08

vizro-core/examples/scratch_dev/app.py
petar-qb: Both `[[` and `\[` works fine on the UI.

However, I'm getting the following SyntaxWarning when I do `\[`: 
```
scratch_dev/app.py:296: SyntaxWarning: "\[" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\["? A raw string is also an option.
  text="""Set \['setosa', 'versicolor'\]""",
```

vizro-core/examples/scratch_dev/app.py
petar-qb: Here’s how the selection works on my side:

- **Click** -> selects a single row (cannot unselect).  
  _This behaves the same as `clickData` in graphs._

- **Cmd + Click** -> selects or unselects individual rows.  
  _Functionally similar to Shift + Click in graph UIs as you can build a non-contiguous selection (e.g., rows 1, 3, 5)._

- **Shift + Click** -> selects a continuous range from the first selected row to the clicked row.  
  _Comparable to using a "Select box" in charts._

https://github.com/user-attachments/assets/21980fb5-8443-4092-a3c6-e8d73b95a4fa

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -168,11 +168,27 @@ def pre_build(self):
...
+
         value = cast(_SupportsSetControl, self._parent_model)._get_value_from_trigger(self.value, _trigger)
 
+        # If value is None then reset control to original value.
+        if value is None:
+            value = _vizro_controls_store[self.control]["originalValue"]
+
+        # Normalize returned value based on target selector type.
+        selector = cast(ControlType, model_manager[self.control]).selector
+        is_multi = getattr(selector, "multi", isinstance(selector, Checklist))
```
maxschulz-COL, line 185: Might bite as if we have another multi selector, but probably ok as unlikely right?

vizro-core/src/vizro/static/js/models/graph.js
maxschulz-COL, line 1: This is fairly hard to understand when new to it, but I assume you guys checked it thoroughly

vizro-core/src/vizro/models/_components/graph.py
antonymilne: Maybe it was when a column name had a `.` in it? If that was it then I think we can just ignore that anyway.

vizro-core/src/vizro/models/_components/graph.py
antonymilne: Cool, that sounds good thank you!

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,97 +1,434 @@
...
+        ),
+        vm.AgGrid(id=f"{pre}table", figure=dash_ag_grid(px.data.iris())),
+    ],
+    controls=_create_filters(prefix=pre),
+)
+
+pre = "p5_"
+pre_target = "p6_"
+page_5 = vm.Page(
+    title="Drill-through source page",
```
maxschulz-COL, line 304: 2 completely unrelated things, but worth noting:
- do we ever mention the word drill-through in docs? I think only in the (outdated) BI comparison
- why do we change page, is that actually built into `set_control` when targeting a model not on the same page? (sorry for not knowing)

vizro-core/src/vizro/models/_components/graph.py
antonymilne: Oh yes, weird... I see this when I try it out:
```
'customdata': [['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor'], ['versicolor']]
```

Not sure whether that's a plotly bug or a feature 🤔 

Is it safe to assume in these cases that it's the same value repeated many times?

No need to change the code either way, I think what you have is fine. Though might be nice to add a comment saying that for mysterious reasons e.g. `px.blah` sometimes it's a list and seems to be the same repeated value or something like that.

vizro-core/src/vizro/actions/_set_control.py
petar-qb: Yepp, you're right.

vizro-core/src/vizro/static/js/models/graph.js
petar-qb: I will give another round of testing/refactoring and will adjust the code comments

vizro-core/examples/scratch_dev/app.py
petar-qb: > - do we ever mention the word drill-through in docs? I think only in the (outdated) BI comparison

Here's the section that explains it in the docs -> https://vizro.readthedocs.io/en/stable/pages/user-guides/graph-table-actions/#cross-filter-between-pages

> - why do we change page, is that actually built into set_control when targeting a model not on the same page? (sorry for not knowing)

Yes. When `set_control` is used to target a control that lives on a different page, the system automatically navigates to that page before updating the control’s value. In other words, `drill-through` is built into how `set_control` works.

vizro-core/src/vizro/models/_components/graph.py
petar-qb: > Not sure whether that's a plotly bug or a feature :thinking_face:

I'm not sure either.. Every time I bumped into this situation all values were same.

> Is it safe to assume in these cases that it's the same value repeated many times?

Let me put a comment here. :+1: 
```py
                return sorted(
                    {
                        # Certain grouped charts (e.g. pie charts) return custom data as a list of values. These values
                        # appear to be the same, so flatten it to a single value to avoid returning a nested object.
                        val[0] if isinstance((val := point[key]), list) else val
                        for point in trigger_box
                    }
                )
```

vizro-core/src/vizro/models/_components/graph.py
petar-qb: Unfortunately this happens more often. You can find the `px.scatter` chart example with the `size` defined in the first page. This one contains `"marker.size"` key in the `selectedData`.
 
image

vizro-core/examples/scratch_dev/app.py
petar-qb: @antonymilne What dash-ag-grid version do you use? I use the latest one: `dash-ag-grid==32.3.2`

> I wonder whether we should after all turn checkboxes on by default if AgGrid contains one of more set_control actions. I don't think would be too hard to do. wdyt?

If `Cmd + Click` and `Shift + Click` work for the rest of the team with the latest `dash-ag-grid` version, I would proceed without adding checkboxes by default, but I would definitely explain in the docs how it can be turned on.

vizro-core/examples/scratch_dev/app.py
maxschulz-COL: @stichbury As for my first point - maybe worth updating? We seem to never mention the (known) concept of drill-through, only actually in the BI comparison (which is outdated) to say you cannot do it in Vizro. Maybe you could take a look?

petar-qb: Here's @antonymilne's rough [ticket](https://github.com/McK-Internal/vizro-internal/issues/1614) that groups many backend stuff which documentation should be improved/added. There are several mentions about the "drill-through", so I guess he's aware of that. He told me he'll revisit it in January and eventually solve many things in one bigger PR.

vizro-core/src/vizro/models/_components/graph.py
petar-qb: I don't know why, but I thought that I have to visit the `lookup = value` even when the `lookup = f"customdata[{index}]"` fails. So, I decided to "refactor it like this" 😄. I'm fixing it..

vizro-core/src/vizro/static/js/models/graph.js
petar-qb: Great catch! There's nothing outside the `clickData["points"]`/`selectedData["points"]` Vizro users might be interested in. Let me clean it and add a comment about the expected `_trigger` format in the `_get_value_from_trigger` functions (for graph/ag-grid).

vizro-core/src/vizro/models/_components/graph.py
petar-qb: I like the suggestion 👍 Did it.

vizro-core/src/vizro/models/_components/graph.py
antonymilne: Add comment to `BoxList` line where the exception would be raised without `default_box=True`.

vizro-core/examples/scratch_dev/app.py
petar-qb: Great catch, Antony! I identified the root cause of the bug. It comes from the way `selectedData` changes when the Graph is refreshed. In contrast, `clickData` remains unchanged. Because `selectedData` is modified during Graph's refresh, the actions attached to the `vm.Graph` are triggered unintentionally.

Here is a minimal Dash example that demonstrates the issue: [PyCafe example](https://py.cafe/petar-qb/dash-clickData-vs-selectedData-on-graph-refresh)

And here is the screen recording showing the behaviour:

https://github.com/user-attachments/assets/a1fdaa6c-b2ec-4b83-9fd3-0bdd68adf9a4

#### Solution:
When the graph is recreated, its `selectedData` is reset to `{"points": []}`. To prevent unintended triggers in this state, we chose to silence the update in `graph.js` by returning `dash_clientside.no_update`. The side effect is that we also ignore the case where a user intentionally clears the selection (for example by using the "Select Box"). This is a minor tradeoff and acceptable for our use case.

vizro-core/examples/scratch_dev/app.py
petar-qb: I tried to customise keyboard shortcuts for Graph/AgGrid, but unfortunately that's not possible. So, we leave it as it is.

petar-qb:
> Here's my idea for how to do docs: [McK-Internal/vizro-internal#2392](https://github.com/McK-Internal/vizro-internal/issues/2392). Let me know what you think! I would probably put this in a different PR (which could go into the same release when this feature comes out or a different one, I don't mind).
> 
> One small annoyance I found with AgGrid while working on this: doing a multi-select with checkboxes works great, but if you click on a single row then it resets the selection to just that row. This is quite easy to do if you have clumsy fingers like me... Is there some way to change the behaviour of clicking on a single row to match clicking on a checkbox, i.e. it adds to the selection rather than replacing it? Even if it's possible, maybe that would cause other problems, I'm not sure... Up to you to see whether it's possible and whether it would be an improvement!

I pushed [the change](https://github.com/mckinsey/vizro/pull/1515/commits/1a838b6510dee74b9dd749f1d5bccb0cc5a7e593) that sets the checkboxes by default if a `set_control` action is attached to the `vm.AgGrid` model.

vizro-core/src/vizro/static/js/models/graph.js
petar-qb: >  Does !isGraphSelectable imply !isSelectedDataTriggered?

No, `isGraphSelectable` represents whether the graph contains "Select Box" which is interpreted as whether multiple graph points can be selected or not. 

> Also, can you remind me why someone might want to set clickmode="event"?  Is it just to avoid the "clicking doesn't work if you don't wait 0.3 seconds" Dash bug?

Yes that's one of the reasons. Another reason could be if a user doesn't like when a single point is clicked that this point is highlighted.

However, I completely refactored the `graph.js` function and it looks much better and easier to digest now. 😃

vizro-core/src/vizro/static/js/models/graph.js
petar-qb: The `graph.js` is completely refactored now and should be easier to understand.

huong-li-nguyen (approved):

petar-qb:
FYI @antonymilne @l0uden 

There are two more TODOs regarding this PR that should be done this or early next year. 
- E2E tests: I will prepare a ticket with detailed description for @l0uden, so you can start working on that depending on your capability.
- Document changes and also explain how `clickmode(E+S)` can be disabled. @antonymilne, will this be documented as part of another batch of documentation changes, or separately? If it’s separate, I can handle it early next year.

antonymilne:
The docs ticket is here and separate from the big batch of changes so ready to work on whenever: https://github.com/McK-Internal/vizro-internal/issues/2392

Unfortunately I don't think the change in https://github.com/mckinsey/vizro/pull/1515/commits/1a838b6510dee74b9dd749f1d5bccb0cc5a7e593 is quite right because it won't work for custom table functions that don't have the `_set_checkboxes` argument (or allow `**kwargs`). This example from the [docs on custom tables](https://vizro.readthedocs.io/en/stable/pages/user-guides/custom-tables/#steps-to-create-a-custom-table) will give:
```
TypeError: my_custom_aggrid() got an unexpected keyword argument '_set_checkboxes'
```

Options are:
1. Try another solution that does `figure = self.figure(**kwargs)` and then modifies `figure.dashGridOptions` if those arguments haven't already been set explicitly. This is what I had imagined we would do here, although I don't know if it's possible.
2. Set `dashGridOptions` dynamically in `dash_ag_grid` using `model_manager`: let's not do this for now but might be a good option for future. Would need to work for custom functions as well somehow, which is not possible unless they explicitly call `dash_ag_grid`.
3. Add a new public argument to `dash_ag_grid` and don't set it dynamically (user needs to set it themselves)
4. Inspect `self.figure._arguments` (also need to work for `**kwargs?`) to work out whether to pass the argument
5. Enforce some new rule that custom grid functions need to have `**kwargs` - breaking change we could do in future but I don't think I like this anyway
6. Just don't add the feature at all so that user needs to specify `dashGridOptions` manually (same as but more awkward syntax)
7. Deliberately leave the bug there since no one complained yet and it's an interesting signal of how users use custom AgGrid functions, which is something I've been thinking about recently anyway...

Please could you try out option 1? 🙏  If that doesn't work then I'm not sure which of the other options I prefer, will need to think about it.

PR #1522, opened 2025-12-10
[Docs] Replace MkDocs with Zensical

stichbury:
Closes https://github.com/McK-Internal/vizro-internal/issues/2316 by transitioning Vizro, Vizro-MCP and Vizro-AI docs over from MkDocs to Zensical.

## Description
**Ready for review! You can see the builds on RtD**:

* [vizro-core](https://vizro--1522.org.readthedocs.build/en/1522/)
* [vizro-MCP](https://vizro--1522.org.readthedocs.build/projects/vizro-mcp/en/1522/)
* [vizro-AI](https://vizro--1522.org.readthedocs.build/projects/vizro-ai/en/1522/)

While I have created a new `zensical.toml` file (rather than just use the existing `mkdocs.yml`) we have had to revert to use the `mkdocs.yml` for now because of an issue with PyCafe. I will update as soon as the fix is in place from Zensical because using the toml format means we can pick up new features as they're added.

The settings are almost identical, but I added the feature `header.autohide` to increase viewport when scrolling down the page and added the feature `navigation.path` for breadcrumbs. We also now have light/dark theme support.

### Issues

(a) There are [some unsupported settings](https://zensical.org/docs/setup/basics/?h=strict#unsupported-settings) which we used -- like `strict` building. I think this is important but it's not here, so that means we could break an internal link and it won't be flagged. That's not ideal IMO but we may be able to catch these with a linkchecker instead, and it's [in the backlog for Zensical to provide it](https://github.com/zensical/zensical/issues/175).

(b) <s>PyCafe isn't working at present but I have identified the reason (probably) and [raised an issue](https://github.com/zensical/zensical/issues/193)</s>. Resolved by using `mkdocs.yml` while awaiting fix for the issue.

(c) <s>The llms-md buttons aren't working at present and that's another priority to investigate</s>. Now fixed

## Notice

### DIFF
--- a/vizro-core/docs/pages/user-guides/custom-components.md
+++ b/vizro-core/docs/pages/user-guides/custom-components.md
@@ -156,7 +156,7 @@ vm.Parameter.add_type("selector", TooltipNonCrossRangeSlider)
 
 You may want to use this strategy to:
 
-- Create a new component that does not exist as a [built-ninVizro model][vizro.models].
+- Create a new component that does not exist as a [built-in Vizro model][vizro.models].
 - Make extensive changes to an existing component.
 - Combine multiple components into a single higher-level component, similar to a Dash [All-in-One component](https://dash.plotly.com/all-in-one-components).
 


stichbury:
@maxschulz-COL here are some insights from Cursor about the reason why PyCafe isn't working at present. Please could you take a look and I can send it over to PyCafe if you agree with the analysis?

<details>

<summary>Keeping for the record but no longer relevant</summary>
<s>We're converting our docs from MkDocs to Zensical, which requires mapping our mkdocs.yml to a new toml config. We're running into a compatibility issue when trying to use the existing extension approach to enable PyCafe linking in our docs. 

## The Problem

The current mkdocs-pycafe architecture uses a factory pattern that works great with YAML but breaks with TOML:

**Current YAML (works):**
```yaml
custom_fences:
  - name: python
    validator: !!python/name:mkdocs_pycafe.validator
    format: !!python/object/apply:mkdocs_pycafe.formatter
      kwds:
        type: vizro
        requirements: vizro==0.1.49
```

**TOML (doesn't work):**
```toml
custom_fences = [
  { name = "python", 
    validator = "mkdocs_pycafe.validator",
    format = "mkdocs_pycafe.formatter",  # String reference, not a call
    kwds = { type = "vizro", requirements = "vizro==0.1.49" }  # Can't apply these
  }
]
```

The issue is that TOML has no equivalent to YAML's `!!python/object/apply:` tag, which calls the factory function with the kwds. TOML can only reference the function as a string, but can't express "call this function with these arguments."

### Potential Solutions

Would you consider one of these approaches to make mkdocs-pycafe more TOML-friendly?

#### Option 1: Pre-instantiated Formatters (Simplest)

Provide pre-configured formatters for common frameworks:

```python
# In mkdocs_pycafe/formatters.py or mkdocs_pycafe/presets.py
from functools import partial
from .core import _formatter

# Pre-configured formatters
solara = partial(_formatter, pycafe_type="solara", requirements="", ...)
vizro = partial(_formatter, pycafe_type="vizro", requirements="vizro==0.1.49", ...)
streamlit = partial(_formatter, pycafe_type="streamlit", requirements="", ...)
```

Then in TOML:
```toml
format = "mkdocs_pycafe.formatters.vizro"  # Direct reference, no calling needed!
```

#### Option 2: Configuration-aware Formatter

Create a formatter that can read configuration from the fence definition:

```python
def configurable_formatter(source, language, css_class, options, md, **kwargs):
    """Formatter that reads pycafe config from kwargs."""
    pycafe_type = kwargs.get('type', 'solara')
    requirements = kwargs.get('requirements', '')
    # ... rest of implementation
```

Then the kwds would naturally flow through:
```toml
format = "mkdocs_pycafe.configurable_formatter"
kwds = { type = "vizro", requirements = "vizro==0.1.49" }
```

#### Option 3: MkDocs Plugin Wrapper

Add a simple plugin interface alongside the extension:

```python
class PyCafePlugin(BasePlugin):
    config_scheme = (
        ('type', Type(str, default='solara')),
        ('requirements', Type(str, default='')),
    )
    
    def on_config(self, config):
        # Set up the custom_fences with properly instantiated formatter
        # based on plugin config
```

This would allow:
```toml
[project.plugins.pycafe]
type = "vizro"
requirements = "vizro==0.1.49"
```

### Our Preference

**Option 1** (pre-instantiated formatters) would be the simplest and most flexible:
- Minimal code changes
- Backward compatible
- Works with both YAML and TOML
- Users can still use the factory pattern for custom configs
- Easy to add new presets

### Impact

This would benefit not just our project, but anyone using:
- Zensical (growing MkDocs/TOML tool)
- Any other TOML-based MkDocs configurations
- Users who want simpler configuration

We'd be happy to contribute a PR if you're open to one of these approaches!

**Project context:**
- Vizro: https://github.com/mckinsey/vizro
- Our use case: https://vizro.readthedocs.io/ (currently using YAML)
- mkdocs-pycafe: https://mkdocs.py.cafe/

Thanks for considering this! 
---

## Why This Approach Could Work

1. **PyCafe controls their API** - They can add these helpers without breaking changes
2. **Option 1 is minimal** - Just adding some pre-instantiated formatters in a new module
3. **Backward compatible** - Doesn't affect existing YAML users
4. **Benefits everyone** - Not just Vizro, but all TOML users
5. **They're approachable** - As you mentioned, they're more responsive than Zensical
</s>
</summary>
</details>

vizro-ai/mkdocs.yml
```diff
@@ -121,11 +131,10 @@ hooks:
...
-validation:
-  omitted_files: warn
-  absolute_links: warn
-  unrecognized_links: warn
-  anchors: warn
+#validation:
+#  omitted_files: warn
+#  absolute_links: warn
+#  unrecognized_links: warn
+#  anchors: warn
```
petar-qb: What's happening here?

vizro-ai/hatch.toml
```diff
@@ -75,14 +74,14 @@ detached = true
 installer = "uv"
 
 [envs.docs.scripts]
-build = "mkdocs build --strict"
```
petar-qb, line 78: Do we lose much by omitting the `--strict` flag? It's used in `mkdocs` to turn warnings into errors.

vizro-ai/mkdocs.yml
huong-li-nguyen: Shall we remove if not needed?

vizro-core/mkdocs.yml
```diff
@@ -126,126 +148,119 @@ markdown_extensions:
...
-  omitted_files: warn
-  absolute_links: warn
-  unrecognized_links: warn
-  anchors: warn
+  - /javascripts/extra.js
+# validation:
+# omitted_files: warn
+# absolute_links: warn
+# unrecognized_links: warn
+# anchors: warn
```
huong-li-nguyen: Same comment here

vizro-core/hatch.toml
```diff
@@ -132,15 +129,15 @@ installer = "uv"
...
-build = "mkdocs build --strict"
+build = "zensical build --clean"
 # Disable warnings on the linkcheck so that HTTP redirects are accepted. We could ignore only that warning and specify
 # more advanced settings using a linkcheckerrc config file.
 link-check = "linkchecker site --check-extern --no-warnings --ignore=404.html --ignore-url=127.0.0.1 --ignore-url=https://vizro.readthedocs.io/ --ignore-url=https://platform.ploomber.io/register --ignore-url=https://huggingface.co/join --ignore-url=https://www.ag-grid.com/"
 # Check if link to QB logo is not deleted from the docs.
 logo-check = "grep -q 'logo_watermark_extended.svg' docs/pages/user-guides/install.md"
 pip = '"{env:HATCH_UV}" pip {args}'
 pycafe-links = "../tools/pycafe/docs_links_filtering.sh"
-serve = "mkdocs serve --open {args}"
```
petar-qb, line 143: Should we keep the `{args}` (same for other occurrences)?

vizro-core/zensical.toml.backup
```diff
@@ -0,0 +1,306 @@
+[project]
+docs_dir = "docs"
+extra_css = ["/stylesheets/extra.css"]
+extra_javascript = ["/javascripts/extra.js"]
+# Make sure any new page added here is also added to the llmstxt plugin below (around line 198)
```
huong-li-nguyen:
```suggestion
# Make sure any new page added here is also added to the llmstxt plugin below (around line 170)
```

vizro-core/zensical.toml.backup
```diff
@@ -0,0 +1,306 @@
...
+
+[project.markdown_extensions.admonition]
+
+[project.markdown_extensions.attr_list]
+
+[project.markdown_extensions.footnotes]
+
+[project.markdown_extensions.md_in_html]
+
+[project.markdown_extensions.pymdownx.arithmatex]
```
huong-li-nguyen, line 110: I find this file a bit hard to read, probably just the format of it 😄 Does it help if we tidy it a bit? Remove everything we don't need?

E.g. these don't do anything right?
```
[project.markdown_extensions.pymdownx.inlinehilite]

[project.markdown_extensions.pymdownx.keys]

[project.markdown_extensions.pymdownx.mark]

[project.markdown_extensions.pymdownx.snippets]

[project.markdown_extensions.pymdownx.superfences]
```

vizro-mcp/hatch.toml
```diff
@@ -40,8 +40,7 @@ test-unit-coverage = [
 
 [envs.docs]
 dependencies = [
-  "mkdocs>=1.6",
-  "mkdocs-material",
+  "zensical",
   "mkdocs-git-revision-date-localized-plugin>=1.2.5",
   "mkdocstrings[python]",
```
petar-qb, line 45: There are still 83 occurrences of “mkdocs” in the Vizro project. We should review all of them and determine whether they can be removed.

Image

vizro-core/mkdocs.yml
```diff
@@ -71,21 +80,20 @@ nav:
...
+  language: en
+  variant: classic
+  favicon: assets/images/favicon.png
   font:
     text: Google sans
     code: Source Code Pro
   icon:
     repo: fontawesome/brands/github
   features:
+    - header.autohide
```
petar-qb, line 104: I like that the `header.autohide` is turned on. However, it would be even better to me if the left-side and right-side "navigation" follow the header when it disappears. Currently, the navigation stays in place when the header disappears.

https://github.com/user-attachments/assets/ea9e6aea-336c-4379-8b81-2ac436ff195d

vizro-core/mkdocs.yml
petar-qb: Should we remove this here and in other places?

vizro-core/zensical.toml.backup
stichbury: They do something :)  They switch on the extensions, which don't have any additional settings, for zensical to pick up.

We aren't using the toml file at present as we're waiting on a release of zensical with a fix in it for superfences. But I shall remove `mkdocs.yml` and switch on the `toml` build just as soon as zensical 0.0.12 is out and pycafe linking is confirmed to work.

vizro-core/mkdocs.yml
stichbury: I feel we should leave these as Zensical have confirmed they will implement validation in the near future, so we should be able to re-employ the checking soon.

vizro-core/zensical.toml.backup
petar-qb, line 1: Is this what's created automatically from the `mkdocs.yml`?

vizro-core/mkdocs.yml
stichbury: It's also a visual reminder that the checking currently isn't in place, so we need to be careful not to break internal links.

vizro-ai/mkdocs.yml
stichbury: See note below.

petar-qb (approved):
First of all, I’m impressed by how fast the documentation is built with Zensical!! 🚀

@stichbury, huge thanks for such a quick and smooth migration to Zensical. Really impressive work 🥇 ❤️ 

I’m not an expert when it comes to documentation tooling and yml/toml files behind it, so this is a humble, high-level review. That said, the docs definitely look and feel better, especially the search experience, and the build times are noticeably faster.

Zensical is still quite new, but it already shows a lot of promise and will likely evolve faster than our previous mkdocs setup.

vizro-ai/mkdocs.yml
petar-qb: Ah, I see you already answered this in another comment.

vizro-core/mkdocs.yml
petar-qb: Duplicate

vizro-ai/hatch.toml
```diff
@@ -75,14 +74,14 @@ detached = true
...
+build = "zensical build --clean"
 # Disable warnings on the linkcheck so that HTTP redirects are accepted. We could ignore only that warning and specify
 # more advanced settings using a linkcheckerrc config file. openai.com doesn't seem to work well with linkchecking,
 # throwing 403 errors, but these are not real errors.
 link-check = "linkchecker site --check-extern --no-warnings --ignore=404.html --ignore-url=127.0.0.1 --ignore-url=https://vizro.readthedocs.io/ --ignore-url=https://platform.openai.com/docs/models --ignore-url=openai.com --ignore-url=https://openai.com/"
 pip = '"{env:HATCH_UV}" pip {args}'
 pycafe-links = "../tools/pycafe/docs_links_filtering.sh"
-serve = "mkdocs serve --open"
+serve = "zensical serve --open"
```
maxschulz-COL: I naively tried to run this locally, and it gives me the following error:

```
hatch run docs:serve
Serving /Users/Maximilian_Schulz/Documents_no_backup/Python/Vizro/vizro/vizro-core/site on http://localhost:8000
Build started
Error: AttributeError: Failed to initiate extension 'mkdocstrings': module 'mkdocstrings' has no attribute 'makeExtension
```

vizro-core/mkdocs.yml
maxschulz-COL: If we keep it lets keeep it with a comment

vizro-core/hatch.toml
```diff
@@ -132,15 +129,15 @@ installer = "uv"
...
+serve = "zensical serve --open"
```
maxschulz-COL: I naively tried to run this locally, and it gives me the following error:

```
hatch run docs:serve
Serving /Users/Maximilian_Schulz/Documents_no_backup/Python/Vizro/vizro/vizro-core/site on http://localhost:8000
Build started
Error: AttributeError: Failed to initiate extension 'mkdocstrings': module 'mkdocstrings' has no attribute 'makeExtension
```

vizro-core/mkdocs.yml
maxschulz-COL, line 1: Many of the settings here are shuffled around, why is that? Makes it a little harder to review to see what is new and what just changed place

vizro-core/zensical.toml.backup
maxschulz-COL: Can we remove it? Why do the other two docs don't have anything related

vizro-ai/hatch.toml
stichbury: It's really important and we are losing it for sure, but unfortunately it's not supported by zensical right now. IT is on their backlog though.

vizro-core/hatch.toml
stichbury: You will be able to run locally but still need the mkdocstrings package as a dependency.

vizro-core/mkdocs.yml
stichbury: Yeah, I'm not sure. I think it's OK (personally) as it means you can navigate away without scrolling, but it does mean it looks a bit strange. This is really intended to help on small devices rather than desktops, so we could ask for a tweak that it doesn't apply except for smaller screens (or customise the CSS ourselves perhaps). WDYT?

vizro-core/mkdocs.yml
stichbury: OK, done.

vizro-mcp/hatch.toml
stichbury: There are still many dependencies on mkdocs because zensical is basically still reliant on a lot of the extensions and plugins. I will check them all but suspect most/many are necessary at present.

vizro-core/zensical.toml.backup
stichbury: No, it took me ages to port it across. I'm now far better at TOML and using Cursor to debug than I was at the start of the week :)

stichbury:
> I definitely see some issues in the rendered docs (atm checking on RtD) - see below:
> 
> image

@maxschulz-COL My friend, Cursor, and I have worked on this. This is the analysis:

> Zensical doesn't currently support the autorefs plugin, which is part of the mkdocstrings ecosystem in MkDocs. This plugin automatically resolves reference-style links like [vizro.models.Graph] to the correct API documentation URLs.
Since Zensical is still in early development (version 0.0.11), it hasn't yet implemented support for all MkDocs plugins, including autorefs.

I have prevailed on Cursor to make content changes across the docs to accommodate this issue, and they are all in a single commit.

[See this page rebuilt now for example](https://vizro--1522.org.readthedocs.build/en/1522/pages/user-guides/graph/)

vizro-core/hatch.toml
maxschulz-COL: What does that mean? Do I need to install it manually? Should this not work with the hatch command?

vizro-core/zensical.toml.backup
maxschulz-COL: But what is the purpose of this file? Why do the other two docs do not not have it?

vizro-core/docs/pages/user-guides/graph-table-actions.md
maxschulz-COL, line 1: I found some more occurences. These originate from the docstrings itself. This was found by just clicking around..

image

vizro-ai/mkdocs.yml
```diff
@@ -122,8 +132,7 @@ hooks:
 extra_css:
   - stylesheets/extra.css
 
-# Strictest settings possible, and will be elevated to ERROR when run with --strict.
-# See https://www.mkdocs.org/user-guide/configuration/#validation.
+# Keep these for now as strict building will be enabled by Zensical soon
```
maxschulz-COL, line 135: Should we not also keep the old comment?

stichbury: I don't think we need it -- we aren't explaining all the other settings in the file and some of those are far less evident from their naming.

stichbury:
Setting this PR aside for now.

Note that I reverted this commit ([e6078819773f90d740f0172d37668669eb0d646b](https://github.com/mckinsey/vizro/pull/1522/changes/9fedb6bb9cdac05c8683f3e30fc0ba0d22f2d57e)) to deal with the fact that autorefs is not supported. It means that API refdocs aren't correctly linked, but we should ideally hold out for that support to be added before we adopt Zensical.

The other ideals would be:

* Python 3.14 support
* `kwds` argument support for superfences so that we can use the .toml file to specify our docs build and still enable PyCafe linking
* strict building and other validation

vizro-core/hatch.toml
stichbury: This was down to a Python versioning issue.

PR #1537, opened 2026-01-08
[Bug] Fix meaning of control vs control action

petar-qb:
closes https://github.com/McK-Internal/vizro-internal/issues/2421


## Description
This PR draws a clear line between the **`vm.Filter` / `vm.Parameter` models** and the **`filter_action` / `parameter_action` actions**. Today these are conflated, which confuses both dashboard creators and even us developers. 

**The distinction:**
  
  - `vm.Filter` / `vm.Parameter` are **models**. Their job is to validate config, set defaults (`_filter_action` and `_parameter_action` too), and declare that a selector's value must be applied whenever its targets are refreshed, by *any* action (`on_page_load`, another `filter_action` / `parameter_action`), or that selector's value must be applied when targets' data is exported.                                                                                                                                                    
  - `filter_action` / `parameter_action` are **actions**. They do one thing: **refresh** the declared targets. So, they are identical to `on_page_load` and `filter_interactions`.

                                                                                                                                                                                                               
  **The bug this fixes:** on `main`, overwriting `vm.Filter.selector.actions` with for example some custom action silently stops the filter from filtering (which is wrong in my opinion). A `vm.Filter` represents filtering logic and must always be applied when its targets are refreshed (or its targets data is exported), regardless of what its selector's actions do.
                                                                                                                                                                                                                 
  **The corrected behaviour:** `vm.Filter.selector.actions` defines only what happens when the selector's default trigger (`value`) changes. So, exactly like any other model. The default is to refresh the targeted figures, but this is just a default. Overwriting it with the custom actions should run and in this PR's version it really runs the custom behaviour instead. The filtering itself is *always* applied when the targets are refreshed.

This also unlocks lazy/deferred application: a user can define `Filter` / `Parameter` with empty selector actions so that changing the selector does **not** refresh the targets on its own, yet the value is still honoured when the targets are refreshed by something else (e.g. a button click). This is what the Kedro pipeline example needs using native Vizro features, and it lays groundwork for a future `vm.Form` component.

Finally, this PR prepares the ground for replacing the private `_filter_action`, `_parameter_action`, `_on_page_load` with the new public `update_figures` API (see #1538), which becomes the shared default action for `Filter`, `Parameter`, and `Page` (`on_page_load`).

## Screen recording from the `main`:

❌ `vm.Filter` is NOT taken into account when `p11_graph` is updated from `on_page_load` or `parameter_action`. 

https://github.com/user-attachments/assets/18d9faac-4770-454a-9fc7-68a8e055918e

## Screen recording from the `bug/fix_meaning_of_control_vs_control_action`:

✅ `vm.Filter` is taken into account when `p11_graph` is updated from `on_page_load` or `parameter_action`. 

https://github.com/user-attachments/assets/1ef95fa2-b3d5-46d6-9494-445c0356c3db


## Notice


vizro-core/changelog.d/20260108_150147_petar_pejovic_fix_meaning_of_control_vs_control_action.md
```diff
@@ -0,0 +1,46 @@
...
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+### Fixed
+
+- Fix that `Filter` and `Parameter` logic is always applied on target figure refresh, even if custom selector actions are defined. ([#1537](https://github.com/mckinsey/vizro/pull/1537))
```
huong-li-nguyen, line 39:
```suggestion
- Custom actions for `Filters` and `Parameters` now run in addition to the default behavior, rather than replacing it. Filtering and parameter updates will always be applied as expected. ([#1537](https://github.com/mckinsey/vizro/pull/1537))
```

vizro-core/examples/scratch_dev/app.py
```diff
@@ -1,170 +1,72 @@
...
-                ),
-            ],
+        vm.Graph(
+            id="p11_graph",
+            figure=px.scatter(
+                df, x="sepal_width", y="sepal_length", color="species", color_discrete_map=SPECIES_COLORS
+            ),
         ),
+        vm.Text(id="p11_text", text="Placeholder"),
+    ],
```
huong-li-nguyen, line 53: Do we need to add some tests for this?

huong-li-nguyen (commented):
I might be missing a bit of context, but based on the PR description I thought the behaviour is now: Custom action will be appended to the default action of filtering for the Filter for example. 

Based on that understanding, i would have thought that the Filter would apply automatically with the custom action added on top. However, in the scratch app, the filtering is only applied after the parameter action is also applied, so it's not the behaviour I would have expected 🤔 

Is this intended? What I would have expected is that the filter is ALWAYS filtering and then adding any custom action on top. So in the scratch dev app, I would have expected it to filter regardless whether i click on the parameter action or not.

antonymilne:
Just a note to say that I've seen this PR and https://github.com/mckinsey/vizro/pull/1538. I'm going to think through both of them properly next week as well as finally go back to my notes to find the answers to these questions you asked:
* could you check is there a reason you haven't made the filter_action public?
* could you find your notes about syncing/cascading controls?

I spent a looooong time last year thinking about about implementing something very similar to #1538 and ultimately decided not to. Possibly that was the wrong decision, but I'd like to go back and find my earlier thoughts on it to understand what I was thinking before anyway!

vizro-core/changelog.d/20260108_150147_petar_pejovic_fix_meaning_of_control_vs_control_action.md
antonymilne: This was resolved but not applied in the pushed version anyway.

antonymilne (approved):

PR #1538, opened 2026-01-13
[Feat] Unify private `OPL`/`_filter`/`_parameter` actions under public `update_targets` action

petar-qb:
closes https://github.com/McK-Internal/vizro-internal/issues/2320
closes https://github.com/McK-Internal/vizro-internal/issues/2055
closes https://github.com/McK-Internal/vizro-internal/issues/1602
closes https://github.com/McK-Internal/vizro-internal/issues/1601

### About previous PR
This is the natural follow-up to #1537. That PR established that `Filter` / `Parameter` values must be applied whenever their targets are refreshed, independent of the selector's action.

### About this PR
This PR provides the mechanism to do that refreshing explicitly and on the user's demand by defining a single public action, `update_targets`.

`update_targets` unifies the private `on_page_load`, `_filter_action`, and `_parameter_action` into one public interface for refreshing figures, and becomes the shared default action for `Filter`, `Parameter`, and `Page` (OPL). (`_on_page_load` is kept only as a thin subclass of `update_targets` so that a page (re)load still writes its header to the DevTools action log; it adds no other behaviour.)

Exposing it as public API enables users to:

- **Reload (arbitrary) figures on a button click, or as a pre-/post-action of any other action.** This is what the Kedro pipeline example needs if we want to implement it only by utilising **native** Vizro features (this is what motivated exposing the API here and reusing it internally rather than keeping it private). So, in the Kedro example there's a use case where a user selects a single `AgGrid` row (one row = one scenario), and then reloads the charts below on the same page by clicking the `"Load scenario"` button.

- **Refresh controls (dynamic filters) alone.** A dynamic filter is a valid target too, so `update_targets(targets=["my_dynamic_filter"])` recomputes just that filter's `options`/`min/max` from the current data, without redrawing any figure. This is useful when something changes the underlying data (e.g. a parameter that resizes/reshapes the data or realtime data changes) and you want the dependent filters' available options to stay in sync on demand, while deferring the more expensive figure redraw until the user explicitly applies. P.S. A bare `update_targets()` (no targets) refreshes the whole page: every figure plus every dynamic filter, mirroring an on-page-load / page reload.

## Screen recording:

https://github.com/user-attachments/assets/c22d6985-3a10-43e1-9dba-72d3c6923417

## TODO:
- [x] - Docs

## Notice


petar-qb:
@lingyielia Is this PR solving the problem you mentioned [here](https://github.com/McK-Internal/vizro-internal/issues/2055#issuecomment-3228815953)?

vizro-core/changelog.d/20260803_120000_petar_pejovic_update_targets_action.md
```diff
@@ -0,0 +1,46 @@
...
+-->
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+### Added
+
+- Add the public `update_targets` action (`vizro.actions.update_targets`) to refresh a page's figures on demand, for example on a `Button` click or as part of another action's chain. ([#1538](https://github.com/mckinsey/vizro/pull/1538))
```
antonymilne: You need to change anything here, but just to check my understanding:
* this can update than just `figures` because it includes `dynamic_filter_ids_on_page`
* so the name `update_targets` is technically correct here rather than just `update_figures`
* in the changelog it could say "figures and dynamic filters" rather than just figures so it's technically a bit incomplete as a description

If my understand is correct, this is totally fine - we don't want to write out "figures and control targets" every time given that 95% of the figures are the ones that matter. It's fine to just say "figures". Maybe if you prefer the name `update_figures` it's fine to call it that also just as a kind of abbreviation. I don't mind though, `update_targets` is also fine.

vizro-core/src/vizro/models/_controls/filter.py
```diff
@@ -584,15 +585,14 @@ def pre_build(self):  # noqa: PLR0912
...
-                    id=f"{FILTER_ACTION_PREFIX}_{self.id}",
-                    column=self._filter_column,
-                    filter_function=self._filter_function,
-                    targets=self.targets,
-                ),
-            ]
+        # The default action refreshes the filter's targets. The filtering logic itself (self._filter_function applied
+        # to self._filter_column) is reapplied whenever the targets are refreshed, independently of this action, so
+        # overwriting selector.actions changes what happens on selector change without disabling the filtering.
+        # We check model_fields_set (not falsiness) so that an explicit `selector=X(actions=[])` is honored as "do
```
antonymilne: Can we set `selector=X(actions=None)`? Just feels a bit more natural than `actions=[]` given that outside an actions chain you can now just specify a singleton `actions=action`. Possibly we should even go so far as to ban `[]` and just use `None`?

Don't change if you think it's a bad idea or makes the validation much more complicated though.

antonymilne (approved):

vizro-core/changelog.d/20260803_120000_petar_pejovic_update_targets_action.md
petar-qb: > You need to change anything here, but just to check my understanding:
> * this can update than just `figures` because it includes `dynamic_filter_ids_on_page`

Correct. So:
```py
        vm.Button(text="Refresh Filter", actions=update_targets(targets=["p41_filter"])),  # updates only filter
        vm.Button(text="Refresh Graph", actions=update_targets(targets=["p41_graph"])),  # updates only graph
        vm.Button(text="Refresh everything on the page", actions=update_targets()),  # updates filter, graph and every other figure or dynamic filter on the page
```

> * so the name `update_targets` is technically correct here rather than just `update_figures`

Yep, that's why I renamed it.

> * in the changelog it could say "figures and dynamic filters" rather than just figures so it's technically a bit incomplete as a description

Yes, good point

>If my understand is correct, this is totally fine - we don't want to write out "figures and control targets" every time given that 95% of the figures are the ones that matter. It's fine to just say "figures". Maybe if you prefer the name `update_figures` it's fine to call it that also just as a kind of abbreviation. I don't mind though, `update_targets` is also fine.

I slightly prefer `update_targets`, so let's keep like this as your don't have preference 😃

vizro-core/src/vizro/models/_controls/filter.py
petar-qb: I agree. I’ve enabled `selector=X(actions=None)` and documented it as the preferred way to express this. I’ve silently kept `actions=[]` working as well to preserve backwards compatibility, and because it’s also a fairly natural thing for users to try.

PR #1539, opened 2026-01-13
[Bug] Fix defaulting of the AgGrid `rowSelection` if actions defined

petar-qb:
closes https://github.com/McK-Internal/vizro-internal/issues/2423

## Description
@antonymilne 

Thanks for spotting the issue! I missed an obvious bug here..

I solve the problem successfully with the first approach you mentioned in [this comment](https://github.com/mckinsey/vizro/pull/1515#issuecomment-3739732893). I moved the default `rowSelection` configuration from the `dash_ag_grid` to `AgGrid.__call__` and only applied when at least one action is defined.

As a side effect (and an intended improvement), when no actions present, `rowSelection` will no longer set by default, meaning row selection is disabled in the UI, which matches the expected behaviour. 👍

Question: Should we turn on `clickmode=("event+select")` in a similar way only if the Graph has `set_control` defined? 

## Notice


vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -146,8 +146,30 @@ def __call__(self, **kwargs):
         if "data_frame" not in kwargs:
             kwargs["data_frame"] = data_manager[self["data_frame"]].load()
 
-        # Enable checkboxes in the AgGrid if any of the actions is a `set_control` action.
-        figure = self.figure(_set_checkboxes=any(isinstance(action, set_control) for action in self.actions), **kwargs)
+        figure = self.figure(**kwargs)
+
+        # Ensure "dashGridOptions" property exists as it's needed for setting "rowSelection".
+        if not hasattr(figure, "dashGridOptions"):
```
antonymilne: When could this happen? Is it whenever the user doesn't explicitly specify `dashGridOptions`?

vizro-core/src/vizro/models/_components/ag_grid.py
```diff
@@ -146,8 +146,30 @@ def __call__(self, **kwargs):
...
+        # Set the AgGrid "rowSelection" if `filter_interaction` action is used and if it's not already set.
+        # The "rowSelection" has to be set so that the AgGrid's actions can be triggered.
+        # TODO: Delete setting the "rowSelection" for filter_interaction after
+        #  https://github.com/McK-Internal/vizro-internal/issues/2402
+        if any(isinstance(action, filter_interaction) for action in self.actions):
+            row_sel = figure.dashGridOptions.setdefault("rowSelection", {})
+            row_sel.setdefault("mode", "singleRow")
+            row_sel.setdefault("enableClickSelection", True)
+            row_sel.setdefault("checkboxes", False)
+            row_sel.setdefault("headerCheckbox", False)
```
antonymilne: Given `filter_interaction` is already deprecated I would just not bother with this at all. 

```suggestion
```

antonymilne:
FYI I just fixed a small inaccuracy in the old changelog: [7cd8e5e](https://github.com/mckinsey/vizro/pull/1539/commits/7cd8e5eb3a7de82db090636224a34dde4baffdb8). As far as I understand, there was nothing specific to `plotly.express` about the change.

vizro-core/src/vizro/models/_components/ag_grid.py
petar-qb: I thought that too, but it’s obviously not true (as the scratch example shows) 😄 Thanks for pointing it out. I’ve removed it.

vizro-core/src/vizro/models/_components/ag_grid.py
petar-qb: I simplified the code. As a result:
1. `rowSelection` is not set when actions are not defined, so rows can’t be selected from the UI which feels correct.
2. If actions are defined, `rowSelection` and its values are defaulted. These defaults are intended to improve UX when `set_control` is used, but they can be overridden if users don’t want, for example, checkboxes for an attached action. Custom attached actions still work, and `filter_interaction` continues to work as well.

antonymilne:
* I think we already need a different `multi` mode that's distinct from `single`. Options are same as in https://github.com/McK-Internal/vizro-internal/issues/2402#issuecomment-3750881701 but I rule out option 1. Option 3 maybe possible by adding `is_multi` property/function in `set_control` (but then how would it work for custom action?). Would need to get `prebuild` order right also. Probably option 2 is best given we'll need to do it for pivot table in future also - automatic detection seems more complex than it's worth doing.
* single behaviour should probably just be same as AgGrid default so keep `enableClickSelection=False`
* multi behaviour should probably just be same as AgGrid so keep `enableClickSelection=False` especially given bug/annoying feature here
* even with option 2 can still set some mode automatically as soon as there's some action defined (not sure whether it should use single or multi behaviour [as now - probably good since matches Dropdown default], maybe whatever matches graph)

Separate:
* should we remove select all tick in header? how should this work? It's currently confusing but probably no better solution other than removing it.

Question: 
* would we do similarly for graph and have separate single/multi-select modes? Not sure but @petar-qb current idea seems like a good start to define `clickmode` only when action is defined.
* should multi argument go in `AgGrid/Graph` itself rather than in figure function?

Alternative: just leave as it is now.

vizro-core/changelog.d/20260113_143142_petar_pejovic_fix_custom_ag_grid_rendering.md
```diff
@@ -0,0 +1,46 @@
...
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+### Fixed
+
+- Fix AgGrid `rowSelection` defaults when `actions` argument is defined. ([#1539](https://github.com/mckinsey/vizro/pull/1539))
```
huong-li-nguyen: I think we could rephrase this to be more user centered, because from a user-perspective i wouldn't understand the implications of this from the sentence above 🤔  

Maybe something like: "Row selection checkboxes are now hidden by default in AgGrid, but can still be configured".

petar-qb:
Nice catch :+1:

Overall, I think the currently open PR is fine and doesn’t degrade the UX. Do you agree? If so, I’d suggest we merge it, and then open (and hopefully merge shortly after) a follow-up PR where `clickmode(E+S)` is set only when `Graph.actions` is defined.

I completely agree there’s room for improvement around the AgGrid/Graph `set_control` behaviour (especially after reading your comment from above). I’d start by addressing the `cellClicked` issue first. At that point, I’d revisit your (very valid) comment and try to cover all relevant use cases.

Until then, I don’t think it’s a big problem if we occasionally serve “incorrect” `dashGridOptions` defaults as they can always be overridden manually by the dashboard creator side..
After that PR, I still have a couple of follow-ups planned to further improve the set_control UX and code (there are probably more related tickets, but I haven’t dug them up yet :smile:):
- Allow all selectors to be outputs of set_control
- Streamline actions: triggers, inputs/outputs, and the extraction function
- (hard and probably not urgent) `set_controls` (multiple controls)

vizro-core/changelog.d/20260113_143142_petar_pejovic_fix_custom_ag_grid_rendering.md
antonymilne: Agreed, although this is a somewhat technical/niche bug fix rather than a user-facing feature.

Maybe this?

```suggestion
- Fix `rowSelection` checkboxes for custom AgGrid with `actions`. ([#1539](https://github.com/mckinsey/vizro/pull/1539))
```

antonymilne:
> Nice catch 👍
> 
> Overall, I think the currently open PR is fine and doesn’t degrade the UX. Do you agree? If so, I’d suggest we merge it, and then open (and hopefully merge shortly after) a follow-up PR where `clickmode(E+S)` is set only when `Graph.actions` is defined.
> 
> I completely agree there’s room for improvement around the AgGrid/Graph `set_control` behaviour (especially after reading your comment from above). I’d start by addressing the `cellClicked` issue first. At that point, I’d revisit your (very valid) comment and try to cover all relevant use cases.
> 
> Until then, I don’t think it’s a big problem if we occasionally serve “incorrect” `dashGridOptions` defaults as they can always be overridden manually by the dashboard creator side.. After that PR, I still have a couple of follow-ups planned to further improve the set_control UX and code (there are probably more related tickets, but I haven’t dug them up yet 😄):
> 
> * Allow all selectors to be outputs of set_control
> * Streamline actions: triggers, inputs/outputs, and the extraction function
> * (hard and probably not urgent) `set_controls` (multiple controls)

Definitely agree the current PR is an improvement compared to the current bug, so yes I think this is a very practical proposal 👍 I'm still not even sure myself whether the behaviour is actually ok now or whether we should still try to improve it further.

The reason I came up with this is because I wanted us to focus on the docs ticket https://github.com/McK-Internal/vizro-internal/issues/2392 next, and when I started looking at the docs again and thinking about how to rewrite them I immediately realised that the single-select behaviour is now pretty confusing. I don't think there's any point doing the docs ticket until we're happy with this, otherwise we'll have to repeatedly rewrite it and redo gifs etc. But also I'd like to try and get the docs done soon if possible.

So yes, I would suggest:
1. merge the current PR as you have it
2. discuss the `clickmode` change but probably it's a good one, then implement (no ticket for this, create one if you like, I don't mind)
3. handle `cellClicked` issue: https://github.com/McK-Internal/vizro-internal/issues/2402
4. discuss any existing problems we're aware of with AgGrid/Graph and try to discover if there's any more by thinking about how we'd document all this: https://github.com/McK-Internal/vizro-internal/issues/2424
5. documentation: https://github.com/McK-Internal/vizro-internal/issues/2392

petar-qb:
> The reason I came up with this is because I wanted us to focus on the docs ticket https://github.com/McK-Internal/vizro-internal/issues/2392 next, and when I started looking at the docs again and thinking about how to rewrite them I immediately realised that the single-select behaviour is now pretty confusing. I don't think there's any point doing the docs ticket until we're happy with this, otherwise we'll have to repeatedly rewrite it and redo gifs etc. But also I'd like to try and get the docs done soon if possible.

Very valid points @antonymilne! 
I have some ideas on how to overcome the `weird single-select behaviour`. Let me try to fix it quickly (as the part of this PR) If I don't succeed to fix it quickly, I'll merge this PR as it is now and just continue doing tickets one by one as you suggested in the comment above.

vizro-core/changelog.d/20260113_143142_petar_pejovic_fix_custom_ag_grid_rendering.md
petar-qb: Will rephrase it as Antony suggested here -> https://github.com/mckinsey/vizro/pull/1539#discussion_r2694266070

PR #1544, opened 2026-01-16
[Bug] Fix defaulting of the Graph `clickmode` and `modebar` if actions defined

petar-qb:
closes https://github.com/McK-Internal/vizro-internal/issues/2425

## Description

Similarly like defaulting `rowSelection` for the AgGrid only if actions defined ([PR](https://github.com/mckinsey/vizro/pull/1539)), we need to default `clickmode="event+select"` for Graph model only if actions defined. This improves the UX as wouldn't highlight clicked points/bars when no actions defined (which is misleading).

This PR also defaults `modebar_remove=["select2d", "lasso2d"]` if actions are not defined (removing "Select Box" and "Lasso" from chart's modebar).

## Notice

### DIFF
--- a/vizro-core/examples/scratch_dev/app.py
+++ b/vizro-core/examples/scratch_dev/app.py
@@ -84,6 +84,18 @@ page_graphs_with_overwritten_clickmode_and_modebar = vm.Page(
         vm.Tabs(
             tabs=[
                 vm.Container(
+                    title="Graph with actions",
+                    components=[
+                        vm.Graph(
+                            title="Expected: ✅clickmode='event+select' set + ✅Select/Lasso in modebar",
+                            figure=px.scatter(
+                                df, x="sepal_width", y="sepal_length", color="species", custom_data=["species"]
+                            ),
+                            actions=va.set_control(control=f"{pre}filter_1", value="species"),
+                        )
+                    ],
+                ),
+                vm.Container(
                     title="Graph with no actions",
                     components=[
                         vm.Graph(


vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -293,6 +292,15 @@ def _optimise_fig_layout_for_dashboard(fig):
...
             fig.update_layout(margin_t=64)
 
+        if self.actions and not fig.layout.clickmode:
+            # Set clickmode to "event+select" if there are actions defined but clickmode is not already set.
+            fig.update_layout(clickmode="event+select")
+
+        if not self.actions and not fig.layout.modebar.to_plotly_json():
+            # Remove selection tools from modebar if there are no actions defined and the modebar is not already set.
+            # TODO: Consider applying the same logic when Graph actions are set_control that target single selectors.
+            fig.update_layout(modebar_remove=["select2d", "lasso2d"])
```
antonymilne, line 301: I thought we'd need to do `config.modeBarButtonsToRemove` which is more awkward since it's out `fig`, so this is great that you found a way to do it inside `layout` 👍 

I have this TODO recorded in https://github.com/McK-Internal/vizro-internal/issues/2424. If we decide that there a good way to proceed with `set_control` different behaviour for `multi=True/False` for AgGrid then we could apply the same pattern here for Graph.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -293,6 +292,15 @@ def _optimise_fig_layout_for_dashboard(fig):
             # Reduce `margin_t` if not explicitly set.
             fig.update_layout(margin_t=64)
 
+        if self.actions and not fig.layout.clickmode:
+            # Set clickmode to "event+select" if there are actions defined but clickmode is not already set.
```
antonymilne:
```suggestion
            # Set clickmode to "event+select" if there are actions defined and clickmode is not already set.
```

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -293,6 +292,15 @@ def _optimise_fig_layout_for_dashboard(fig):
...
+            fig.update_layout(clickmode="event+select")
+
+        if not self.actions and not fig.layout.modebar.to_plotly_json():
```
antonymilne: Maybe we should be checking `fig.layout.modebar.remove` here if that's possible? And then possibly you don't need the `to_plotly_json`?

vizro-core/changelog.d/20260116_141244_petar_pejovic_fix_clickmode_when_actions.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
```
antonymilne:
```suggestion
- Simplify chart UI when graph is not an action trigger. ([#1544](https://github.com/mckinsey/vizro/pull/1544))
```

PR #1546, opened 2026-01-20
[Docs] Implement mkdocstring to automatically create pydantic docsstrings

maxschulz-COL:
Closes https://github.com/McK-Internal/vizro-internal/issues/1345

## Description
PR is now based on a custom `jinja` template, essentially overwriting a fair bit of what the `griffe-pydantic` extension does/should do.

For lack of time, most of this was originally LLM created, but the results look stable and good.

New in the PR
- custom template for table for fields, similar to normal docstring parameters
- switching of docstring parameter style to `table`, which is (in my eyes) a lot clearer

Some quirks:
- had to write custom code to remove fields from TOC on the right hand side
- had to write custom code to detect `default_factory` and the default of fields that have been excluded from the schema

## For review
Mainly check the API docs reference - models and actions have changed the most.

## Not soved
I could not get the table to highlight on hover for the pydantic fields, unlike the normal table for docstring parameters

## Notice


vizro-core/schemas/0.1.51.json
maxschulz-COL, line 1: Look how much nicer our schema has become :D

antonymilne:
Just a quick comment: "If fields do not have a description, they do not appear in attributes (maybe ok), and so they don't appear in TOC on the right" - this sounds good to me actually, since it prevents `type` from appearing presumably? I think it's fair enough to say that a field without a description shouldn't go in docs.

maxschulz-COL:
> Just a quick comment: "If fields do not have a description, they do not appear in attributes (maybe ok), and so they don't appear in TOC on the right" - this sounds good to me actually, since it prevents `type` from appearing presumably? I think it's fair enough to say that a field without a description shouldn't go in docs.

Hm maybe... I just don't like the mismatch with the TOC on the right hand side. If you scan that you could think that this attribute just doesn't exist. But I guess that's just taste. 

The type I got rid of completely with the the extension

vizro-core/mkdocs.yml
antonymilne: Does this still work? Doesn't it need the reference to change to include `DynamicallyInspect` now that there's multiple classes in there? Compare to line 243.

vizro-core/mkdocs.yml
```diff
@@ -57,6 +57,7 @@ nav:
       - Table functions: pages/API-reference/table-callables.md
       - Figure functions: pages/API-reference/figure-callables.md
       - Kedro integration: pages/API-reference/kedro-integration.md
+      - Constants: pages/API-reference/constants.md
```
antonymilne: Why do we want this to be public?

antonymilne (commented):
This is super exciting 😁 Just skimmed through and left a couple of random comments and will review properly tomorrow!

vizro-core/mkdocs.yml
maxschulz-COL: It is so that the defaults link through to an actual number. I selectively only included the constants that needed a hyperlink. All else is still hidden

vizro-core/mkdocs.yml
maxschulz-COL: Yes it still worked, but you are right, we can actually take out line 243!

vizro-core/mkdocs.yml
antonymilne: Hmmm yes I see. Would you be horribly upset if we just removed these constants and hardcoded them in all the relevant places? Having looked at where they're used, it's only in `Grid` and `Flex`. At the moment I don't much like the extra level of indirection this introduces into our docs where you have to click on a link to a different page just to understand that a value is 12px. So overall I would prefer to just remove the constants altogether and not need the constants page in our docs.

vizro-core/docs/griffe_extensions.py
```diff
@@ -33,3 +34,53 @@ def on_instance(self, *, obj: griffe.Object, **kwargs):  # noqa: D102
         logger.info("Dynamically inspecting '%s'", obj.path)
         inspected_module = griffe.inspect(obj.module.path, filepath=obj.filepath)
         obj.parent.set_member(obj.name, inspected_module[obj.name])
+
+
+class FilterPrivateAttrs(griffe.Extension):
```
antonymilne: Is it not possible to do this already just with [some of these settings](https://mkdocstrings.github.io/python/usage/configuration/members/)?

I would have thought that e.g. `filters = ["!^_", "^(?!type$)"]` in models.md would do the job?

maxschulz-COL: I didnt manage..

vizro-core/mkdocs.yml
maxschulz-COL: Done, it comes now as a byproduct of the new template/direct inspection anyway haha

vizro-core/docs/griffe_extensions.py
maxschulz-COL: It may be to do with the extension, at some point I gave up trying

vizro-core/pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
   "pydantic>=2.7.0",
   "flask_caching>=2",
   "wrapt>=1",
-  "black",
+  "black>=26.1.0",
```
antonymilne: I don't think we should change this unless you have a good reason to do so? https://github.com/mckinsey/vizro/pull/1502#issuecomment-3784982335

vizro-core/docs/griffe_extensions.py
```diff
@@ -1,8 +1,11 @@
 """Extensions we have written to help generate API reference docs."""
 
 import griffe
+from pydantic import BaseModel
+from pydantic.fields import FieldInfo, PydanticUndefined
 
 logger = griffe.get_logger("griffe_dynamically_inspect")
+pydantic_logger = griffe.get_logger("griffe_pydantic_docs")
```
antonymilne:
```suggestion
pydantic_logger = griffe.get_logger("griffe_pydantic_docs_cleaner")
```

vizro-core/mkdocs.yml
```diff
@@ -210,29 +210,41 @@ plugins:
...
           options:
             show_bases: false
             show_source: false #currently doesn't show source at correct level, and not for pydantic models
             docstring_style: google
             merge_init_into_class: true
-            docstring_section_style: list
+            docstring_section_style: table
             separate_signature: true
-            #   filters: ["!^_"]
+            unwrap_annotated: true # check effect
```
antonymilne: What is the effect?!

vizro-core/pyproject.toml
maxschulz-COL: Sure, just annoying for local setups because one needs to prune the environment. And it would mean that there can we two different outcomes to `_to_python` based on which version of black one has installed. Are you ok with that?

vizro-core/mkdocs.yml
maxschulz-COL: It extract the type from the  annotation. it works mostly well and I think its a good choice because otherwise we would find ourselves with the type field too crowded

vizro-core/pyproject.toml
maxschulz-COL: Done

maxschulz-COL:
> Ahh, so exciting! 🥳 So happy, if we don't need to keep these Args docstring in sync anymore. I didn't check the code, mostly how it renders. I love the new table format, a lot cleaner now. Before I approve just a few questions:
> 
> 1. The majority of these methods, we actually don't want to display. Only the "custom_header" is a public method, the others were not shown before, and I assume we want to keep it like that?

I added them specifically in :) They are the non-private validators. I think (while we were not super careful in determining what is private and what not there) that it is useful to users to read in docs how further (beyond the type) their entry gets validated.

Unfortunately the extension does not pick up on validators defined via `Annotated` - which is how we define most field validators. That is something we could also consider, either to move things out of `Annotated` OR to fix that bug ourselves

EDIT: Given that this seemed to cause confusion, I have removed the validators for now, but I think we might want to add them in the future

> 2. Does this "Show json schema" come for free or did we consciously decide to add it everywhere now? I don't know if it's decluttering too much, I don't know when users would like to see that?

Yes comes for free, so we can easily disable

> 3. One thing J. always asked about, can we display the `id` inside the arguments tables automatically, too? This wasn't possible before, so no worries if it's not possible, but if it's now easy to do, that would be nice to enable.

Good question, it does show on the `VizroBaseModel`, but not on the derived models... Let me check!

EDIT: Done

huong-li-nguyen (approved):

vizro-core/mkdocs.yml
```diff
@@ -210,29 +212,41 @@ plugins:
...
+      custom_templates: docs/templates
       handlers:
         python:
           options:
             show_bases: false
             show_source: false #currently doesn't show source at correct level, and not for pydantic models
             docstring_style: google
             merge_init_into_class: true
-            docstring_section_style: list
+            docstring_section_style: table
```
antonymilne: Did you also try `spacy`? Just judging by the mkdocstrings docs I would have thought that might be even for us since many of our field descriptions and type annotations are quite long.

vizro-core/docs/griffe_extensions.py
```diff
@@ -33,3 +36,180 @@ def on_instance(self, *, obj: griffe.Object, **kwargs):  # noqa: D102
...
+
+
+class PydanticDocsCleaner(griffe.Extension):
+    """Clean up Pydantic model documentation.
+
+    This extension:
+    1. Filters out private members (PrivateAttr, 'type' field, underscore-prefixed methods)
+    2. Filters out validators (after griffe_pydantic labels them)
+    3. Excludes pydantic fields from TOC while keeping them renderable in tables
+    """
```
antonymilne, line 48: Would be nice to explain what `pydantic_model.html.jinja` is for also and whether it's copy and pasted from something or based off something or just made from scratch.

vizro-core/src/vizro/models/_components/form/radio_items.py
```diff
@@ -61,21 +47,22 @@ class RadioItems(VizroBaseModel):
...
+            to see all available arguments. [Not part of the official Vizro schema](../explanation/schema.md) and the
+            underlying component may change in the future."""
 
     _dynamic: bool = PrivateAttr(False)
     _in_container: bool = PrivateAttr(False)
     _inner_component_properties: list[str] = PrivateAttr(dbc.RadioItems().available_properties)
 
     # Reused validators
-    _validate_options = model_validator(mode="before")(validate_options_dict)
+    # _validate_options = model_validator(mode="before")(validate_options_dict)
```
antonymilne: This doesn't look right...

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -123,7 +108,7 @@ class Dropdown(VizroBaseModel):
     _inner_component_properties: list[str] = PrivateAttr(dcc.Dropdown().available_properties)
 
     # Reused validators
-    _validate_options = model_validator(mode="before")(validate_options_dict)
+    # _validate_options = model_validator(mode="before")(validate_options_dict)
```
antonymilne:
```suggestion
    _validate_options = model_validator(mode="before")(validate_options_dict)
```

vizro-core/src/vizro/models/_components/button.py
```diff
@@ -58,32 +43,35 @@ class Button(VizroBaseModel):
...
                 default={},
-                description="""Extra keyword arguments that are passed to `dbc.Button` and overwrite any
-            defaults chosen by the Vizro team. This may have unexpected behavior.
-            Visit the [dbc documentation](https://www.dash-bootstrap-components.com/docs/components/button/)
-            to see all available arguments. [Not part of the official Vizro schema](../explanation/schema.md) and the
-            underlying component may change in the future. Defaults to `{}`.""",
             ),
         ]
     ]
+    # Using the description field does not show the below in rendered docs
```
antonymilne: Do you know why? Is it just too long or something?

vizro-core/src/vizro/models/_components/form/radio_items.py
maxschulz-COL: Thats really embarassing, I could have sworn i switched them all back... weird

maxschulz-COL: The real worry is, why did our tests not catch that!

vizro-core/mkdocs.yml
maxschulz-COL: I initially didnt like it, but a second look has changed my mind.I have now implemented an adapted version of this for the pydantic fields. It really does make things more legible

vizro-core/src/vizro/models/_components/form/radio_items.py
maxschulz-COL: Done

maxschulz-COL:
> 1. Some validators are still commented out - please do search for these since I didn't check every file carefully so there might be others I didn't comment on.

Done

> 2. In `vizro.models.types`, the docstrings for `OutputsType` and `ActionsType` shouldn't say "defaults to []" (not caused by your changes) - please remove this.

I left it in because those type do not show with a default, what do you think?

> 3. Can we always show inherited fields? You've done this manually for `id` but the fields inherited in [`update_notification`](https://vizro--1546.org.readthedocs.build/en/1546/pages/API-reference/actions/#vizro.actions.update_notification) aren't shown. Maybe it's possible to change the logic just to include all fields from any superclass that's `VizroBaseModel`. Alternatively we could just [link to the base class](https://mkdocstrings.github.io/python/usage/configuration/general/#show_bases).

Ok let me investigate

antonymilne:
> I left it in because those type do not show with a default, what do you think?

Isn't the default given where the field is used, e.g. https://vizro--1546.org.readthedocs.build/en/1546/pages/API-reference/models/#vizro.models.Action shows the default is `[]`? I don't see how the type can have a default value associated with it; it's the field that has a default, no?

vizro-core/src/vizro/models/_components/button.py
maxschulz-COL: Fixed that now in the template

PR #1552, opened 2026-01-22
[Bug] Fix AgGrid styling

huong-li-nguyen:
## Description
Closes https://github.com/McK-Internal/vizro-internal/issues/2441

The new dash-ag-grid version seems to have introduced new AgGrid variables with new assignments. This PR fixes the major hickups, but I would recommend to do a complete review of the new AgGrid CSS variables and eventually start from scratch to make sure we don't have any redundant / old CSS that's not needed.

- Remove AgGrid header background color 
- Assign missing CSS variables

TO DO in other PRs:
- Calendar styling still looks a bit off, but can be done in a separate PR
- Minor edits to icon colors
- Fix screenshots with error message of WebGL @l0uden 

## Screenshot
Screenshot 2026-01-22 at 17 39 21
Screenshot 2026-01-22 at 17 39 16

## Notice

### DIFF
--- a/vizro-core/src/vizro/static/css/aggrid.css
+++ b/vizro-core/src/vizro/static/css/aggrid.css
@@ -5,7 +5,7 @@
   --ag-background-color: transparent;
   --ag-odd-row-background-color: transparent;
   --ag-data-color: var(--bs-heading-color);
-  --ag-header-background-color: transparent;
+  --ag-header-background-color: var(--bs-body-bg);
   --ag-icon-font-family: aggridquartz;
   --ag-icon-size: 1rem;
   --ag-row-height: 48px;


vizro-core/pyproject.toml
```diff
@@ -17,7 +17,7 @@ classifiers = [
 dependencies = [
   "dash>=3.4.0,<4",  # temporary upper bound to avoid breaking changes in Dash v4
   "dash_bootstrap_components>=2",  # 2.0.0 needed to support dash>=3.0.0
-  "dash-ag-grid>=31.3.1,<33",  # 31.3.1 needed to support dash>=3.0.0; temporary upper bound until we fix https://github.com/McK-Internal/vizro-internal/issues/2441
+  "dash-ag-grid>=31.3.1",  # needed to support dash>=3.0.0;
```
petar-qb, line 20: Is it still working properly for the `dash-ag-grid==31.3.1` and other versions? Could you test it and adjust the lower bound if needed?

huong-li-nguyen: Yeees, tested it with the current lower bound and it still works fine, because I didn't remove any of our old CSS 👍

vizro-core/src/vizro/static/css/aggrid.css
```diff
@@ -5,9 +5,9 @@
   --ag-background-color: transparent;
   --ag-odd-row-background-color: transparent;
   --ag-data-color: var(--bs-heading-color);
-  --ag-header-background-color: var(--right-side-bg-iterated);
+  --ag-header-background-color: transparent;
```
nadijagraca: The background color was updated when we tried to implement sticky headers. I left this change intentionally in the PR to support a smoother process for users applying sticky headers via custom css. If the background were reverted to transparent, users would also need to explicitly define a background color for the header to avoid visual issues caused by transparent background of the header.

huong-li-nguyen: Ah good to know, but then let me apply the actual bg color of the right side, because it should be white in the light theme! :)

PR #1555, opened 2026-01-23
[QA] Fix webgl problem

l0uden:
## Description

Fixed problem with some charts were not loaded in `headless` mode in Chrome and Firefox

## Notice

### DIFF
--- a/vizro-core/changelog.d/20260123_140211_runner.md
+++ b/vizro-core/changelog.d/20260123_140211_runner.md
@@ -1,0 +1,48 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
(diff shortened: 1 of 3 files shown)


l0uden:
> Why not add the `svg` argument to the other graphs across the tests?
> 
> I’m not convinced that setting `render_mode="svg"` will deterministically solve the issue (P.S: I haven’t dug deeply into the root cause or the impact of this change).
> 
> That said, I think we should merge it and let real usage tell us whether this fully addresses the problem or not.

I've found this solution [on plotly forums](https://community.plotly.com/t/webgl-is-not-supported-even-though-its-enabled/14855/11). I've applied it only to problematic charts, hope it will sufficient, because other chart don't have such a problem.

PR #1558, opened 2026-01-26
[QA] Set control multi select tests

l0uden:
## Description

Closes https://github.com/McK-Internal/vizro-internal/issues/2403

## Notice

### DIFF
--- a/vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
+++ b/vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
@@ -1035,9 +1035,7 @@ def test_self_filtered_graph(dash_br):
     dash_br.click_at_coord_fractions(
         scatter_point_path(cnst.SCATTER_SET_CONTROL_SELF_FILTER, point_number=21, trace_index=1), 0, 0
     )
-    # after first click delay needed.
-    # sometimes without delay test stuck on the first click
-    # (maybe because graph is not fully loaded and test can't click on the appropriate point using coordinates)
+    # after first click delay is needed to ensure that the graph is loaded so its point can be clicked again.
     time.sleep(0.5)
     # this click does not apply "self-filtering" and it only highlights clicked point.
     dash_br.click_at_coord_fractions(


vizro-core/tests/e2e/vizro/dashboards/default/pages/set_control_multi_select_pages.py
```diff
@@ -0,0 +1,226 @@
...
+
+iris = px.data.iris()
+
+iris_unique_species = px.data.iris()
+unique_species = iris_unique_species["species"].unique()
+n_species = len(unique_species)
+n_rows = len(iris_unique_species)
+iris_unique_species["species"] = [unique_species[i % n_species] for i in range(n_rows)]
+
+gapminder = px.data.gapminder()
```
petar-qb: Is the `iris_unique_species` same as `iris` but where species go like: `"setosa" (row 1)` -> `"versicolor" (row 2)` -> `"virginica" (row 3)` -> `"setosa" (row 4)`... ? If so, here's a suggestion:

```suggestion
iris = px.data.iris()
unique_species = iris["species"].unique()

iris_species_cycle = iris.copy()
iris_species_cycle["species"] = np.resize(unique_species, len(iris_species_cycle))
```

P.S. I didn't check the code.

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -88,9 +98,7 @@ def test_set_control_cross_filter_graph(dash_br):
 
     # click on the 'versicolor' data in scatter graph and check result for box graph
     dash_br.click_at_coord_fractions(
-        f"div[id='{cnst.SCATTER_SET_CONTROL_CROSS_FILTER_ID}'] g[class^='trace']:nth-of-type(2) path:nth-of-type(20)",
-        0,
-        1,
+        scatter_point_path(cnst.SCATTER_SET_CONTROL_CROSS_FILTER_ID, point_number=20), 0, 1
```
petar-qb: I like the refactoring! Makes the code much easier to understand 🚀

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
 
     # check that graph 'x' axis changed to 'petal_length'
     dash_br.wait_for_element('text[class="xtitle"][data-unformatted="petal_length"]')
+
+
+@pytest.mark.filterwarnings("ignore::DeprecationWarning")
```
petar-qb: Why do we need ignore::DeprecationWarning here?

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+def test_set_control_clickmode_event_select(dash_br):
+    """Test cross filter between two graphs with multi select."""
+    accordion_select(dash_br, accordion_name=cnst.ACTIONS_ACCORDION)
+    page_select(
+        dash_br,
+        page_name=cnst.SET_CONTROL_MULTI_SELECT_PAGE,
+    )
+
+    # select virginica
```
petar-qb:
```suggestion
    # select virginica - this causes both: checklist and radioitems to be set
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+    dash_br.click_at_coord_fractions(scatter_point_path(cnst.SCATTER_SET_CONTROL_EVENT_SELECT, point_number=21), 0, 1)
+
```
petar-qb, line 511: Let's check targeted control values and chart here as well

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+    # select versicolor with SHIFT key
```
petar-qb:
```suggestion
    # select versicolor with SHIFT key - this causes only checklist to be set and radioitems stays unchanged as multiple values are selected and radioItems is single-select control
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+    modifier_click(
+        dash_br, selector=scatter_point_path(cnst.SCATTER_SET_CONTROL_EVENT_SELECT, point_number=22), key=Keys.SHIFT
+    )
+
+    # Check y axis max value is '2.4'
```
petar-qb:
```suggestion
    # Check filters target graph is correct - y axis max value is '2.4'
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+        dash_br,
+        component_id=cnst.RADIOITEMS_SET_CONTROL_MULTI_SELECT_FILTER,
+        options_value_status=[
+            {"value": 1, "selected": False, "value_name": "setosa"},
+            {"value": 2, "selected": False, "value_name": "versicolor"},
+            {"value": 3, "selected": True, "value_name": "virginica"},
+        ],
+    )
+
+    # unselect all values
```
petar-qb:
```suggestion
    # unselect all values - this causes target control values to reset
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+        dash_br,
+        component_id=cnst.RADIOITEMS_SET_CONTROL_MULTI_SELECT_FILTER,
+        options_value_status=[
+            {"value": 1, "selected": False, "value_name": "setosa"},
+            {"value": 2, "selected": False, "value_name": "versicolor"},
+            {"value": 3, "selected": True, "value_name": "virginica"},
+        ],
+    )
+
+    # select versicolor with SHIFT key
```
petar-qb:
```suggestion
    # select versicolor with SHIFT key - this causes that only versicolor is selected as source graph has clickmode="event"
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+    check_selected_categorical_component(
+        dash_br,
+        component_id=cnst.RADIOITEMS_SET_CONTROL_MULTI_SELECT_FILTER,
+        options_value_status=[
+            {"value": 1, "selected": False, "value_name": "setosa"},
+            {"value": 2, "selected": True, "value_name": "versicolor"},
+            {"value": 3, "selected": False, "value_name": "virginica"},
+        ],
+    )
+
```
petar-qb, line 640: Let's add the "unclicking" test step here. For the clickmode="event" graphs, unclicking the last selected point will not cause controls reseting.

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+
+def test_set_control_clickmode_aggrid_checkbox_click(dash_br):
```
petar-qb, line 642: Let's adjust these tests to match the `test_set_control_clickmode_event_select`. So:
1. Click virginica (test filters and graph)
2. Click versicolor (test filters and graph)
3. Unclick both (test filters are reset and graph)

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+        component_id=cnst.RADIOITEMS_SET_CONTROL_MULTI_SELECT_FILTER,
+        options_value_status=[
+            {"value": 1, "selected": False, "value_name": "setosa"},
+            {"value": 2, "selected": True, "value_name": "versicolor"},
+            {"value": 3, "selected": False, "value_name": "virginica"},
+        ],
+    )
+
+
+def test_set_control_clickmode_aggrid_command_click(dash_br):
```
petar-qb: I agree with Copilot comment here. I would remove this test completely as it uses the `Keys.COMMAND` and it tests pretty much the same scenario as the previous test that uses row checkboxes that is not Mac-specific.

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+    )
+
+
+def test_set_control_button_none(dash_br):
+    """Test set control with 'None' by clicking button."""
+    accordion_select(dash_br, accordion_name=cnst.ACTIONS_ACCORDION)
+    page_select(
+        dash_br,
+        page_name=cnst.SET_CONTROL_BUTTON_CARD,
+    )
```
petar-qb, line 804: Let's merge this and the following three tests (related to the `cnst.SET_CONTROL_BUTTON_CARD` page), and click components in the following order:
1. cnst.SET_CONTROL_CARD_SINGLE_VALUE (sets viriginica on both filters),
2. cnst.SET_CONTROL_CARD_MULTI_VALUE, (sets [versicolor, virginica] on checklist, leaves virginica on radioitems)
3. cnst.SET_CONTROL_BUTTON_EMPTY_LIST, (sets [] on checklist, leaves viriginica on radioitems)
4. cnst.SET_CONTROL_BUTTON_NONE, (resets both filters to original values)

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+            {"value": 1, "selected": True, "value_name": "setosa"},
+            {"value": 2, "selected": False, "value_name": "versicolor"},
+            {"value": 3, "selected": False, "value_name": "virginica"},
+        ],
+    )
+
+
+@pytest.mark.filterwarnings("ignore::DeprecationWarning")
+def test_filtered_graph_trigger_filter_for_aggrid(dash_br):
+    """Test filter ag_grid by selecting data in graph."""
```
petar-qb:
```suggestion
def test_filtered_graph_does_not_trigger_set_control(dash_br):
    """Test filtered graph does not trigger the set_control again."""
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+        dash_br,
+        page_name=cnst.FILTERED_GRAPH_AGGRID_TRIGGER_SET_CONTROL_PAGE,
+    )
+
+    # select versicolor in scatter graph
+    dash_br.click_at_coord_fractions(
+        scatter_point_path(cnst.FILTERED_SCATTER_TRIGGER_SET_CONTROL_ID, point_number=21), 0, 1
+    )
+
+    # unselect all in checklist for scatter as a source graph
```
petar-qb:
```suggestion
    # unselect all in checklist that filters a set_control source graph
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+    dash_br.multiple_click(select_all_path(elem_id=cnst.CHECKLIST_FT_GRAPH_SET_CONTROL), 1)
```
petar-qb: In this test we try to prove that set_control is not triggered from the source graph when the sourced graph is filtered with another filter. So, I think we should add `sleep(0.X)` here to leave the space for the mistake to happen. Is there any other way how it can be achieved?

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+        options_value_status=[
+            {"value": 1, "selected": False, "value_name": "setosa"},
+            {"value": 2, "selected": True, "value_name": "versicolor"},
+            {"value": 3, "selected": False, "value_name": "virginica"},
+        ],
+    )
+
+
+def test_filtered_aggrid_trigger_filter_for_aggrid(dash_br):
+    """Test filter ag_grid by selecting data in another aggrid."""
```
petar-qb:
```suggestion
def test_filtered_aggrid_does_not_trigger_set_control(dash_br):
    """Test filtered ag_grid does not trigger the set_control again."""
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+    accordion_select(dash_br, accordion_name=cnst.ACTIONS_ACCORDION)
+    page_select(
+        dash_br,
+        page_name=cnst.FILTERED_GRAPH_AGGRID_TRIGGER_SET_CONTROL_PAGE,
+    )
+
+    # navigate to AgGrid tab
+    dash_br.multiple_click("a[role='tab'][class='nav-link']", 1)
+
+    # select versicolor by clicking on cell
```
petar-qb:
```suggestion
    # select virginica by clicking on cell
```

or change the `row_index` to `row_index=1` to align with the graph test above. 333

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+    dash_br.multiple_click(
+        table_ag_grid_cell_path_by_row(cnst.FILTERED_AGGRID_TRIGGER_SET_CONTROL_ID, row_index=2, col_id="sepal_length"),
+        1,
+    )
+
```
petar-qb, line 986: Same as above: Should we add the `sleep` line here?

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +501,527 @@ def test_default_property_controls(dash_br):
...
+    # unselect all in checklist for aggdrid as a source graph
```
petar-qb:
```suggestion
    # unselect all in checklist that filters a set_control source ag_grid
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
l0uden: Because `click_at_coord_fractions` contains a bit old pytest method to select the coordinates

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
l0uden: Does it make much sense? I'm unselecting one of the scatter points later and checking all of this.

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
l0uden: But it checks the code you programmed to use COMMAND no SHIFT.

vizro-core/tests/e2e/vizro/dashboards/default/pages/set_control_multi_select_pages.py
l0uden: Worked, thanks!

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
l0uden: Does previous step where I'm checking that click + shift not working not prove that this logic is not working? Is it the different code or smth?

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
l0uden: added delay=0.5 after click

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
l0uden: Added 3rd step. About my concerns for the checks after first step left a comment above

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
l0uden: Changed the description. Good to have different values

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
l0uden: Done

l0uden:
@petar-qb , huge thanks for the review!

Here's also one thing. In the test `test_self_filtered_graph` we need `three` clicks on the dot in graph too fully unselect it. As i remember you've said it should be only `two`. Could you please take a look?

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
petar-qb: Is there an alternative function we could use as a replacement? If not, can we recreate this behavior and implement it ourselves? We’ll likely need to do that anyway once this method becomes obsolete in newer pytest versions.

Would you mind taking a quick look into this? If there’s no straightforward solution, I suggest narrowing `"ignore::DeprecationWarning"` to include part of the actual deprecation message, so it’s clearer what exactly is being deprecated when we revisit this later.

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
petar-qb: You're right. Let's keep it as it is 😃

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
petar-qb: hmm.. it only partially proves that unselecting won't work. However, let's keep it as it is so we don't overcomplicate the test.

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
petar-qb: I'm not sure, but maybe the `modifier_click` could be used to replace deprecated `click_at_coord_fractions`.

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
petar-qb: It's a nit picking I know, but could we select `virginica` and then `versicolor` instead of `setosa` and `versicolor`?

That way we’d properly test the `radioitems` after the values change. Currently, we can’t tell whether `set_control` is actually working from `ag_grid` to `radioitems` when a single row is selected, since the `radioitems` value remains `setosa` both before the test and after selecting two rows.

This change would also make the test more consistent with the previous ones.

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
petar-qb: Makes sense!

Could you just adjust this test so it matches the previous test (after applying [this comment](https://github.com/mckinsey/vizro/pull/1558/changes#r2753521917)). 

I would rename this test to `test_set_control_clickmode_aggrid_command_and_shift_click` and:
1. click on virginica row.
2. shift + click on versicolor row (to select two rows)
3. command + click on virginica to unselect virginica.
4. command + click on versicolor to unselect versicolor and reset controls. 
(something like that 😄)

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
petar-qb: @l0uden thanks for making an effort and resolving this comment. But, this parametrised solution doesn't really solve the problem. Your first test is `set_control_button_none` and it checks that `checklist` has `all values` selected and that `radioitems` has `setosa` selected. This actually doesn't test that the `set_control` is applied as these values are original control values before the test starts.

To solve this problem, I suggested to merge these four tests into a single test case (not parametrised) and to apply steps in the way I suggested in this comment above (and in the issue ticket too). By applying the steps in the way I suggested, all four combinations will be valid as the state of the controls will change after every step.

petar-qb:
> Here's also one thing. In the test test_self_filtered_graph we need three clicks on the dot in graph too fully unselect it. As i remember you've said it should be only two. Could you please take a look?

Great catch. I just checked and yes, three clicks are intended and expected to work. Let's just adjust the `TODO` comment in the test.

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
l0uden: Investigated it more deeply and the problem was with the coordinates which was pointing to coordinates of the element starting with top-left corner. After pytest 4.3 same method will be pointing to the center of the element. So i've changed coordinates to click on the center in future and on to the top-left corner now. After pytest update this code will still work and we don't need to skip the warning anymore. Thanks!

petar-qb: Awesome @l0uden!!

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
l0uden: Done

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
l0uden: Done

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
l0uden: Now I understood!) Done

vizro-core/tests/tests_utils/e2e/vizro/constants.py
```diff
@@ -278,6 +278,35 @@
...
 
+SET_CONTROL_MULTI_SELECT_PAGE = "set-control-multi-select-page"
+SCATTER_SET_CONTROL_EVENT_SELECT = "scatter_set-control-event-select"
+SCATTER_SET_CONTROL_EVENT = "scatter_set-control-event"
+TABLE_SET_CONTROL_MULTI_SELECT = "table_set-control-multi-select"
+BOX_SET_CONTROL_TARGET_MULTI_SELECT = "box_set_control_target_multi-select"
+CHECKLIST_SET_CONTROL_MULTI_SELECT_FILTER = "checklist_set_control_multi_select_filter"
+RADIOITEMS_SET_CONTROL_MULTI_SELECT_FILTER = "radioitems_set_control_multi_select_filter"
+
+SET_CONTROL_BUTTON_CARD = "set-control-button-card-page"
```
petar-qb:
```suggestion
SET_CONTROL_BUTTON_CARD_PAGE = "set-control-button-card-page"
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +494,559 @@ def test_default_property_controls(dash_br):
...
+    # select virginica in scatter graph 3 times to unselect it on the graph
+    dash_br.click_at_coord_fractions(
+        scatter_point_path(cnst.SCATTER_SET_CONTROL_SELF_FILTER, point_number=21, trace_index=1), 0, 0
+    )
+    dash_br.click_at_coord_fractions(
+        scatter_point_path(cnst.SCATTER_SET_CONTROL_SELF_FILTER, point_number=21, trace_index=1), 0, 0
+    )
+    dash_br.click_at_coord_fractions(
+        scatter_point_path(cnst.SCATTER_SET_CONTROL_SELF_FILTER, point_number=21, trace_index=1), 0, 0
+    )
```
petar-qb:
```suggestion
    # select virginica in scatter graph 3 times to unselect it on the graph. 
    # number of clicks needed for reseting such a graph will eventually be reduced to 2.
    # this click results with additional "self-filtering".
    dash_br.click_at_coord_fractions(
        scatter_point_path(cnst.SCATTER_SET_CONTROL_SELF_FILTER, point_number=21, trace_index=1), 0, 0
    )
    # this click does not apply "self-filtering" and it only highlights clicked point.
    dash_br.click_at_coord_fractions(
        scatter_point_path(cnst.SCATTER_SET_CONTROL_SELF_FILTER, point_number=21, trace_index=1), 0, 0
    )
    # this step "unclicks" selected point and applies "self-filtering" with reset filter value.
    dash_br.click_at_coord_fractions(
        scatter_point_path(cnst.SCATTER_SET_CONTROL_SELF_FILTER, point_number=21, trace_index=1), 0, 0
    )
```

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +494,559 @@ def test_default_property_controls(dash_br):
...
+        dash_br,
+        component_id=cnst.CHECKLIST_SET_CONTROL_SELF_FILTER,
+        checklist=True,
+        select_all_status=True,
+        options_value_status=[
+            {"value": 1, "selected": True, "value_name": "setosa"},
+            {"value": 2, "selected": True, "value_name": "versicolor"},
+            {"value": 3, "selected": True, "value_name": "virginica"},
+        ],
+    )
```
petar-qb, line 1060: If possible, check that the graph is really reset to the original y-axis (not critical).

l0uden: Done

vizro-core/tests/e2e/vizro/test_dom_elements/test_actions.py
```diff
@@ -503,3 +496,573 @@ def test_default_property_controls(dash_br):
...
+    # select virginica in scatter graph 3 times to unselect it on the graph.
+    # number of clicks needed for resetting such a graph will eventually be reduced to 2.
+    # this click results with additional "self-filtering".
+    dash_br.click_at_coord_fractions(
+        scatter_point_path(cnst.SCATTER_SET_CONTROL_SELF_FILTER, point_number=21, trace_index=1), 0, 0
+    )
+    # after first click delay needed.
+    # sometimes without delay test stuck on the first click
+    # (maybe because graph is not fully loaded and test can't click on the appropriate point using coordinates)
+    time.sleep(0.5)
```
petar-qb: This is true, great catch! 🚀 
```suggestion
    # after first click delay is needed to ensure that the graph is loaded so its point can be clicked again.
    time.sleep(0.5)
```

PR #1559, opened 2026-01-27
[Bug] Fix visibility of theme-specific logos

huong-li-nguyen:
## Description
Closes https://github.com/McK-Internal/vizro-internal/issues/2447
- Theme-specific logos were not visible anymore as we've wrapped into `a` tags and these were still hidden

@l0uden - could you add new screenshot tests for these? 

- Add logo_dark.xx and logo_light.xx to the assets folder
- Then switch themes and take screenshots

## Notice


huong-li-nguyen:
> Good catch! But I don't understand why this wasn't working before or why this would be the right fix? What's the problem with the logic in `_get_logo_images`? Is `hidden = not None` hiding things when it shouldn't?

No, it's working correctly, but we once wrapped the logo images inside some `a` tags (to make the logos clickable and go back to the homepage), which affected the CSS hack to hide / show the logos based on the theme switch.

We had this before:
So the anchor tag was hidden, and even though the logos were visible after the theme-switch, their parent which is the `a` tag was still hidden :) Remember we have this hack to first hide all theme-specific logos and then only display the one relevant to the theme.

```
a:has(#logo-light),
a:has(#logo-dark),
#logo-dark,
#logo-light {
  display: none;
}

❌ Only shows the <img>, not the <a>
[data-bs-theme="dark"] #logo-dark {
  display: inline; 
}

[data-bs-theme="light"] #logo-light {
  display: inline;
}
```

PR #1566, opened 2026-02-03
[UX] Improve visual layout when using filled containers

huong-li-nguyen:
## Description
Closes https://github.com/McK-Internal/vizro-internal/issues/2475

Ready for review for @stephkaiser , for devs it might make more sense, after @stephkaiser has finalized design.

### Changes
1. **Typography Hierarchy**: Reduced component titles (graphs, tables, AG Grid) from H3 to **H4** (if embedded inside container only), keeping container titles at H3 for clear visual hierarchy. X-axis and Y-axis titles inside charts are 14px instead of 16px (regardless whether the chart is embedded in a container or not).
2. **Container separation**: Added horizontal line below container titles to make them stand out
3. **Color refinement**: Incorporated switch of right-side and colored container backgrounds. Switch back left-side bg color for light theme to grey to maintain consistency.

## Visual Impact

- Container titles (H3) are now clearly larger than component titles and separated by a horizontal line
- Component titles are appropriately sized for containers (H4), stay as H3 if used without a container
- Clear visual hierarchy: page → container → component
- Y-axis and X-axis titles of charts have the same font-size as the rest of the chart (14px), otherwise they are too large, if the chart title becomes H3 inside a container
- Light and dark themes now follow the same structural and visual logic, reducing cognitive load when switching between them

## Open Questions

- **Filled container borders**: Filled containers don't look great without a border. Should borders be included by default? Should the variant be called "filled+outlined" or just "filled"?
- **Left-side background color**: What should the left-side bg color be (with and without colored containers)? I like the current version actually (with the update on the light theme), but just want to make sure you are happy too @stephkaiser 
- **Dropdown background colors**: What should dropdown bg colors be? Just FYI, we're diverting from the QBDS because we have a smaller subset of CSS variables. But for now, don't feel restricted by it, and just let me know what you think looks best and then we can see how we can implement it. 

General review of bg colors required for both dark and light:
- left-side bg color
- header bg color
- dropdown bg color
- dropdown label bg color

Screenshot 2026-02-05 at 17 40 13

## Notice


vizro-core/examples/scratch_dev/assets/css/custom.css
```diff
@@ -1,4 +1,13 @@
...
+    height: 30px;
+}
+
+#page-header {
+    display: none;
+}
+
+/* Does a filled container need a border? */
+.bg-container {
+    border: 1px solid var(--bs-border-color);
```
huong-li-nguyen: @stephkaiser , @antonymilne , by default the filled containers do not have a border. We have three variants:
- filled
- outlined
- plain

What you see in the demo is currently filled+outlined 😄 But I am wondering whether that should be the default for filled, or if we should add a new variant.

antonymilne: No strong feelings from me as to whether we prefer it with a border or not. If @stephkaiser and you prefer to give `filled` an outline in general then I would just modify the style for `filled` to include the border though - I don't think it's worth giving the extra flexibility of a whole new option `filled+outlined` to do it. Most people are unlikely to use something that "sounds complicated" like that, so better instead just to build it into `filled` for them.

One thing worth considering though is how this compares to `Card` which I think currently behaves like this:
* in light theme it has no border but does have shadow
* in dark theme it has fill but no border and no shadow

Should these be consistent with a filled container? I don't know.

antonymilne:
Generally looks great to me, although I'm not going to be able to understand exactly what's changed for background colors without you showing me 😅 I've lost track of all the different possibilities there and what we chose to do before and why. In general though, I'm personally a big fan of achieving symmetry between light and dark themes, by which I mean there's a strict one to one correspondence between colors in light and dark themes. i.e. if X on screen is color A in light theme and color B in dark theme then everything that's color A in light theme becomes color B in dark theme. The theme switching logic then just swaps out color A for color B everywhere it sees it and doesn't change the logic of the "color hierarchy" in any way.

The typography hierarchy and keyline under the container title looks much better to me now 👍 

> * **Filled container borders**: Filled containers don't look great without a border. Should borders be included by default? Should the variant be called "filled+outlined" or just "filled"?

Answered in the comment, but I have no strong feelings on borders on or off by default. Whatever we choose as default I would just build into `filled`.

> * **Left-side background color**: What should the left-side bg color be (with and without colored containers)? I like the current version actually (with the update on the light theme), but just want to make sure you are happy too 

Looks good to me!

> * **Dropdown background colors**: What should dropdown bg colors be? Just FYI, we're diverting from the QBDS because we have a smaller subset of CSS variables. But for now, don't feel restricted by it, and just let me know what you think looks best and then we can see how we can implement it.

No strong feelings here, so long as we ideally follow my symmetry logic. So if we choose to make it the same as e.g. filled container background color, it takes whatever is the filled container background color in both light and dark theme.

vizro-core/examples/scratch_dev/assets/css/custom.css
huong-li-nguyen: @stephkaiser and i decided to go without the outline, so it's just filled :) In the current branch, you see that the Card is actually consistent. It's filled and has a box-shadow in both dark and light theme, however, the box-shadow has just never been visible in the dark theme 😬 Steph is aware, she's going to revise all bg colors here (left-side, right-side, card, dropdown labels, dropdown menu, etc.)

vizro-core/changelog.d/20260205_170535_90609403+huong-li-nguyen_colored_container.md
```diff
@@ -0,0 +1,48 @@
...
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
```
antonymilne: We put this in the release notes in https://github.com/mckinsey/vizro/pull/1005/ so probably we should put it in here.

```suggestion
### Added

- Export png button in modebar inside `Graph`. ([#1566](https://github.com/mckinsey/vizro/pull/1566))

### Changed

- Chart backgrounds are no longer transparent. ([#1566](https://github.com/mckinsey/vizro/pull/1566))

```

huong-li-nguyen:
> Looks good! Much simpler logic than before which always makes me happy 💯
> 
> If we think the restriction of "you shouldn't use outlined and filled containers in the same page" is going to remain even after @stephkaiser has looked at the background colors then it might be worth adding to the docs?

I added an admonition to the docs, but in general it's not forbidden and I think some would probably also say that it doesn't look horrible 😄 So I just said that "we recommend it"

PR #1571, opened 2026-02-04
[VDC] New markdown component

lingyielia:
Closes https://github.com/McK-Internal/vizro-internal/issues/2422

After review:

- [x] Move the vizro-core portion of code to a new branch
- [x] add release CI: https://github.com/mckinsey/vizro/pull/1625

After this PR merge:

- [x] Require status checks to include new CIs to pass before merging 
- [ ] Test with https://test.pypi.org/legacy/ - Merge this PR first with the test workflow included, trigger it from main
- [ ] Remove the dry run CI in the actual release PR
- [ ] Pypi setup (test in https://test.pypi.org/ first)
- [ ] Create release PR and push to Pypi

## Notice

### DIFF
--- a/vizro-dash-components/webpack.config.js
+++ b/vizro-dash-components/webpack.config.js
@@ -113,6 +113,16 @@ module.exports = function (env, argv) {
         },
       },
     },
+    performance: {
+      hints: "error",
+      maxAssetSize: 400 * 1024, // 400 KB default for all assets
+      maxEntrypointSize: 50 * 1024, // 50 KB for the main entry point
+      assetFilter(assetFilename) {
+        // MathJax is a known large dependency (~2 MB); exempt from the default limit.
+        if (assetFilename === "async-mathjax.js") return false;
+        return assetFilename.endsWith(".js");
+      },
+    },
     plugins: [
       new WebpackDashDynamicImport(),
       new webpack.ProvidePlugin({


vizro-core/src/vizro/models/_components/text.py
```diff
@@ -26,7 +26,7 @@ class Text(VizroBaseModel):
             dict[str, Any],
             Field(
                 default={},
-                description="""Extra keyword arguments that are passed to `dcc.Markdown` and overwrite any
+                description="""Extra keyword arguments that are passed to `vdc.Markdown` and overwrite any
```
antonymilne: Let's just leave these `description` saying `dcc.Markdown`. Otherwise we have to write docs for `vdc` and publish them etc. which it's not worth doing now.

antonymilne (commented):
Looks great, nice work! I pushed a whole bunch of commits - all various tidying, but please do check them out to make sure they make sense. The only functional change I made was to use `InlineCodeHighlight` too.

I'd like to briefly discuss:
- linting
- ~dmc dependency~ realised this was actually a big problem but should now be fixed - see below
- testing
- publishing

Then once the changes to `vizro-core` have been reverted I think we can merge this PR 🎉 

Edit: I realised there were actually some bigger problems here we didn't think of before: the bundle `vizro_dash_components/vizro_dash_components.js` was absolutely massive (3.6MB) and inefficient. Please take a look through these commits to make sure you understand how it's been fixed:
* [9851abb](https://github.com/mckinsey/vizro/pull/1571/commits/9851abb4aa889eac9c692497ba9925ff97596daf) - MathJax took up 2MB of the bundle. In `dcc.Markdown` this is only downloaded when required (`mathjax=True`) whereas we were downloading it all the time
* https://github.com/mckinsey/vizro/pull/1571/commits/5d8351191fd2641f9768d110957bdc46a9030d26 - we were bundling all of `dmc`, which meant there were duplicated dmc bundles. I've changed to a pattern so that it's now a Python dependency and we then reuse the same dmc bundle. This isn't very conventional in Dash (there's no good pattern for this sort of thing) but I think it should work well.

As a result we could also implement the full `dcc.Link` behaviour by following the same reuse rather than duplicate pattern: [c2b3591](https://github.com/mckinsey/vizro/pull/1571/commits/c2b359196f317c0fe3a94d7638e1d8a45c7e1d94).

The overall result here is we should have gone from always downloading 3.6MB to the following:
* whenever `vizro_dash_components` imported, download `vizro_dash_components.js` (8.2 KB)
* whenever `vdc.Markdown` used, download `async-markdown.js` (310 KB)
* whenever `mathjax=True`, download `async-mathjax.js` (2.0 MB)
* `vizro_dash_components` shouldn't be responsible for any other downloads since `dcc` and `dmc` handle them and we now reuse them

vizro-core/src/vizro/static/css/code.css
antonymilne, line 1: I think (hope) there might be a better solution here but let's save it for the new PR that modifies `vizro-core`.

vizro-core/src/vizro/models/_components/text.py
antonymilne: Actually I've now written some more information in the README.md, so we can just link to that I think.

vizro-core/src/vizro/static/css/code.css
antonymilne: I think the changes I've made now should make most/all of this stuff redundant.

lingyielia: This is great. The "double CSS load" problem is very annoying. With the fix we can remove all the css reapplying!

vizro-dash-components/CLAUDE.md
```diff
@@ -0,0 +1,68 @@
...
+- JSDoc comments on the component and its props become Python docstrings
+- `defaultProps` on the component set Python default values
+
+### Architecture notes
+
+**Lazy loading**: The Markdown component follows the same lazy loading pattern as `dash-core-components` (DCC). It uses `@plotly/dash-component-plugins` `asyncDecorator` and `@plotly/webpack-dash-dynamic-import` to create separate webpack chunks (`async-markdown.js`, `async-mathjax.js`) that are loaded on demand. See DCC's `Markdown.react.js` for the reference implementation.
+
+**DMC integration**: Code highlighting uses `dash-mantine-components` (DMC) components (`CodeHighlight`, `InlineCodeHighlight`) accessed via `window.dash_mantine_components` global. This avoids bundling Mantine twice and ensures consistent theming with the host app's `MantineProvider`. The host Dash app must wrap its layout in `dmc.MantineProvider` for proper styling.
+
+## Testing with vizro-core
```
maxschulz-COL, line 55: I have wondered in the past whether there is a clever way to make this more streamlines. What would be the best practice to test automatically when developing a dependency?

This must be done lots of times, just thinking of libraries like `pydantic` and `pydantic-core`. Is there a better way to test this?

vizro-dash-components/CLAUDE.md
maxschulz-COL, line 1: Refer to it in main CLAUDE.md?

vizro-dash-components/hatch.toml
```diff
@@ -0,0 +1,31 @@
...
+scripts = {add = "scriv create --add", collect = ["scriv collect --add", "- pre-commit run --files=CHANGELOG.md > /dev/null"]}
+
+[envs.default]
+dependencies = [
+  "pre-commit",
+  "dash-iconify",  # just for the example usage.py
+  "dash[dev,testing]>=4,<5",  # provides dash-generate-components + dash_duo fixture
+  "chromedriver-autoinstaller>=0.6.4",
+  "pytest",
+  "vizro"
```
maxschulz-COL: Could this be a place where we take the current branch rather than the released version? Not sure... See other comment

vizro-dash-components/tests/test_vizro_integration.py
```diff
@@ -0,0 +1,33 @@
+"""Test vizro_dash_components works inside a Vizro dashboard as a custom Dash component."""
+
+import pandas as pd
+import vizro.models as vm
```
maxschulz-COL: Let's think once hard on how to best determine the Vizro version here! I would love this to go both ways, ie changes here and changes in Vizro always get tested against the latest version on `main` AND the latest released version. I think this is important to save ourselves future head-aches. Remember all the trouble with `vizro-ai` :)

vizro-dash-components/vizro_dash_components/__init__.py
maxschulz-COL, line 1: Is this from Boilerplate? this looks like a crazy `__init__` to me :)

vizro-dash-components/CLAUDE.md
```diff
@@ -0,0 +1,68 @@
...
+### Writing Dash components
+
+Each `.tsx` file in `src/ts/components/` becomes a Dash component. Key patterns:
+
+- Extend `DashComponentProps` for the required `id` and `setProps` props
+- Call `setProps({prop: newValue})` to trigger Dash callbacks
+- JSDoc comments on the component and its props become Python docstrings
+- `defaultProps` on the component set Python default values
+
+### Architecture notes
```
maxschulz-COL: Is that even needed at a high level CLAUDE.md, or should there be one closer to the Markdown component?

Can individual components actually be in folders, I think this would be easier to digest

vizro-dash-components/README.md
maxschulz-COL, line 1: I guess we eventually want docs? @stichbury

vizro-dash-components/usage.py
maxschulz-COL, line 1: Should this not be in its own folder?

vizro-dash-components/vizro_dash_components/__init__.py
lingyielia: Yes, it's auto generated by boilerplate :)

vizro-dash-components/CLAUDE.md
lingyielia: Great point, moved it to source code comments

vizro-dash-components/usage.py
lingyielia: This is also produced by boilerplate. Let's keep it there for now

vizro-dash-components/CLAUDE.md
lingyielia: I think this is a right concern. The most streamlined approach probably would be:

  1. vizro-dash-components should test itself standalone (with plain Dash, no Vizro). It's a general-purpose Dash component library. The current dash_duo tests already do this.
  2. The integration test belongs in vizro-core, which is the consumer. When vizro-core adds vizro-dash-components as a dependency, vizro-core's own test suite should verify its components work within Vizro dashboards.

The current test creates a circular-feeling dependency: vizro-dash-components lists vizro as a test dependency in hatch.toml, but vizro-core will depend on vizro-dash-components.

Could we solve by removing `test_vizro_integration.py` from here. In vizro-core's hatch.toml, point the vdc to `points to ../vizro-dash-components` to test against the branch version. Then during testing, we need to run test from `vizro-dash-components` and from `vizro-core` @antonymilne ydwt?

vizro-dash-components/hatch.toml
lingyielia: This line does look suspicious. see comment above.

vizro-dash-components/tests/test_vizro_integration.py
lingyielia: see comment above.

vizro-dash-components/CLAUDE.md
maxschulz-COL: yes I think that is a good idea. I think we should do it like that. To replay what you said just to check understanding:

- here we only test vdc as Dash components, because that is what they do
- however to not release something that then doesn't work with `vizro-core`, we will have integration tests on Vizro core that also run on changes in `../vizro-dash-components`. All changes in `vizro-core` will also run these integration tests
- this leaves only the local debug case right? like say the CI fails, how can I _easily_ verify things are working/fixed

vizro-dash-components/hatch.toml
maxschulz-COL: Yes I think this one definitely needs to go

vizro-dash-components/CLAUDE.md
lingyielia: Locally, we need to do what's mentioned here in CLAUDE.md:

`vizro-core` already depends on the released `vizro-dash-components` package. To test local changes:

1. `vizro-dash-components/`: Build with `hatch run generate-components`
1. `vizro-core/`: Override the installed package with your local version: `hatch run examples:pip install -e ../vizro-dash-components`
1. `vizro-core/`: Add the component under test to `scratch_dev/app.py`
1. `vizro-core/`: Run the example with `hatch run example`

maxschulz-COL: But you agree with my summary of the first two points?

lingyielia: Yes!

lingyielia: @antonymilne @maxschulz-COL  I removed `test_vizro_integration.py` and the vizro dependency in `hatch.toml` from this pr now. 👍

vizro-dash-components/hatch.toml
```diff
@@ -7,14 +7,17 @@ scripts = {add = "scriv create --add", collect = ["scriv collect --add", "- pre-
 [envs.default]
 dependencies = [
   "pre-commit",
-  "dash-iconify", # just for the example usage.py
-  "dash[dev]>=3.0.0"  # provides dash-generate-components, required by `npm run build`
+  "dash-iconify",  # just for the example usage.py
```
antonymilne: Is this still needed for the example? I think we should go for an example usage.py that doesn't need it.

vizro-dash-components/package.json
```diff
@@ -1,51 +1,51 @@
...
+  "version": "0.1.0-dev0",
+  "description": "Vizro Dash Components are used by the Vizro framework but can be used in a pure Dash app",
+  "main": "index.ts",
+  "scripts": {
+    "build:js": "webpack",
+    "build:backends": "dash-generate-components ./src/ts/components vizro_dash_components -p package-info.json  --ignore \\.test\\.",
+    "build": "npm run build:js && npm run build:backends",
+    "watch": "webpack --mode development -- --watch",
+    "lint:js": "npx @biomejs/biome check src/",
+    "fmt:js": "npx @biomejs/biome check --write src/"
```
antonymilne: Are these lines still needed now it's all done in pre-commit?

.github/workflows/test-integration-vizro-dash-components.yml
```diff
@@ -0,0 +1,57 @@
...
+        run: npm ci --legacy-peer-deps
+
+      - name: Install Hatch
+        run: pip install hatch
+
+      - name: Generate Python components from TypeScript source
+        # Generated Python code and JS bundles are gitignored, so we must rebuild before testing.
+        run: hatch run generate-components
+
+      - name: Show dependency tree
```
antonymilne: Let's put this straight after the `Install hatch` stage.

vizro-dash-components/tests/test_markdown.py
antonymilne, line 1: Did you check whether there was any way to check bundle size easily with `dash_duo`? If there's not an easy way then no problem, we can forget about it. But if it's easy to do then would be nice to add.

vizro-dash-components/hatch.toml
```diff
@@ -23,5 +26,5 @@ prep-release = [
   'echo "Now raise a PR to merge into main with title: [Release] Release of vizro-dash-components $(hatch version)"'
 ]
 pypath = "python -c 'import sys; print(sys.executable)'"
-generate-components = "npm run build"
-
+test = "pytest tests --headless {args}"
+test-unit = "pytest tests --headless {args}"
```
antonymilne: Let's remove `test-unit` given we don't separate integration vs unit tests here.

.github/workflows/test-vizro-dash-components.yml
antonymilne, line 1: This should be just `test-vizro` without the integration. Similar for its `name`.

.github/workflows/release-if-needed.yml
```diff
@@ -62,6 +62,18 @@ jobs:
...
+        uses: actions/setup-node@v6
+        with:
+          node-version: 20
+      - name: Generate Python components for vizro-dash-components
+        if: needs.check-version.outputs.package_name == 'vizro-dash-components'
+        # Generated Python code and JS bundles are gitignored, so we must rebuild before packaging.
+        run: |
+          cd vizro-dash-components
+          npm ci --legacy-peer-deps
+          hatch run generate-components
```
antonymilne, line 76: This is run here and in the `test-release-vizro-dash-components` workflows, is that right?

antonymilne (approved):
Looks good! Nice work 🙌 

Don't forget to make sure the workflows are added to our required ones on github also once this has merged!

vizro-dash-components/CLAUDE.md
antonymilne: Yes, this is definitely the right thing to do! We should only be testing `vdc` as a standalone set of Dash components here. The tests that they integrate correctly into Vizro should be done in `vizro-core`.

As for how to test that releases of `vdc` don't break Vizro, there are a few options, and any are fine by me I think:
* do what @maxschulz-COL suggests and trigger the vizro integration tests when vdc is changed
* just pin the vdc dependence in vizro and then manually bump the vizro requirement when we release vdc. This is ok because we [control both packages](https://iscinumpy.dev/post/bound-version-constraints/#when-is-it-okay-to-set-an-upper-limit)
* just don't worry about vdc breaking vizro because it seems pretty far-fetched that something could pass the Dash tests but fail the Vizro ones

vizro-dash-components/hatch.toml
lingyielia: removed

.github/workflows/release-if-needed.yml
lingyielia: yes, in `test-release-vizro-dash-components` we just don't have release notes extraction or GitHub tagging, the rest are the same as a dry run

vizro-dash-components/tests/test_markdown.py
lingyielia: Ah I forgot to check that. dash_duo doesn't seem to have direct test for this. But here are some possible approaches we can take:
- the `hatch run generate-components` command already produces webpack output including the asset size like
```
asset async-mathjax.js 1.99 MiB [compared for emit] [minimized] (name: async-mathjax) (id hint: async)
asset async-markdown.js 310 KiB [compared for emit] [minimized] (name: async-markdown) (id hint: async) 1 related asset
asset vizro_dash_components.js 8.26 KiB [emitted] [compared for emit] [minimized] (name: main)
```

I'm adding some very simple guardrails for now to fail `hatch run generate-components` if unexpected large asset is produced. 

For more granular control, maybe it makes more sense to write a custom lint tool to check all assets' size?

PR #1574, opened 2026-02-05
[Feat] Conditional Notifications for custom actions

petar-qb:
closes https://github.com/McK-Internal/vizro-internal/issues/2414

## Description
This PR introduces a conditional notifications system through the new`"notifications"` **custom action** property in Vizro that extends the actions with first-class support for showing user notifications based on action outcomes (success or exception/fail) and execution context. Actions (conditional) notifications is set of predefined notifications where if looked like this:
```py
@capture("action")
def custom_action(action_input_parameter_1):
    sleep(3)
    match random.randint(0, 5):
        case 0:
            return "Text output"
        case 1:
            return "Text output", "my_success"
        case 2:
            return "Text output", ("success", "Action duration: 3.15s")
        case 3:
            sleep(1)
            return "Text output", ("my_success", "Action duration: 4.16s")
        case 4:
            raise Exception("Random error occurred!")
        case 5:
            return Exception("Random error occurred!", ("error", "Action duration 3.16s"))

vm.Action(
    function=custom_action("action_input_parameter_1"),
    output="output_text_id",
    notifications = { 
        "progress": "Action is running. Input argument: {{action_input_parameter_1}}",
        "success": "Action finished successfully. {{result}}",
        "error": "Action failed.\n\n{{error_mgs}}\n\n{{result}}",
        "my_success": "Info that action finished. {{result}}"
    }
)
```
they will behave like this:
1. `"progress"` notification will be displayed before the action triggers. It will be a default `variant="progress"` notification. Optionally, you can template its text with the dynamic argument by specifying action function parameter name like `{{action_input_parameter_1}}`. 
2. `case 0` -> return "Text output" => "success"  `variant="success"` notification will be displayed with defined text.
3. `case 1` -> return "Text output", "my_success" => "my_success"  `variant="info"` notification will be displayed with defined text.
4. `case 2` -> return "Text output", ("success", "Action duration: 3.15s") => "success"  `variant="success"` notification will be displayed with defined + templated `{{result}}` text.
5. `case 3` -> return "Text output", ("my_success", "Action duration: 4.16s") => "my_success"  `variant="info"` notification will be displayed with defined + templated `{{result}}` text.
6. `case 4` ->  raise Exception("Random error occurred!") => "error"  `variant="error"` notification will be displayed with defined + templated `{{error_msg}}` text.
7. `case 5` -> return Exception("Random error occurred!", ("error", "Action duration 3.16s")) => "error"  `variant="error"` notification will be displayed with defined + templated `{{error_msg}}` + templated `{{result}}` text.

It's important to add that:
1. There are three built-in notification keys: `"progress"`, `"success"`, and `"error"`. If these are defined, they are shown automatically before the action starts ("progress") or after it finishes (either successfully ("success") or with an exception ("error")), even if the notification key is not explicitly returned from the action function. Any custom notification keys (e.g., "my_success") must be explicitly returned from the action function, either as a string or as a tuple if the action result is needed for templating.
2. Every custom action has default `notifications={"error": "Action failed."}`. This can simply be overwritten with `notifications={"error": None}` if needed.
3. All exceptions (except the `PreventUpdate`) are now logged in the server console.
4. If an action raises an exception, and there's no "error" (or a custom error key) defined (or its value is `None` like `{"error": None}`), the exception will be propagated to the dash debug UI tool. Otherwise the error ("Action failed!" or a custom one) will be displayed.
5. You can define notifications dict value as `show_notification`/`update_notification` object instead of pure text if you need for example to change default `variant`, `title`, `icon`, `auto_close` (so anything that's not defaulted)
6. When you trigger the action again, the new `"progress"` notification will replace the old notification (deliberately built on the client-side and works only for conditional notifications).
7. Templates {{result}}, {{error_msg}} will be replaced with empty string if not provided.
8. The template variable {{progress_parameter_name}} is not replaced with an empty string if it’s missing. This was intentionally changed to make debugging and dashboard creation easier. This is different from {{result}} and {{error_msg}}, which may be missing because Python functions can return them dynamically. In contrast, `progress_parameter_name` is a fixed string and once it's set, it will always be available.
9. Returning `no_update` from the action is considered as a success. The next action in the chain will be triggered.
10. Raising `PreventUpdate` from the action is considered as a `success`. This will prevent the actions chain from continuing to execute though.
11. Raising any other exception (except `PreventUpdate`) from the action is considered as an `error`. This will prevent the actions chain from continuing to execute.

## Screen recording:

https://github.com/user-attachments/assets/61a88dda-0a72-4d19-a0a2-e6de70bf3a6d

## Notice


vizro-core/changelog.d/20260312_093331_petar_pejovic_conditional_notifications.md
petar-qb, line 1: How should we announce this change, under Added or Highlights section? 

Here's what's changed:
- If a `custom action` fails the  `"Action failed."` `error` notification is shown.
- If an `export_data` fails, the `"Exporting failed."` `error` notification is shown.
- If a `set_control` finishes successfully, the `"The control value is set."` `success` notification is shown.
- All actions exceptions are now logged in the server terminal
- If an action exception results with the notification, the exception will not be propagated to the Dash `In-App Error Reporting`

vizro-core/src/vizro/actions/_export_data.py
```diff
@@ -32,12 +32,14 @@ class export_data(_AbstractAction):
...
     type: Literal["export_data"] = "export_data"
     targets: list[ModelID] = Field(
-        default=[],
+        default_factory=list,
         description="List of target component ids for which to download data. If none are given then "
         "download data from all components on the page.",
     )
     file_format: Literal["csv", "xlsx"] = Field(default="csv", description="Format of downloaded files.")
 
+    notifications: ActionNotificationType  # type: ignore[misc]
```
petar-qb: I added the `notifications` property to all builtin actions except to `show_notification` / `update_notification` actions as it semantically doesn't make much sense (like which one will show first and which second). This resulted with adding it to every other builtin action separately, and also a few places where I needed to add the following check: `if hasattr(action, "notifications"): ...`. 

If we decide to revoke this decision, then I'll need to handle the `vizro-notifications` output a bit differently due to fact that this output is already used in the `show_notification` / `update_notification` actions. 

P.S. While writing this comment, I realised this feels like the right approach 😄

vizro-core/src/vizro/actions/_export_data.py
```diff
@@ -70,6 +72,15 @@ def pre_build(self):
...
         """Exports data after applying _controls."""
+        # TODO PP IMPORTANT: REMOVE SLEEP AND EXCEPTION AFTER TESTING
+        import random
+        from time import sleep
+
+        sleep(2)
+
+        if random.random() > 0.5:
+            raise Exception("Random error occurred during data export!")
+
```
petar-qb: PP: Remove this and similar for `_parameter_action.py` and `set_control.py`  before merging this PR.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -304,41 +363,127 @@ def _action_callback_function(
...
+
+        # --- List outputs ---
         elif isinstance(outputs, list):
+            # TODO PP NOW: Cover with unit-tests
+            # Allow: `(rv_1, ..., rv_N)`,  `(rv_1, ..., rv_N, notification_payload)`,
+            # `[rv_1, ..., rv_N]`, and `([rv_1, ..., rv_N], notification_payload)`.
+            *return_value, notification_payload = self._split_trailing_notification_payload(
+                return_value=return_value,
+                expected_lengths=[2, len(outputs) + 1],
+            )
```
petar-qb: PP: Improve it now and replace `expected_lengths: list[int]` with `expected_length: int`.

vizro-core/src/vizro/static/js/models/action.js
```diff
@@ -46,9 +46,73 @@ function guard_action_chain(trigger_value, guard_data, trigger_component_id) {
...
+function show_progress_notification(
+  trigger,
+  notificationObject,
+  actionParameters,
+  ...actionRuntimeArguments
+) {
+  console.debug("Showing progress notification");
+
+  // Clear existing notification to enable showing the notification with the same ID again.
+  dash_mantine_components.appNotifications.api.hide(notificationObject[0].id);
```
petar-qb, line 93: We might introduce the following so users can hide it explicitly. `hide_notification(notification=modelId)`
    
Or even better would might be to implicitly hide the notification before every `show_notification`. This would solve the problem generally across the app.

vizro-core/src/vizro/actions/_export_data.py
```diff
@@ -32,11 +32,16 @@ class export_data(_AbstractAction):
 
     type: Literal["export_data"] = "export_data"
     targets: list[ModelID] = Field(
-        default=[],
+        default_factory=list,
```
petar-qb: Is this ok? It removes the `"default": []` from the schema.

antonymilne: It's ok but not quite as good as `default=[]` for the reason you mention. I don't see the point of the change though?

vizro-core/src/vizro/actions/_export_data.py
```diff
@@ -32,11 +32,16 @@ class export_data(_AbstractAction):
...
         description="List of target component ids for which to download data. If none are given then "
         "download data from all components on the page.",
     )
     file_format: Literal["csv", "xlsx"] = Field(default="csv", description="Format of downloaded files.")
+    error_text: str | None = Field(
```
antonymilne: This is exactly the sort of thing I had in mind as something we might want to add in future but shouldn't have a field for now. 🙂 I don't think we should have any new fields in any of our built-in actions for now - we can always add later if there's demand for it. For now we can just hardcode the message into the `notifications` property.

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -107,6 +110,14 @@ def function(self):
...
         return [notification]
 
+    @property
+    def _dash_components(self) -> list[Component]:
+        return [] if self._is_conditional else super()._dash_components
+
+    def _define_callback(self):
+        if not self._is_conditional:
+            super()._define_callback()
+
```
antonymilne, line 128: What's this bit for?

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -207,3 +220,7 @@ def outputs(self):  # type: ignore[override]
         if self._same_page:
             return self.control
         return ["vizro_url.pathname", "vizro_url.search"]
+
+    @property
+    def notifications(self):  # type: ignore[override]
+        return _normalize_action_notifications({"success": self.success_text, "error": None})
```
antonymilne: Let's maybe discuss with @stephkaiser and play around with some example apps to figure out what, if any, notifications this and our other built-in actions should do. 

As before, I'd maybe err on the side of too many notifications just to advertise it as a feature, so long as it's not annoying. e.g. not sure what it would feel like if we had notifications for parameter/filter also 🤔

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -417,16 +585,54 @@ def _define_callback(self):
 
         @callback(output=callback_outputs, inputs=callback_inputs, prevent_initial_call=True)
```
petar-qb, line 633: Since callback itself can’t be tested directly, should I extract its logic into a separate function to make it testable?

vizro-core/src/vizro/actions/_notifications.py
petar-qb: We don’t treat `_is_conditional` notifications as standard actions. This means we intentionally avoid creating action util components and action util callbacks for them, since that would make them behave like regular actions (one action one http req).

Instead, conditional notifications follow a different pattern. For them we introduce additional dedicated internal callback output and we return an additional result value from the `action_callback` to handle them separately.

vizro-core/src/vizro/actions/_export_data.py
petar-qb: Let's continue this discussion here -> https://github.com/mckinsey/vizro/pull/1574#discussion_r2948317299

vizro-core/src/vizro/actions/_set_control.py
petar-qb: Before I list options and my preferences, I'm sharing summary of how errors are currently handled:

```mermaid
flowchart LR
    A[Error] --> B{Notification raised?}
    B -->|Yes| C[Show notification and no error in Dash Debug Panel]
    B -->|No| D{debug mode?}
    D -->|True| E[Show error in Dash Debug Panel only]
    D -->|False| F[Fails silently]
```
P.S. All exceptions (except PreventUpdate) are logged to server's stdout.

So, while a dashboard is under development (`debug=True`), the dashboard creator always sees on the UI when exception happens. However, dashboards users in the production (`debug=False`) will only see error notifications, and unhandled exceptions will silently fail for the users and leave a trace in the server's console for dashboard creator.

---

Here are the options from my perspective:
1. Expose full api through the `notifications` field for `export_data` and `set_control`. Set default behaviour for them (error msg for export_data and success msg for set_control).
2. Expose only `error_text` for `export_data` and `success_text` for `set_control`. So, exposing only fields to control those that are set by default.
3. Not exposing any API, but set default.
4. Not exposing any API, not setting any defaults

All approaches share the following:
 - Nothing is broken (it's obvious but worth mentioning)
 - Errors that happen in the OPL/Filer/Parameter/set_control, while developing dashboard, will be visible as a stacktrace in the Dash debug panel. I found this really useful when I tested different approaches, because seeing only the "Action failed." notification when OPL or DF-Parameter fail is bad UX to me.

---

### My view per approach: 

1. I prefer exposing full api through the `notifications` field for `export_data` and `set_control` actions. Also, I would default an error msg for export_data and a success msg for set_control (like it is in the current PR version). I would skip showing any notification for private actions. I think we can currently skip exposing the `notifications` field for private actions. 
Advantages of this approach:
- Users will eventually see default notifications when using for example `set_control`.
- Users don't need to deal with `notifications` for builtin actions but full flexibility is available. It's available through the same API that's used for custom actions too which is, in my opinion, less confusing than learning how to use new `error_text` and `success_text`.. It works the same as for custom actions and serves full flexibility.

2. I don't like exposing `error_text` for `export_data` and `success_text` for `set_control` approach as it requires mental effort to learn what these fields are and how they affect notifications (maybe the naming is wrong). In that case we will eventually have to expose `progress_text` in one moment. Learning what fields exist for which action, documenting and maintaining these three fields per builtin action is harder than the same for only one: `notifications` that's already learnt, documented and maintained in custom actions.

But, if we skip exposing any kind of api for handling notifications in builtin actions, then only two options left:

3. set defaults (that can't be overwritten) - This could easily annoy our users as, in terms to change the defaults, they need to inherit an entire action.
4. not setting defaults at all - This is bad because we neither show new feature by default nor give our users space for using notifications in builtin actions. This could make our users feel that our builtin actions are less usable than custom actions (when they learn how to use notifications in custom actions they will try to do the same for builtin actions too). I see this option as postponing of finding the solution for the problem. We will eventually expose this for builtin actions with one of the above mentioned approaches. 

@l0uden would you mind sharing us your opinion too as you used conditional notifications quite a lot?

petar-qb: I removed everything related to condional notifications support for builtin actions. This will be done separately, and here the ticket for that -> https://github.com/McK-Internal/vizro-internal/issues/2657

petar-qb:
@maxschulz-COL, @huong-li-nguyen hey hey 👋 could you take a look at the conditional notifications PR?

Do it in a way you prefer, but I suggest checking the description, watching the screen recording, and trying to run the scratch_dev app to see how it feels. The most important file to review is `_action.py`, and the rest is mostly JS and tests.

This PR supports conditional notifications only for custom actions. Support for builtin actions will come later as described in [this conversation](https://github.com/mckinsey/vizro/pull/1574#discussion_r2948317299) and as agreed with AM. 

I will start working on the documentation for this PR early next week, which could be reviewed separately by someone else.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -304,41 +351,134 @@ def _action_callback_function(
...
         if not outputs:
+            # Allow: `None` and `notification_payload`.
+            if self._is_notification_payload(return_value):
+                notification_payload, return_value = return_value, None
             if return_value is not None:
-                raise ValueError("Action function has returned a value but the action has no defined outputs.")
+                raise ValueError(
+                    "Action function has returned a value but the action has no defined outputs. "
+                    "If you want to return a notification, make sure the returned value matches the notification key "
+                    "in the action's notifications."
```
huong-li-nguyen:
```suggestion
                    """Action function has returned a value but the action has no defined outputs. 
                    If you want to return a notification, make sure the returned value matches the notification key
                    in the action's notifications."""
```

huong-li-nguyen: Could we also add an example in this message saying something like "make sure xxxx e.g. {}"

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -304,41 +351,134 @@ def _action_callback_function(
...
+            # Allow: `return_value_mapping` and `(return_value_mapping, notification_payload)`.
+            return_value, notification_payload = self._split_trailing_notification_payload(return_value=return_value)
+
             if not isinstance(return_value, Mapping):
                 raise ValueError(
                     "Action function has not returned a dictionary-like object "
-                    "but the action's defined outputs are a dictionary."
+                    "but the action's defined outputs are a dictionary. "
+                    "If you want to return a notification, make sure the returned value matches the notification key "
+                    "in the action's notifications."
```
huong-li-nguyen: Might just be preference, but I normally prefer to use single quotes for one line and multiple for text going above one line :)
```suggestion
                    """but the action's defined outputs are a dictionary. 
                    If you want to return a notification, make sure the returned value matches the notification key
                    in the action's notifications."""
```

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -304,41 +351,134 @@ def _action_callback_function(
...
+                # Matches: `([rv_1, ..., rv_N], notification_payload)` for N defined outputs.
+                first_element_matches_output = (
+                    len(return_value) == 2  # noqa: PLR2004
+                    and hasattr(return_value[0], "__len__")
+                    and len(return_value[0]) == len(outputs)
+                )
+                if tuple_has_extra_output or first_element_matches_output:
+                    return_value, notification_payload = self._split_trailing_notification_payload(
+                        return_value=return_value
+                    )
```
huong-li-nguyen: I thought this part was soo confusing, because I couldn't imagine the specific use case when this takes effect. Do we need to support lists and dict? I assume we do, because we do it for all other actions? Or can we simplify this?

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -304,41 +351,134 @@ def _action_callback_function(
...
+            return [no_update] * len(outputs_spec)
+        if isinstance(outputs_spec, dict):
+            return dict.fromkeys(outputs_spec, no_update)
+        return no_update
+
+    def _render_notification(
+        self,
+        notification_key: str,
+        notification_result: Any | None = None,
+        error_msg: str | None = None,
```
huong-li-nguyen:
```suggestion
        error_msg: str | BaseException | None = None,
```

Is this always either a str or None? or can this also just return the live `BaseException` form the `except` block?

vizro-core/src/vizro/models/types.py
```diff
@@ -727,6 +771,14 @@ class _OptionsDictType(TypedDict):
 a list of strings, or a dictionary mapping strings to strings. Each output can be specified as
 `<model_id>` or `<model_id>.<argument_name>` or `<component_id>.<property>`. Defaults to `[]`."""
 
+
+ActionNotificationType = Annotated[
+    "dict[str, str | show_notification | update_notification | None]",
+    AfterValidator(_normalize_action_notifications),
+    Field(default_factory=dict, description="Conditional notifications", validate_default=True),
```
huong-li-nguyen: Can we add more details to the description here? This type has a lot more info implicitly contained than just conditional notifications right? :)

huong-li-nguyen (commented):
Hi @petar-qb - amazing work! I can only imagine how much effort you put into this! I've tested the `scratch_dev` and added my own examples too, with several error messages and it worked like a charm! I have to say that the code was a bit hard to digest, the logic and edge cases seemed very complex at first 😅  I'll try to take another look tomorrow.

I am really looking forward to the docs, as I think it will make the use cases a lot clearer. Heads up for navigating through all of this! 👏  


Have a few comments and questions:

> 
> case 0 -> return "Text output" => "success" variant="success" notification will be displayed with defined text.
> case 1 -> return "Text output", "my_success" => "my_success" variant="info" notification will be displayed with defined text.

This case I don't fully understand, why does it default to variant=success  if you just return "Text output", but if you return a tuple "Text output", "my_success" it  defaults to variant="info"? For me, these two cases don't differ that much except for returning additional text or maybe I don't understand these cases 🤔 


> There are three built-in notification keys: "progress", "success", and "error". If these are defined, they are shown automatically before the action starts ("progress") or after it finishes (either successfully ("success") or with an exception ("error")), 

Can't we have `warning` as a built-in notification key too? Pretty much everytime a UserWarning or something like this is raised?

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -107,6 +110,18 @@ def function(self):
         }
         return [notification]
 
+    @property
+    def _dash_components(self) -> list[Component]:
```
huong-li-nguyen, line 114: Could you explain this? I am not sure if I fully understand it

vizro-core/src/vizro/actions/_notifications.py
```diff
@@ -107,6 +110,18 @@ def function(self):
...
+        # Skip adding components for conditional notifications, which are typically required for the actions chain.
+        # These notifications are handled separately and are not part of the actions chain execution.
+        return [] if self._is_conditional else super()._dash_components
+
+    def _define_callback(self):
```
huong-li-nguyen, line 121: Same here - I didn't fully get this part

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -293,7 +311,36 @@ def _transform_output(output):
 
         return {output_name: _transform_output(output) for output_name, output in self._validated_outputs.items()}
 
-    def _action_callback_function(
+    def _is_notification_payload(self, value: Any) -> bool:
```
huong-li-nguyen, line 347: What's a payload? 😄

vizro-core/src/vizro/models/types.py
```diff
@@ -100,6 +101,49 @@ def _coerce_to_list(value: Any) -> Any:
...
+    """Normalize action notifications dict.
+
+    Input value: dict[str, str | show_notification | update_notification | None]
+    Return value: dict[str, show_notification | update_notification | None] with the following transformations:
+
+    - If "error" missing, default to "Action failed." (show vs update depends on presence of a "progress" notification).
+    - Convert existing string dict values to show_notification or update_notification:
+        * "progress" string key -> show_notification(variant="progress")
+        * other string keys -> update_notification(...) if progress exists, else show_notification(...)
+    - Mark all non-None notifications as conditional.
```
huong-li-nguyen, line 114: This was very helpful! 👏

vizro-core/src/vizro/models/types.py
```diff
@@ -100,6 +101,49 @@ def _coerce_to_list(value: Any) -> Any:
...
+    # If present, ensure that the "progress" is a notification action and not a string to extract its ID.
+    if isinstance(progress := value.get("progress"), str):
+        progress = value["progress"] = show_notification(text=progress, variant="progress")
+
+    # Convert all string notifications to actions and set _is_conditional=True.
+    for notif_key, notif_value in value.items():
+        if isinstance(notif_value, str):
+            defaults = {
+                "text": notif_value,
+                "variant": notif_key if notif_key in {"progress", "success", "error"} else "info",
```
huong-li-nguyen, line 130: Is there a reason we don't include warning?

vizro-core/src/vizro/models/types.py
```diff
@@ -100,6 +101,49 @@ def _coerce_to_list(value: Any) -> Any:
...
+            notification_action: show_notification | update_notification
+            if notif_key != "progress" and progress is not None:
+                notification_action = update_notification(**defaults, notification=progress.id)  # type: ignore[arg-type]
+            else:
+                notification_action = show_notification(**defaults)  # type: ignore[arg-type]
+
+            value[notif_key] = notification_action
+
+        if value[notif_key] is not None:
+            value[notif_key]._is_conditional = True
```
huong-li-nguyen, line 142: In which case is it None? If it is just a normal notification?

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -293,7 +311,36 @@ def _transform_output(output):
...
+        """Determine whether the given value represents a valid notification payload returned from the action.
+
+        The value is considered as the notification payload if it is either a string that matches a key in
+        `self.notifications`, or a tuple whose first element is a string that matches a key in `self.notifications`.
+        """
+        if isinstance(value, str):
+            key = value
+        elif isinstance(value, tuple) and len(value) == 2 and isinstance(value[0], str):  # noqa: PLR2004
```
huong-li-nguyen: Shall we make this a specific type rather than combining these checks?

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen, line 462: Can you just remind me when this happens? How does the user configure it? And why would they configure it like this?

vizro-core/src/vizro/models/_action/_action.py
petar-qb: ### 1. 
image

image

### 2. 
image

image

### 3. 
image

image

---

I only succeeded to get the clean error message output when using single quotes (third option). Even though the code looks prettier with the first option, I'd prefer doing the third option by paying a price of having less readable code.

vizro-core/src/vizro/models/_action/_action.py
petar-qb: I totally understand your confusion. I spent a long time trying not to break anything, to cover all possible use cases and still to preserve readable code as much as possible (this time I payed a price of having quite readable code).

When `Action.outputs` is defined, e.g: 
`Action.outputs = ["card_1", "text_2"]`

we already support these legacy return patterns:
- `return "asd", "qwe"`
- `return ["asd", "qwe"]`

To allow adding a `notification_payload` without changing existing returns, I introduced:
- `return "asd", "qwe", "my_success"`
- `return ["asd", "qwe"], "my_success"`

>Do we need to support lists and dict?

The `notification_payload` must be the last element of a **tuple** for any `Action.outputs` definition, so these are not allowed:
- `return ["asd", "qwe", "my_success"]`
- `return {"a": "asd", "b": "qwe", "notification_payload": "my_success"}`

Supporting both old and new patterns required a few not-so-pretty lines, but it keeps everything backward compatible.

I would like to simplify the code, if it doesn't require breaking the current use cases, but haven't found a way to do so. If you have any recommendations, please share here 😃

vizro-core/src/vizro/actions/_notifications.py
petar-qb: You probably haven't pulled the latest changes. I added this comment:
```py
    @property
    def _dash_components(self) -> list[Component]:
        # Skip adding components for conditional notifications, which are typically required for the actions chain.
        # These notifications are handled separately and are not part of the actions chain execution.
        return [] if self._is_conditional else super()._dash_components
```

I've improved this message further and will push the changes soon.

vizro-core/src/vizro/actions/_notifications.py
petar-qb: You probably haven't pulled the latest changes. I added this comment:
```py
    def _define_callback(self):
        # Skip defining callbacks for conditional notifications, which are typically required for the actions chain.
        # These notifications are handled separately and are not part of the actions chain execution.
        if not self._is_conditional:
            super()._define_callback()
```

I've improved this message further and will push the changes soon.

vizro-core/src/vizro/models/_action/_action.py
petar-qb: You probably haven't pulled the latest changes. I added this docstring for this function:
```py
        """Determine whether the given value represents a valid notification payload returned from the action.

        The value is considered as the notification payload if it is either a string that matches a key in
        `self.notifications`, or a tuple whose first element is a string that matches a key in `self.notifications`.
        """
```

vizro-core/src/vizro/models/types.py
petar-qb: There's a detailed conversation about "special/predefined" notification keys I suggested in [this comment](https://github.com/McK-Internal/vizro-internal/issues/2414#issuecomment-3758870659). At the end of this comment I explained why I would limit only on these three keys ("progress", "success" and "error"):
(I also recommend you reading the rest of the conversation if you're interested in.)
```
P.S. We should not predefine "on_info" or "on_warning" keys as we cannot know when these would be showed, unlike:

"in_progress": "Kedro pipeline is running"  # If specified, it's showed before the action is triggered
"on_success": "Kedro pipeline finished successfully!"  # If specified, it's showed when action finishes successfully (it doesn't raise an exception) and it doesn't return other notification flag 
"on_failure": "Exception happened"  # If specified, it's showed when action raises an exception.
```

So, if an additional `notification_payload` is not returned from the action's function, we know to distinguish only two cases: "success" (something returned) vs "error" (exception raised).

vizro-core/src/vizro/models/types.py
petar-qb: Good question. We treat `notifications={"error": None}` the same as `notifications={}`. Allowing a notification key to be set to `None` lets users override an action’s default notifications.

All custom actions for example have a default like `notifications={"error": "Action failed."}`, which dashboard creators can override or disable by setting it `{"error": None}`.

In the future, some built-in actions will also have default notifications, for example `export_data.notifications={"error": "Exporting failed."}`. Users will still be able to override these messages or to disable them with `{"error": None}`.

Read more about it in [this comment](https://github.com/McK-Internal/vizro-internal/issues/2414#issuecomment-3767813658).

vizro-core/src/vizro/models/_action/_action.py
petar-qb: Thank you very much for this comment! 🚀 

It opened my eyes, so I refactored the code by introducing the `NotificationPayload` model. Now, the code is easier to understand and modify in the future 😃

vizro-core/src/vizro/models/_action/_action.py
petar-qb: Does [this comment](https://github.com/mckinsey/vizro/pull/1574#discussion_r3063177696) answer your question?

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: I see, thanks for checking though!

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: No, all good! The explanations above already helped 👍

vizro-core/src/vizro/actions/_notifications.py
huong-li-nguyen: No, I did read the docstring! But I still didn't get it, sorry 😓 Let me be more specific: why do we even need to skip _dash_components for conditional notifications and handle them separately from the actions chain?

vizro-core/src/vizro/actions/_notifications.py
huong-li-nguyen: Let me close this discussion and continue in the other thread. I was reviewing based on the changes here, so it showed me the latest, I just didn't understand why we need to treat them differently and why these notifications need to be handled separately? Might be a stupid question 😄

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: No, I did read it! But I still didn't fully understand it until I saw your new type :)

Maybe we do something like this?

"""Determine whether a value is a valid notification payload.

A notification payload is a value returned by an action that signals a notification should be triggered.
It can be either:
- A string corresponding to a key in `self.notifications`, or
- A tuple whose first element is the notification name (matching a key in `self.notifications`),
  with optional additional data following it.

The optional additional data provides context for the notification handler, such as parameters,
metadata, or identifiers used to customize the notification's behavior or content.

Examples:
    "email_sent"
        -> Trigger the "email_sent" notification with default behavior.

    ("email_sent", {"user_id": 123})
        -> Trigger "email_sent" and pass a user ID to personalize the message.

    ("retry_failed", {"attempt": 2, "error": "timeout"})
        -> Trigger "retry_failed" with details about the failure.

    ("show_alert", "Something went wrong")
        -> Trigger "show_alert" with a message string.
"""

vizro-core/src/vizro/models/types.py
huong-li-nguyen: Got it, thanks!!

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: I like the new version a lot!

vizro-core/src/vizro/models/_action/_action.py
huong-li-nguyen: Yes, wondering if we should add these as comments in here? Or do you think it will convolute the code?

huong-li-nguyen (approved):
Hi @petar-qb - thanks for addressing some of the comments! 

I left some remaining questions, but they are all minor. Approving from my side and looking forward to the docs 👍

vizro-core/src/vizro/models/_action/_action.py
petar-qb: Thanks for the input! I added it

vizro-core/src/vizro/models/_action/_action.py
petar-qb: Thanks for the input, I reshaped the comment 😃

vizro-core/src/vizro/models/types.py
```diff
@@ -100,6 +101,49 @@ def _coerce_to_list(value: Any) -> Any:
...
+                "variant": notif_key if notif_key in {"progress", "success", "error"} else "info",
+            }
+
+            notification_action: show_notification | update_notification
+            if notif_key != "progress" and progress is not None:
+                notification_action = update_notification(**defaults, notification=progress.id)  # type: ignore[arg-type]
+            else:
+                notification_action = show_notification(**defaults)  # type: ignore[arg-type]
+
+            value[notif_key] = notification_action
```
antonymilne, line 139: Maybe this:
```suggestion
            notif_value = value[notif_key] = notification_action
```

and then use `notif_value` on L141 and L142?

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -293,7 +344,53 @@ def _transform_output(output):
...
+        except ValidationError:
+            return False
+
+        return notification_payload.key in getattr(self, "notifications", {})
+
+    def _split_trailing_notification_payload(self, return_value: Any) -> tuple[Any, Any | None]:
+        """If return_value is a tuple whose last element is a notification payload, split it off."""
+        if not isinstance(return_value, tuple) or return_value == ():
+            return return_value, None
+
```
antonymilne, line 383: It's fine how it is but maybe you prefer this? Very minor, up to you.

```suggestion
*external_return, notification_payload = return_value

```

Then on L387 `external_return` would be a list and you could do `[external_return] = external_return` in the `len == 1` case.

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -304,41 +401,131 @@ def _action_callback_function(
...
+            # Valid cases for `return_value` with a `notification_payload`:
+            #  - (rv_1, ..., rv_N, notification_payload): one extra element in a tuple beyond defined outputs
+            #  - ([rv_1, ..., rv_N], notification_payload): first element of a 2-tuple is list matching defined outputs
+            if isinstance(return_value, tuple):
+                # Matches: `(rv_1, ..., rv_N, notification_payload)` for N defined outputs.
+                tuple_has_extra_output = len(return_value) == len(outputs) + 1
+                # Matches: `([rv_1, ..., rv_N], notification_payload)` for N defined outputs.
+                first_element_matches_output = (
+                    len(return_value) == 2  # noqa: PLR2004
+                    and hasattr(return_value[0], "__len__")
```
antonymilne, line 456:
```suggestion
                    and isinstance(return_value[0], Sized)
```

Does this code work as expected when `return_value[0]` is a string that happens to be the same length as the number of outputs?

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -304,41 +401,131 @@ def _action_callback_function(
...
+        else:
+            # Allow: `Any` and `(Any, notification_payload)`.
+            # Breaking change: if a single output returns a tuple whose last element looks like `notification_payload`,
+            # it will be split into (return_value[:-1], notification_payload).
+            return_value, notification_payload = self._split_trailing_notification_payload(return_value=return_value)
+
+        return {
+            "external_return": return_value,
+            # The `notification_payload` matches the notification payload format or is None (default it to "success").
+            "notification_payload": NotificationPayload.model_validate(
```
antonymilne, line 487: This is very fancy using `model_validate`, I like it!

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -304,41 +401,131 @@ def _action_callback_function(
...
+        """Renders the notification based on the notification_key, notification_result and error_msg."""
+        # Skip setting `vizro_notification` output if the action does not support notifications.
+        if not (action_notifications := getattr(self, "notifications", None)):
+            return None
+
+        # Return no_update if the notification key does not exist, or its value is None.
+        if (notification_model := action_notifications.get(notification_key)) is None:
+            return no_update
+
+        # Template {{result}} with the empty string if the `result` is not provided.
```
antonymilne: I should have noticed this when I looked at the docs earlier but missed it. I think we discussed this before but am not sure what we said now...

Is there any reason we're doing this with double braces rather than single braces? Single braces feel much more natural to me since it matches f-strings and KPI cards for example.

If single braces makes sense to you too then I'd do like this:
```
# action definition 
notifications = {"success": "your result is {result}"}

# here
 def _render_notification(
        self,
        notification_key: str,
        **kwargs
        ):

injected_values = defaultdict(str, kwargs)
msg.format_map(injected_values)
```

vizro-core/src/vizro/static/js/models/action.js
```diff
@@ -46,9 +46,80 @@ function guard_action_chain(trigger_value, guard_data, trigger_component_id) {
...
 
+/**
+ * Replaces template variables in the format {{key}} within the given text with corresponding values from valuesMap.
+ *
+ * @param {string} text - The text containing template variables.
+ * @param {Object} valuesMap - An object mapping keys to their replacement values.
+ * @returns {string} The text with template variables replaced by their corresponding values.
+ */
+function replaceTemplateVariables(text, valuesMap) {
+  return text.replace(/\{\{(\w+)\}\}/g, (match, key) => {
```
antonymilne: If you like the idea of changing double braces to single ones for the outcome notifications then we should change here too.

antonymilne (approved):
Amazing work Petar! This was a big and complicated change and the result is great. I pushed some very small changes directly and have left several small comments. The only one that's really important is the use of double braces vs. single. I can't remember whether we discussed this before and decided on double - if so then leave it as is. But if you agree that single braces are better then let's change to that.

I couldn't find any unresolved conversations still requiring my attention (very good idea to make the notifications for built-in actions a separate issue so we don't get held up trying to solve it here 👍) but that might just be because the github UI has changed and I can't find the right button to press... So just let me know if I missed something.

vizro-core/src/vizro/models/_action/_action.py
petar-qb: `*external_return, notification_payload = return_value` converts `return_value[:-1]` to the list, but `external_return = return_value[:-1]` preserves the type (tuple if tuple, list if list..)

let's keep it as it is

vizro-core/docs/pages/user-guides/custom-actions.md
```diff
@@ -430,3 +431,573 @@ For example, let's alter the [above example](#trigger-with-a-runtime-input) of a
     === "Result"
 
         ![](../../assets/user_guides/actions/custom-actions4.png)
+
+## Notifications
+
+You can display notifications to show the progress and outcome of a custom action using the built-in  `"progress"`, `"success"` and `"error"` keys in the action's `notifications` argument:
```
stichbury:
```suggestion
You can use notifications to show the progress and outcome of a custom action with the built-in  `"progress"`, `"success"` and `"error"` keys in the action's `notifications` argument:
```

vizro-core/src/vizro/models/_action/_action.py
petar-qb: Very well spotted Antony!! 🦅 🚀 

You are right, but I have to say: "It's not a bug, it's a feature" 😄 

A while ago when we handled action outputs (without notifications) we found that Dash handles `"ab"` for two callback outputs in a way to return the `"a"` to the first output and the `"b"` to another. So, we decided not to have a special case for strings and to treat them the same way as tuples and lists (just like Dash does). 

So (lines copy pasted from old `test_action.py` tests), 
`("ab", [Output("component_1", "property"), Output("component_2", "property")])`
is treated as well as this:
`(["value_1", "value_2"], [Output("component_1", "property"), Output("component_2", "property")])`

Hence, I just continued working that way and I haven't introduced any special treatment for the strings. So in this PR, I've introduced the following valid test parameter:

```
            (
                ("ab", "notification_key"),
                "ab",
                "notification_key",
                [Output("component_1", "property"), Output("component_2", "property")],
            ),
```

(It's treated the same as:)

```
            (
                (["value_1", "value_2"], "notification_key"),
                ["value_1", "value_2"],
                "notification_key",
                [Output("component_1", "property"), Output("component_2", "property")],
            ),
```

vizro-core/docs/pages/user-guides/custom-actions.md
```diff
@@ -430,3 +431,573 @@ For example, let's alter the [above example](#trigger-with-a-runtime-input) of a
...
+    === "Result"
+
+        [![CustomNotificationKeys]][customnotificationkeys]
+
+
+### Dynamic notification content
+
+Notification messages can include _template variables_ that are filled with values from the action at runtime. This enables you to display dynamic information, such as results or error details, directly in the notification.
+
+Two templates are supported to make it possible to provide more informative and contextual feedback to users without hardcoding the message content:
```
stichbury:
```suggestion
There are two templates to provide more informative and contextual feedback to users without hardcoding the message content:
```

stichbury (approved):
LGTM! Two very small changes requested that are insignificant, feel free to ignore or include. 🌟 🚀

vizro-core/src/vizro/models/_action/_action.py
petar-qb: I don't remember why I used double braces.. Probably a short circuit in my head as I started with them from the beginning 😄. I have rewritten the code so single braces are used everywhere now.

PR #1576, opened 2026-02-10
[Feat] Enable horizontal navigation

nadijagraca:
## Description

closes: https://github.com/McK-Internal/vizro-internal/issues/2319

- Added new `position` argument to `NavBar` that can take either `left` or `top` values. It defaults to `left`. 

Current behaviour:
- NavBar(position="left") - default
   - always shows just icon + tooltip, no text is possible
   - label is mandatory and gives tooltip
   - icon is optional and gives filter_* etc. if not specified, otherwise gives specified icon

-  NavBar(position="top") - new option that makes NavLink behave as follows:
   - always shows just text, no tooltip or icon is possible
   - label is mandatory and gives text
   - icon is optional. If specified it raises error since you can't put an icon in horizontal nav? 

TO-DO:

- [x] Refactor code & relevant css
- [x] Add validations

## Notice


huong-li-nguyen:
@nadijagraca - hope it's fine I pushed myself. I am not sure how much time @stephkaiser has this week to take a look at this, so for now, I just implemented the same horizontal nav as Brix has it. That one already works for us. But to do that styling, I had to simplify some of the layout changes here too. If you add all the content to the header-left instead, you don't need all that additional custom CSS etc. :) Please check if I didn't break anything, and feel free to refactor the code properly. For now, I would say we go with this style unless @stephkaiser has time this week.

vizro-core/src/vizro/models/_navigation/nav_bar.py
```diff
@@ -37,11 +36,14 @@ class NavBar(VizroBaseModel):
         Field(default={}, description="Mapping from name of a pages group to a list of page IDs/titles."),
     ]
     items: list[NavLink] = []
+    position: str = "left"
```
huong-li-nguyen: Should probably be a Literal["left", "top"] with some field description

vizro-core/docs/pages/user-guides/navigation.md
```diff
@@ -422,9 +422,158 @@ You can alter the icons used by specifying the name of the icon in the [Google M
 
         [![CustomIcons]][customicons]
 
+## Use a horizontal navigation bar
```
stichbury:
```suggestion
## Horizontal navigation
```

vizro-core/docs/pages/user-guides/navigation.md
```diff
@@ -422,9 +422,158 @@ You can alter the icons used by specifying the name of the icon in the [Google M
...
+            ),
+        )
+
+        Vizro().build(dashboard).run()
+        ```
+
+    === "app.yaml"
+
+        ```yaml
+        # Still requires a .py to add data to the data manager and parse YAML configuration
```
stichbury, line 475: This is meant to be here? It is a bit confusing, no?

vizro-core/docs/pages/user-guides/navigation.md
```diff
@@ -422,9 +422,158 @@ You can alter the icons used by specifying the name of the icon in the [Google M
...
+              - My third page
+        ```
+
+    === "Result"
+
+        [![BasicHorizontalNavigation]][basichorizontalnavigation]
+
+### Grouped pages with horizontal navigation
+
+You can still group multiple pages under a single navigation item in horizontal mode either by specifying `pages` as a dictionary or providing `NavLink` models in `items`.
```
stichbury:
```suggestion
You can group multiple pages under a single navigation item in horizontal mode either by specifying `pages` as a dictionary or providing `NavLink` models in `items`.
```

vizro-core/changelog.d/20260223_145451_nadija_ratkusic_graca_horizontal_navigation.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Added `position` argument to `vm.NavBar` model to enable horizontal navigation. See the [user guide on horizontal navigation](https://vizro.readthedocs.io/en/stable/pages/user-guides/navigation/#use-a-horizontal-navigation-bar) ([#1576](https://github.com/mckinsey/vizro/pull/1576))
```
petar-qb:
```suggestion
- Add `position` argument to `vm.NavBar` model to enable horizontal navigation. See the [user guide on horizontal navigation](https://vizro.readthedocs.io/en/stable/pages/user-guides/navigation/#use-a-horizontal-navigation-bar) ([#1576](https://github.com/mckinsey/vizro/pull/1576))
```

vizro-core/src/vizro/models/_navigation/nav_bar.py
```diff
@@ -37,28 +36,25 @@ class NavBar(VizroBaseModel):
...
+        for nav_link in self.items:
+            nav_link._nav_position = self.position
 
         for position, item in enumerate(self.items, 1):
-            # The default icons are named filter_1, filter_2, etc. up to filter_9.
-            # If there are more than 9 items, then the 10th and all subsequent items are named filter_9+.
-            icon_default = f"filter_{position}" if position <= 9 else "filter_9+"  # noqa: PLR2004
-            item.icon = item.icon or icon_default
+            # The default icons are added only when position == "left"
+            if self.position == "left":
```
petar-qb: Let's switch these two lines with the outer for loop so the code loops only `if self.position == "left"`

vizro-core/src/vizro/models/_navigation/nav_bar.py
```diff
@@ -83,6 +79,24 @@ def build(self, *, active_page_id=None) -> _NavBuildType:
             # Active page is not in navigation at all, so hide navigation panel.
             nav_panel = dbc.Nav(id="nav-panel", className="d-none invisible")
 
-        # `flex-column` ensures that we return a vertical NavBar. In the future, we could use that className
-        # to create a horizontal NavBar.
-        return html.Div(children=[dbc.Navbar(id="nav-bar", children=nav_links, className="flex-column"), nav_panel])
+        # `flex-column` ensures that we return a vertical NavBar. For top position, we use `navbar-top` className.
+        navbar_class = "navbar-top" if self.position == "top" else "flex-column"
```
petar-qb: Really really minor thing so feel free to ignore. I prefer questioning `if self.position == "left"` as the same condition is in the `pre_build`, and because `flex-column` is first mentioned in the comment above, so I find it just a bit easier to follow 😄 
```suggestion
        navbar_class = "flex-column" if self.position == "left" else "navbar-top"
```

vizro-core/src/vizro/models/_navigation/nav_link.py
```diff
@@ -31,38 +31,51 @@ class NavLink(VizroBaseModel):
...
                     placement="right",
                     target=f"{self.id}-tooltip-target",
                 ),
-            ],
+            ]
+            if is_left_nav
+            else [
+                html.Span(self.label),
+            ]
+        )
```
petar-qb, line 71: `is_left_nav` is only used in one place, so I'd remove it. 
```suggestion
        nav_link_children = [
            html.Span(self.icon, className="material-symbols-outlined", id=f"{self.id}-tooltip-target"),
            dbc.Tooltip(
                self.label,
                placement="right",
                target=f"{self.id}-tooltip-target",
            ),
        ] if self._nav_position == "left" else [html.Span(self.label)]
```

petar-qb (commented):
Looks good so far 🏅I only reviewed half of the PR and the rest will do after some meetings. 

btw, is it still WIP?

vizro-core/src/vizro/models/_navigation/nav_link.py
```diff
@@ -50,19 +56,24 @@ def build(self, *, active_page_id=None):
...
                     self.label,
                     placement="right",
                     target=f"{self.id}-tooltip-target",
                 ),
-            ],
+            ]
+            if self._nav_position == "left"
+            else [
+                html.Span(self.label),
+            ]
```
petar-qb:
```suggestion
            else [html.Span(self.label)]
```

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -117,6 +117,19 @@ class Dashboard(VizroBaseModel):
         ),
     ]
 
+    @property
+    def _is_left_navigation(self) -> bool:
+        """Returns True if navigation is positioned on the left side."""
+        if self.navigation is None:
+            raise ValueError("Navigation must be initialized")
```
petar-qb: Let's extract this in the `pre_build` method so it's checked only once.

Also, this case should be covered with a unit test.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -117,6 +117,19 @@ class Dashboard(VizroBaseModel):
...
+
+        nav_selector = self.navigation.nav_selector
+
+        if nav_selector is None or isinstance(nav_selector, Accordion):
+            return True
+
+        return nav_selector.position == "left"
```
petar-qb:
```suggestion
        nav_selector = self.navigation.nav_selector

        return nav_selector is None or isinstance(nav_selector, Accordion) or nav_selector.position == "left"
```

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -425,10 +422,21 @@ def _outer_page(self, inner_page: _InnerPageContentType) -> _OuterPageContentTyp
...
+        left_hidden = _all_hidden([header_left])
+        right_hidden = _all_hidden([header_right])
+        both_hidden = left_hidden and right_hidden
+
+        header = html.Div(
+            id="header",
+            children=[header_left, header_right],
+            hidden=both_hidden,
+            className="no-left" if left_hidden else "",
+        )
```
petar-qb: I would stick with the previous implementation as `_all_hidden` is called twice in both cases and there are three variables more in your solution. 

```suggestion
        header = html.Div(
            id="header",
            children=[header_left, header_right],
            hidden=_all_hidden([header_left, header_right]),
            className="no-left" if _all_hidden(header_left) else "",
        )
```

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -448,9 +456,20 @@ def _arrange_page(self, outer_page: _OuterPageContentType):
...
-        page_main = html.Div(id="page-main", children=[nav_bar, collapse_left_side, collapse_icon_outer, right_side])
+        page_main_children = (
+            [nav_bar, collapse_left_side, collapse_icon_outer, right_side]
+            if self._is_left_navigation
+            else [collapse_left_side, collapse_icon_outer, right_side]
+        )
+        page_main = html.Div(
+            id="page-main",
+            children=page_main_children,
+        )
```
petar-qb: In my opinion, this file contains too many utility variables, so I suggest removing those that are not reused.
Don't apply this if you don't agree. I haven't worked with this model for a while and can't say with certainty what's easier to read/maintain.
```suggestion
        page_main = html.Div(
            id="page-main",
            children=[nav_bar, collapse_left_side, collapse_icon_outer, right_side]
            if self._is_left_navigation
            else [collapse_left_side, collapse_icon_outer, right_side],
        )
```

vizro-core/tests/unit/vizro/models/_navigation/test_nav_bar.py
```diff
@@ -76,6 +78,13 @@ def test_items_with_with_pages_icons(self, pages_as_dict):
         assert nav_bar.items[0].icon == "home"
         assert nav_bar.items[1].icon == "filter_2"
 
+    def test_items_top_position(self, pages_as_dict):
+        nav_bar = vm.NavBar(pages=pages_as_dict, position="top")
+        nav_bar.pre_build()
+        assert all(isinstance(item, vm.NavLink) for item in nav_bar.items)
+        assert all(item.icon == "" for position, item in enumerate(nav_bar.items, 1))
+        assert all(item._nav_position == "top" for position, item in enumerate(nav_bar.items, 1))
```
petar-qb:
```suggestion
    def test_items_top_position(self, pages_as_dict):
        nav_bar = vm.NavBar(pages=pages_as_dict, position="top")
        nav_bar.pre_build()
        assert all(isinstance(item, vm.NavLink) for item in nav_bar.items)
        assert all(item.icon == "" for item in nav_bar.items)
        assert all(item._nav_position == "top" for item in nav_bar.items)
```

vizro-core/tests/unit/vizro/models/_navigation/test_nav_bar.py
```diff
@@ -198,7 +210,53 @@ def test_nav_bar_not_active_pages_as_list(self, pages_as_list):
...
+        nav_bar = vm.NavBar(pages=pages_as_dict, position="top")
+        nav_bar.pre_build()
+        nav_bar.items[0].id = "nav-link-1"
+        built_nav_bar = nav_bar.build(active_page_id=page_1_id)
+        expected_navigation = dbc.Navbar(
+            children=[
+                dbc.NavLink(
+                    children=[
+                        html.Span("Group"),
+                    ],
```
petar-qb: Apply this in other places too where there's only one element in the list and where it looks good.
```suggestion
                    children=[html.Span("Group")],
```

petar-qb (approved):
Looks good to me! 🚀 I left several (mostly minor) comments, hence I already approved the PR. 🥇

petar-qb:
FYI @nadijagraca 
e2e `test_url_params_encoding_and_page_refresh` and `test_url_params_decoding_and_navigate_to_page` are flaky tests that are not broken by this PR. This will be addressed by the end of the week.

vizro-core/src/vizro/static/css/vizro_overwrites.css
```diff
@@ -66,14 +66,15 @@ All the HEX values starting with --text-code are taken from the Github code high
   background: var(--bs-tertiary-bg) !important;
   border: none;
   flex-wrap: nowrap;
+  min-height: 0 !important;
```
huong-li-nguyen: Do we need the `min-height`?

vizro-core/src/vizro/models/_navigation/nav_bar.py
antonymilne: Very unimportant but I agree, just because `position` is `left` by default.

vizro-core/src/vizro/models/_navigation/nav_link.py
```diff
@@ -31,13 +31,19 @@ class NavLink(VizroBaseModel):
         Field(default="", description="Icon name from Google Material icons library."),
     ]
     _nav_selector: Accordion = PrivateAttr()
+    _nav_position: Literal["left", "top"] = PrivateAttr(default="left")
```
antonymilne, line 34:
```suggestion
    _nav_position: Literal["left", "top"] = PrivateAttr()
```

Probably not needed since it's always set by `NavBar`?

vizro-core/src/vizro/models/_navigation/nav_link.py
```diff
@@ -31,13 +31,19 @@ class NavLink(VizroBaseModel):
...
 
     @_log_call
     def pre_build(self):
         from vizro.models._navigation.accordion import Accordion
 
         self._nav_selector = Accordion(pages=self.pages)  # type: ignore[arg-type]
 
+        if self.icon and self._nav_position == "top":
+            raise ValueError(
+                "You cannot use icons with top navigation. Icons are only supported for the left navigation."
```
antonymilne:
```suggestion
                'You cannot use icons when the `vm.NavBar` has `position="top"`. Icons are only supported for `position="left"`.'
```

vizro-core/src/vizro/models/_navigation/nav_link.py
```diff
@@ -50,19 +56,22 @@ def build(self, *, active_page_id=None):
...
                 html.Span(self.icon, className="material-symbols-outlined", id=f"{self.id}-tooltip-target"),
                 dbc.Tooltip(
                     self.label,
                     placement="right",
                     target=f"{self.id}-tooltip-target",
                 ),
-            ],
+            ]
+            if self._nav_position == "left"
+            else [html.Span(self.label)]
```
antonymilne:
```suggestion
            else self.label
```

Probably no span or list needed here? You might need to adjust some CSS if you rely on it though.

vizro-core/src/vizro/models/_dashboard.py
antonymilne: When can the `None` case even happen? I think it's probably not even possible when this function actually runs?

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -117,6 +117,16 @@ class Dashboard(VizroBaseModel):
...
 
+    @property
+    def _is_left_navigation(self) -> bool:
+        """Returns True if navigation is positioned on the left side."""
+        if self.navigation is None:
+            raise ValueError("Navigation must be initialized")
+
+        nav_selector = self.navigation.nav_selector
+
+        return nav_selector is None or isinstance(nav_selector, Accordion) or nav_selector.position == "left"
```
antonymilne: Maybe simpler if we make this detect whether it's top navigation instead, given that's the simpler case to check? Then the check would just become (no need to check for `None`)
```
getattr(nav_selector, "position", None) == "top"
```

And we can do `_is_navbar_position_top` instead.

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -425,10 +419,17 @@ def _outer_page(self, inner_page: _InnerPageContentType) -> _OuterPageContentTyp
...
+            children=[header_left, header_right],
+            hidden=_all_hidden([header_left, header_right]),
+            className="no-left" if _all_hidden(header_left) else "",
+        )
+
         return html.Div(
             [
                 header,
-                nav_bar,
+                nav_bar if self._is_left_navigation else None,
```
antonymilne: Is this check really needed? What happens if you just include `nav_bar` regardless? That's what we did before this change and it worked even when there was no navbar, so I feel like it should still work now?

vizro-core/docs/pages/user-guides/navigation.md
```diff
@@ -422,9 +422,158 @@ You can alter the icons used by specifying the name of the icon in the [Google M
 
         [![CustomIcons]][customicons]
 
+## Horizontal navigation
+
+For a more compact layout, you can position the navigation bar horizontally at the top of your dashboard.
+
+### Basic horizontal navigation
+
+To create a horizontal navigation bar, set the position argument to `top` in your `NavBar` model. In this scenario, individual page titles are displayed as navigation links across the top of the dashboard..
```
antonymilne:
```suggestion
To create a horizontal navigation bar, set `position="top"` in your `NavBar` model. In this scenario, individual page titles are displayed as navigation links across the top of the dashboard.
```

vizro-core/docs/pages/user-guides/navigation.md
```diff
@@ -422,9 +422,158 @@ You can alter the icons used by specifying the name of the icon in the [Google M
...
+            components=[
+                vm.Card(text="My text here"),
+            ],
+        )
+
+        dashboard = vm.Dashboard(
+            pages=[page_1, page_2, page_3],
+            navigation=vm.Navigation(
+                nav_selector=vm.NavBar(
+                    pages=["My first page", "My second page", "My third page"],
```
antonymilne:
```suggestion
```

I think this can probably be removed and it will work the same? In which case `hl_lines="31"` needs to change also.

vizro-core/docs/pages/user-guides/navigation.md
```diff
@@ -422,9 +422,158 @@ You can alter the icons used by specifying the name of the icon in the [Google M
...
+        # See yaml_version example
+        # pages defined as in default example
+        navigation:
+          nav_selector:
+            type: nav_bar
+            position: top
+            pages:
+              - My first page
+              - My second page
+              - My third page
```
antonymilne:
```suggestion
```

vizro-core/docs/pages/user-guides/navigation.md
```diff
@@ -422,9 +422,158 @@ You can alter the icons used by specifying the name of the icon in the [Google M
...
+
+    === "Result"
+
+        [![BasicHorizontalNavigation]][basichorizontalnavigation]
+
+### Grouped pages with horizontal navigation
+
+You can group multiple pages under a single navigation item in horizontal mode either by specifying `pages` as a dictionary or providing `NavLink` models in `items`.
+
+- By using a dictionary: When you provide pages as a dictionary to the `NavBar` model, the dictionary keys become the horizontal navigation item labels, and the associated pages appear in an accordion in the side panel below:
```
antonymilne:
```suggestion
- By using a dictionary: When you provide `pages` as a dictionary to the `NavBar` model, the dictionary keys become the horizontal navigation item labels, and the associated pages appear in an accordion in the side panel below.
```

vizro-core/changelog.d/20260223_145451_nadija_ratkusic_graca_horizontal_navigation.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+
+### Added
+
+- Add `position` argument to `vm.NavBar` model to enable horizontal navigation. See the [user guide on horizontal navigation](https://vizro.readthedocs.io/en/stable/pages/user-guides/navigation/#use-a-horizontal-navigation-bar) ([#1576](https://github.com/mckinsey/vizro/pull/1576))
```
antonymilne:
```suggestion
- Add `position` argument to `vm.NavBar` model to enable horizontal navigation. See the [user guide on horizontal navigation](https://vizro.readthedocs.io/en/stable/pages/user-guides/navigation/#use-a-horizontal-navigation-bar). ([#1576](https://github.com/mckinsey/vizro/pull/1576))
```

vizro-core/changelog.d/20251110_152146_90609403+huong-li-nguyen_make_themes_public.md
antonymilne, line 1: This shouldn't be here :)

stephkaiser:
design looks good to me, no comments :)

vizro-core/src/vizro/models/_navigation/nav_link.py
nadijagraca: Unit tests are failing once I remove the `default`.  @huong-li-nguyen can you take a look during the review?

vizro-core/docs/pages/user-guides/navigation.md
```diff
@@ -422,9 +422,153 @@ You can alter the icons used by specifying the name of the icon in the [Google M
...
+                vm.Card(text="My text here"),
+            ],
+        )
+
+        dashboard = vm.Dashboard(
+            pages=[page_1, page_2, page_3],
+            navigation=vm.Navigation(
+                nav_selector=vm.NavBar(
+                    position="top",
+                )
```
huong-li-nguyen: Can you just double-check the formatting in all the code snippets? I think it doesn't need to be on the next line in many cases
```suggestion
            navigation=vm.Navigation(nav_selector=vm.NavBar(position="top")
```

vizro-core/schemas/0.1.52.json
```diff
@@ -1316,6 +1316,13 @@
           },
           "title": "Items",
           "type": "array"
+        },
```
huong-li-nguyen: I think only one json file needs to change?

vizro-core/src/vizro/models/_navigation/nav_link.py
petar-qb: The `_nav_position` is only used in `NavLink.pre_build` and `NavLink.build`, and is set in `NavBar.pre_build`. However, there’s no guarantee that the `_nav_position` is set (`NavBar.pre_build` executed) before `NavLink.pre_build` is executed. Aside from filters, that are "pre-built" first, there's no order for the rest of the models (see Vizro.pre_build from _vizro.py).

I’d suggest the following:
1. Remove the condition `if self.icon and self._nav_position == "top":` completely. From a user perspective, validation like this is restrictive (bad UX in my opinion) as changing `NavBar.position` from `left` to `top` should not require removing all icons before rerunning the app.
2. Remove the default value of `_nav_position` and set it explicitly in tests where needed. If that’s not feasible, keep the default as `"left"`.

~Alternatively, we could remove the default value and enforce execution order by ensuring `NavBar.pre_build` runs before `NavLink.pre_build` (by adjusting `Vizro.pre_build`).~

~So currently, this code:~

```py
        if self.icon and self._nav_position == "top":
            raise ValueError(
                'You cannot use icons when the `vm.NavBar` has `position="top"`. '
                'Icons are only supported for `position="left"`.'
            )
```

~cannot be tested deterministically, as "in the real world" setting the `NavLink._nav_position` to `"top"`, only sometimes happen before the mentioned code execution.~

---

EDIT:

I just noticed that the `NavLink.pre_build` is called **again** from the `NavBar.pre_build` which ensures that the mentioned validation always raises the ValueError 👍. 
However, I still think this validation is unnecessary. If you disagree, I'm ok with leaving this validation. Also, try to set the `_nav_position` from the tests if possible. Otherwise leave the `default="left"` too.

vizro-core/src/vizro/static/css/vizro_overwrites.css
```diff
@@ -229,6 +229,7 @@ All the HEX values starting with --text-code are taken from the Github code high
...
```
huong-li-nguyen:
```suggestion
```
Not sure if you checked whether we really needed it?

vizro-core/src/vizro/static/css/layout.css
```diff
@@ -361,3 +362,30 @@ Current logic:
...
   border: 1px solid var(--bs-border-color);
 }
+
+#header-left .navbar-top {
+  border: none;
+  display: flex;
+  gap: 2rem;
+  height: 100%;
+  padding: 0;
+  padding-left: 24px;
```
huong-li-nguyen, line 371:
```suggestion
  display: flex;
  gap: 2rem;
  padding: 0;
  padding-left: 24px;
```

I think not needed?

vizro-core/src/vizro/models/_dashboard.py
```diff
@@ -117,6 +117,13 @@ class Dashboard(VizroBaseModel):
         ),
     ]
 
+    @property
+    def _is_top_navigation(self) -> bool:
+        """Returns True if navigation is positioned on the top side."""
+        nav_selector = getattr(self.navigation, "nav_selector", None)
+
+        return getattr(nav_selector, "position", None) == "top"
```
huong-li-nguyen: I think I would make it more specific, as only the NavBar can have the position argument. 

```suggestion
    def _is_top_navigation(self) -> bool:
        """Returns True if navigation is positioned on the top side."""
        if self.navigation is None:
            return False
        nav_selector = self.navigation.nav_selector
        return isinstance(nav_selector, NavBar) and nav_selector.position == "top"

```

vizro-core/src/vizro/models/_navigation/nav_link.py
petar-qb: @nadijagraca I just talked with Antony about this, I think the last open question that should be solved before this PR is merged. We agreed that this validation in `NavLink.pre_build` is unnecessary and that we should show a warning instead, similarly like it's done in the `graph.py` and other files:

```py
            warnings.warn(
                "Using the `title` argument in your Plotly chart function may cause misalignment with "
                "other component titles on the screen. To ensure consistent alignment, consider using "
                "`vm.Graph(title='Title', ...)`.",
                UserWarning,
            )
```

So, let's replace this validation with something like this: `warnings.warn("Your warning msg", UserWarning)` instead. 

P.S. You'll probably need to "silent" this warning in tests that raises it.

Will that enable to remove the `default="left"`?

PR #1581, opened 2026-02-17
[Feat] Add `variant` argument to dropdown

nadijagraca:
## Description
closes: https://github.com/McK-Internal/vizro-internal/issues/2374

Add `variant` argument to `vm.Dropdown`. Options are `["filled", "plain"]` Default option is `filled`. 

- `variant="filled"` - everything stays the same, our default styling
- `variant="plain"` - new style (transparent box with bottom border)

## Notice


vizro-core/src/vizro/static/css/vizro_overwrites.css
```diff
@@ -137,3 +137,19 @@ div.form-check.form-switch {
 legend {
   font-size: inherit;
 }
+
+.dropdown--plain {
```
antonymilne: Is this meant to have `--` rather than `-`?

vizro-core/examples/scratch_dev/app.py
antonymilne, line 1: Let's have a changelog entry for this 🙂

petar-qb:
I pushed one more page that stress tests different dropdowns with the variant prop defined. Here's how it looks like:

image

image

@nadijagraca could you take a look and see whether the CSS has to be fixed for any case?

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -57,6 +57,10 @@ class Dropdown(VizroBaseModel):
     title: str = Field(default="", description="Title to be displayed")
     # TODO: ideally description would have json_schema_input_type=str | Tooltip attached to the BeforeValidator,
     #  but this requires pydantic >= 2.9.
+    variant: Literal["plain", "filled"] = Field(
+        default="filled",
+        description="Predefined styles to choose from. Options are `filled` or `plain`.",
+    )
     description: Annotated[
         Tooltip | None,
         BeforeValidator(coerce_str_to_tooltip),
```
petar-qb: Let's move the `"# TODO: ideally description..."` below `variant` definition.

vizro-core/src/vizro/static/css/vizro_overwrites.css
```diff
@@ -95,7 +95,7 @@ All the HEX values starting with --text-code are taken from the Github code high
   padding-left: 0;
 }
 
-.container-controls-panel .dropdown {
+.container-controls-panel .dropdown, .dropdown-plain {
```
petar-qb: By changing this, the last page of the new scratch_dev will look much better 😃. 
```suggestion
.container-controls-panel .dropdown, .dropdown--plain {
```

Or rename `.dropdown--plain` -> `.dropdown-plain` as AM said.

vizro-core/src/vizro/models/_components/form/dropdown.py
```diff
@@ -111,6 +115,7 @@ def _action_inputs(self) -> dict[str, _IdProperty]:
         return {"__default__": f"{self.id}.value"}
 
     def __call__(self, options):
+        variants = {"filled": "dropdown", "plain": "dropdown--plain"}
```
petar-qb: If we do this: 
```suggestion
        variants = {"filled": "dropdown", "plain": "dropdown dropdown-plain"}
```

I guess we wouldn't need the following change anymore:
`.container-controls-panel .dropdown` -> `.container-controls-panel .dropdown, .dropdown-plain` in the `vizro_overwrites.css`. 

If so, then I prefer that syntax as the code becomes easier to maintain (any change for dropdown will automatically be applied to all variants)

huong-li-nguyen: I've changed it to `.container-controls-panel .dash-dropdown`  and `variants = {"filled": "", "plain": "dropdown-plain"}` :) 

It will have the same effect, but we are more consistent then (similar logic as in container)

huong-li-nguyen:
I pushed some changes, because we are deviating more from the qbds design, so had to find some middle ground until Steph comes back :) Sorry, we couldn't provide a proper design for you to implement just yet 😅 

This is good to go for me, after resolving the remaining comments 👍

vizro-core/examples/scratch_dev/app.py
huong-li-nguyen: Let's add some docs then too @nadijagraca. Then we can link to that in the changelog entry. Basically just adding a quick example that a variant can be configured :)

vizro-core/src/vizro/static/css/vizro_overwrites.css
```diff
@@ -95,8 +95,7 @@ All the HEX values starting with --text-code are taken from the Github code high
   padding-left: 0;
 }
 
-.container-controls-panel .dropdown, .dropdown-plain {
-  max-width: 240px;
```
nadijagraca, line 97: We need to leave the max-width limit of 240px because in case of long values it will take up most of the horizontal space in the container.

huong-li-nguyen: Maybe let's do 280 then? Seems to be the default dropdown width in the qbds

petar-qb: @huong-li-nguyen @nadijagraca Any qbds quidelines for min/max width for sliders? 
I think that the `min-width=200px` is currently set for sliders, and I find it pretty narrow especially RangeSlider that has two numerical inputs). 

image

What do you think?

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -25,6 +25,62 @@ If you have binary data (such as `False`/`True` or `0`/`1`), you might prefer to
 
     The later is required if you want to provide different display labels to your option values or in case you want to provide boolean values as options. In this case, you need to provide a string label for your boolean values as boolean values cannot be displayed properly as labels in the underlying Dash components.
 
+### Styled dropdown
+
+There are two predefined dropdown styles that can be customized using the variant argument. If no `variant` is specified, the default style applied is `variant="filled"`.
```
stichbury:
```suggestion
You can customize two predefined dropdown styles using the variant argument. If no `variant` is specified, the default style applied is `variant="filled"`.
```

vizro-core/src/vizro/static/css/vizro_overwrites.css
huong-li-nguyen: In the QBDS Figma files, sliders are set to 240px and dropdowns to 280px, so you can use those if it looks better to you.

However, the QBDS doesn't specify any min/max sizes. The components just stretch to fit their container. Those widths are just examples, not official guidelines :)

vizro-core/docs/pages/user-guides/selectors.md
```diff
@@ -25,6 +25,62 @@ If you have binary data (such as `False`/`True` or `0`/`1`), you might prefer to
...
+
+        iris = px.data.iris()
+
+        page = vm.Page(
+            title="Styled dropdown",
+            components=[
+                vm.Graph(figure=px.scatter(iris, x="sepal_length", y="petal_width", color="species")),
+            ],
+            controls=[
+                vm.Filter(column="species", selector=vm.Dropdown(variant="plain")),
```
huong-li-nguyen: I would show both so it's clearer what the difference is
```suggestion
               vm.Filter(column="species", selector=vm.Dropdown(title="Filled")),
                vm.Filter(column="species", selector=vm.Dropdown(variant="plain", title="Plain")),
```

PR #1626, opened 2026-02-26
[CI] Exclude breaking `virtualenv!=21.0.0`

petar-qb:
closes https://github.com/McK-Internal/vizro-internal/issues/2510

## Description
[Hatch does not work with virtualenv 21.0.0](https://github.com/pypa/hatch/issues/2193)

PR that fixes this bug on their side: https://github.com/pypa/hatch/pull/ 2195

Consider to remove the `'virtualenv!=21.0.0'` from all places when pypa team fixes it in the new `virtualenv==21.0.1`.

Here's a [ticket](https://github.com/McK-Internal/vizro-internal/issues/2511) for that so we don't forget to do so.

## Notice

### DIFF
--- a/vizro-ai/.readthedocs.yaml
+++ b/vizro-ai/.readthedocs.yaml
@@ -11,7 +11,7 @@ build:
     python: "3.13"
 
   commands:
-    - pip install hatch
+    - pip install hatch 'virtualenv!=21.0.0'
     - cd vizro-ai/ && hatch run docs:pip tree
     - cd vizro-ai/ && hatch run docs:build
     - mkdir --parents $READTHEDOCS_OUTPUT
--- a/vizro-core/.readthedocs.yaml
+++ b/vizro-core/.readthedocs.yaml
@@ -10,7 +10,7 @@ build:
   tools:
     python: "3.13"
   commands:
-    - pip install hatch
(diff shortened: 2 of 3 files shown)


.github/workflows/build-vizro-whl.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         with:
           python-version: ${{ env.PYTHON_VERSION }}
       - name: Install Hatch
-        run: pip install hatch
+        run: pip install hatch 'virtualenv!=21.0.0'
```
maxschulz-COL, line 32:
```suggestion
        run: pip install hatch 'virtualenv~=20.0'
```
Would this not be a better change?

petar-qb: I think we want to include `21.0.1` as [the issue](https://github.com/pypa/hatch/issues/2193) escalated quite widely. `pypa` team is going to release the fix very soon (https://github.com/pypa/hatch/pull/2195) and then we can remove this constraint or replace it with the 'virtualenv>21.0.0' or leave it as it is.

maxschulz-COL: ok dokes

PR #1629, opened 2026-02-26
[Feat] Allow configuring icons for accordion-headers

huong-li-nguyen:
## Description
Closes https://github.com/McK-Internal/vizro-internal/issues/2453
**Context:** We considered switching the accordion to dmc to support icons, so I compared these two, and the dmc version didn't add much functionality compared to the dbc version (e.g. it also didn't fix the the collapse/uncollapse behaviour @lingyielia mentioned once). The main benefit would have been icons, which we can support with dbc still and by adding a bit of configuration. This PR adds that. 

API is not fixed yet, need to align with @antonymilne . I found the different nav selector combinations make it easy for things to get confusing. Another question is whether we allow icons on per-page nav links, or other components, which might also influence the API

**Accordion**
- `icons`: optional mapping from group name → Material icon name (accordion header). Validated: keys must be page group names; values must be non-empty.

**NavLink**
`accordion`: optional `Accordion` to configure the nested accordion (e.g. icons). Pages still come from the NavLink; icon keys are validated after sync.

**Navigation**
No API change. Same Accordion model for default accordion and nav_selector=vm.Accordion(...).

**Examples:** 

- scratch_dev (accordion-only with group icons), 
- dev (NavBar + accordion=vm.Accordion(icons={...})).

## Screenshot
Screenshot 2026-02-26 at 14 06 11
Screenshot 2026-02-26 at 13 45 53

## Notice

### DIFF
--- a/.github/workflows/build-vizro-whl.yml
+++ b/.github/workflows/build-vizro-whl.yml
@@ -29,7 +29,7 @@ jobs:
         with:
           python-version: ${{ env.PYTHON_VERSION }}
       - name: Install Hatch
-        run: pip install hatch
+        run: pip install hatch 'virtualenv!=21.0.0'
       - name: Build vizro-core package
         run: hatch build -t wheel
       - name: Upload vizro-core artifact
--- a/.github/workflows/checks-vizro-ai.yml
+++ b/.github/workflows/checks-vizro-ai.yml
@@ -38,7 +38,7 @@ jobs:
           python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Install Hatch
-        run: pip install hatch
+        run: pip install hatch 'virtualenv!=21.0.0'
 
       - name: Show dependency tree
         run: hatch run pip tree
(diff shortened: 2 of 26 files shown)


huong-li-nguyen:
I think if we want an icon on every title, then maybe we need a Title model with different config

petar-qb:
> I think if we want an icon on every title, then maybe we need a Title model with different config

Very interesting thought and I like the idea 👍 

Could we introduce such a model now without breaking any changes? Probably we can along with deprecating the `icon` prop in models that are not Title. 🤔

petar-qb:
I haven’t found the issue or PR where we discussed this topic. If anyone comes across it, please share the link here as there may be conclusions and reasons behind them that we shouldn’t overlook now

huong-li-nguyen:
> I haven’t found the issue or PR where we discussed this topic. If anyone comes across it, please share the link here as there may be conclusions and reasons behind them that we shouldn’t overlook now

I think there are several PRs 😄 But I found these ones: https://github.com/mckinsey/vizro/pull/1053 and https://github.com/mckinsey/vizro/pull/1092
We did initially consider using a `Title` model to enable the info icon, but then went for the `description` argument. I guess we have the same issue now 😄

huong-li-nguyen:
Decided with @antonymilne that we won't implement this for now

PR #1650, opened 2026-03-09
[Tidy] Extend `colors` and change defaults for waterfall and KPI cards

huong-li-nguyen:
## Description
Closes https://github.com/McK-Internal/vizro-internal/issues/2377

Updates default semantic colors across Vizro themes by:

- setting waterfall increasing / decreasing / totals colors to the new semantic palette values,
- aligning KPI positive/negative indicator colors with the same semantics,
- exposing these semantic colors in theme color/palette definitions for reuse.

## Screenshot
Run `hatch run example themes`
Screenshot 2026-03-09 at 13 38 30
Screenshot 2026-03-09 at 13 38 21

## Notice


vizro-core/src/vizro/themes/_colors.py
```diff
@@ -81,6 +81,12 @@
     transparent="rgba(0, 0, 0, 0)",
     white="#FFFFFF",
     black="#000000",
+    increase="#097DFE",  # Equivalent to blue/blue-500
+    decrease="#EA5748",  # Not equivalent to any color in the palette, same as warning
+    neutral="#3E495B",  # Equivalent to gray/gray-700
+    success="#00B5A9",  # Not equivalent to any color in the palette
+    warning="#FFC107",  # Not equivalent to any color in the palette, same as decrease
+    error="#EA5748",  # Not equivalent to any color in the palette
```
huong-li-nguyen: @stephkaiser - I am naming these increase/decrease/neutral/success etc here, because for example for `yellow` we already have a different hex code assigned. Are you happy with these namings?

The hex codes for `decrease`, `success`,`warning` and `error` are currently not in our color codes (so don't match up with any of the colors [here](https://github.com/mckinsey/vizro/blob/main/vizro-core/src/vizro/themes/_colors.py)). Do we want to change these to any of the color values we already have? Otherwise, fine for me if they are not equivalent to any of the colors in the palette.

stephkaiser: @huong-li-nguyen yep agreed, the naming is good 👍 

the success is from our colours (darkgreen-500) but is from one of the duplicate sequential palettes we removed. We could publish all the sequentials (all the duplicate colours) if that makes things easier/more aligned. I tried using our colours for the others (decrease, warning, error) but we didnt have a red and the others didnt quite work or look good, for example the yellow. So they'll have to be new colours unfortunately

huong-li-nguyen: All good! I'll leave it as is then!

vizro-core/src/vizro/themes/_colors.py
```diff
@@ -81,6 +81,12 @@
     transparent="rgba(0, 0, 0, 0)",
     white="#FFFFFF",
     black="#000000",
+    increase="#097DFE",  # Equivalent to blue/blue-500
+    decrease="#EA5748",  # Not equivalent to any color in the palette, same as warning
```
antonymilne: This isn't the same as warning?

antonymilne: The comment on `warning` also lies.

antonymilne (commented):
If this is what we really want then it's ok by me, but I must admit I don't quite understand it 🤔 I'm not sure whether the docs colors.png is correctly up to date though so that might be confusing me?

* increase/decrease colours for waterfall look good but wouldn't it be better to use the same colours for the KPI cards up/down? Since it basically means the same thing. Just from the semantic names, increase/decrease seems much more suitable for the KPI cards than success/danger like you've changed it to?
* error and decrease colours look very similar, should we just make them the same?

vizro-core/src/vizro/themes/_colors.py
huong-li-nguyen: Not warning, the same as error! Sorry, got confused! 😄 Fixed

huong-li-nguyen:
> If this is what we really want then it's ok by me, but I must admit I don't quite understand it 🤔 I'm not sure whether the docs colors.png is correctly up to date though so that might be confusing me?

No, it's definitely up-to-date! :)

> * increase/decrease colours for waterfall look good but wouldn't it be better to use the same colours for the KPI cards up/down? Since it basically means the same thing. Just from the semantic names, increase/decrease seems much more suitable for the KPI cards than success/danger like you've changed it to?

I'm leaving the first question for @stephkaiser to answer :) In the KPI cards, we use the bootstrap variables. There is no bootstrap variable for increase/decrease. So in the CSS files, we would have to use the success/error bs variables anyway. Also a reminder, previously it was just `--bs-blue` and `--bs-pink`, so I think in the css files, the semantics are not very relevant. Also independent of this increase/decrease color change, we would have added new colors for warning, info, success and error! :) So another alternative would be that we have another variable called `increase_alternative` 😅 because the success color can also be used for increases. But I thought that might be redundant.

> * error and decrease colours look very similar, should we just make them the same?

They are the same, but I thought it still makes sense to assign them two times, because it's semantically clearer. We need colors for warning, error, success and info anyway. Maybe let's just discuss tomorrow in our call :) Will be easier

antonymilne:
> > If this is what we really want then it's ok by me, but I must admit I don't quite understand it 🤔 I'm not sure whether the docs colors.png is correctly up to date though so that might be confusing me?
> 
> No, it's definitely up-to-date! :)

OK cool, but then I'm confused by what happens in this screenshot?
image

That looks like a dark green for the positive KPI card? But then (also light theme) this green looks much lighter?
image

Are my eyes tricking me or is there some inconsistency here?

> > * increase/decrease colours for waterfall look good but wouldn't it be better to use the same colours for the KPI cards up/down? Since it basically means the same thing. Just from the semantic names, increase/decrease seems much more suitable for the KPI cards than success/danger like you've changed it to?
> 
> I'm leaving the first question for @stephkaiser to answer :) In the KPI cards, we use the bootstrap variables. There is no bootstrap variable for increase/decrease. So in the CSS files, we would have to use the success/error bs variables anyway. Also a reminder, previously it was just `--bs-blue` and `--bs-pink`, so I think in the css files, the semantics are not very relevant. Also independent of this increase/decrease color change, we would have added new colors for warning, info, success and error! :) So another alternative would be that we have another variable called `increase_alternative` 😅 because the success color can also be used for increases. But I thought that might be redundant.

OK, I think I see. Are the success/danger colours the same in dark and light theme?

> > * error and decrease colours look very similar, should we just make them the same?
> 
> They are the same, but I thought it still makes sense to assign them two times, because it's semantically clearer. We need colors for warning, error, success and info anyway. Maybe let's just discuss tomorrow in our call :) Will be easier

Definitely that's semantically clearer, I like it. You said the comment about decrease being the same as warning was wrong and meant to say that decrease is the same as error, which is exactly what I was suggesting, so all good here!

huong-li-nguyen:
> Are my eyes tricking me or is there some inconsistency here?

Ah, the first screenshot is a pure dash app with the dbc.BOOTSTRAP theme, so it's using the bootstrap variables for success and error from the dbc.BOOTSTRAP, whereas the screenshot below is a vizro app, hence using our bootstrap theme and our bootstrap variables. The success/error colors are the same for dark and light in the vizro app, but of course they differ from the success/error colors from the dbc.BOOTSTRAP

> > > * increase/decrease colours for waterfall look good but wouldn't it be better to use the same colours for the KPI cards up/down? Since it basically means the same thing. Just from the semantic names, increase/decrease seems much more suitable for the KPI cards than success/danger like you've changed it to?
> > 
> > 
> > I'm leaving the first question for @stephkaiser to answer :) In the KPI cards, we use the bootstrap variables. There is no bootstrap variable for increase/decrease. So in the CSS files, we would have to use the success/error bs variables anyway. Also a reminder, previously it was just `--bs-blue` and `--bs-pink`, so I think in the css files, the semantics are not very relevant. Also independent of this increase/decrease color change, we would have added new colors for warning, info, success and error! :) So another alternative would be that we have another variable called `increase_alternative` 😅 because the success color can also be used for increases. But I thought that might be redundant.
> 
> OK, I think I see. Are the success/danger colours the same in dark and light theme?

Yes, they will be the same for dark and light.

vizro-core/src/vizro/themes/_colors.py
antonymilne: @stephkaiser Does our `qualitative.light_green` color not work well here for success? Fine if not, I'm happy with how it is but somehow the current success colour just feels a little bit too bluey to me 🤔 Especially when compared to the `info` color on notifications.

antonymilne (approved):
This makes much more sense now, thanks for the explanation!

Looking at https://github.com/mckinsey/vizro/pull/1650#discussion_r2907006077 and the colour used for `increase` and `success` I feel like there's potentially room for improvement here but I'm not sure whether it's possible since maybe there's too many conflicting "nice to haves" 🤔 

1. IMO it would be nice if `increase` color on waterfall = `positive` on KPI cards. We already have `decrease` color on waterfall = `negative` on KPI cards (which also = `error`). So if you look at a waterfall chart next to a set of KPI cards it just feels a little off that -ve direction colors match and the +ve doesn't.
2. would be nice if these +ve and -ve colors matched the diverging color palette
3. presumably `success` needs to be green rather than blue to distinguish it from `info` (this is good now though could maybe be even greener IMO - see https://github.com/mckinsey/vizro/pull/1650#discussion_r2907006077)

Currently this PR does `positive = success = green` and `increase = blue`. I think other options are:
* `increase = positive = success = green`. Compared to now this would mean increase becomes green which I guess might conflict with 2 above if our diverging palette goes from red to blue rather than red to green? Not sure what the latest plan is for diverging palettes though
* `increase = positive = blue` and `success = green` but then it's messy with bootstrap variables as @huong-li-nguyen explained

So overall I'm not sure what's best. I leave it up to you two to decide @huong-li-nguyen and @stephkaiser. I'm happy with how it is here, just it would be even nicer if we could somehow figure out a way to match up the increase and positive colors if you also think it makes sense to do so.

vizro-core/src/vizro/themes/_colors.py
huong-li-nguyen: I think the reason why it's a little bluey is that it is color-blind proof. Otherwise, you might run into the red-green issue. But we do have the upward/downward arrows, so might not be a huge issue. I'll leave that for @stephkaiser to decide though 👍

huong-li-nguyen:
> So overall I'm not sure what's best. I leave it up to you two to decide @huong-li-nguyen and @stephkaiser. I'm happy with how it is here, just it would be even nicer if we could somehow figure out a way to match up the increase and positive colors if you also think it makes sense to do so.

I would keep the current green-shade, even though it might seem a bit blue-ish, because I think that's one of the reasons why it passes the color-blind test in combination with red. I am indifferent between these two solutions:

1) Keep it as is
2) Change increase color in waterfall chart to green, so then the increase/decrease colors are exactly the same as in the KPI card. I don't think it's an issue that it doesn't match with our diverging palette.

Number 2 also looks nice when combined with other charts:

Screenshot 2026-03-10 at 10 15 01

So I'll leave the final decision to @stephkaiser :)

vizro-core/src/vizro/themes/_colors.py
stephkaiser: @antonymilne the light_green was not colourblind safe with the red unfortunately, but the dark_green is more bluey and thats why it worked. I think its important we use colourblind safe colours since there might not always be up/down arrows

stephkaiser:
> * increase/decrease colours for waterfall look good but wouldn't it be better to use the same colours for the KPI cards up/down? Since it basically means the same thing. Just from the semantic names, increase/decrease seems much more suitable for the KPI cards than success/danger like you've changed it to?

@antonymilne In this ticket (https://github.com/McK-Internal/vizro-internal/issues/2377#issuecomment-3999034522), I suggested we use the blue/red option as our default because if charts don't have a negative value, then the single colour I think should be blue. I created the green/red option as an extra option for users if they required the green/red combination. 

I'm happy to make it the same for both KPI cards and waterfall, but if we go with the blue/red as our default, then the blue clashes with the info blue as well. So maybe its best we just go with the following: (for both KPI cards and waterfall). wdyt?

I was also thinking of adding this to our diverging + sequential palettes as well 

Screenshot 2026-03-10 at 11 01 35

huong-li-nguyen:
> Screenshot 2026-03-10 at 11 01 35

I like! Will implement and merge like that

Issue #1587, opened 2026-02-18
[Docs] Update docs screenshots - batch 4


huong-li-nguyen:
## Guidelines
1. Read through the parent ticket: https://github.com/mckinsey/vizro/issues/1583
2. Update the **Result** screenshots in:

    - Layout: https://vizro.readthedocs.io/en/stable/pages/user-guides/layouts/#grid-basic-example
    - Themes: https://vizro.readthedocs.io/en/stable/pages/user-guides/themes/
    - Data: https://vizro.readthedocs.io/en/stable/pages/user-guides/data/
    - Custom charts: https://vizro.readthedocs.io/en/stable/pages/user-guides/custom-charts/

PR #1652, opened 2026-03-10
[Docs] Update docs screenshots - batch 4

nadijagraca:
## Description
Closes https://github.com/mckinsey/vizro/issues/1587
## Notice


vizro-core/docs/assets/user_guides/layout/flex_advanced.png
petar-qb, line 1: This example I guess should show how setting the `wrap=True` affects the layout. If so, let's add one more chart here so it's clear what happens. See how old example moves the last chart below.

huong-li-nguyen: I would just reduce the window size and then redo the screenshot instead of adding more config :) But I agree with @petar-qb ! Maybe just double-check that all the screenshots look the same in terms of layout

huong-li-nguyen: Same for flex_with_grid_combined.png

PR #1705, opened 2026-04-28
[New] Chat component in Vizro Experimental

lingyielia:
## Notice


vizro-experimental/pyproject.toml
```diff
@@ -0,0 +1,87 @@
...
+description = "Experimental Vizro features in incubation, including a chat component and floating chat popup."
+dynamic = ["version"]
+license-files = {paths = ["LICENSE.txt"]}
+name = "vizro-experimental"
+readme = "README.md"
+requires-python = ">=3.10"
+
+[project.optional-dependencies]
+# Needed to use add_chat_popup() with the default auto-agent.
+agent = [
```
antonymilne, line 31: Is it definitely worth separating this out rather than just having it as a core requirement? (No strong feelings from my side, but I think we should only do optional dependencies if there's a really good reason for it.)

vizro-experimental/src/vizro_experimental/chat/actions/_base_chat_action.py
```diff
@@ -0,0 +1,448 @@
...
+        # Horrible hack to restore chat history when you change page and return.
+        page = model_manager._get_model_page(self)
+
+        @callback(
+            [
+                Output(f"{self._chat_id}-hidden-messages", "children", allow_duplicate=True),
+                Output(
+                    "vizro_version", "children", allow_duplicate=True
+                ),  # Extremely horrible hack we should change, just done here to make
+                # sure callback triggers (must have prevent_initial_call=True).
```
antonymilne: @petar-qb can maybe think of a slightly less horrible hack here? But the current one is ok if we can't.

vizro-experimental/src/vizro_experimental/chat/models/chat.py
```diff
@@ -0,0 +1,301 @@
...
+            convention. Defaults to `[]`.
+        placeholder (str): Placeholder text for the input field. Defaults to `"How can I help you?"`.
+        file_upload (bool): Enable file upload functionality. Defaults to `False`.
+        example_questions (list[str]): List of example questions to show in a popup menu. Defaults to `[]`.
+
+    """
+
+    type: Literal["chat"] = "chat"
+    actions: Annotated[list[_BaseChatAction], BeforeValidator(_coerce_chat_actions)] = Field(
+        default_factory=list,
```
antonymilne: This should just be `default=[]`. GenAI always gets this wrong with pydantic... Same for `example_questions`.

vizro-experimental/src/vizro_experimental/chat/models/chat.py
```diff
@@ -0,0 +1,301 @@
...
+    CSS_ROOT,
+    ICON_BUTTON_SIZE,
+)
+from ..actions._base_chat_action import _BaseChatAction
+
+
+def _coerce_chat_actions(value: Any) -> Any:
+    """Wrap a single action in a list to match Vizro's action-chain field shape."""
+    if value is None:
+        return []
```
antonymilne: I don't think L38-39 are required. If `value=None` then it will fail validation, which is fine.

vizro-experimental/src/vizro_experimental/chat/models/types.py
```diff
@@ -0,0 +1,65 @@
...
+
+import json
+from typing import Any, Literal, TypeAlias, TypedDict
+
+from typing_extensions import NotRequired
+
+Role: TypeAlias = Literal["user", "assistant"]
+
+
+class Message(TypedDict):
```
antonymilne, line 13: Should we just turn this into a pydantic model (subclass `BaseModel`, not `VizroBaseModel`)? Then the below validation could be built into the type itself with field types and validators. It would still be easy to serialise to/from json.

vizro-experimental/src/vizro_experimental/chat/popup/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+    app = Vizro()
+    app.build(dashboard)
+
+    add_chat_popup(app, title="Analytics Assistant")
+    app.run()
+"""
+
+
+def __getattr__(name):
```
antonymilne, line 24: This would be worth a comment since it looks pretty weird. I guess it's just due to optional dependencies?

vizro-experimental/src/vizro_experimental/chat/popup/popup.py
```diff
@@ -0,0 +1,431 @@
...
+        from .dashboard_agent import create_dashboard_agent, make_generate_response
+
+        agent, _context = create_dashboard_agent(model=model, reasoning_effort=reasoning_effort)
+        generate_response = make_generate_response(agent)
+
+    chat_panel = _build_chat_panel(chat_id, placeholder, title, streaming)
+    toggle_button = _build_toggle_button(chat_id, color=color)
+
+    # Append to the MantineProvider's first child (dashboard-container)
+    app.dash.layout.children[0].children.append(chat_panel)
```
antonymilne: This isn't very robust but is the best we can do for now. In future we could potentiall give `MantineProvider` in vizro core an `id` and use that here. We'd still need a big warning in vizro-core to not change the id though.

vizro-experimental/src/vizro_experimental/chat/popup/popup.py
```diff
@@ -0,0 +1,431 @@
...
+)
+from ..actions.streaming_chat_action import (
+    _register_streaming_chunk_callback,
+    _register_streaming_endpoint,
+    _StreamingRequest,
+)
+from ..models.types import _parse_store_messages
+
+
+def add_chat_popup(  # noqa: PLR0913
```
antonymilne, line 58: Can we get rid of `app` and just use `dash.get_app()` as elsewhere?

vizro-experimental/examples/scratch_dev/app.py
```diff
@@ -0,0 +1,40 @@
...
+
+class EchoAction(ChatAction):
+    """Echo the most recent user message back as the assistant reply."""
+
+    def generate_response(self, messages: list[Message], **_: Any) -> str:
+        """Return the last user message verbatim, prefixed with 'You said:'."""
+        return f"You said: {messages[-1]['content']}"
+
+
+vm.Page.add_type("components", Chat)
```
antonymilne: I probably asked this before and it didn't work for some reason, but I just tried it out now and it seemed to...

Can't we just put this line in `chat.py` itself so the user never needs to do it in `app.py`?

vizro-experimental/docs/pages/chat/chat-component.md
```diff
@@ -0,0 +1,56 @@
+# How to use the Chat component
+
+This guide shows you how to add a [`Chat`][vizro_experimental.chat.Chat] component to your dashboard.
+
+The `Chat` component renders a message history, a textarea, and a send button. It is decoupled from any specific LLM — you bring the backend by subclassing [`ChatAction`][vizro_experimental.chat.ChatAction] and overriding `generate_response`.
+
+## Add a chat to a page
+
+To add a `Chat`, register it as an allowed page component with `vm.Page.add_type` and pass an action. The simplest possible action just echoes the user's last message.
```
antonymilne: If we manage to get rid of the need for a user to do this, remember to change here.

vizro-experimental/docs/pages/chat/chat-component.md
```diff
@@ -0,0 +1,56 @@
...
+
+            def generate_response(self, messages: list[Message]) -> str:
+                return f"You said: {messages[-1]['content']}"
+
+
+        vm.Page.add_type("components", Chat)
+
+        page = vm.Page(
+            title="Echo chat",
+            components=[Chat(actions=[EchoAction()])],
```
antonymilne: Any `actions=[one_element]` should just be `actions=one_element` now

antonymilne (commented):
This is awesome, amazing work! I wasn't expecting all this documentation and testing as well but that makes it even better.

I already tested functionality on the internal version so didn't test it out properly here other than to open the scratch app. So it would be good if @petar-qb could do a review of the functionality and also skim through the code (I already looked through and it's ok for me - it doesn't need to be up to vizro-core standard since it's in vizro-experimental).

One thing I did notice is that `hatch run example` on scratch_dev and chat_component don't look styled correctly to me:
image

Note to myself for future: it would still be nice if user could write just `@capture` function rather than subclass at least for cases without state required. Remember there's no real overhead in recreating AI client and could be set outside function anyway. This looks feasible but not worth spending time on now.

@lingyielia I know you said you couldn't get this working with `vdc.Markdown` but could you just briefly document here where you got to in case we want to pick it up again in the future? My comment from before was this:

> nice to have: replace the markdown with vdc.Markdown. I tried doing this and it worked great for non-streaming but streaming mode didn't work well. I didn't try hard to fix but would be great if you could spend ~1 hour on this. It might need some fixes on the vdc.Markdown or here, I'm not sure

antonymilne (approved):

vizro-experimental/examples/chat_component/app.py
```diff
@@ -0,0 +1,252 @@
...
+
+# Page 1: Simple Echo (no API key required)
+page_echo = vm.Page(
+    title="Simple Echo (No API Key)",
+    components=[
+        Chat(actions=simple_echo()),
+    ],
+)
+
+# Page 2: Example Questions Demo
```
petar-qb, line 32: Here's a few inconsistencies I found regarding the chat UI/UX:
- [x] - 1. Client and server messages on the same side of the chat (maybe that's ok).
- [x] - 2. Too much vertical space between messages from both sides.
- [x] - 3. No clear separation between the response from the previous answer and the new question in the chat.
- [x] - 4. `example_questions` when opened overlap with  chat messages and chat input text.
- [x] - 5. There's no automatic chat scroll down when the new response arrives.
- [x] - 6. THere's no loading icon (or any text) that shows that chat waits for the answer. 
- [x] - 7. How about to add the “clear chat” button that clears the context?

https://github.com/user-attachments/assets/7bdaaa3e-3c8e-4315-ba5d-b3f0ba242477

@lingyielia please tick all inputs from above you find irrelevant or you solve.

vizro-experimental/examples/chat_component/app.py
```diff
@@ -0,0 +1,252 @@
...
+page_example_questions = vm.Page(
+    title="Example Questions",
+    layout=vm.Grid(grid=[[0], [1], [1], [1], [1]]),
+    components=[
+        vm.Card(
+            text="""
+## Example Questions Demo
```
petar-qb, line 39: There's no clear error message when something goes wrong. For example when the OPENAI_API_KEY is expired.

https://github.com/user-attachments/assets/b500c4ed-9650-4d33-9a4a-43f188ca4ee0

vizro-experimental/examples/chat_component/app.py
```diff
@@ -0,0 +1,252 @@
...
+                "Explain quantum computing in simple terms",
+                "Write a haiku about programming",
+                "What are the benefits of TypeScript over JavaScript?",
+            ],
+        ),
+    ],
+)
+
+# Page 3: Example Questions + File Upload Combined
+page_example_with_upload = vm.Page(
```
petar-qb, line 70: Attached file is lost from the context after first response.

https://github.com/user-attachments/assets/4fdd6854-1197-466b-990e-7aaa48aa35ef

vizro-experimental/examples/chat_component/app.py
```diff
@@ -0,0 +1,252 @@
...
+    title="Examples + Upload",
```
petar-qb, line 71: Generated chart code uses custom function with `go.Figure` instead of simple `px.bar`. Maybe that's ok IDK. 

Image

vizro-experimental/docs/pages/chat/chat-component.md
petar-qb, line 1: Here's another comments (this time about the docs) that gathers a few recommendations in a single comment:
 
- [ ] - 1. Can’t navigate from `vizro-core` or `vizro-ai` documentation to `vizro-experimental` documentation (but can do it in opposite way)
- [ ] - 2. Would you mind removing `em-dash` (—) characters from the docs. This is the low handing fruit in terms to "fake" that a human wrote the docs 😄 
- [ ] - 3. Missing example `.gifs`. 
- [ ] - 4. Extract "API references" into another header tab title like it’s in `vizro-core` and other libs.
- [ ] - 5. I would add a hint on how to add `OPENAI_API_KEY` and `OPENAI_API_BASE` (similar like it's described in the `vizro-ai`)

petar-qb (commented):
@lingyielia, this is incredible piece of work!! 🔥 🔥

I tested the `chat_components` example and skimmed through the documentation and left several comments. 

Now, I'm going to review (not in detail) the code and will leave the second batch of the comments.

vizro-experimental/examples/chat_component/app.py
petar-qb: I guess that the "Actions failed" (error) notification will show in these cases after you pull the changes from the `main`. Could you double check that?

vizro-experimental/examples/chat_component/actions.py
```diff
@@ -0,0 +1,517 @@
...
+
+    """
+
+    model: str = Field(default="gpt-4.1-nano", description="OpenAI model name.")
+
+    @property
+    def client(self):  # noqa: D102
+        if OpenAI is None:
+            raise ImportError("OpenAI package not installed. Install with: pip install openai")
+        return OpenAI()
```
petar-qb, line 147: Would applying singleton approach help here? In that case OpenAI model would be created once when dashboard is built. However, this potentially could limit multi-user operations.

vizro-experimental/examples/chat_component/actions.py
```diff
@@ -0,0 +1,517 @@
...
+        Args:
+            messages: Parsed history (``role`` and ``content`` per item).
+
+        Returns:
+            Generated text response.
+
+        """
+        if Anthropic is None:
+            raise ImportError("Anthropic package not installed. Install with: pip install anthropic")
+        client = Anthropic()
```
petar-qb, line 200: Let's align with the `openai_streaming_chat` property approach

vizro-experimental/examples/chat_component/actions.py
```diff
@@ -0,0 +1,517 @@
...
+                style={"color": "red"},
+            )
+
+        prompt = messages[-1]["content"] if messages else ""
+
+        if not uploaded_files:
+            return html.P("Please upload a data file first!", style={"color": "#1890ff"})
+
+        # Get the first uploaded file
+        uploaded_data = uploaded_files[0]["content"]
```
petar-qb: Any limitations on the size of the uploaded file?

vizro-experimental/examples/chat_component/actions.py
```diff
@@ -0,0 +1,517 @@
...
+            elif uploaded_filename and uploaded_filename.endswith((".xls", ".xlsx")):
+                df = pd.read_excel(io.BytesIO(decoded))
+            else:
+                return html.P(
+                    f"Unsupported file type: {uploaded_filename}. Please upload CSV or Excel files.",
+                    style={"color": "red"},
+                )
+        except Exception:
+            logger.exception("Error parsing uploaded file")
+            return html.P("Error parsing file. Please check the file format and try again.", style={"color": "red"})
```
petar-qb: Exceptions are handled differently here an in other actions. Maybe we should align how exceptions are handled for all chat actions. 

Let's align how the `client` model is created (in the `generate_response` or as a `property`)

vizro-experimental/examples/popup/app.py
```diff
@@ -0,0 +1,71 @@
...
+
+    from vizro_experimental.chat.popup import add_chat_popup
+
+    app = Vizro()
+    app.build(dashboard)
+
+    add_chat_popup(
+        app,
+        title="Chat Assistant",
+        placeholder="Ask me anything about the data...",
```
petar-qb, line 65: The chat popup looks amazing! 
What the chat know about and how it can help dashboard creator/user?

vizro-experimental/src/vizro_experimental/chat/actions/_base_chat_action.py
```diff
@@ -0,0 +1,448 @@
...
+    if payload_extras:
+        for key, value in payload_extras.items():
+            if key == "uploaded_files":
+                continue
+            if key in params:
+                out[key] = value
+    return out
+
+
+def _loading_bubble() -> html.Div:
```
petar-qb, line 106: Can we use this in every new action?

vizro-experimental/src/vizro_experimental/chat/actions/_base_chat_action.py
petar-qb: @lingyielia If you remove `vizro_version` Output and `prevent_initial_call=True`, would it work then?

petar-qb: If `allow_duplicate=True` prevents you to do that, then leave it as it is. I have solved several problems similarly in the past (by targeting `_content_pages.id` or something like that).

vizro-experimental/src/vizro_experimental/chat/popup/popup.py
```diff
@@ -0,0 +1,431 @@
...
+        prevent_initial_call=True,
+    )
+
+    clientside_callback(
+        ClientsideFunction("vizroChatComponent", "setupEnterKeyHandler"),
+        Output(f"{chat_id}-chat-input", "id", allow_duplicate=True),
+        Input(f"{chat_id}-chat-input", "value"),
+        State(f"{chat_id}-chat-input", "id"),
+        prevent_initial_call=True,
+    )
```
petar-qb, line 365: Let's add `hidden=True` to all `clientside_callback` and (server-side) `callback` added in this PR. This will remove them from the callback architecture visible in the Dash Dev Tools.

vizro-experimental/pyproject.toml
lingyielia: Keeping it optional because add_chat_popup takes a user-supplied generate_response and Chat itself never needs an agent — pydantic-ai is the default-agent implementation, not a feature requirement. I would keep it optional for now, especially if we expect more experimental features coexist here.

vizro-experimental/examples/chat_component/app.py
lingyielia: Ah, the user messages and the example_questions opened are supposed to have a background. But in release 0.1.55, the `var(--surfaces-bg-card)` was removed from the bootstrap. Now replaced var(--surfaces-bg-card) with var(--bs-tertiary-bg)

vizro-experimental/src/vizro_experimental/chat/actions/_base_chat_action.py
lingyielia: replaced with `prevent_initial_call="initial_duplicate",` and that works from my test so far. Any unexpected side effect you can think of?
https://github.com/plotly/dash/blob/v4.1.0/dash/_validate.py#L562-L584

vizro-experimental/src/vizro_experimental/chat/models/types.py
lingyielia: Because this is public, maybe current TypeDict is more convinent. `Message` is the type users see in `generate_response(messages)` and they access it as msg["role"] / msg["content"], which is exactly the OpenAI/Anthropic dict-message convention.

vizro-experimental/src/vizro_experimental/chat/popup/popup.py
lingyielia: Great point! Using dash.get_app() instead

vizro-experimental/examples/chat_component/app.py
lingyielia: - 1. Let's wait for the official DS guidance
 - 2. Let's wait for the official DS guidance
 - 3. fixed
 - 4. fixed
 - 5. Yes, I want this feature too! Let's address it in future prs.
 - 6. There is loading (see those non-streaming example. for pure text streaming, the streaming signal comes too fast so the loading icon got stripped right away)
 - 7. Let's wait for the official DS guidance

vizro-experimental/examples/chat_component/app.py
lingyielia: error messages like expired keys are not allowed by those security rules. But agreed, a more general error message could help here 👍

vizro-experimental/docs/pages/chat/chat-component.md
lingyielia: - 1. This is because the docs hasn't been published yet
 - 2. yes!
 - 3. Let's skip it for now
 - 4. This is intended per @stichbury's docs restructuring plan
 - 5. yes!

vizro-experimental/examples/chat_component/actions.py
lingyielia: These are examples and we intended to show different patterns.

vizro-experimental/examples/chat_component/actions.py
lingyielia: These are examples and we intended to show different patterns.

vizro-experimental/examples/popup/app.py
lingyielia: It has access to data manager and model manager

vizro-experimental/examples/chat_component/actions.py
lingyielia: These are examples and we intended to show different patterns.

vizro-experimental/examples/chat_component/actions.py
lingyielia: Not in this example. This could be an improvement area for this example

vizro-experimental/src/vizro_experimental/chat/actions/_base_chat_action.py
lingyielia: Already automatic used. _register_loading_indicator_callback is invoked once per Chat instance in _BaseChatAction.pre_build, and once for the popup in add_chat_popup. The send-button callback inside it always appends _loading_bubble() after the user message, so every action gets the dots indicator without per-action wiring.

stichbury (approved):
I'm happy to merge these docs and I can work on changes after the fact, and set up the RTD project.

PR #1723, opened 2026-05-04
[Feat] Syncing controls on the same page

petar-qb:
closes https://github.com/McK-Internal/vizro-internal/issues/2193
closes https://github.com/McK-Internal/vizro-internal/issues/2973

closes https://github.com/McK-Internal/vizro-internal/issues/2054
closes https://github.com/McK-Internal/vizro-internal/issues/2955

Syncing controls across the pages PR that's merged into this one -> https://github.com/mckinsey/vizro/pull/1853

## Description
- Enables control syncing: changing a control applies its value to synced controls on the same page immediately or on other pages when opened
- Relaxes drill-through so the targeted controls doesn't have to have `show_in_url`.
- Enables `set_control` to target multiple controls with the same value at the same time.

### API: 
- A `Filter` / `Parameter` can now list a control on the same or on the different page in its `targets`, keeping the two or more in sync.                                    
- Drill-through (a `set_control` triggered from a `Graph`, `AgGrid`, `Button`, or `Card`) can now target a control on another page **without** that control needing `show_in_url=True`.
- `set_control.control` can now be a list of control ids.

### Good use cases:
1. Filter targets a hidden parameter that targets graph's title.
2. Global controls implementation
3. For multi-tabs page where container's controls are synced.

### Implementation:
If we have two controls `A` and `B` that are "synced" (see the API above), their configuration defaults to:
`A.selector.actions=[set_control(control="B", value=None), update_targets(targets=["graph_1"])]`
`B.selector.actions=[set_control(control="A", value=None), update_targets(targets=["graph_2"])]`

There are 4 https requests that happen when a single filter values changes (for example `A.value` is set to `X`):  
1. `A.set_control` triggers and sets `B.value` to `X`.
2. `A.update_targets` triggers and updates the `graph_1`. (**this goes in parallel with the action below**)
3. `B.set_control` triggers that DOES NOT set the `A.value` to `X` as the `A.value` is already set to `X`. **This is important as it breaks the infinite loop of cross set_control triggering.**
5. `B.update_targets` triggers and updates the `graph_2`.

### Number of HTTP requests:

for the following schema (F-filter, G-graph):

```mermaid
flowchart LR
    F1 --> F2
    F1 --> G1
    F2 --> F1
    F2 --> G2
    F2 --> F3
    F3 --> G2
``` 

Dashboard creator defines filter targets in the following way:
1. F1.targets=[F2, G1]
2. F2.targets=[F1, F3, G2]
3. F3.targets=[G2]

For the given targets, we define actions like (SC - set_control, UT - update_targets):
1. F1.selector.actions=[SC(F2), UT(G1)]
2. F2.selector.actions=[SC(F1), SC(F3), UT(G2)]
3. F3.selector.actions=[UT(G2)]

At first, it maybe looks like when F1 changes that only 2 http requests will fire and that for example G2 will never been updated. However, due to the internal actions chaining here's what happens when each of these changes.

F1 change:
- F1: SC(F2) 
- F1: UT(G1)
- F2: SC(F1) - doesn't trigger F1 actions again as the F1 value doesn't change
- F2: SC(F3)
- F2: UT(G2)
- F3: UT(G2)

F2 change - Same as F1 but in the different order:
F3 change - UT(G2) 

So, it means that changing F1 causes 6 http requests, F2 fires 6 http requests too and F3 only 1 http.

### Potential optimisation of the number of HTTP requests:

**Ideally, every "mesh" of controls affecting figures or other controls can be solved with 2 http requests.**
For the defined filter targets from above, we can perform search through tree of filters' targets and cask the following default actions:
1. F1.selector.actions=[`set_controlS(F2, F3, stop_internal_action_chaining=True)`, `UT(G1, G2, G3)`]
2. F2.selector.actions=[`set_controlS(F1, F3, stop_internal_action_chaining=True)`, `UT(G1, G2, G3)`]
3. F3.selector.actions=[`UT(G2)`]

There are three features we should implement in terms to achieve the above:
1. Support for targeting multiple controls with `set_control` action. https://github.com/McK-Internal/vizro-internal/issues/2955 
2. Introduce the new `stop_internal_action_chaining: bool = False` action's attribute. -> https://github.com/McK-Internal/vizro-internal/issues/2956
3. Implement search mechanism through filters' targets that detect all controls and figures that have to be set/updated. -> https://github.com/McK-Internal/vizro-internal/issues/2957

Optionally, we can implement clientside `set_control` to reduce the number of HTTP requests from 2 -> 1.

## Navigating to the targeted page rules

| # | Source (trigger) | Target location(s) | Same-page target | Cross-page target | Navigates? |
|---|------------------|--------------------|------------------|-------------------|------------|
| 1 | Selector (sync) | same page only | direct | — | No |
| 2 | Selector (sync) | cross page, 1 page | — | store | No |
| 3 | Selector (sync) | cross page, N pages | — | store | No |
| 4 | Selector (sync) | mixed: same + cross | direct | store | No |
| 5 | Figure (drill-through) | same page only | direct | — | No |
| 6 | Figure (drill-through) | cross page, 1 page | — | store | **Yes → that page** |
| 7 | Figure (drill-through) | cross page, N pages | — | store | No |
| 8 | Figure (drill-through) | mixed: same + cross (1 cross page) | direct | store | No |

### Syncing on the same page:

https://github.com/user-attachments/assets/870c20a6-938d-4d28-9979-c277cad0adb8

### Syncing across the pages:

https://github.com/user-attachments/assets/d9a751ba-1106-410e-8124-c66e60684e26

## Notice


antonymilne:
I didn't check the code on this one since it's still WIP - just looking at the description. 

tl;dr: I'm not a big fan but I think it's ok. I do think there's some value in syncing controls on the same page but it's not that great. It's the extension to cross-page syncing that is really the interesting part here, and obviously doing same page is a necessary prerequisite for that (or at least comes for free).

The API looks good. The thing I don't like so much is the implementation using action chains. It's probably the right way to do it still, since any other approach is likely to be much more complex to implement, but possibly you can think of some other implementation that solves some of the problems of this implementation without being horribly complicated. I'd rather go for the current implementation than something slicker but horribly complicated.

The part I don't like so much is this flow:
```mermaid
flowchart LR
    A_SC["A.set_control"]
    A_UF["A.update_targets"]
    G1["update graph_1"]

    B_SC["B.set_control"]
    B_UF["B.update_targets"]
    G2["update graph_2"]

    A_SC --> A_UF --> G1
    A_SC --> B_SC --> B_UF --> G2
```

Due to the no-op `B.set_control`, the `update_graph_2` call is delayed. On localhost this is fine but on a real server I think it might feel a bit laggy (try testing in dev mode with throttling to see). Take the example that both synced controls target the same graph, which is not so unrealistic with the use case "Filter targets a hidden parameter that targets graph's title"). You might get graph loading state ➡️ updates ➡️ loading state ➡️ refreshes. Which is not amazing but probably ok.

Given `B.set_control` must be done serverside, I don't see any easy way to shortcut it so suspect there's not much we can do about this without changing to a much more complicated system, which is likely not worth it.

vizro-core/changelog.d/20260504_134102_petar_pejovic_syncing_controls_spa.md
petar-qb, line 1: Revisit after merging MPA into this branch

vizro-core/docs/pages/user-guides/controls.md
petar-qb, line 1: Revisit after merging MPA into this branch

vizro-core/docs/pages/user-guides/controls.md
```diff
@@ -99,92 +107,196 @@ To refresh the targets on demand, add a [`Button`][vizro.models.Button] that run
...
+
+    === "Result"
+
+        [![ControlGroup]][controlgroup]
+
+## Sync controls
+
+You can keep two controls in sync so that changing one automatically applies the same value to the other. This is useful, for example, when you need to both filter and parametrize a chart from the same user selection, such as filtering a chart down to one species and also using that species to set the chart's title.
+
+To sync controls, add another control's `id` to the `targets` of a [filter](filters.md) or [parameter](parameters.md). Whenever the control changes, Vizro sets the targeted control to the same value (using the [`set_control` action][vizro.actions.set_control] behind the scenes) and then refreshes that both controls figure targets. All combinations work: filter and filter, parameter and parameter, and filter and parameter.
```
petar-qb: Add note that the type of controls must match. So, slider can't sync with switch

vizro-core/src/vizro/actions/_set_control.py
```diff
@@ -123,6 +126,8 @@ class set_control(_AbstractAction):
         description="Filter or Parameter component id to be affected by the trigger. "
         "If the control is on a different page to the trigger, then it must have `show_in_url=True`."
     )
+
+    # TODO AM-PP: How about making it optional with default=None.
```
petar-qb, line 131: Set `None` as default. Validate and raise exception/warning depending on the source `set_control` model. Learn users how to set set_control.value through these exceptions/warnings

vizro-core/src/vizro/models/_controls/_controls_utils.py
```diff
@@ -90,6 +89,90 @@ def get_control_parent(control: ControlType) -> Page | Container | None:
...
+            f"Control '{control.id}' lists control target(s) {targeted_controls} in `targets`, but its selector has "
+            f"explicit `actions`, so these targets are not kept in sync automatically. Add a `set_control` action to "
+            f"the selector's `actions` for each one to sync them, and remove them from `targets`.",
+            UserWarning,
+        )
+
+
+def build_default_control_selector_actions(
+    selector: SelectorType, targeted_controls: list[ModelID], update_targets_id: str, update_targets: list[str]
+) -> None:
```
petar-qb, line 157:
```suggestion
def build_default_control_selector_actions(
    selector: SelectorType, targeted_controls: list[ModelID], update_targets_id: str, targeted_figures: list[str]
) -> None:
```

vizro-core/src/vizro/models/_controls/parameter.py
```diff
@@ -23,10 +25,13 @@
...
-        raise ValueError(
-            f"Invalid target {target}. Targets must be supplied in the form <target_component>.<target_argument>"
-        )
-    elif target.split(".")[1] == "figure":
+        return target
+
+    # Use the first argument segment (not the full remainder) so that nested addressing like
+    # "<component>.figure.<arg>" is rejected too, not only the exact "<component>.figure".
+    targeted_argument = target.split(".")[1]
+
```
petar-qb, line 33: double check

vizro-core/src/vizro/models/_models_utils.py
```diff
@@ -163,8 +163,11 @@ def make_actions_chain(self):
         # Every action has to know about the model action trigger to properly set the action's builtin arg "_trigger".
         action._first_in_chain_trigger = model_action_trigger
 
-        # The actions chain guard should be called only for on page load.
-        action._prevent_initial_call_of_guard = not isinstance(action, _on_page_load)
+        # The actions chain guard should be called only for on page load. We detect this by checking the action's
+        # own wired trigger against the page's on-page-load trigger, rather than the action's class or id: this way
+        # it doesn't matter whether the first action in the chain is the built-in `_on_page_load` or a user-supplied
+        # replacement (e.g. `Page(actions=[va.update_targets(...)])`), it is still recognized as on-page-load.
+        action._prevent_initial_call_of_guard = not action._trigger.startswith(ON_PAGE_LOAD_ACTION_PREFIX)
```
petar-qb, line 170: Check if possible to set _prevent_initial_call_of_guard if parent model is Page. If so, this would enable overwriting Page.actions and if so, document and test that too

vizro-core/docs/pages/user-guides/controls.md
```diff
@@ -99,92 +107,200 @@ To refresh the targets on demand, add a [`Button`][vizro.models.Button] that run
...
+
+    === "Result"
+
+        [![ControlGroup]][controlgroup]
+
+## Sync controls
+
+You can keep two controls in sync so that changing one automatically applies the same value to the other. This is useful, for example, when you need to both filter and parametrize a chart from the same user selection, such as filtering a chart down to one species and also using that species to set the chart's title.
+
+To sync controls, add another control's `id` to the `targets` of a [filter](filters.md) or [parameter](parameters.md). Whenever the control changes, Vizro sets the targeted control to the same value (using the [`set_control` action][vizro.actions.set_control] behind the scenes) and then refreshes both controls' figure targets. All combinations work: filter and filter, parameter and parameter, and filter and parameter.
```
nadijagraca: The last part of the second sentence sounds a bit clunky. Maybe change to something like: 
Whenever the control changes, Vizro sets the targeted control to the same value (using the [set_control action][vizro.actions.set_control] behind the scenes) and then refreshes the figure targets of both controls. All combinations work: filter with filter, parameter with parameter, and filter with parameter.

vizro-core/docs/pages/user-guides/controls.md
```diff
@@ -99,92 +107,200 @@ To refresh the targets on demand, add a [`Button`][vizro.models.Button] that run
...
+
+!!! note "Synced controls must hold compatible values"
+
+    Syncing copies one control's value straight into the other, so the two controls must understand the same kind of value. Pair a categorical selector ([`Dropdown`][vizro.models.Dropdown], [`RadioItems`][vizro.models.RadioItems], [`Checklist`][vizro.models.Checklist]) with another categorical selector, a numeric [`Slider`][vizro.models.Slider] / [`RangeSlider`][vizro.models.RangeSlider] with another numeric one, a range selector with another range selector, and so on. Mismatched pairings do not sync: for example a [`Slider`][vizro.models.Slider] cannot sync with a [`Switch`][vizro.models.Switch], and a multi-select selector syncing into a single-select one is ignored once more than one value is selected.
```
nadijagraca: I would move this section into its own paragraph so that it can read better, and also add a sentence describing what happens when second value is added in multi-to-single value filter sync. 

Mismatched pairings do not sync: for example a [`Slider`][vizro.models.Slider] cannot sync with a [`Switch`][vizro.models.Switch], and a multi-select selector syncing into a single-select one is ignored once more than one value is selected. Once a second value is added, the sync is skipped and the target keeps showing its last synced value.

nadijagraca (approved):
This is an amazing feat, amazing job Petar. 🚀  🙌 
I just have two small suggestions for the docs, but feel free to merge either way.

vizro-core/src/vizro/models/_models_utils.py
petar-qb: Will be considered and potentially done as the part of this ticket -> https://github.com/McK-Internal/vizro-internal/issues/2984

l0uden (approved):
Awesome work!

PR #1737, opened 2026-05-12
[Skill] Add skill for actions

lingyielia:
## Description

Add a new skill `wiring-vizro-actions`

## Screenshot

Eval set 1: (wiring-vizro-actions only)
Screenshot 2026-05-12 at 4 40 18 PM
Screenshot 2026-05-12 at 4 40 30 PM
Screenshot 2026-05-12 at 4 40 38 PM
Screenshot 2026-05-12 at 4 40 56 PM
Screenshot 2026-05-12 at 4 41 07 PM

Eval set 2: (e2e from design to build)
Screenshot 2026-05-12 at 8 22 18 PM
Screenshot 2026-05-12 at 8 29 39 PM
Screenshot 2026-05-12 at 8 22 42 PM

## Notice


vizro-e2e-flow/evals/dashboard-build/evals.json
```diff
@@ -48,7 +58,13 @@
         "Interactive options were presented during the design phase",
         "app.py uses vizro.plotly.express (not plain plotly.express)",
         "Agent ran validate_aggregation.py and produced a PASS result (bar/line charts use @capture('graph') with pre-aggregated data)",
-        "KPI cards use built-in kpi_card or kpi_card_reference (not rebuilt as custom charts)"
+        "KPI cards use built-in kpi_card or kpi_card_reference (not rebuilt as custom charts)",
```
huong-li-nguyen: They don't need to be built-in only right? it could also be a custom figure, just not a custom chart.

Like this is still alowed:
```
@capture("figure")
my_custom_kpi_card(...)
```

vizro-e2e-flow/skills/dashboard-build/SKILL.md
```diff
@@ -35,6 +35,7 @@ IMPORTANT: Each step produces a spec file in the `spec/` directory to document r
...
+- **Interactions / actions**: For cross-filter, cross-highlight, drill-through, or data export, load the **wiring-vizro-actions** skill. Check the `interactions:` section in `spec/2_interaction_ux.yaml` for which patterns to implement. Key gotchas: `import vizro.actions as va`; use `va.set_control` (NOT deprecated `filter_interaction`); add `custom_data` for non-positional graph cross-filter; use `visible=False` and `"NONE"` in selector options for cross-highlight Parameters; use `show_in_url=True` plus `layout=vm.Flex(direction="column")` and a back `vm.Button(href=...)` for cross-page drill-through; add a short `header` hint on any interactive Graph/AgGrid. For highlight-aware custom charts, see [custom_charts_guide.md](./references/custom_charts_guide.md). **Before declaring actions wiring done, walk through the "Common Implementation Mistakes" table at the bottom of [actions-reference.md](../wiring-vizro-actions/references/actions-reference.md) and check your code against every row** — these are silent failures (callback 500s, drill-through doing nothing) that don't surface until a user actually clicks.
```
huong-li-nguyen: Do we also need to load the **wiring-vizro-actions** skill for the built-in data export? Also general question is it generally better to explain all of this in text rather than having some example code here and referencing these snippets here? Maybe even just referencing our docs?

`"use `show_in_url=True` plus `layout=vm.Flex(direction="column")` and a back `vm.Button(href=...)` for cross-page drill-through; add a short `header` hint on any interactive Graph/AgGrid. "`

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
```diff
@@ -126,6 +126,66 @@ Tertiary metric: Smaller chart or table row
...
+**Problem**: 3+ cross-filters on one page — users lose track of what affects what.
+
+**Solution**: Limit to 1–2 interaction patterns per page. Match each interaction to a named pattern from the **wiring-vizro-actions** skill; if you cannot name the pattern, the interaction probably isn't justified.
+
+### Feature-First Thinking
+
+**Problem**: "Let's add cross-filter because we can" without a clear user need.
+
+**Solution**: Start from user task. What decision does this interaction enable? If a standard sidebar filter covers it, don't add a click-driven cross-filter.
+
```
huong-li-nguyen, line 140: I think I would even be stricter, I would say we always default to the sidebar filters unless the user explicitly asks for "chart interactions". I want "chart a to be clickable and filter chart b"

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
```diff
@@ -126,6 +126,66 @@ Tertiary metric: Smaller chart or table row
...
+**Problem**: Adding cross-filter for fewer than ~5 groups, where a dropdown would be just as fast.
+
+**Solution**: Cross-filter pays off when the data has hierarchy (Pattern 1 / 2) or when click-to-explore is more natural than dropdown selection. Otherwise prefer the sidebar filter.
+
+### No Visual Affordance on Interactive Source
+
+**Problem**: User has no idea a chart or table is clickable.
+
+**Solution**: Add a short, action-oriented `header` on every Graph/AgGrid with `actions=` (e.g. `header="Click a bar to filter by region"`). Avoid verbose or vague hints.
```
huong-li-nguyen: i don't think it needs to be in the header (too prominent for me), it can just be inside the `description` argument, hence inside the tooltip on the graph/table title?

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
```diff
@@ -126,6 +126,66 @@ Tertiary metric: Smaller chart or table row
...
+**Problem**: "Let's add cross-filter because we can" without a clear user need.
+
+**Solution**: Start from user task. What decision does this interaction enable? If a standard sidebar filter covers it, don't add a click-driven cross-filter.
+
+### Cross-Filter When a Filter Suffices
+
+**Problem**: Adding cross-filter for fewer than ~5 groups, where a dropdown would be just as fast.
+
+**Solution**: Cross-filter pays off when the data has hierarchy (Pattern 1 / 2) or when click-to-explore is more natural than dropdown selection. Otherwise prefer the sidebar filter.
+
```
huong-li-nguyen, line 146: Same here, let's just default to not using cross-filters unless the user asks for it?

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
```diff
@@ -126,6 +126,66 @@ Tertiary metric: Smaller chart or table row
...
+**Problem**: User has no idea a chart or table is clickable.
+
+**Solution**: Add a short, action-oriented `header` on every Graph/AgGrid with `actions=` (e.g. `header="Click a bar to filter by region"`). Avoid verbose or vague hints.
+
+### Missing Back Button on Drill-Through Target
+
+**Problem**: User navigates to a detail page (Pattern 1) and feels trapped.
+
+**Solution**: The drill-through target page MUST include a back `vm.Button(text="← Back", href="/source-page", variant="outlined")` as the first component. The page MUST also use `layout=vm.Flex(direction="column")` so the button takes natural height — Grid would waste a 140px+ row.
```
huong-li-nguyen: Does adding config or a screenshot here help? Or is it better to be verbose? I am actually wondering because I don't know

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
```diff
@@ -126,6 +126,66 @@ Tertiary metric: Smaller chart or table row
...
+
+**Problem**: Cross-page drill-through silently fails — no error, just no filtering on the target page.
+
+**Solution**: The Filter on the target page must have `show_in_url=True`.
+
+### Cross-Filter When You Wanted to Highlight
+
+**Problem**: Click filters data out, removing comparison context — but the user just wanted to emphasize one entity.
+
+**Solution**: Use Pattern 3 (Comparison Spotlight) — Parameter targeting a custom chart's `highlight_X` argument, `visible=False`, `"NONE"` in selector options. The data stays, only the styling changes.
```
huong-li-nguyen: I feel like we should just add code snippets for all of these with examples, then we dont' need to explain all of this right? Or even just referencing the api docs?

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
```diff
@@ -0,0 +1,168 @@
+---
+name: wiring-vizro-actions
+description: Use this skill when adding cross-filter, cross-highlight, drill-through, or data export interactions to a Vizro dashboard — both for choosing the right interaction pattern during design and for implementing actions in code. Activate when the user wants clicks to filter or highlight other components, needs cross-page navigation with pre-set filters, or wants users to download data.
```
huong-li-nguyen: "highlight other components" is too broad right? cross-filter, cross-highlight only works from graph -> table and vice versa, also maybe graph-> figure and vice versa. So in my mind it's just graph, table and figure (basically all components that can take on a df). For example, the use case of button -> graph doesn't make sense as the button doesn't have any data to display that could be clicked

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
```diff
@@ -0,0 +1,168 @@
...
+---
+
+# Wiring Vizro Actions
+
+## Core concept: Source → Control → Target
+
+All advanced interactions follow the same shape:
+
+1. A **source** component (Graph, AgGrid) triggers `va.set_control` when the user clicks it.
```
huong-li-nguyen: Can a Figure also be a source component? Yes  right?

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
```diff
@@ -0,0 +1,168 @@
...
+`import vizro.actions as va`. Built-in actions are passed directly — do **not** wrap them in `vm.Action`.
+
+```python
+# Correct
+vm.Button(text="Export data", actions=va.export_data())
+vm.Graph(actions=va.set_control(control="region_filter", value="y"))
+
+# Wrong — do NOT wrap built-ins
+vm.Graph(actions=[vm.Action(function=va.set_control(...))])
+```
```
huong-li-nguyen: In theory it's not wrong right? It's just verbose. I would just leave it out

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
```diff
@@ -0,0 +1,168 @@
...
+└─ NO → continue
+
+Source chart needs visual confirmation of click AND must filter elsewhere?
+├─ YES → Pattern 5
+└─ NO → continue
+
+Users need to download data?
+├─ YES → Pattern 6
+└─ NO → Standard filters/parameters are sufficient
+```
```
huong-li-nguyen, line 67: I would actually be a lot stricter 😄 I don't want any fancy dashboard interactions enabled unless I asked for them. Especially if some of them are only visible after clicking etc.

I would say we always default to just simple filters unless the user asks for it? @stephkaiser what would you recommend here in terms of UX? applied UX?

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
```diff
@@ -0,0 +1,168 @@
...
+
+## When NOT to use advanced interactions
+
+- View-only dashboards (executive summary — users glance, not explore)
+- Simple filtering needs (a sidebar dropdown covers it)
+- Small datasets (< ~5 groups — a filter is simpler and clearer)
+- More than 2 interaction patterns per page (becomes confusing)
```
huong-li-nguyen, line 76: I think our default should be to not use advanced interactions (cross-filtering, cross-highlighting, drill-downs are all nice-to-have but functionally you cover 80% of all dashboards with just a simple filter navbar)

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
```diff
@@ -0,0 +1,168 @@
...
+    header="Click a cell to filter the table below",
+    figure=px.density_heatmap("appointments", x="day", y="time_slot"),
+    actions=[
+        va.set_control(control="day_filter", value="x"),
+        va.set_control(control="time_filter", value="y"),
+    ],
+)
+vm.Filter(id="day_filter", column="day", targets=["appointments_table"])
+vm.Filter(id="time_filter", column="time_slot", targets=["appointments_table"])
+```
```
huong-li-nguyen: Ah yes! This is what I was expecting! I assume in the other files we don't reference these code snippets but instead explain it in words, because we don't want this entire file to be read in?

vizro-e2e-flow/skills/dashboard-design/SKILL.md
```diff
@@ -114,6 +114,28 @@ Tier 3: Component-level
 
 Load the **designing-vizro-layouts** skill for grid system, component sizing, filter placement, and selector rules. Use the [wireframe templates](references/wireframe_templates.md) when building ASCII wireframes for user approval.
 
+### Interaction Design
+
+Beyond standard sidebar filters, Vizro supports advanced interactions where clicking a chart or table affects other components. Load the **wiring-vizro-actions** skill for the 6 named interaction patterns (Hierarchical Drill-Down, Single-Page Drill-Down, Comparison Spotlight, Multi-Dimensional Slice, Select & Explore, Data Export) with wireframes, spec entries, and code.
+
+All advanced interactions follow **Source → Control → Target**: a source component (Graph, AgGrid) sets an intermediate control (Filter or Parameter, always with an explicit `id`), which updates target components.
```
huong-li-nguyen: target components can also only be graph, aggrid, figures, right?

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
```diff
@@ -0,0 +1,936 @@
...
+Page 1 (Overview) ──click rep row──> sets rep_filter (show_in_url=True)
+                                              │
+                                              ▼
+                                       Page 2 (Detail), filtered to that rep
+                                              │
+                                              └──> [← Back] returns to Page 1
+                                              └──> [Export] downloads filtered data
+```
+
+### Wireframe — Page 1: Overview
```
huong-li-nguyen, line 64: Ahh, I think I disagree that this should happen 😅  I am in general, not a fan of hierarchical drill-downs because it's a lot of clicking and confusing for users. I generally like to keep the dashboard as simple and intuitive as possible, and i think this is also what our default should be. Just because there is hierarchy inside the data, I do not think that we by default should apply hierarchical drill-downs to all of the charts 😬 @stephkaiser , curious to hear what you think!

But in this case, I would just prefer to have a standard filter sidebar and be able to filter by region and then see the charts cut by different categories. I think that's a more natural user interaction flow, then the user clicking on the same chart multiple times to drill to the details and then drill out again to see another region.

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
```diff
@@ -0,0 +1,936 @@
...
+|  |          |  |  header: "Click a row to view details"                          |    |
+|  |          |  |  [click: drill-through → rep_filter → Detail page]              |    |
+|  |          |  +----------------------------------------------------------------+    |
++--+----------+------------------------------------------------------------------------+
+```
+
+### Wireframe — Page 2: Detail (Flex layout)
+
+```
++--+----------+------------------------------------------------------------------------+
```
huong-li-nguyen, line 95: this is fine for me 👍  Even though I wonder if this is the right position for the Back button? @stephkaiser

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
```diff
@@ -0,0 +1,936 @@
...
+|  |FILTERS   |  |  [Horizontal Bar]                                               |    |
+|  |          |  |  [click: cross-filter → dept_filter]                            |    |
+|  |          |  +----------------------------------------------------------------+    |
+|  |          |                                                                        |
+|  |          |  CONTAINER: Employee Detail [filtered by dept_filter] (variant=filled) |
+|  |          |  +----------------------------------+  +-----------------------------+  |
+|  |          |  |  CHART: Compensation Distribution|  |  CHART: Performance Ratings  |  |
+|  |          |  +----------------------------------+  +-----------------------------+  |
+|  |          |  +----------------------------------------------------------------+    |
+|  |          |  |  TABLE: Employees [sortable]                                    |    |
```
huong-li-nguyen: Same here. I think a simple filter is fine? Why don't we just let the agent or the user decide when to use advanced interactions? My preference would always be to keep it as simple as possible and any advanced customization should be triggered by the user, and not be our default

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
```diff
@@ -0,0 +1,936 @@
...
+
+| Decision | Choice | Reasoning |
+| --- | --- | --- |
+| Filter scoped to container | Yes | Only the Employee Detail container is filtered; the bar chart at top stays unfiltered |
+| Container `variant="filled"` | Recommended | Visually groups filtered components |
+| Filter visible | Yes | User can change department from a dropdown without re-clicking |
+
+---
+
+## Pattern 3: Comparison Spotlight (cross-highlight)
```
huong-li-nguyen, line 334: This is fine for me - should only be triggered if the user explicitly asks to highlight something in another chart

huong-li-nguyen (commented):
Hi @lingyielia ,

thanks for adding this in! Looks like a lot of work, so I don't want to be a blocker. Whatever you guys think is best, I'll go with that. The only point I want to raise is that I kind of disagree that some of these advanced interactions should be triggered by default 😬  I kind of disagree with some of the "when to use" sections, because in my opinion, we should keep the dashboard as simple and intuitive as possible. And a lot of the advanced interactions are not necessarily needed if you just do all the filtering with the sidepanel, which is more intuitive. Curious to hear what @stephkaiser thinks, but I think I would add less instructions on when to use these, and default more towards our basic filtering system and only trigger the advanced actions / interactions if a user explicitly asks for it (not just when a certain hierarchy in the data is detected).

But i'll leave that ultimately for you guys to decide 👍  I didn't test out the skill as I think @petar-qb will do a much better job in this than I do, so I only looked at the logic

vizro-e2e-flow/evals/dashboard-build/evals.json
petar-qb, line 1: The "expectations" across the three use cases look very similar. No need to address it in this PR, but just wondering whether is it ok with you?

If you think it doesn't make much sense, we could revisit and tailor the expectations more closely to each use case (I'm happy to open a ticket if you agree).

vizro-e2e-flow/evals/wiring-vizro-actions/evals.json
```diff
@@ -0,0 +1,94 @@
...
+      "expected_output": "A single-page Vizro dashboard implementing Pattern 2 (Single-Page Drill-Down): clicking a bar in the department chart filters the employee container.",
+      "files": [],
+      "expectations": [
+        "Dashboard launches without errors and has no browser console errors",
+        "`import vizro.actions as va` appears in app.py",
+        "Department bar chart uses `va.set_control` with the Filter's `id` as `control` and `value='y'` (horizontal bar) or `value='x'` (vertical bar) matching the orientation",
+        "Bar chart is horizontal (orientation='h') — matches the project default for bar charts",
```
petar-qb, line 13: Isn't the vertical orientation the default one 😄

vizro-e2e-flow/evals/wiring-vizro-actions/evals.json
```diff
@@ -0,0 +1,94 @@
...
+        "Filter has an explicit `id` so `set_control` can reference it",
+        "Employee section is grouped inside a `vm.Container`, and the Filter is scoped to that container (either declared as a container control, or with explicit `targets`)",
```
petar-qb: Looking at the input prompt, I would not expect the employee section to be grouped inside a container. How did you build these use cases and evals? Did you sketch the input prompt using agents, and then have another agent build the evals based on the results?

petar-qb: Also, should we expect that the filter is invisible according to this sentence from the input prompt? -> "I don't want a separate dropdown — clicking the bar IS the filter"

vizro-e2e-flow/evals/wiring-vizro-actions/evals.json
```diff
@@ -0,0 +1,94 @@
...
+        "Department bar chart has a short action-oriented `header` (e.g. `header='Click a bar to filter by department'`)",
+        "No `filter_interaction` (deprecated API) anywhere in app.py",
+        "Spec file `spec/2_interaction_ux.yaml` includes an `interactions:` block referencing Pattern 2 / Single-Page Drill-Down, with source → control → targets documented"
+      ]
```
petar-qb, line 21: Should we add more evals that expect ag-grid table and other listed figures on the page?

vizro-e2e-flow/evals/wiring-vizro-actions/evals.json
```diff
@@ -0,0 +1,94 @@
...
+      "expected_output": "Two-page Vizro dashboard implementing Pattern 1 (Hierarchical Drill-Down) with cross-page drill-through, back button, and data export.",
+      "files": [],
+      "expectations": [
+        "Dashboard launches without errors and has no browser console errors",
+        "Page 1 reps table (AgGrid) uses `va.set_control` referencing a Filter on Page 2",
+        "Page 2 Filter has `show_in_url=True` — required for cross-page drill-through",
+        "Page 2 uses `layout=vm.Flex(direction='column')` at the page level (not a Grid) so the back button doesn't waste a full row",
+        "First component on Page 2 is a back `vm.Button` with `href='/<page-1-slug>'` and `variant='outlined'` — not an action; uses navigation",
+        "Page 2 includes a `vm.Button(text='Export data', actions=va.export_data())` (or similar)",
+        "Remaining Page 2 content (KPIs + charts + table) is wrapped in a `vm.Container` with its own `vm.Grid` layout for precise positioning",
```
petar-qb: I don't think we should test page's layout that deeply in the wiring-vizro-actions "tests". Also, there's no inputs in the prompt about the components positioning and layout for this use case as I can see (I'm wondering how these tests pass 😄).

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
lingyielia: Right now it's already been triggered quite rarely, looks like the agent is making a good judgement. 😃

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
lingyielia: Right now it's already been triggered quite rarely, looks like the agent is making a good judgement. 😃

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
lingyielia: this prompt got picked up quite well. for model, it's just one line setence

vizro-e2e-flow/evals/wiring-vizro-actions/evals.json
```diff
@@ -0,0 +1,94 @@
...
+      "expected_output": "Single-page dashboard implementing Pattern 3 (Comparison Spotlight) — cross-highlight from bar chart into a custom bump chart.",
+      "files": [],
+      "expectations": [
+        "Dashboard launches without errors and has no browser console errors",
+        "Bump chart is implemented as a custom `@capture('graph')` function with a `highlight_country=None` parameter",
+        "Inside the custom chart: fades all traces (`update_traces(opacity=0.3, ...)`) when `highlight_country is None`, and bolds the matching trace (e.g. `update_traces(selector={'name': highlight_country}, opacity=1, line_width=3)`) when a country is selected",
+        "A `vm.Parameter` targets the custom chart's `highlight_country` argument (target string `<chart_id>.highlight_country`) and has `visible=False`",
+        "The Parameter's selector includes `'NONE'` as an option (typically first) so users can reset to no highlight",
```
petar-qb: I feel most of these "expectations" are too restrictive ("overfitted") and could potentially fail even if we really improve skills. By "too restrictive" I mean how agents can exactly guess that for example opacity is `0.3` or that custom graph parameter name is exactly "highlight_country".

petar-qb: Same goes for the two evals below (which I just skimmed and haven't investigated in detail 😄)

vizro-e2e-flow/skills/dashboard-build/references/custom_charts_guide.md
petar-qb, line 1: Now I see from where "too restrictive" expectations in the evaluations above come from 😄. 

At first, I don't think having too detailed skills is the perfect move for us because it might cause overfitting. I will test  the new skill later today/tomorrow morning and will back to you and to this comment with results. 

[ ] - @petar-qb Describe testing results here. Check what happens if user asks for multi-select highlighting and similar.

vizro-e2e-flow/skills/dashboard-build/SKILL.md
```diff
@@ -97,8 +98,11 @@ Navigate the running dashboard to catch two types of errors that code review alo
 1. Using your chosen tool, perform these checks:
 
     - Navigate to dashboard URL (e.g., `http://localhost:8050`)
-    - Click through all pages
-    - Check browser console for errors
+    - Click through all pages and check the browser console for errors
+    - **If `app.py` contains any `actions=` (cross-filter, cross-highlight, drill-through, export), exercise each one**: grep your app for `actions=` and for every match, click the corresponding source (a point on the scatter, a bar, a row in the AgGrid, the export Button, etc.). Then check:
+        - the browser console for `Callback error updating ...` messages
+        - the network tab for any 500 response on `_dash-update-component`
+        - the action's intended effect actually happened (URL updated for `show_in_url=True`, target filter changed, highlight applied, file downloaded, etc.)
```
petar-qb, line 105: Like this a lot!

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
```diff
@@ -126,6 +126,66 @@ Tertiary metric: Smaller chart or table row
...
+
+### Missing Back Button on Drill-Through Target
+
+**Problem**: User navigates to a detail page (Pattern 1) and feels trapped.
+
+**Solution**: The drill-through target page MUST include a back `vm.Button(text="← Back", href="/source-page", variant="outlined")` as the first component. The page MUST also use `layout=vm.Flex(direction="column")` so the button takes natural height — Grid would waste a 140px+ row.
+
+### Missing `show_in_url=True` for Cross-Page
+
+**Problem**: Cross-page drill-through silently fails — no error, just no filtering on the target page.
```
petar-qb: I'm quite sure it doesn't fail silently. So, I would or reword or even remove this section completely (up to you).

Image

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
```diff
@@ -126,6 +126,66 @@ Tertiary metric: Smaller chart or table row
...
+**Problem**: Click filters data out, removing comparison context — but the user just wanted to emphasize one entity.
+
+**Solution**: Use Pattern 3 (Comparison Spotlight) — Parameter targeting a custom chart's `highlight_X` argument, `visible=False`, `"NONE"` in selector options. The data stays, only the styling changes.
+
+### Self-Highlight Without Visual Contrast
+
+**Problem**: Clicked bar/point looks identical to the others — no feedback.
+
+**Solution**: The custom chart must clearly differentiate highlighted vs non-highlighted (opacity, color, line width, marker border). See the **wiring-vizro-actions** skill's Pattern 3 / 5 and `custom_charts_guide.md` ("Highlight-Aware Custom Charts").
+
```
petar-qb: I wouldn't force the "custom self-highlighting" as it now should be natively supported. 
See screen recording from the https://github.com/mckinsey/vizro/pull/1515 description.

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
petar-qb: I don't think we should avoid hierarchical drill-downs by any price. If a proxy filter is visible then users will be able to do any: set the filter value directly or through the "parent hierarchy" source cross-filter figure.

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
petar-qb: I saw a rule that the "Back" button **must** always exist in the target-drill-through page in several different places in this PR. I would relax that rule and "force" to do that only if there are more than 3 pages on the dashboard (or something like that). If there are only two/three pages, users will figure out quickly how to get back to the source-drill-through page.

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
```diff
@@ -0,0 +1,936 @@
...
+### Wireframe
+
+```
++--+----------+------------------------------------------------------------------------+
+|  |NAV       |                                                                        |
+|  |          +------------------------------------------------------------------------+
+|  |HR        |  +----------------------------------------------------------------+    |
+|  |          |  |  CHART: Avg Tenure by Department                                |    |
+|  |----------|  |  header: "Click a bar to filter by department"                  |    |
+|  |FILTERS   |  |  [Horizontal Bar]                                               |    |
```
petar-qb: Should this filter be moved in the container instead?

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
```diff
@@ -0,0 +1,936 @@
...
+### When NOT to use
+
+- User wants to filter data out (use Pattern 1, 2, or 4).
+- Fewer than ~5 entities (highlighting adds little when everything is already legible).
+- No clear "compare one vs all" mental model.
+
+### Interaction flow
+
+```
+[Source chart with many entities] ──click entity──> sets highlight_param (visible=False)
```
petar-qb: Should you emphasise that it's a parameter in this flow?

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
```diff
@@ -0,0 +1,936 @@
...
+| Decision | Choice | Reasoning |
+| --- | --- | --- |
+| Two separate Filters | Required | Each dimension needs its own control so it can be reset independently |
+| Actions chain | Required | Both filters must be set on one click |
+| Both filters visible | Recommended | Users can see the active slice |
+| Target is the table only | Yes | Heatmap itself does not need to react to the click |
+
+---
+
+## Pattern 5: Select & Explore (self-highlight + cross-filter)
```
petar-qb: I think I've commented this somewhere above: self-highlight should be included by default for all graphs that source set_control. 

[ ] - @petar-qb double check the statement with the example below

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
petar-qb: Yes, but currently it's not that useful as it can only propagate/send a hardcoded value (same as button for example). So it can't fetch and propagate a dynamic value that's built based on the data_frame (and filters). So:

```py
vm.FIgure(
    figure=kpi_card(...), 
    actions=set_control(control="control-id", value="setosa"),  # sends "setosa" to the targeted filter
)
```

(Unfortunately, this limitation is not easy to overcome..)
I suggest not mentioning that the Figure can source the `set_control` action. (same as for buttons)

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
petar-qb: Actually, it's incorrect and it raises an exception. I think it's done like that deliberately, so that users don't mix `vm.Action.outputs` / `vm.Action.inputs` with the builtin action, and so that they don't mix predefined input builtin action arguments (like `file_format` from the `va.export_data`) with the `vm.Action`. 

Hence, I suggest leaving this section.

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
petar-qb: I’m finding myself agreeing with this more the more I come across similar comments.

petar-qb (commented):
@lingyielia thanks for the great PR 🚀 This feels like a big step forward toward a solid text-to-dashboard experience. I left quite a few comments, but feel free to ignore and resolve anything you don’t find useful 😄

Quick question: are there already tickets for developing skills (and evals) around custom actions (vm.Action) and conditional notifications?

I’ll try the skill out in practice tomorrow and follow up with more detailed feedback.

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
lingyielia: Hmm tooltip might be too hidden... Also checked the https://huggingface.co/spaces/vizro/BI-dashboard, most of them are using headers. @stephkaiser could you give us some guidance here?

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
lingyielia: In practice so far, it's not been triggered very often. And for ones added, it's actually quite cool and interesting to use. And on the other side, this feature is so hidden that probably very few people know its existence and no one have time to read the docs to enable it. Would be cool to see someone can actually use this feature.
 
So I would push for keep the current level and see how it goes. It will be a easy update if this turns out to be not useful

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
lingyielia: update with more strict "When to use" / "When NOT to use", to refuse what simple filter should handle

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
lingyielia: update with more strict "When to use" / "When NOT to use", to refuse what simple filter should handle

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
lingyielia: updated with more strict "When to use" / "When NOT to use", to refuse what simple filter should handle

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
lingyielia: great catch! updated

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
lingyielia: In which case should we recommend using this special case self-highlighting? https://vizro.readthedocs.io/en/latest/pages/user-guides/graph-table-actions/#self-highlight-a-graph

If this is a redundant feature, shall we remove it from the docs?

lingyielia: I now dropped the pattern 5 entirely

vizro-e2e-flow/skills/dashboard-build/references/custom_charts_guide.md
lingyielia: removed this section entirely

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
lingyielia: removed

vizro-e2e-flow/evals/wiring-vizro-actions/evals.json
lingyielia: loosed what's mentioned here. You can try with `/skill-creator run the xyz evals with-skill vs. without-skill` and see which makes more sense.

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
lingyielia: Sadly even 2 pages, this could be quite confusing how to get back. 🤣 We need a prescribed flow ideally end to end

vizro-e2e-flow/skills/dashboard-build/SKILL.md
lingyielia: updated it to only reference the docs rather than listing out all details

vizro-e2e-flow/evals/wiring-vizro-actions/evals.json
lingyielia: Ah, that's a preference in skill. Most of time the horizontal bars renders more nicely

lingyielia:
> Quick question: are there already tickets for developing skills (and evals) around custom actions (vm.Action) and conditional notifications?

No, feel free to create and pick up these 2 tickets 👍

vizro-e2e-flow/evals/dashboard-build/evals.json
lingyielia: There are two main considerations when writing expectations:
1. there are some universal requirements we want they are met across all eval runs. This would make all cases have duplicates to some extend
2. there are very specific ask only in one eval.

In terms of number of evals, we need to consider the tradeoffs. More evals could help determine how robust the current pr is, but this also means more token and time cost. 👍

vizro-e2e-flow/evals/wiring-vizro-actions/evals.json
lingyielia: Happy to address this improvement in a separate pr 👍

vizro-e2e-flow/evals/wiring-vizro-actions/evals.json
lingyielia: The overall skill suggest using grid always for more precise control of layout. The button is an odd case, because it would take too much space in grid.

vizro-e2e-flow/evals/wiring-vizro-actions/evals.json
lingyielia: Now it's default to be visible, but can to invisible if requested

lingyielia: the evals are usually created in the early phase, and all are written by agent, based on developer's high level goals. But surely worth tuning based on what often go wrong and where we care most.

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
lingyielia: Great catch! updated elsewhere to reference the skill instead

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
lingyielia: updated to be more specific

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
lingyielia: rephrased and simplified items here to keep them as more general design mistakes

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
petar-qb: Just to add that I created [the issue](https://github.com/McK-Internal/vizro-internal/issues/2745) to handle highlighting figure's interactiveness natively in Vizro 😃

huong-li-nguyen (approved):
Approving under the condition that we just incorporate whatever @stephkaiser thinks is best in terms of UX on the remaining open questions 👍 I am fine with whatever she decides

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
petar-qb: cool works for me @lingyielia

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
petar-qb: Great, I will remove it from the docs.

petar-qb (approved):
An amazing work @lingyielia!!! 🚀 I tested it thoroughly with dashboard prompts I made and it really works better with the newly introduced skill 👍

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
```diff
@@ -0,0 +1,709 @@
...
+- The control is **always explicit** with an `id`. You never connect source to target directly.
+- The source uses `va.set_control` (a built-in action). The control then drives the target.
+- Targets are inferred from the control: a Filter's `targets` (or its container scope), a Parameter's `targets`.
+- This shape composes: multiple sources can write to the same control; one source can chain multiple `set_control` calls.
+
+## Pattern 1: Hierarchical Drill-Down (cross-page)
+
+**Domain example**: Sales Pipeline (Region → Sales Rep → Deals).
+
+### When to use
```
stephkaiser, line 39: very nice :)

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
stephkaiser: this is probably the best place for a back button with our current Vizro design. Its the same position as what we've used in the past

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
stephkaiser: I agree that dashboards are best when they are simple and communicate the data clearly - but I also know that depending on the client use case, sometimes hierarchical drill-downs are necessary like in this example where an overview and individual rep view is required. However, i do agree that it is sometimes clearer if users can filter in the filter sidebar, rather than clicking on the chart (which doesnt always look clickable/interactive) and using the chart as a filter (or a parameter like in pattern 3: cross-highlight). Maybe we can add a flexible rule like: _"Don't rely on charts/graphs for actions like filters or parameters, always default to assigning the correct control in the sidebar instead. Only override this if the user has specified something else."_ 

I think the "When to use/When NOT to use" does define the pattern decision making well (although I have not tested this skill yet).

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
stephkaiser: In this scenario/pattern, i'd say we dont need the side bar (only the header since there is only one page so no need for navigation, and controls should be in a container ideally above the chart (as it is currently in this wireframe :)). 

Saying this because the wireframe looks like there is a nav sidebar

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
```diff
@@ -0,0 +1,709 @@
...
+                vm.Graph(id="perf", figure=px.box("employees", x="department", y="rating")),
+                vm.AgGrid(id="emp_table", figure=dash_ag_grid("employees")),
+            ],
+            controls=[vm.Filter(id="dept_filter", column="department")],
+        ),
+    ],
+)
+```
+
+### Key decisions
```
stephkaiser, line 323: probably also need to have a reset button for single-page dashboard patterns like this (styled like the Export button)

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
stephkaiser: not sure I fully understand your question @huong-li-nguyen, lets discuss when youre back.

Separate comment - users will need to combine some patterns together, for example 1 and 5. Is this defined anywhere in the skill?

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
stephkaiser: agree with Li

vizro-e2e-flow/skills/dashboard-design/references/common_mistakes.md
```diff
@@ -126,6 +126,48 @@ Tertiary metric: Smaller chart or table row
...
+**Problem**: Click filters data out, removing comparison context — but the user just wanted to emphasize one entity.
+
+**Solution**: Use Pattern 3 (Comparison Spotlight) — Parameter targeting a custom chart's `highlight_X` argument, `visible=False`, `"NONE"` in selector options. The data stays, only the styling changes.
+
+### Invisible Interactions With No Reset Path
+
+**Problem**: Hidden controls leave users with no visible state and no obvious reset.
+
+**Solution**: Keep controls visible by default; hide one only when the chart's own state communicates the selection (Pattern 3 highlight is the canonical case).
```
stephkaiser: and add a reset button when there are no controls/sodebar and controls are in the chart - for single page dashboards only

petar-qb:
It would be awesome to update the `vizro-e2e-flow/README.md` file as well.

vizro-e2e-flow/skills/wiring-vizro-actions/SKILL.md
lingyielia: Updated the flow to make 5 independent from other interactions

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
lingyielia: added one line to emphasis on this default recommendation

vizro-e2e-flow/skills/wiring-vizro-actions/references/actions-reference.md
lingyielia: Pattern 2 has "visible=True" default. updated in the common mistakes to further cover the edge case

PR #1782, opened 2026-06-29
[Feat] Add Vizro actions log to dash dev tools

nadijagraca:
closes https://github.com/McK-Internal/vizro-internal/issues/2744

## Description
This PR adds Vizro actions log to dash dev panel when `debug=True`. 
## Screenshot
Screenshot 2026-06-29 at 12 04 52

## Notice


vizro-core/changelog.d/20260618_181901_nadija_ratkusic_graca_devtool_logs.md
```diff
@@ -0,0 +1,48 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category with a link to the relevant PR at the end of your entry, e.g. Enable feature XXX. ([#1](https://github.com/mckinsey/vizro/pull/1))
```
petar-qb: Let's mention the change here, and add a sentence or two about the new feature somewhere in the docs

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -670,7 +674,24 @@ def action_callback(external: list[Any] | dict[str, Any], internal: dict[str, An
                 )
                 notification_key, notification_result = notification_payload.key, notification_payload.result
 
-            return_value = {"internal": {"action_finished": action_finished, "action_progress_indicator": no_update}}
+            timestamp = datetime.now(tz=timezone.utc).strftime("%H:%M:%S.%f")[:-3]
+            if error_msg is None:
+                log_text = f"[{timestamp}] ===== Running action with id {self.id}, function {self._action_name} ====="
+            else:
+                log_text = (
```
petar-qb: Add a comment similar to one below that explains how it’s not possible to log an error and raise the exception as stacktrace in debug more at the same time. So action error is logged only if error is handled with the notification.

comment from below:
```
            # It's not possible to both propagate the error to the UI and show an error notification at the same time.
            # Therefore, we either display an error notification or propagate the exception to avoid failing silently.
```

vizro-core/src/vizro/models/_action/_action.py
```diff
@@ -670,7 +674,24 @@ def action_callback(external: list[Any] | dict[str, Any], internal: dict[str, An
...
+                    f"[{timestamp}] ===== FAILED action with id {self.id!r}, "
+                    f"function={self._action_name!r}  error={error_msg!r}"
+                )
+            action_log = Patch()
+            action_log.append(log_text + "\n")
```
petar-qb: Let's move this into an util function and call it something like:
```py
            return_value = {
                "internal": {
                    "action_finished": action_finished,
                    "action_progress_indicator": no_update,
                    "action_log": _get_action_log(id=self.id, name=self._action_name, error_msg=error_msg),
                }
            }
```

vizro-core/src/vizro/models/_page.py
```diff
@@ -230,6 +230,7 @@ def build(self) -> _PageBuildType:
         components_container.children.extend(
             [
                 *action_components,
+                dcc.Store(id="vizro_logs_store", data=[], storage_type="local"),
```
petar-qb: I don’t like `storage_type="local"` as it keeps logs even when app tab is closed and then reopened again later. Let’s use `storage_type="session"` and please double check if everything works fine

vizro-core/src/vizro/_vizro.py
```diff
@@ -195,6 +199,64 @@ def run(self, **kwargs: Any):
...
+            id="vizro_logs_offcanvas",
+            placement="bottom",
+            scrollable=True,
+            backdrop=False,
+            is_open=False,
+            children=[
+                dbc.Button("Clear logs", id="vizro_logs_clear", size="sm", color="secondary", class_name="mb-2"),
+                html.Pre(id="vizro_logs", children=[]),
+            ],
+        )
```
petar-qb: Is there any way to move the “Clear logs” button up to the offcanvas's header? 

Image

vizro-core/src/vizro/_vizro.py
```diff
@@ -195,6 +199,64 @@ def run(self, **kwargs: Any):
...
+        # Sync store and panel: vizro_logs_store lives in the page layout so Patch() on the store is safe.
+        # vizro_logs is in the DevTools overlay (optional=True) so this no-ops when not in debug mode.
+        # A plain list assignment avoids the Redux-store lookup that Patch() would require for a DevTools component.
+        clientside_callback(
+            "function(data) { return data || []; }",
+            Output("vizro_logs", "children"),
+            Input("vizro_logs_store", "data"),
+            optional=True,
+        )
```
petar-qb: Honestly, I don't understand what this callback is for. Would you mind explaining it?

vizro-core/src/vizro/models/_page.py
petar-qb: I haven’t tested this, but I think it should be possible to make logs work per page by adjusting the store ID to something like `id="vizro_logs_store_{self.id}"`, and by returning the `action_log` from the callback in `_action.py` only for the current page.

However, I’m not sure I prefer that approach over dashboard-level logs for two reasons:

1. It seems better to track the user’s full activity across the dashboard.
2. It’s unclear where cross-page actions should be logged.

petar-qb: Do you think we should log `reset_controls` too, but without using the same stdout string structure, so it’s clear that resetting controls does not trigger a new HTTP request?

petar-qb: If we decide to keep the dashboard-level logs I think it would be awesome if opening a new page is recognised and logged differently like `\n` followed by `...=== Page “My second page” is opened ===…`  or whatever, followed by `on_page_load` action log. If you also think this would be nice ui improvement, then let's add that change too 😃

petar-qb (commented):
Amazing piece of work!! 🚀

vizro-core/src/vizro/models/_action/_action.py
petar-qb: What do you think about this code styling improvement?

vizro-core/src/vizro/_vizro.py
petar-qb: unresolving this

vizro-core/src/vizro/actions/_on_page_load.py
```diff
@@ -41,3 +41,11 @@ def outputs(self):  # type: ignore[override]
...
             for target in self.targets
         }
+
+    @property
+    def _log_header(self) -> str:
+        # Prepended to this action's DevTools log entry so a page (re)load stands out above the action line. Note this
+        # also fires on refresh and after a control reset, since both re-run the on-page-load action - hence "loaded"
+        # rather than "opened".
+        page = model_manager._get_model_page(self)
+        return f'\n=== Page "{page.title}" loaded ===\n'
```
petar-qb, line 51: Nice solution!! 🥇

vizro-core/src/vizro/_vizro.py
```diff
@@ -195,6 +199,80 @@ def run(self, **kwargs: Any):
...
+
+        for component in [button, offcanvas]:
+            c = component.to_plotly_json()
+            hooks.devtool(
+                namespace=c["namespace"],
+                component_type=c["type"],
+                props=c["props"],
+            )
+
+        clientside_callback(
```
petar-qb, line 251: Not a huge mistake (if at all), so just to let you know that opened logs panel closes when page refreshes

https://github.com/user-attachments/assets/bbb042ee-4e21-494e-8a91-47c75d148a86

vizro-core/src/vizro/_vizro.py
```diff
@@ -195,6 +199,80 @@ def run(self, **kwargs: Any):
...
+            "function(_, is_open) { return !is_open; }",
+            Output("vizro_logs_offcanvas", "is_open"),
+            Input("open_vizro_logs", "n_clicks"),
+            State("vizro_logs_offcanvas", "is_open"),
+            optional=True,
+        )
+        # Sync store and panel: vizro_logs_store lives in the page layout so Patch() on the store is safe.
+        # vizro_logs is in the DevTools overlay (optional=True) so this no-ops when not in debug mode.
+        # A plain list assignment avoids the Redux-store lookup that Patch() would require for a DevTools component.
+        clientside_callback(
```
petar-qb, line 263: Actions logs content "clears" when DevTools is collapsed and then expended. To reload an old state you have to refresh the page or to run some action. 

https://github.com/user-attachments/assets/e6ee5ccf-8729-4041-8613-113e799d68a7


It would be nice if we are able to collapse the DevTools panel without even closing the actions logs panel, as there are already two ways to do so: by clicking on the "X" button, or by clicking on the "Vizro logs" button again.

vizro-core/src/vizro/_vizro.py
```diff
@@ -195,6 +199,80 @@ def run(self, **kwargs: Any):
...
+        # Log control resets. Registered once globally (not per page) since reset-button and vizro_logs_store are
+        # shared ids. Deliberately uses a different format from action logs - no "=====" wrapping and a coarser
+        # HH:MM:SS timestamp - to make clear the reset runs client-side and is not itself a server callback; any
+        # on-page-load actions it triggers afterwards produce their own "=====" entries. optional=True no-ops this
+        # when there are no controls (reset-button absent). A plain array assignment is used because clientside
+        # callbacks can't use Patch(); it lands before the downstream action Patches, so nothing is clobbered.
+        clientside_callback(
+            # UTC HH:MM:SS to match the action log timestamps, which are generated server-side in UTC.
+            "function(_, data) { const t = new Date().toISOString().slice(11, 19); "
+            "return [...(data || []), `[${t}] ----- Reset controls -----\\n`]; }",
```
petar-qb: Would you mind adjusting the timestamp so it shows milliseconds like in other logs. Also, let's use `=` instead of `-` as it's used in other logs too.

Image

petar-qb (commented):
This version definitely looks better. I left a few additional comments and kept the Copilot comments unresolved as well. Would you mind taking another pass through all unresolved comments (my and copilot) and either addressing them or adding a note on why they may not apply?

vizro-core/src/vizro/_vizro.py
nadijagraca: This callback connects vizro_logs_store and off canvas component where vizro logs are displayed. Actions write log entries to vizro_logs_store using Patch() (only place where Patch() works safely), but the visual panel (vizro_logs) lives in the DevTools overlay where Patch() doesn't work. So this callback copies the full store data over to the panel. 

(I'm sorry I though I replied on this in the earlier iteration.)

vizro-core/src/vizro/models/_action/_action.py
nadijagraca: I don't see huge benefit in putting it in an util function as it's ever going to be used once in the code. Although, it would be easier to test this part of the code in unit tests. If you still want me to move it into an util function let me know, and I'll make the changes.

vizro-core/src/vizro/models/_page.py
nadijagraca: I agree, I prefer the dashboard-level logs.

vizro-core/src/vizro/_vizro.py
nadijagraca: It is tricky to achieve that behaviour as the button and offcanvas are both injected via hooks.devtool, so they live inside the Dash debug-menu content. Collapsing the menu unmounts that whole subtree, so the offcanvas is removed from the DOM (not just hidden) and its open state is lost; re-expanding mounts a fresh, closed one. To keep it open, the panel has to be mounted somewhere that survives the collapse.
We have two options:
1. Move only the offcanvas into the persistent dashboard layout; keep the toggle button in the debug menu. The panel then survives collapse (only "X" or the button close it). 
2.  Move the whole thing (button + offcanvas) into our own debug-only overlay, fully decoupling it from the debug-menu lifecycle. More self-contained, but we give up the native debug-menu button styling/placement.

vizro-core/src/vizro/models/_action/_action.py
petar-qb: yepp, let's do it if you don't mind. It's easier to test and it declutters a bit already complex `action_callback`.

vizro-core/src/vizro/_vizro.py
petar-qb: I'd avoid the second option and stick with the first. I think we might even be able to handle [this case](https://github.com/mckinsey/vizro/pull/1782#discussion_r3536348068) with the same setup.

petar-qb (approved):
Amazing @nadijagraca!! 🚀 

If possible and easy, let's add auto scrolling, so that new logs are always visible without having to manually scroll the logs panel. Other than that, the new feature works smoothly👌

vizro-core/src/vizro/static/js/models/dashboard.js
```diff
@@ -145,11 +145,76 @@ function collapse_nav_panel(n_clicks, is_open) {
...
+    return;
+  }
+
+  const scrollContainer = document
+    .getElementById("vizro_logs")
+    ?.closest(".offcanvas-body");
+  if (scrollContainer) {
+    scrollContainer.scrollTop = scrollContainer.scrollHeight;
+  }
+}
```
petar-qb, line 163: I think "vizro_logs_end" always exists when the `scroll_vizro_logs_to_end` is called. If you agree let's simplify its implementation. 

```suggestion
function scroll_vizro_logs_to_end() {                                                                                                                                                                          
    document.getElementById("vizro_logs_end")?.scrollIntoView({ block: "end" });                                                                                                                                 
  }
```

PR #1806, opened 2026-07-23
[Tidy] Implement UX/UI changes

nadijagraca:
## Description
closes https://github.com/McK-Internal/vizro-internal/issues/2845
Changes in this PR:

- [x] - position of collapse/expand button
- [x] - theme toggle icon
- [x] - Design and position of Reset controls button
- [x] - Background colors (it will be address in the following PR)

## Screenshot

Before:

image

Current:

image

## Notice


petar-qb:
Thanks @nadijagraca , for handling this.

I wanted to share some initial thoughts. I haven't reviewed the PR in detail yet, but I will do so once it is complete. I also added “before” and “current” screenshots to the PR description.

So, I really like the new the `Reset all` button. However, its hover state appears initially and then disappears after the button is clicked.

I also like the new theme button. 👍 

Regarding the **Show/Hide control panel** button:
- I am not sure how intuitive it will be for users to find this button. I can imagine someone opening a Vizro app on a small screen and being surprised not to see either the controls or an obvious way to open them. By "obvious way", I mean that users will look at where they expect the controls but not seeing them, neither the button that expands the control panel. 
- The new icon also looks faded compared with the theme toggle button. Are the colour and opacity values the same for both icons? To me, the Show/Hide icon looks disabled in both themes.
- It would also be helpful if the icon's arrow changes the direction based on the control panel state. For example, it could use ◀️ when the control panel is visible and ▶️ when it is hidden.

One more question: Why is the container background colour in the current version different from before when using the same code?

huong-li-nguyen:
Let me know when this is ready for review :) Are all cross-combinations from @stephkaiser design in the scratch dev? I think we have to add screens with all components on the right side (table, graph, kpi card, card), inside and outside of a container, controls left and right / inside - outside of a container etc.

@stephkaiser - are we sure about removing the borders from the left-side? Seems very subtle, but I'll trust whatever you decide in the end.

Screenshot 2026-07-24 at 14 48 27

vizro-core/src/vizro/themes/_templates.py
```diff
@@ -1,10 +1,14 @@
...
 
+from vizro.themes._colors import colors
+
 dashboard_overrides = go.layout.Template(
     layout=go.Layout(
         margin_b=16,
         margin_l=24,
         margin_t=24,
+        paper_bgcolor=colors.transparent,
+        plot_bgcolor=colors.transparent,
```
huong-li-nguyen: We can't make the charts transparent 😬 Because otherwise if you export the charts the screenshots look very off

huong-li-nguyen: Screenshot 2026-07-28 at 09 39 54

This is what you get when you click on the export png button on the chart (dark theme)

huong-li-nguyen:
Not sure about the dropdown bg color inside container. There doesn't seem to be enough contrast in the dark theme, whereas in the light theme there is a lot more contrast. WDYT? @stephkaiser  
I don't want to open a new can of worms. If this gets too complicated/complex, I would leave the bg-colors untouched in this PR

Screenshot 2026-07-28 at 09 52 21
Screenshot 2026-07-28 at 09 52 15

huong-li-nguyen:
@nadijagraca - if you can, I would split out the changes related to bg-colors and the rest. The rest we can easily merge in this PR. I think we might need more discussion on the changes related to the bg-colors

huong-li-nguyen:
> Not sure about the dropdown bg color inside container. There doesn't seem to be enough contrast in the dark theme, whereas in the light theme there is a lot more contrast. WDYT? @stephkaiser I don't want to open a new can of worms. If this gets too complicated/complex, I would leave the bg-colors untouched in this PR
> 

Actually, @nadijagraca , can you double-check the designs? I just checked the Figma designs from Steph and the dropdown has different bg-colors there (menu when open etc.). Maybe double-check with @stephkaiser what bg-color the dropdowns (open menu etc.) should have when placed on the left-side. I don't think there are designs / examples for that in @stephkaiser figma wireframe

stephkaiser:
> > Not sure about the dropdown bg color inside container. There doesn't seem to be enough contrast in the dark theme, whereas in the light theme there is a lot more contrast. WDYT? @stephkaiser I don't want to open a new can of worms. If this gets too complicated/complex, I would leave the bg-colors untouched in this PR
> 
> Actually, @nadijagraca , can you double-check the designs? I just checked the Figma designs from Steph and the dropdown has different bg-colors there (menu when open etc.). Maybe double-check with @stephkaiser what bg-color the dropdowns (open menu etc.) should have when placed on the left-side. I don't think there are designs / examples for that in @stephkaiser figma wireframe

The dropdown menu colour has changed from QBDS so we are aligning with that - they also dont change colour anymore depending on their position, they stay the same colour regardless if its in the panel or in the main content area on the right. 

I'll review the colours in the next PR

stephkaiser:
Thank you @nadijagraca! 

My comments below:

@nadijagraca can we add an example showing both navigation and controls in the left panel so I can see + review them

1) I agree with Petar's points below - can we check the icon colours? and I also love the idea of the direction of the arrow changing (I can supply this icon for you, if you need) 

> The new icon also looks faded compared with the theme toggle button. Are the colour and opacity values the same for both icons? To me, the Show/Hide icon looks disabled in both themes.

> It would also be helpful if the icon's arrow changes the direction based on the control panel state. For example, it could use ◀️ when the control panel is visible and ▶️ when it is hidden.

2) I'm sorry for changing my mind :D but I think I prefer the previous design of the reset button (keep the same text, button is not full width of the panel - just a normal button). Main reason is after seeing it implemented i think its too much/too big, and I remember we wanted a more subtle reset button instead. 

3) Maybe a bug? when i click on the reset button, it works and resets content, but when i go back the hover state isn't appearing. Button still works, just doesnt have the hover state.

4) Can we add a rule where if there is a title and primary button, can they be grouped together at the top so the button is default on top right corner and aligned with the title of the page? We can still allow users to override this rule. Wdyt?
Screenshot 2026-07-31 at 16 21 19

5) Can we reduce the logo size as well? 
Screenshot 2026-07-31 at 16 52 11



everything else works really well and looks great!! Thank you!

nadijagraca:
> Thank you @nadijagraca!
> 
> My comments below:
> 
> @nadijagraca can we add an example showing both navigation and controls in the left panel so I can see + review them
> 
> 1. I agree with Petar's points below - can we check the icon colours? and I also love the idea of the direction of the arrow changing (I can supply this icon for you, if you need)
> 
> > The new icon also looks faded compared with the theme toggle button. Are the colour and opacity values the same for both icons? To me, the Show/Hide icon looks disabled in both themes.
> 
> > It would also be helpful if the icon's arrow changes the direction based on the control panel state. For example, it could use ◀️ when the control panel is visible and ▶️ when it is hidden.
> 
> 2. I'm sorry for changing my mind :D but I think I prefer the previous design of the reset button (keep the same text, button is not full width of the panel - just a normal button). Main reason is after seeing it implemented i think its too much/too big, and I remember we wanted a more subtle reset button instead.
> 3. Maybe a bug? when i click on the reset button, it works and resets content, but when i go back the hover state isn't appearing. Button still works, just doesnt have the hover state.
> 4. Can we add a rule where if there is a title and primary button, can they be grouped together at the top so the button is default on top right corner and aligned with the title of the page? We can still allow users to override this rule. Wdyt?
> 
> Screenshot 2026-07-31 at 16 21 19
> 5. Can we reduce the logo size as well?
> 
> Screenshot 2026-07-31 at 16 52 11
> everything else works really well and looks great!! Thank you!

Hi Steph, 
Thank you for your review. 
1. Implemented. 
2. Reverted the styling. 
3. Resolved. 
4. Design wise, I love the idea, however, it would probably be a breaking change for users with Grid layout. @huong-li-nguyen what do you think? 
5. Implemented.

stephkaiser:
Thanks @nadijagraca, new changes look good! 

some more comments:

1. I'm still getting the bug - here's a recording if thats useful:
https://github.com/user-attachments/assets/0c554212-7e63-4e00-92f5-1b54290dbbfe

2. Thanks for reverting the style of the reset button! one thing i noticed - can we add 24px padding on the bottom please. Current vs Design:
Screenshot 2026-08-05 at 17 22 22
Screenshot 2026-08-05 at 17 34 27

3. For the navigation part, can we remove the divider at the bottom? Current vs Design:
Screenshot 2026-08-05 at 17 30 14
Screenshot 2026-08-05 at 17 38 37

4. Can we check spacing between icons is 24px? i cant check it myself sorry :) it looks a bit wider but maybe its just my eyes! Current vs Design:
Screenshot 2026-08-05 at 17 00 18
Screenshot 2026-08-05 at 17 21 13

huong-li-nguyen:
4) I wouldn't do. We would need lots of conditional logic. E.g. what happens if we have several primary buttons? I think it's too intrusive, and where could they even override it? Would have to rethink API. Probably not worth implementing at this stage.

Let me know when i should review the PR, but it seems like @stephkaiser is still reviewing design first

stephkaiser (approved):
I still have the bug, but everything else looks good! thank you @nadijagraca

huong-li-nguyen:
@stephkaiser @nadijagraca - are we sure about this behaviour? Doesn't seem super consistent. Everything else is working as expected and fine for me to merge, as soon as we align on an answer to below. Can you ask @petar-qb to stress-test this too? He's normally great in capturing all kinds of edge cases 😄 

**How it is currently**
- If there is only the Navpanel, the toggle switch and the collapse/expand button are on the top right
- If there is the Navpanel + NavBar, the toggle switch stays on the top right and the collapse/expand button moves to the NavBar (this doesn't make sense to me? If the collapse/expand button moves to the navbar, why doesn't the toggle switch move?)

General question: Is the collapse / expand button to far away from the side panel? 

Screenshot 2026-08-07 at 16 37 16
Screenshot 2026-08-07 at 16 36 37

stephkaiser:
> @stephkaiser @nadijagraca - are we sure about this behaviour? Doesn't seem super consistent. Everything else is working as expected and fine for me to merge, as soon as we align on an answer to below. Can you ask @petar-qb to stress-test this too? He's normally great in capturing all kinds of edge cases 😄
> 
> **How it is currently**
> 
> * If there is only the Navpanel, the toggle switch and the collapse/expand button are on the top right
> * If there is the Navpanel + NavBar, the toggle switch stays on the top right and the collapse/expand button moves to the NavBar (this doesn't make sense to me? If the collapse/expand button moves to the navbar, why doesn't the toggle switch move?)
> 
> General question: Is the collapse / expand button to far away from the side panel?

The reason why the toggle switch stays on the top right header, and the hide/show panel button moves to the nav bar is because the toggle switch is a global action which affects the entire application so makes sense to stay at header level - The hide/show panel button should be positioned in context with the action its doing - so in this case it makes more sense for the hide/show panel button to be in the nav bar. When there is no nav bar, ideally it should be in the nav panel but without doing a full redesign, the next best place would be to place it with the toggle switch in the top right header. 

So your question "Is the collapse / expand button to far away from the side panel?" Maybe, but the top right is the best place to put it for now without doing further design improvements to the panels. In my opinion, it is also not a big problem because naturally the header should be the next place users would look to find this button.

huong-li-nguyen (approved):

PR #1832, opened 2026-08-10
Add beeswarm chart to visual vocabulary

amritpalsingh85419:
## Notice


nadijagraca:
Hi @amritpalsingh85419 👋 
Thanks so much for picking this up. 

Before we can merge this, there are a couple of things to address:
1. The chart itself isn't quite a beeswarm yet. `beeswarm.py` currently uses `px.strip`, which just randomly jitters points but they can still overlap. A true `beeswarm` packs points so they never overlap, which is what actually reveals the distribution's density. 
Plotly has no native beeswarm trace, so this needs more than a one-line px call. See implementation of `dumbbell.py` or `cumulative_curve.py` for inspiration of how to implement charts that don't have direct plotly express match. This could be custom `@capture("graph")` function built on `go.Scatter `with some extra logic for computing non-overlapping point positions. 
2. `README.md `in visual-vocabulary/ also needs updating.

Happy to help if you have questions. Let us know if you'd like to keep working on this or if you'd prefer we close it for now and you reopen when you have time. No pressure either way, and thanks again for the contribution!

amritpalsingh85419:
Hi @nadijagraca 

Thanks for the detailed feedback! I'd definitely like to keep working on this. I'll look at dumbbell.py and cumulative_curve.py and rework the implementation to use a proper non-overlapping beeswarm layout with go.Scatter. I'll also update the visual-vocabulary README.

vizro-core/examples/visual-vocabulary/pages/examples/beeswarm.py
```diff
@@ -0,0 +1,64 @@
+import pandas as pd
+import plotly.graph_objects as go
+import vizro.plotly.express as px
+from vizro.models.types import capture
+
+tips = px.data.tips()
```
nadijagraca: Let's maybe use iris dataset for this chart. Iris works better because many flowers share the exact same measurement (like several setosa flowers all at 1.5cm), giving the chart real overlapping points to spread apart, tips amounts are almost all slightly different, so there's not much to demonstrate.

vizro-core/examples/visual-vocabulary/pages/examples/beeswarm.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+@capture("graph")
+def beeswarm(data_frame: pd.DataFrame, x: str, y: str, point_spacing: float = 0.12) -> go.Figure:
```
nadijagraca: Lets add another argument `y_tolerance_fraction: float = 0.02` that will be use to calculate overlap. 
Also, set `point_spacing` to 0.05. With 0.05 the points will fan out properly instead of pile up.

vizro-core/examples/visual-vocabulary/pages/examples/beeswarm.py
```diff
@@ -0,0 +1,64 @@
...
+    def offset_candidates():
+        yield 0.0
+        level = 1
+
+        while True:
+            yield level * point_spacing
+            yield -level * point_spacing
+            level += 1
+
+    def overlaps(candidate, value, placed):
```
nadijagraca, line 30: Decouple x and y tolerances, `point_spacing` is doing double duty here as both the x-offset step and the y-collision tolerance. 
The problem is those are different scales because x-offsets are category-index units (0, 1, 2...), while value/placed_value are raw data. With the default 0.12, two points only "collide" in y if they're within 0.12 of that raw value  which is far too tight for most real columns, so the algorithm ends up placing almost everything at offset 0 instead of spreading it out.

**Suggestion**: derive a separate y-tolerance from the column's own range, e.g. 
`y_tolerance =(data_frame[y].max() - data_frame[y].min()) * y_tolerance_fraction ` 

This makes the "closeness" check adapt to whatever data is plotted.

Overlaps would look something like below:
```
y_tolerance =(data_frame[y].max() - data_frame[y].min()) * y_tolerance_fraction 

def overlaps(candidate, value, placed):
    return any(
        abs(candidate - placed_offset) < point_spacing and abs(value - placed_value) < y_tolerance
        for placed_offset, placed_value in placed
    )
```

vizro-core/examples/visual-vocabulary/pages/examples/beeswarm.py
```diff
@@ -0,0 +1,64 @@
...
+import plotly.graph_objects as go
+import vizro.plotly.express as px
+from vizro.models.types import capture
+
+tips = px.data.tips()
+
+
+@capture("graph")
+def beeswarm(data_frame: pd.DataFrame, x: str, y: str, point_spacing: float = 0.12) -> go.Figure:
+    def offset_candidates():
```
nadijagraca, line 21: This generator has no upper bound, so a dense cluster of points can end up spread arbitrarily wide. Since categories sit one unit apart, an offset bigger than 0.5 actually crosses into the neighbouring category's column.  
**Suggestion**: cap it at a safe width, and fall back to stacking at the edge once that width is full, rather than growing forever: `max_offset = 0.5 - point_spacing`

```
max_offset = 0.5 - point_spacing

def offset_candidates():
    yield 0.0
    level = 1

    while level * point_spacing <= max_offset:
        yield level * point_spacing
        yield -level * point_spacing
        level += 1
```

vizro-core/examples/visual-vocabulary/pages/examples/beeswarm.py
```diff
@@ -0,0 +1,64 @@
...
+    categories = data_frame[x].unique()
+
+    for category_index, category in enumerate(categories):
+        category_data = data_frame[data_frame[x] == category].sort_values(y)
+        placed = []
+
+        x_positions = []
+        y_positions = []
+
+        for value in category_data[y]:
```
nadijagraca, line 45: Add an `else` on the `for offset in offset_candidates()` to handle the "ran out of room" case:

```
for value in category_data[y]:
    for offset in offset_candidates():
        if not overlaps(offset, value, placed):
            placed.append((offset, value))
            x_positions.append(category_index + offset)
            y_positions.append(value)
             break
     else:
        offset = max_offset if len(placed) % 2 == 0 else -max_offset
        placed.append((offset, value))
        x_positions.append(category_index + offset)
        y_positions.append(value)
```

vizro-core/examples/visual-vocabulary/pages/examples/beeswarm.py
```diff
@@ -0,0 +1,64 @@
...
+        fig.add_trace(
+            go.Scatter(
+                x=x_positions,
+                y=y_positions,
+                mode="markers",
+                name=str(category),
+            )
+        )
+
+        fig.update_xaxes(
```
nadijagraca: This is inside the for `category_index, category in enumerate(categories) loop`, but the arguments (`tickvals, ticktext, title_text`) don't change between iterations so we can move it  outside the loop, next to f`ig.update_yaxes` below, to avoid the redundant calls.

nadijagraca (changes requested):
Hi @amritpalsingh85419, 
Thanks for addressing the comments. Page, README and chart_groups.py now look good, and the beeswarm implementation is looking better as well.

I did leave a couple of comments on the beeswarm implementation. The suggestions address a few things that affect how the chart actually renders, so it would be good to address those before merging. 

Happy to walk through any of the comments in more detail if helpful. Once they’re addressed, feel free to request a re-review. Thanks again for the contribution!

amritpalsingh85419:
> Hi @amritpalsingh85419, Thanks for addressing the comments. Page, README and chart_groups.py now look good, and the beeswarm implementation is looking better as well.
> 
> I did leave a couple of comments on the beeswarm implementation. The suggestions address a few things that affect how the chart actually renders, so it would be good to address those before merging.
> 
> Happy to walk through any of the comments in more detail if helpful. Once they’re addressed, feel free to request a re-review. Thanks again for the contribution!

Thanks @nadijagraca! I've addressed the remaining beeswarm implementation feedback, including the Iris dataset, separate y-tolerance, bounded offsets, fallback placement, and moving the x-axis update outside the loop. I've pushed the changes and would appreciate a re-review

vizro-core/examples/visual-vocabulary/pages/examples/beeswarm.py
amritpalsingh85419: Done, I've switched the example to the Iris dataset so the overlapping measurements demonstrate the beeswarm behavior more clearly.

nadijagraca (approved):
Hi @amritpalsingh85419 , 
Thank you addressing the comments, the chart looks good now. 
Please address co-pilot comments and update main, after that feel free to merge. :)
Thank you again for your contribution. 🚀

amritpalsingh85419:
> Hi @amritpalsingh85419 , Thank you addressing the comments, the chart looks good now. Please address co-pilot comments and update main, after that feel free to merge. :) Thank you again for your contribution. 🚀

Thanks @nadijagraca , I've addressed the requested changes and updated the branch with the latest main. The implementation and test checks are now passing. Thanks for the review and guidance!

nadijagraca (approved):
Hi @amritpalsingh85419, 
Thank you for addressing all the comments, really great job! 🙌 🚀 

Can you please resolve conflicts and then we can merge. 
Also, please make sure to add yourself to the list of contributors, you can find it in 
`vizro-core/docs/pages/explanation/authors.md‎`

amritpalsingh85419:
Hi @nadijagraca, I've resolved the conflicts with the latest main and added myself to vizro-core/docs/pages/explanation/ authors. md as requested. The branch is updated and ready for merge. Thanks again for the review!

PR #1837, opened 2026-08-18
[Fix] Time filter raises on a time-of-day column that is not datetime64

### LINKED ISSUE #1836: [Bug] Time filter raises on a time-of-day column that is not datetime64
### Which package?

vizro-core

### Package version

0.1.61.dev0 (`main` at 99634b8)

### Description

A `Filter` on a time-of-day column raises when the column is not `datetime64`. The failure lands in `_coerce_temporal`, in the `_is_time` branch: the series is only converted when it is `datetime64`, but the precision normalisation immediately after runs over whatever the column actually holds.

```python
# vizro-core/src/vizro/models/_controls/filter.py
if _is_time:
    if is_datetime64_any_dtype(series):
        series = series.dt.time          # only for datetime64

    if normalize_precision:
        series = series.map(lambda v: v.replace(microsecond=0))   # runs for every dtype
```

For a column of `"HH:MM"` strings that calls `str.replace(microsecond=0)`. For a `datetime.time` column containing a null it calls `None.replace(...)`. Where it does not raise, it ends up comparing `str` against `datetime.time`.

The string case is the one that bites without any user interaction. A column of `"09:00"` strings is classified `categorical` (`infer_dtype` returns `string`, not `time`), so a `Dropdown` is auto-selected, and its default value is every option — all of which match the time regex inside `_coerce_temporal`. So the filter takes the `_is_time` path on page load and the `_on_page_load` callback returns HTTP 500 before the user touches anything.

A `datetime.time` column auto-selects a `TimePicker`, whose default is the empty range `['', '']`, and that short-circuits. That path raises when a value is actually selected.

Measured behaviour of the two filter functions:

| column | `_filter_isin` | `_filter_between` |
| --- | --- | --- |
| `datetime.time`, no nulls | ok | ok |
| `datetime.time` with `None` | `AttributeError: 'NoneType' object has no attribute 'replace'` | ok |
| `"HH:MM"` strings | `TypeError: str.replace() takes no keyword arguments` | `TypeError: '>=' not supported between instances of 'str' and 'datetime.time'` |
| `datetime64` | ok | ok |

`datetime64` columns, which are the common case, are unaffected.

### How to Reproduce

No dashboard needed:

```python
import pandas as pd
from datetime import time
from vizro.models._controls.filter import _filter_isin, _filter_between

_filter_isin(pd.Series(["09:00", "10:00"]), ["09:00"])
# TypeError: str.replace() takes no keyword arguments

_filter_between(pd.Series(["09:00", "10:00"]), ["08:00", "09:30"])
# TypeError: '>=' not supported between instances of 'str' and 'datetime.time'

_filter_isin(pd.Series([time(9, 0), None, time(10, 0)]), ["09:00"])
# AttributeError: 'NoneType' object has no attribute 'replace'
```

End to end, a page with a `Filter` on a string time column returns HTTP 500 from the page-load callback, with the traceback running `_get_modified_page_figures` → `_apply_filters` → `_apply_filter_controls` → `_filter_isin` → `_coerce_temporal`.

### Output

```
File ".../vizro/models/_controls/filter.py", line 145, in <lambda>
    series = series.map(lambda v: v.replace(microsecond=0))
TypeError: str.replace() takes no keyword arguments
```

I have a fix and tests ready and will open a PR referencing this issue.

winklemad:
## Description

Fixes #1836.

`_coerce_temporal` only converts the series when it is `datetime64`, then normalises precision over whatever the column actually holds:

```python
if _is_time:
    if is_datetime64_any_dtype(series):
        series = series.dt.time          # only for datetime64

    if normalize_precision:
        series = series.map(lambda v: v.replace(microsecond=0))   # runs for every dtype
```

So a time-of-day column that is not `datetime64` either raises or compares mismatched types:

| column | `_filter_isin` before | `_filter_between` before |
| --- | --- | --- |
| `"HH:MM"` strings | `TypeError: str.replace() takes no keyword arguments` | `TypeError: '>=' not supported between instances of 'str' and 'datetime.time'` |
| `datetime.time` with a null | `AttributeError: 'NoneType' object has no attribute 'replace'` | ok (nulls already compare `False`) |
| `datetime.time`, no nulls | ok | ok |
| `datetime64` | ok | ok |

The string case needs no user interaction. `infer_dtype` reports `string` rather than `time` for a `"09:00"` column, so it is classified `categorical` and a `Dropdown` is auto-selected. The dropdown's default value is every option, all of which match the time regex in `_coerce_temporal`, so the `_is_time` path runs during `_on_page_load` and the callback returns HTTP 500 on first render.

A `datetime.time` column auto-selects a `TimePicker`, whose default is the empty range `['', '']`; that short-circuits, so it raises when a value is selected rather than on load.

The change adds a small `_to_dt_time` helper and uses it to coerce the series for non-`datetime64` dtypes, then guards the two precision-normalisation maps with an `isinstance` check so nulls pass through untouched and simply do not match.

`datetime64` and clean `datetime.time` columns take exactly the same path as before. Both are covered by regression tests that pass on `main` as well as here, so the diff is provably additive for the common case.

Scope is limited to the `_is_time` branch. I did not touch the `_is_date` / `_is_datetime` branches or the mixed-precision range padding above them.

## Screenshot

Not applicable — no visual change. The user-visible effect is that a page which previously returned HTTP 500 on load now renders and filters correctly.

## Notice


winklemad:
Thanks — the first point was a real bug in my patch, and a nastier one than the crash it was meant to fix.

`pd.Timestamp("2024-01-01").time()` is `00:00`, so a date-only entry was silently becoming midnight and *matching* early-morning filters. Same for a `datetime.date`, and for something like a bare integer. That contradicted the docstring I'd written two lines above it, which promised anything unreadable becomes null.

Now only values that actually carry a time of day are converted:

```python
if isinstance(entry, dt_time):
    return entry
# Checked before dt_date because dt_datetime subclasses it, and only the former has a time of day.
if isinstance(entry, dt_datetime):
    return entry.time()
if isinstance(entry, str) and _TIME_REGEX.match(entry):
    ...
return None
```

So `"2024-01-01"`, `date(2024, 1, 1)` and `5` all become null and don't match, while `datetime(2024, 1, 1, 10, 30)` keeps `10:30`. Three cases added to `test_filter_isin_time` covering exactly that, including the midnight false-positive.

On the second point, the two normalisation maps are now one pass — `strip_seconds` is computed once from `value` rather than re-derived per element. I left the string-only vectorised path alone: it would only help when the column is uniformly strings, and it costs a branch that the entry-by-entry map already handles.

Third point folded in — the docstring now names what counts as a time of day and says explicitly that a date-only value is treated as null rather than midnight.

`pytest tests/unit` is 2069 passed, 27 xfailed; ruff check and format are clean.

petar-qb (approved):
This is such an amazing contribution @winklemad!! 🚀 🏅 Thankss

The code, comments, and PR description made the review process incredibly smooth and easy!

You’d also be very welcome to take a look at the `date` and `datetime` parts and potentially fix any similar issues you spot there! 🙏

PR #1838, opened 2026-08-18
[Fix] Slider step validation raises when max is set without min

winklemad:
## Description

A slider given `max` and `step` but no `min` fails to validate:

```python
import vizro.models as vm

vm.RangeSlider(max=6, step=1)
# TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'
```

`validate_step` checks that `max` is set, then uses `min` regardless:

```python
if info.data["max"] is not None and step > (info.data["max"] - info.data["min"]):
```

`vizro-core/src/vizro/models/_components/form/_form_utils.py`

With `min` unset that subtraction is `6.0 - None`, so the error surfaces during model validation and the dashboard never builds. The guard now requires both bounds before comparing `step` against the range; the existing `ValueError` and its message are untouched for the case where both are supplied.

The three neighbouring shapes already work and stay working: `min` and `step` without `max`, `min` and `max` and `step`, and `step` on its own.

`validate_step` is shared by `Slider` and `RangeSlider`, so tests were added on both sides.

## Screenshot

Not applicable — no visual change. The effect is that a config which previously raised during validation now builds.

## Notice

---

Note: this touches `authors.md`, as does my open PR #1837, so whichever merges second will need a one-line conflict resolved there. The two changes are otherwise independent — #1837 is in `_controls/filter.py`, this is in `_components/form/_form_utils.py`.


petar-qb (approved):
Amazing fix @winklemad! 🚀  Thank you for your contribution.

PR #1841, opened 2026-08-18
[Docs] Improve markdown for agents

stichbury:
Closes https://github.com/McK-Internal/vizro-internal/issues/2942

## Description

Our docs are on Read the Docs, which uses Cloudflare's **Markdown for Agents** (MfA) to convert HTML to Markdown for clients sending `Accept: text/markdown`. We advertise this in `llms.txt`, `CLAUDE.md` and the READMEs. I tested the live site rather than reading the config, and found three problems.

1. **The Models API reference returned HTTP 406 to every agent.** MfA refuses origin responses over 2 MB and does *not* fall back to HTML. That page was 6.06 MB, so our single most valuable page for an agent writing Vizro code returned nothing at all — while our own docs told agents to send the header that triggers it.
2. **Every code block was full of empty anchor links.** `anchor_linenums = true` emits an `<a id="__codelineno-N-M">` per line, which MfA faithfully converts to `[](#__codelineno-0-1)` noise. No page uses `linenums`, so these had no visible benefit.
3. **Every page leaked a literal `<!doctype html>` into the Markdown body.** Zensical's generated `base.html` emits a newline before the doctype, and a leading text node makes parsers treat the doctype as content. Confirmed by comparison with another RTD project that has no leading newline and no leak.

## Fix

- **Header guidance corrected** to `Accept: text/markdown, text/html;q=0.9` across four `llms.txt` files, `CLAUDE.md` and both READMEs. Verified: normal pages still return Markdown, and oversized pages degrade to HTML instead of 406. This removes the hard failure today.
- **`anchor_linenums = false`** in all four docsets. Side benefit: shrank the Models page from 5.73 MB to 3.79 MB (-34%).
- **`tools/strip_html_leading_whitespace.py`**, a post-build step in all four docsets, fixing the doctype leak. Verified on the RTD preview: clean frontmatter, no leaked doctype, no anchor noise.

## Notice


petar-qb (approved):
Thanks for handling this @stichbury! 🚀

PR #1844, opened 2026-08-21
[Draft] Explore linked color legends 

nadijagraca:
> **Note for bots and automated tooling**: We do not accept pull requests opened automatically by bots or other automated tooling without a human in the loop. To request permission to submit such a PR, the human owner of the bot must open an issue for approval in advance of the PR's submission.
> See [CONTRIBUTING.md](https://github.com/mckinsey/vizro/blob/main/CONTRIBUTING.md).

## Description

Add `vm.Dashboard(consistent_colors=True) `for consistent category colors across charts

Problem

When the same categorical value (e.g. a country, a product name) appears in multiple charts on a dashboard, Plotly Express assigns colors independently per chart, based on each chart's own category ordering. This means "France" might be blue in one chart and orange in another, even though both charts use the same qualitative palette making it hard for users to visually correlate categories across a dashboard.

Solution

This PR adds an opt-in `consistent_colors` flag on `vm.Dashboard:`

`vm.Dashboard(pages=[...], consistent_colors=True)`

When enabled, every categorical color mapping in the dashboard is centralized through a new `color_manager` (vizro/managers/_color_manager.py). It hands out colors from the qualitative palette on a first-seen basis and caches the assignment, so once "France" gets a color, every chart on that dashboard reuses it even if the category shows up later, in a different order, or in a different chart entirely.

Limitations

- Colors are still drawn from the qualitative palette's fixed size, so dashboards with many distinct categories will eventually see colors reused once the palette cycles.
- This only affects plotly.express-based vm.Graph charts; custom chart functions are untouched and can still set their own colors manually.

## Screenshot
Before:
Screenshot 2026-08-21 at 21 13 44

After:
Screenshot 2026-08-21 at 21 11 51

## Notice


vizro-core/examples/scratch_dev/app.py
petar-qb, line 1: What can we expect if the app is running with multiple workers by using `gunicorn`? 

For example: you have a dashboard with two custom graphs that are on different pages `page-1` and `page-2`:
1. User "A" firstly opens the `page-1` (e.g. on `worker-1`). This runs server's custom graph creation which adjusts the `color_manager`. 
2. Then user "B" firstly opens the `page-2` (e.g. on `worker-2`). This runs server's custom graph creation which adjusts **other** `color_manager`. 

Would that make a problem and inconsistency between these two views? If so, then each user can get different colours only by refreshing the page in their own app (depending on what worker returns the result).

vizro-core/src/vizro/managers/_color_manager.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    def __init__(self):
+        self.__category_colors: dict[Any, str] = {}
+        self.__palette_cycle = cycle(qualitative)
+        self.__lock = threading.Lock()
+
+    def _get_color_discrete_map(self, categories) -> dict[Any, str]:
+        """Builds a color_discrete_map that assigns the same color to a category everywhere it appears."""
+        with self.__lock:
```
petar-qb, line 25: Let's add a comment of why `with self.__lock:`, so when we expect `_get_color_discrete_map` to be called more than once at the same time. Actually, this is the line that needs the lock I think:
`self.__category_colors[category] = next(self.__palette_cycle)`.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -32,10 +32,69 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+def _consistent_colors_enabled(graph: "Graph") -> bool:
+    """Whether the dashboard containing `graph` has opted in to vm.Dashboard(consistent_colors=True).
+
+    Looks up the specific Dashboard that owns `graph` (via its Page), rather than any Dashboard that
+    happens to be registered in model_manager - multiple Dashboards can coexist in one process (e.g. in a
+    notebook or test suite) without an intervening Vizro._reset().
```
petar-qb, line 45: In general I'm wondering how Vizro works with for example two coexisting Dashboards. Let's try to make two dashboards (one with consistent_colors=True and one with False) running on two different ports and see what happens.

vizro-core/src/vizro/models/_components/graph.py
```diff
@@ -181,6 +240,7 @@ def __call__(self, **kwargs):
         # If the functionality of process_callable_data_frame moves to CapturedCallable then this would move there too.
         if "data_frame" not in kwargs:
             kwargs["data_frame"] = data_manager[self["data_frame"]].load()
+        kwargs = _apply_consistent_colors(self, kwargs)
```
petar-qb, line 243: What if `color_discrete_map` is overwritten statically or dynamically with the Parameter like here https://vizro.readthedocs.io/en/stable/pages/user-guides/parameters/#nested-parameters?

petar-qb (commented):
Generally, it looks good, and I really like the idea of having this feature if it turns out to work well! 🚀 So I have more questions than actual suggestions 😄

You mentioned in the PR description:
> Colors are still drawn from the qualitative palette's fixed size, so dashboards with many distinct categories will eventually see colors reused once the palette cycles.

Isn't this already an issue with `consistent_colors: bool = False`?

The underlying problem is the fixed palette. For example, if you have 11 categories, the `first` and `eleventh` categories will get the same color. This happens regardless of whether `consistent_colors` is enabled.

The difference becomes visible when the data is filtered so that only those two categories (`first` and `eleventh`) remain:

* with `consistent_colors: bool = False`, they will get different colors because Plotly assigns colors dynamically based on the currently visible categories and 
* with `consistent_colors: bool = True`, they will keep their original colors and therefore both have the same color.

So I see the disadvantage of this approach, but I still think it brings more value than it takes away. Given that we're introducing the stable `vizro==1.0.0`, I would even consider accepting the visual breaking change and switching the default from `consistent_colors: bool = False` to `consistent_colors: bool = True`.

vizro-core/src/vizro/models/_components/graph.py
nadijagraca: User defined `color_discrete_map` always takes presedence over our generated one, so this example will work as intended.

vizro-core/src/vizro/models/_components/graph.py
nadijagraca: The `_consistent_colors_enabled()` looks up the specific dashboard owning the graph. Multiple dashboards (one with `consistent_colors=True`, one without) can coexist safely.

vizro-core/examples/scratch_dev/app.py
nadijagraca: Yes, this can happen. Each gunicorn worker is a separate process with its own `color_manager` instance, so if `page-1` and `page-2` are first discovered on different workers, the same category can land on different palette positions in each worker's memory.

In practice the impact is limited as colors are consistent within a worker, so a given user keeps seeing the same colors across requests as long as they stay on that worker. The inconsistency only shows up if two different users happen to be routed to different workers and compare notes, or if --preload isn't used and workers boot at different points.

Maybe we should add a note to the consistent_colors docstring calling this out as a known limitation?

petar-qb: > In practice the impact is limited as colors are consistent within a worker, so a given user keeps seeing the same colors across requests as long as they stay on that worker. The inconsistency only shows up if two different users happen to be routed to different workers and compare notes, 

There is nothing that guarantees that the same user will always receive responses from the same worker. `gunicorn` does not provide worker affinity for a given client, so subsequent requests from the same user may be handled by different workers. Therefore, we have to assume that each client request, such as opening a page, applying a filter, or exporting data, may be handled by a different worker.

> or if --preload isn't used and workers boot at different points.

Unfortunately, the `--preload` flag does not solve the issue either. With `--preload`, `Vizro.build` runs before the `gunicorn` workers are forked, which triggers each `page.build`. However, the `color_manager` color-to-category assignment happens inside `Graph.__call__`, which is invoked for the first time only when a user opens the page containing that graph. At that point, the workers have already been forked, so the assignment can still differ between workers.

---

Given that any production-grade hosted dashboard application is likely to be served by multiple workers, I think this issue is significant enough that we should not proceed with the feature in its current form, and I would close the PR.

If this cannot be fixed, and I don't see a straightforward solution other than preemptively constructing every graph before the workers are forked, I don't think we should proceed with the current approach.

I'd be interested to hear @nadijagraca's and @antonymilne's opinions as well.

petar-qb: Even if we `pre-build` all graphs and pre-fetch all `data_frame`s and recommend using the `--preload` flag, we still cannot guarantee consistent colours after the same user applies something as simple as filtering..

Dynamic data_frame functions can introduce new categories at any point, which may result in different colours being assigned to them by different workers.

Since there doesn't seem to be a reliable way to solve this, I suggest we close this PR with the reasons mentioned here and above and instead document how users can explicitly define colours for categories when needed.

PR #1852, opened 2026-09-02
[QA] Datetimepicker and timepicker as a source for set_control

l0uden:
## Description

Closes https://github.com/McK-Internal/vizro-internal/issues/2963

## Notice


petar-qb (commented):
First of all, thanks a lot for proving the bug and writing the tests upfront! 🚀 

I thought you will inject new use cases (`TimePicker(range=True)` and `DateTimePicker(range=True)`) within existing `test_set_control_cross_filter_graph_non_categorical` and `test_set_control_cross_filter_aggrid_non_categorical` tests (and related pages), instead of creating new tests and pages. Do you see any disadvantages of that approach except that the currently passing tests will fail until I fix the bug.

l0uden:
> First of all, thanks a lot for proving the bug and writing the tests upfront! 🚀
> 
> I thought you will inject new use cases (`TimePicker(range=True)` and `DateTimePicker(range=True)`) within existing `test_set_control_cross_filter_graph_non_categorical` and `test_set_control_cross_filter_aggrid_non_categorical` tests (and related pages), instead of creating new tests and pages. Do you see any disadvantages of that approach except that the currently passing tests will fail until I fix the bug.

It is hard to control or use existing data for this purposes. That is why i decided to create separate ones.

petar-qb:
@l0uden 
>It is hard to control or use existing data for this purposes. That is why i decided to create separate ones.

Maybe I'm missing something, but can't you just add `df["time_column"] = ...` and `df["date_column"] = ...` to the existing `data_frame` and then add those two into the `custom_data` of the source `set_control` graph (same for ag-grid), add two filters with corresponding two `set_control` actions from both graph and ag-grid?

l0uden:
> @l0uden
> 
> > It is hard to control or use existing data for this purposes. That is why i decided to create separate ones.
> 
> Maybe I'm missing something, but can't you just add `df["time_column"] = ...` and `df["date_column"] = ...` to the existing `data_frame` and then add those two into the `custom_data` of the source `set_control` graph (same for ag-grid), add two filters with corresponding two `set_control` actions from both graph and ag-grid?

What I wanted to say is that because of big amount of filters it is hard to control output data. And in the existing tests it is  even harder, because the output is graph. That is why I made output in the new tests as ag_grid. For the precise checking.

petar-qb (approved):
> What I wanted to say is that because of big amount of filters it is hard to control output data. And in the existing tests it is even harder, because the output is graph. That is why I made output in the new tests as ag_grid. For the precise checking.

Still don't get it to be honest 😄 However, let's merge it as it is and then we can refactor it after merging the bugfix PR.

l0uden:
> > What I wanted to say is that because of big amount of filters it is hard to control output data. And in the existing tests it is even harder, because the output is graph. That is why I made output in the new tests as ag_grid. For the precise checking.
> 
> Still don't get it to be honest 😄 However, let's merge it as it is and then we can refactor it after merging the bugfix PR.

Agreed. Let's talk about it later.

